PR #58, opened 2016-07-16
Known telescopes

bhazelton:
Changes to support a known telescope list.

Major changes to underlying structure (but not API) include:
-- creating a UVBase class for all classes that have UVParameters so the code could be reused for a Telescope object
-- moving UVParameter class to a new file to support the UVBase class
-- moving ITRF xyz <-> lat/lon/alt computations to a utils file so they could be reused

New features:
-- Telescope objects
-- known telescope list, currently including PAPER, HERA, MWA
-- method to match UVData telescope name to known telescope list and update missing UVParameters based on the match (added to the end of all read functions)

Open questions/issues I'd like feedback on:
-- is the overall design the right one?
-- should the paper entry be PAPER or PAPER_SA? (The later causes some tests to fail at the moment but those could be fixed if it's the right thing to do)
-- I commented one of Adam's tests in test_uvdata, lines 307-310 because I couldn't figure out how to fix it properly. I need some help to fix it up.


bhazelton:
Suggested changes from telecon:

change match_telescope name
only 1 warning per match
add force keyword to replace all

center_xyz + frame real parameters
lat/lon/alt as properties locked to center_xyz/frame

pass telescope name to get_telescopes & return the one that matches

method to return all valid telescope names

bhazelton:
Ok, I think I've fixed all the things we talked about on the last two telecons.

jpober:
I think everything in here looks reasonable, but I'm afraid I don't understand the changes in enough detail to resolve the merge conflicts.

bhazelton:
I rebased the branch on master so it should be good to go now.

PR #129, opened 2017-02-21
First version of UVCal module and new calfits file format

zakiali:


bhazelton:
Notes from code review discussion:

rename cal.py -> uvcal.py (object & filename should match)

lst_array: should it be a parameter on the cal solution?

flag array:
	is this the flags found in cal or an and of the input & cal derived flags.
	We agree that this should be an *and* of the input array and the flags introduced in flagging. Add to description to be clear.

x_orientation:
	Should this also be added to uvdata?

if cal_type is delay, still need flag array to have length N_freq

in set_gain and set_unknown, fix form of quality_array to have length N_freq


make origin, git_hash (of the calibration code) an optional parameter on uvcal

fix frequency units in calfits

add channel_width as a required parameter on UVCal. Use this in the write method for CDELT
(Do we have a test for how to deal with one frequency and channel_width in uvdata?)

Same thing for integration_time.

Issue with delay type calfits files:
	delays don’t have a frequency axis
	flags do.
	what to do?
		Maybe 2 hdus, but that adds complexity
		convert to gains

check ordering & step size on polarization array to determine CDELT

make a header item to specify if this is a gain or delay style calfits file.

Handle history items properly (FITS style) — see uvfits.py

bhazelton:
Something that just came up in a discussion with Emil has me thinking.

Generally calibration solutions are for the 2 polarizations of the antenna, rather than for the 4 crossed polarizations of the visibilities. So it's not clear to me that polarization should take on the same values for the calibration object that it does for UVData. Miriad apparently makes this distinction by calling the antenna-based ones 'feeds' and the visibility based one 'pols'. I haven't really looked into what uvfits does, but we should think this through carefully.

zakiali:
That's a good point that always slips my mind. The way I was thinking about this was that I would only ever use the 'XX'/'YY' pols when saving calibration solutions which would correspond to X/Y feeds. So we would ignore cross polarizations. But there should probably be a better scheme for specifying feeds instead of polarization products.

bhazelton:
Yes. I think we need to think this through some more, and I'd like to have a larger group on the phone to discuss it because it's subtle and we want to get it right. Zaki, can you set up a when is good poll to find a time next week to talk about this. I'd like to get as many of us as possible together so we can work this out and agree on something.

pyuvdata/uvcal.py
```diff
@@ -0,0 +1,289 @@
...
+                                                   form=('Npols',))
+
+        desc = ('Array of times, center of integration, shape (Ntimes), ' +
+                'units Julian Date')
+        self._time_array = uvp.UVParameter('time_array', description=desc,
+                                           form=('Ntimes',),
+                                           expected_type=np.float,
+                                           tols=1e-3 / (60.0 * 60.0 * 24.0))
+
+        desc = ('Integration time of a time bin. Units in JD')
```
bhazelton: This has different units than the UVData integration_time parameter (which has units of seconds). I think it should match UVData.

pyuvdata/uvcal.py
```diff
@@ -0,0 +1,289 @@
...
+            raise ValueError('Nants_data must be less than or equal to Nants_telescope')
+        return True
+
+    def set_gain(self):
+        """Set cal_type to 'gain' and adjust required parameters."""
+        self.cal_type = 'gain'
+        self._gain_array.required = True
+        self._delay_array.required = False
+        self._quality_array.form = self._gain_array.form
+        self._flag_array.form = self._gain_array.form
```
bhazelton: I don't think you need this line in any of the set_* methods because it's always the same and none of them change it. You need the quality one because it's different for the different cal types

bhazelton:
So the conclusion after talking about cal solution polarizations is to replace polarization_array (which was identical to the UVData polarization_array) with a jones_array which has possible elements for the diagonal terms and off-diagonal terms of the jones matrix (for either linear or circular pol). The object supports having only some of the terms present for calibration codes that only calculate some of the terms. Along with that change, Npol changed to Njones to give the number of jones terms present.

pyuvdata/calfits.py
```diff
@@ -0,0 +1,194 @@
...
+
+        if self.cal_type == 'unknown':
+            raise ValueError("unknown calibration type. Do not know how to"
+                             "store parameters")
+
+        # header ctypes for NAXIS
+        prihdr['CTYPE5'] = ('ANTS', 'Antenna numbering.')
+        prihdr['CUNIT5'] = 'Integer'
+        prihdr['CRVAL5'] = 0
+        prihdr['CDELT5'] = -1
```
bhazelton: I'm a little worried that you're starting this array at zero and then incrementing by -1, are you sure this is what you want? Also, I'm not sure what 'Antenna numbering' in your comment means because you also have a variable called antenna_numbers that I don't think is related to this.

pyuvdata/calfits.py
```diff
@@ -0,0 +1,194 @@
...
+
+        prihdr['CTYPE3'] = ('TIME', 'Time axis.')
+        prihdr['CUNIT3'] = ('JD', 'Time in julian date format')
+        prihdr['CRVAL3'] = self.time_array[0]
+        prihdr['CDELT3'] = self.integration_time
```
bhazelton: How should a calibration solution that is not time dependent but valid for a range of times be handled here? I think it's clear it should have a length of one along this axis, but I'm not sure how someone would specify the range of times it  applies to.

pyuvdata/calfits.py
```diff
@@ -0,0 +1,194 @@
...
+
+        # more checks for polarization. check ordering.
+        prihdr['CTYPE2'] = ('JONES', 'Jones matrix array')
+        prihdr['CUNIT2'] = ('Integer', 'representative integer for polarization.')
+        prihdr['CRVAL2'] = self.jones_array[0]  # always start with first jones.
+        prihdr['CDELT2'] = -1
```
bhazelton: It doesn't look like you're checking that the ordering is what you expect here. I think you probably need to check that they all move by the same step size between the different polarizations and set CDELT2 to that step size. You can look at what I did in on the master branch in uvfits.py at line ~347 to see how I handled this problem (it really became an issue when I added the select function).

pyuvdata/calfits.py
```diff
@@ -0,0 +1,194 @@
...
+
+        col1 = fits.Column(name='ANTNAME', format='8A',
+                           array=self.antenna_names)
+        col2 = fits.Column(name='ANTINDEX', format='D',
+                           array=self.antenna_numbers)
+        cols = fits.ColDefs([col1, col2])
+        ant_hdu = fits.BinTableHDU.from_columns(cols)
+
+        sechdr = fits.Header()
+        secdata = self.flag_array.astype(np.int64) # Can't be bool
```
bhazelton: It doesn't look like you have any axis descriptions for this secondary header which contains the flags. I don't think there's enough information without it is there? Also, when you have a gain cal type, I think the flags should go into the same HDU with the data to cut down on confusion (and so you don't have to repeat all the header stuff). I know that will lead to variable numbers of columns for the different types, but I think that's ok (it certainly happens in uvfits files). I tend to think that the gain type is going to be the most common file type (since you need it to represent a full calibration solution), so I think we want it to optimize for it.

I also don't see anything either here in the write or in the read about the input_flag_array. I think we should write it and read it if it exists.

pyuvdata/calfits.py
```diff
@@ -0,0 +1,194 @@
...
+        sechdu = fits.ImageHDU(data=secdata, header=sechdr)
+        hdulist = fits.HDUList([prihdu, ant_hdu, sechdu])
+        hdulist.writeto(filename, clobber=clobber)
+
+    def read_calfits(self, filename, run_check=True, run_check_acceptability=True):
+        F = fits.open(filename)
+        data = F[0].data
+        hdr = F[0].header.copy()
+
+        antdata = F[1].data
```
bhazelton: You should be checking for the HDU name here, rather than hard coding an ordering. Fits files are explicitly allowed to have any ordering of HDUs after the first one. Check out line 223 in uvfits.py to see how we find the antenna table.

pyuvdata/calfits.py
```diff
@@ -0,0 +1,194 @@
...
+        self.Nants_data = hdr['NANTSDAT']
+        self.Nants_telescope = hdr['NANTSTEL']
+        self.gain_convention = hdr['GNCONVEN']
+        self.x_orientation = hdr['XORIENT']
+        self.cal_type = hdr['CALTYPE']
+        try:
+            self.observer = hdr['OBSERVER']
+        except: pass
+        try:
+            self.pipeline = hdr['CALPIPE']
```
bhazelton: I don't think we have something called pipeline on the uvcal object anymore, so this should be removed.

pyuvdata/calfits.py
zakiali: This is a place holder and has bogus values. In sticking to fits conventions, each NAXIS requires a CTYPE variable to describe it. I could ignore the CRVAL5 and CDELT5 variables and leave it at that. This would avoid confusion with the antenna_numbers variable (which is what CTYPE5 was supposed to capture, but because antenna names are not always contiguous, it is hard to capture with this variable).

pyuvdata/calfits.py
zakiali: That is a good point. There probably needs to be another variable to describe applicable jds. 
Maybe something called self.apply_time_array or something along those lines?

pyuvdata/calfits.py
bhazelton: I see. Maybe call it the antenna axis and refer to the antenna_numbers and antenna_names variables? I think probably ditching everything other than the CTYPE makes sense.

pyuvdata/calfits.py
bhazelton: I was trying to compare this to what we do with frequency when the cal_type is delay and I realized I don't actually understand what's going on there. I see that the comment is that it's 'Valid frequencies to apply delay', but it only grabs the first frequency and then it supplies channel_width for the delta, so I don't see how that gives a range of frequencies.

zakiali: I think I was relying on NFREQS being the the number that gets you the range. However, I think some new variables need to be created to define the range of times/frequencies the solutions apply too. Maybe something like NVALIDFREQS/NVALIDTIMES? These would be integers and we would turn them into a range in the usual way using the RVAL and DELT variables.

bhazelton: I think that could work, but I'd lean toward something more explicit, like a freq_range and time_range with two numbers with units.

In addition to clarity, one reason that I prefer this approach is that it's not clear to me that times in uvdata objects all have to be equally separated (and indeed we have discussed doing cat's eye averaging for HERA data). Of course, if that did happen we'd have to make some other changes to cal objects and files, so this might not be something to worry about now.

zakiali: I added new parameters to hold the ranges of when solutions are valid.

bhazelton:
Todos from telecon:

- [ ] add input flag array to data file (read/write)

- [ ] add to tutorial for cal stuff

- [ ] parameters -> uvdata_parameters

bhazelton:
I don't this this is a blocker for this PR, but the coverage on calfits.py is 87%, which is a little lower than for the other files in the repo. @adampbeardsley could you maybe assist @zakiali with improving the calfits test coverage?

adampbeardsley:
Happy to. Looks like the biggest gaps are due to not testing `Njones > 1` and `input_flag_array`. Do we have any test files that would hit those places? I can make them artificially to test internal consistency, but it would be good to check against an external file.

zakiali:
The current files I have for testing do not have Njones>1 or input_flag_array's. I'm currently trying to get heracal in a state where I can run omnical so that I can generate files with those attributes. Is it worth making them artificially now and then updating once I generate some better data files?

dannyjacobs:
This looks good.  README is updated including a note that the one supported file format, calfits is a new format and still in beta.

Good job done by team Cal: Zaki Ali (lead author), Bryna Hazelton (merge management), and Adam Beardsley (test coverage)

Issue #111, opened 2016-12-14
Add ability to match two files along various axes (e.g. LST locking, flagging, etc)


bhazelton:
Based on ability to select, but may require more work.

bhazelton:
Should write relevant info into history.

bhazelton:
This needs to be better defined, but may be very do-able now that we have select. It might even go in uvtools.

Issue #38, opened 2016-06-01
Select function


jpober:
Method of the object to downselect times/pols/freqs etc.

bhazelton:
I'm working on this on the select branch. I now have methods on UVData to select specific blts, antennas, times, frequencies or polarizations and to update all the appropriate parameters. I also have some tests, although those should probably be expanded. Do we want a method that can take selections on more than one of these axes and then call the appropriate axis-specific method?

jpober:
I think that's a good idea, a uv.select() which calls e.g. select_antennas().  Would the syntax work to have select() able to call multiple axis-specific selectors simultaneously?

bhazelton:
I think right now I'd have it call them sequentially but that would be
hidden from the user, so if you passed restrictions on several axes it'd do
all of them before returning. If that sounds good I'll just need to think a
little bit about what to call the various keywords for the different axes
so that it'll be clear.

bhazelton:
Implemented.

PR #135, opened 2017-03-08
Added a select function (issue #38)

bhazelton:
Includes options for selecting antennas (by name or by number), frequencies, times, polarizations and even specific baseline-times (the latter probably won't be widely used but is needed for some kinds of file matching desired in the future, see issue #111).

The function is fully covered by tests, has a doc string, and I've added some tutorial examples on selecting, although the examples don't cover all the possibilities.

There are warnings (and associated errors in the appropriate write functions) for selections of frequencies or polarizations that will break miriad or uvfits file formats. The errors in the write functions are also covered by the tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ For simple installation, the latest stable version is available via pip using ``
 
 ### Optionally install the development version
 For the development version, clone the repository using
-```git clone https://github.com/HERA-Team/pyuvdata/releases/latest```
+```git clone https://github.com/HERA-Team/pyuvdata.git```
 
 Navigate into the directory and run ```python setup.py install```.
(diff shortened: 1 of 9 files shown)


adampbeardsley:
I just made a few commits to round out the select and tests a little bit.
My only remaining concern is whether we want to use random functions in the tests -- they're pretty simple here, but I worry it may make reproducibility difficult.

bhazelton:
Those improvements look good, Adam, thank you!

bhazelton:
I started with random tests partly to make sure I didn't miss anything, but I agree with your concern about reproducibility, Adam. So I converted the tests not to use a random choice. I also cleaned up the multiple file reads in the tests by copying the objects.

PR #141, opened 2017-03-31
An alternative approach to getting the version programmatically

bhazelton:
@adampbeardsley see if this approach works for you.


bhazelton:
I think this approach has an added benefit over the other approach in that the git description & hash are calculated when the code is run, rather than being calculated the last time setup.py was called.

adampbeardsley:
Generally, I like this solution Bryna. It's more robust and cleaner than mine. However, I have two concerns:
1. I cannot import pyuvdata if my working directory is not pyuvdata. I think this is because of the DATA_PATH import, so I should be able to fix this. I suspect this will also be a problem for the git calls.
2. You mention that the git info is determined when the code is run, rather than when setup.py is called. I'm not sure this is the behavior we want (nor do we want my solution). We want the git info to reflect the code that is actually run. This gets a bit tricky - if I'm in the pyuvdata directory, I think whatever the code is there currently will run, but if I'm outside pyuvdata, whatever was installed will run. I don't have a good answer for how to get this.

Issue #133, opened 2017-03-01
Store pyuvdata version in any file written by pyuvdata


adampbeardsley:
We have seen in other code bases (e.g. FHD, aipy) that it can be useful to store the git hash of the installed software in output files. This is helpful for troubleshooting or just keeping track of what was done to the data.
In order to do this, we will need to store the git hash somewhere at installation. 
Also, we should have a __version__ attribute.
aipy handles all this in its setup.py in case we wanted to borrow some code.

bhazelton:
We do have a version attribute in setup.py. But I agree about the git hash being added to all files. Zaki has included a couple of variables for that kind of information in the cal branch which I think we should flow back into the visibility files. We should take a careful look at what is in cal and make sure that it will meet our needs for the visibility files so that they can be consistent with each other.

adampbeardsley:
I'm starting to look at this in detail, and just wanted to jot some notes down so it's not lost. From what I understand, passing `version` to the setup in setup.py tags the installation so things like pip can check what version it is, but it does not set the `pyuvdata.__version__` in the module. This has to be done in the __init__.py file.

I'm looking at borrowing from astropy to accomplish this and recording the git hash. The basic idea is to have setup.py create a file which has the version and githash, then import those things in __init__. Astropy does it in a pretty sophisticated way, but I think at least for a first go, we can distill it to something pretty easy.

bhazelton:
done in latest PR

PR #144, opened 2017-04-05
Fix version

adampbeardsley:
This branch introduces a version module for handling the version and git information. It should work whether the user is running from a git repo, from a setup.py installed version, or from pip installed.


bhazelton:
This is addressing issue #133

bhazelton:
This branch needs to be rebased before it gets merged. But if the cal branch gets merged first the rebase will have to be redone, so we should wait on the rebase until this PR is signed off on.

PR #146, opened 2017-04-21
Time profiling pull request

aelanman:
Improvements in speed for read/write operations.

### DIFF
--- a/pyuvdata/miriad.py
+++ b/pyuvdata/miriad.py
@@ -26,7 +26,6 @@ class Miriad(UVData):
                   "polarization_array".format(pol=pol))
         return pol_ind
 
-#    @profile
     def read_miriad(self, filepath, correct_lat_lon=True, run_check=True, run_check_acceptability=True):
         """
         Read in data from a miriad file.
@@ -160,6 +159,10 @@ class Miriad(UVData):
                                                 ra, dec]]
                 pol_list.append(uv['pol'])
                 # NB: flag types in miriad are usually ints
+
+        for pol,data in data_accumulator.iteritems():
+            data_accumulator[pol] = np.array(data)
+
         self.polarization_array = np.array(pol_list)
         if len(self.polarization_array) != self.Npols:
             warnings.warn('npols={npols} but found {l} pols in data file'.format(
@@ -195,8 +198,8 @@ class Miriad(UVData):
 #        def resplit(a):
 #            a = a.split('_')
 #            return [float(a[0]), int(a[1]), int(a[2])]
-#
-#        unique_blts = map(resplit, unique_blts)
+
+#        unique_blts = np.array(map(resplit, unique_blts))
 
         self.Nants_data = len(sorted_unique_ants)
 
@@ -290,8 +293,8 @@ class Miriad(UVData):
                 for ant_j in ant_j_unique:
                     if ant_i > ant_j:
                         continue
-               #     if "_".join(map(str,[t, ant_i, ant_j])) not in unique_blts:
-               #         continue
+                    if "_".join(map(str,[t, ant_i, ant_j])) not in unique_blts:
+                        continue
                     t_grid.append(t)
                     ant_i_grid.append(ant_i)
                     ant_j_grid.append(ant_j)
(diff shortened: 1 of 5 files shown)


pyuvdata/miriad.py
```diff
@@ -410,8 +425,8 @@ def read_miriad(self, filepath, correct_lat_lon=True, run_check=True, run_check_
             warnings.warn(str(ve))
 
         # check if object has all required uv_properties set
-        if run_check:
-            self.check(run_check_acceptability=run_check_acceptability)
+#        if run_check:
+#            self.check(run_check_acceptability=run_check_acceptability)
```
bhazelton: Are you sure you want the check lines to be commented out? That seems like a bad idea.

pyuvdata/uvdata.py
```diff
@@ -616,14 +614,15 @@ def phase(self, ra, dec, epoch):
...
+        c_s = const.c.to('m/s').value
         for ind, jd in enumerate(self.time_array):
             # calculate ra/dec of phase center in current epoch
             obs.date, obs.epoch = self.juldate2ephem(jd), self.juldate2ephem(jd)
             precess_pos.compute(obs)
             ra, dec = precess_pos.a_ra, precess_pos.a_dec
 
             # generate rotation matrices
-            m0 = a.coord.top2eq_m(self.lst_array[ind] - obs.sidereal_time(), latitude)
+            m0 = a.coord.top2eq_m(self.lst_array[ind] - obs.sidereal_time(), latitude)    ##TODO This is a bottleneck because top2eq_m involves a matrix inversion.
```
bhazelton: Do you have a plan to address this? I think marking it as a bottleneck is usefully informative, but the TODO label gets grabbed by some editors and should only be used if there's a plan to fix it.

pyuvdata/miriad.py
aelanman: Whoops.... Yes, I cut those from mine because it was running slow and didn't want to address it, but I'll put them back and run the profiler again.

I don't think it'll break, since write_miriad also runs the acceptability check.

pyuvdata/uvdata.py
aelanman: What if I just remove the TODO label? I haven't found a way to address it yet, but maybe someone can? I have another line marked as a bottleneck in read_miriad().

bhazelton: I think that's perfect.

adampbeardsley: I don't have time to do it now, but one way to at least reduce the bottleneck, would be to only calculate m0 for unique times, rather than for all blts.

bhazelton: The other thought I have on this is that we should consider using astropy rather than aipy (which uses pyephem under the hood I think).

adampbeardsley: Aipy doesn't use pyephem for this particular function. In fact, it's a rather simple function which just just inverts eq2top_m, which simply records the matrix using sin and cos. I'd be tempted to write a small util function in pyuvdata to write down the inverse directly, which is equally easy.

adampbeardsley:
Note to self: before merging this branch, we need to check the phasing outputs against master - be sure we haven't changed this because we don't have a valid test yet.

bhazelton:
I think that when this PR is merged, we should put out a new pip package, which requires bumping the version number. Any opinions on whether it should be v1.2 or v1.1.1?

adampbeardsley:
Because this branch is getting close to merging, I went ahead and did a comparison of phasing with master. It's hard to make an actual unit test... so I just used master to read in a couple files, did various phasing and unphasing operations, writing output files throughout. I then used this branch to repeat those, and compare with the save files throughout.

What I learned is that there are a couple places where vectorizing the arithmetic caused intermediate quantities to be cast as float32 instead of float64 (when using uvws, which are float32). This caused small errors in uvws, which were small enough that they passed our tolerance, but translated to unacceptable errors in the data. I was able to make a couple changes locally to make sure everything is cast as float64, and everything works.

This is making me think that we actually want the uvws always stored as float64. I'm going to start a new issue for this, but I don't think it's stopping for the time being. I'll push my local changes, just to match master, but they should go away if we do eventually use float64.

adampbeardsley (approved):

Issue #142, opened 2017-04-05
add select function to uvcal object


bhazelton:

bhazelton:
done in recent PR

PR #150, opened 2017-04-27
Add select function for UVCal

bhazelton:
Essentially mirrors select function for UVData.

This addresses issue #142


bhazelton:
I'd like your opinion on lines 51 & 61 of calfits.py @zakiali

bhazelton:
@adampbeardsley I'm baffled by the build outputs on my latest commit. The PR build passed but the push build failed with a bizarre error I've never seen before. I tried rerunning the build and got the same error. Can you take a look when you get a chance?

adampbeardsley:
We'll find out in a few minutes if that last commit does the trick. But it's definitely something we should have in there. version was not picking up non-annotated tags, so for example pyuvdata.__version__ was getting set to 1.0 even when we were on 1.1.

adampbeardsley:
I'm thinking more about this, and I think my solution is only a band-aid. The real problem is that for this test we do a shallow clone (--depth=50). But this does not always grab a tag. For example, if I do a clone on master with --depth=1, I get the error again. The only reason my --tags fix worked (temporarily) is that including v1.1 brought the latest tag within the clone depth. But if we go a long time without a tag, the problem will reemerge.
I see two solutions, the first being preferable, but I don't know how to do it:
- Put in the test code to somehow fetch tag information, regardless of the depth. 
- Or start nightly/weekly tags to make sure this doesn't come up.
Any other suggestions?

bhazelton:
Thanks for figuring that out, @adampbeardsley! I didn't realize that the version 1.1 tag wasn't annotated, I didn't think I did anything differently with that version than I did with the prior one. I'll do some googling to see if I can figure out how to have the test code fetch more tag information.

docs/tutorial.rst
```diff
@@ -204,41 +206,82 @@ b) Writing a gain calibration file.
       dd = []
       fl = []
       ch = []
-      for ant in antenna_numbers:
+      for ant in ant_array:
           dd.append(gains[pol][ant])
```
adampbeardsley: Are all these steps necessary for a tutorial? The gains, flags, and chisq's are created in a loop, repackaged in a loop, then transposed in a non-transparent way. Because this is just a demonstration, wouldn't it be easier to just create the arrays as random numbers with the appropriate shapes from the beginning?

pyuvdata/calfits.py
```diff
@@ -96,7 +116,7 @@ def write_calfits(self, filename, spoof_nonessential=False,
             prihdr['CTYPE4'] = ('FREQS', 'Frequency.')
             prihdr['CUNIT4'] = ('Hz', 'Units of frequecy.')
             prihdr['CRVAL4'] = self.freq_array[0][0]
-            prihdr['CDELT4'] = self.channel_width
+            prihdr['CDELT4'] = freq_spacing[0]
```
adampbeardsley: If Nfreqs == 1, this will be a scalar and we'll get a no `__getitem__` error.

pyuvdata/calfits.py
```diff
@@ -40,6 +40,26 @@ def write_calfits(self, filename, spoof_nonessential=False,
...
 
+        if self.Nfreqs > 1:
+            freq_spacing = self.freq_array[0, 1:] - self.freq_array[0, :-1]
+            if not np.isclose(np.min(freq_spacing), np.max(freq_spacing),
+                              rtol=self._freq_array.tols[0], atol=self._freq_array.tols[1]):
+                raise ValueError('The frequencies are not evenly spaced (probably '
+                                 'because of a select operation). The calfits format '
+                                 'does not support unevenly spaced frequencies.')
+        else:
+            freq_spacing = self.channel_width  # or should this be the length of the frequency range?
```
adampbeardsley: I'm not sure what is meant by "length of the frequency range," but we should sort it out before merging?

pyuvdata/calfits.py
```diff
@@ -167,19 +187,19 @@ def write_calfits(self, filename, spoof_nonessential=False,
             sechdr['CTYPE3'] = ('TIME', 'Time axis.')
             sechdr['CUNIT3'] = ('JD', 'Time in julian date format')
             sechdr['CRVAL3'] = self.time_array[0]
-            sechdr['CDELT3'] = self.integration_time
+            sechdr['CDELT3'] = time_spacing[0]
```
adampbeardsley: Same problem as above if Ntimes==1

pyuvdata/calfits.py
```diff
@@ -167,19 +187,19 @@ def write_calfits(self, filename, spoof_nonessential=False,
...
 
             sechdr['CTYPE4'] = ('FREQS', 'Valid frequencies to apply delay.')
             sechdr['CUNIT4'] = ('Hz', 'Units of frequecy.')
             sechdr['CRVAL4'] = self.freq_array[0][0]
-            sechdr['CDELT4'] = self.channel_width
+            sechdr['CDELT4'] = freq_spacing[0]
```
adampbeardsley: here too

pyuvdata/calfits.py
```diff
@@ -211,7 +231,7 @@ def write_calfits(self, filename, spoof_nonessential=False,
         prihdr['CTYPE3'] = ('TIME', 'Time axis.')
         prihdr['CUNIT3'] = ('JD', 'Time in julian date format')
         prihdr['CRVAL3'] = self.time_array[0]
-        prihdr['CDELT3'] = self.integration_time
+        prihdr['CDELT3'] = time_spacing[0]
```
adampbeardsley: here too

pyuvdata/uvcal.py
```diff
@@ -231,6 +233,202 @@ def set_unknown_cal_type(self):
...
+            for jd in times:
+                if jd in self.time_array:
+                    time_inds = np.append(time_inds, np.where(self.time_array == jd)[0])
+                else:
+                    raise ValueError('Time {t} is not present in the time_array'.format(t=jd))
+
+            time_inds = list(sorted(set(list(time_inds))))
+            self.Ntimes = len(time_inds)
+            self.time_array = self.time_array[time_inds]
+            # should we adjust the time_range parameter?
```
adampbeardsley: I would say no. If the user intends to declare a different range of validity, I think that's separate from just selecting solutions for certain times.

pyuvdata/uvcal.py
```diff
@@ -231,6 +233,202 @@ def set_unknown_cal_type(self):
...
+
+            time_separation = self.time_array[1:] - self.time_array[:-1]
+            if np.min(time_separation) < np.max(time_separation):
+                warnings.warn('Selected times are not evenly spaced. This '
+                              'will make it impossible to write this data out to '
+                              'some file types')
```
adampbeardsley: Maybe change this to say it's not supported by calfits (the only data format we can write)

pyuvdata/uvcal.py
```diff
@@ -231,6 +233,202 @@ def set_unknown_cal_type(self):
...
+            for f in frequencies:
+                if f in freq_arr_use:
+                    freq_inds = np.append(freq_inds, np.where(freq_arr_use == f)[0])
+                else:
+                    raise ValueError('Frequency {f} is not present in the freq_array'.format(f=f))
+
+            freq_inds = list(sorted(set(list(freq_inds))))
+            self.Nfreqs = len(freq_inds)
+            self.freq_array = self.freq_array[:, freq_inds]
+            # should we adjust the freq_range parameter?
```
adampbeardsley: Err, now that I think about it, I'm not sure I understand the definitions of time_range and freq_range. One could imagine having 1024 channels over 100MHz-200MHz, and saying the freq_range is exactly that. In that case, can the range be anything other than (1e8, 2e8)? I suppose it could be (1.5e8, 2e8) and the first half is known to be garbage.
You could also imagine a single solution which is good for a wider range than a single channel bandwidth.
Now I've confused myself. We should probably resolve this soon though.

pyuvdata/uvcal.py
```diff
@@ -231,6 +233,202 @@ def set_unknown_cal_type(self):
...
+
+            freq_separation = self.freq_array[0, 1:] - self.freq_array[0, :-1]
+            if np.min(freq_separation) < np.max(freq_separation):
+                warnings.warn('Selected frequencies are not evenly spaced. This '
```
adampbeardsley: Should this be changed to use the tolerance bits?

docs/tutorial.rst
bhazelton: I think that would be fine. I didn't actually write this, I just tweaked what @zakiali had, but if he's on board we can significantly simplify it.

pyuvdata/calfits.py
bhazelton: frequency_range is a parameter that gives the range of valid frequencies for this cal solution (this is particularly useful for delay type cal solutions where there isn't a frequency axis but you need to know what range of frequencies the solution applies to). The question here is if there's only one frequency channel after the select, should the delta between the channels (something that is explicitly defined in fits arrays) be equal to the channel width or the width of the frequency range? Since frequency range is already reported in another header item, I'd lean toward the channel_width, but I wanted to flag it for discussion.

pyuvdata/calfits.py
bhazelton: Thanks, I'll fix this.

pyuvdata/calfits.py
bhazelton: Thanks, I'll fix this.

pyuvdata/uvcal.py
bhazelton: Ok. I agree, I just wanted to make sure we were on the same page.

pyuvdata/uvcal.py
bhazelton: Will do.

pyuvdata/uvcal.py
bhazelton: Probably, and if so it should be changed in uvfits.py as well. The question then is what value should be used for the channel separation? The mean? a value that is rounded to the tolerance? This has come up in another discussion with @aelanman on the uvfits side and I'm happy to implement it but I'd like to know what others think about the value to use.

docs/tutorial.rst
zakiali: I'm on board with creating random arrays with the correct shapes from the get go.

pyuvdata/calfits.py
```diff
@@ -40,6 +40,28 @@ def write_calfits(self, filename, spoof_nonessential=False,
...
+        if self.Nfreqs > 1:
+            freq_spacing = self.freq_array[0, 1:] - self.freq_array[0, :-1]
+            if not np.isclose(np.min(freq_spacing), np.max(freq_spacing),
+                              rtol=self._freq_array.tols[0], atol=self._freq_array.tols[1]):
+                raise ValueError('The frequencies are not evenly spaced (probably '
+                                 'because of a select operation). The calfits format '
+                                 'does not support unevenly spaced frequencies.')
+            freq_spacing = freq_spacing[0]
+        else:
+            freq_spacing = self.channel_width  # or should this be the length of the frequency range?
```
zakiali: I think this should stay as is. The frequency range is the range of frequencies where the solutions apply. Ditto with the time_spacing on line 63.

bhazelton: sounds good, I'll take out the comments.

pyuvdata/uvcal.py
bhazelton: I think I should do the same thing we decided for the time_range -- leave it alone and let the user adjust it if desired.

docs/tutorial.rst
bhazelton: ok. I'll see what I can do to make it simpler.

pyuvdata/uvcal.py
zakiali: Maybe keep this as is for now? It's not clear to me what the channel separation should be if we used the tolerance bits.

bhazelton: I actually think we need to fix it because the same logic is causing bugs in uvfits. I just would like to come to a consensus on what it should be.

pyuvdata/uvcal.py
zakiali: So that's how I was thinking about it. In the single solution case, you could imagine having a solution at 1 frequency that is applicable over the whole band. In that case, the solution would not be frequency dependent. However, if there were two solutions, then this gets a bit vague. 
I actually think that this should primarily be used in the delay case where we have one number and you can use it for any frequency that is within the frequency range. And have it either be optional or done away with in the gain case because that gets more confusing. 
However, time_range should be used in either case. (time independent solutions etc...). 
Hopefully all that made sense.

pyuvdata/uvcal.py
zakiali: I would go for rounding to the tolerance in that case.

pyuvdata/uvcal.py
bhazelton: I see. So you're advocating making the freq_range parameter only required if the cal_type is delay. I think that makes good sense.What do you think @adampbeardsley?

If we change this we'll need to add some tests in various places to check that the code behaves if it is not set for gain type solutions.

adampbeardsley: Ok, I think I get the general idea. I'm not sure how difficult it would be to make it a delay-only parameter. It actually sounds quite difficult with how we have the object set up. I'm not sure it's harming anything if it's there for gain type, so I'll leave that up to you two.

bhazelton:
Thanks @zakiali!

I think I've addressed all your comments, @adampbeardsley

pyuvdata/uvcal.py
bhazelton: I think it's pretty easy, I'll try it and see how it goes. I'm most worried about getting the testing right.

bhazelton: We already have cal_type dependent parameters on the object, just like we have parameters on UVData that are required only for certain phase_types.

Issue #112, opened 2016-12-14
All cal types must have methods to convert to gain_array


bhazelton:
so that we can exchange gains more easily.

bhazelton:
done in latest PR

PR #153, opened 2017-05-03
Add method convert_gains to UVCal with tests

bhazelton:
Added a simple method to UVCal to convert delay cals to gain type cals as requested in issue #112.

### DIFF
--- a/pyuvdata/uvcal.py
+++ b/pyuvdata/uvcal.py
@@ -432,6 +432,28 @@ class UVCal(UVBase):
         if run_check:
             self.check(run_check_acceptability=run_check_acceptability)
 
+    def convert_to_gain(self):
+        if self.cal_type == 'gain':
+            raise ValueError('The data is already a gain cal_type.')
+        elif self.cal_type == 'delay':
+            phase_array = np.zeros((self.Nants_data, self.Nspws, self.Nfreqs, self.Ntimes, self.Njones))
+            for si in range(self.Nspws):
+                temp = -2 * np.pi * np.dot(self.delay_array[:, si, :, :, np.newaxis],
+                                           self.freq_array[si, np.newaxis, :])
+                temp = np.transpose(temp, (0, 3, 1, 2))
+                phase_array[:, si, :, :, :] = temp
+
+            amplitude_array = np.ones_like(phase_array)
+            gain_array = amplitude_array * np.exp(1j * phase_array)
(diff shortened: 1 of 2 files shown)


pyuvdata/uvcal.py
```diff
@@ -432,6 +432,28 @@ def select(self, antenna_nums=None, antenna_names=None,
         if run_check:
             self.check(run_check_acceptability=run_check_acceptability)
 
+    def convert_to_gain(self):
+        if self.cal_type == 'gain':
+            raise ValueError('The data is already a gain cal_type.')
+        elif self.cal_type == 'delay':
+            phase_array = np.zeros((self.Nants_data, self.Nspws, self.Nfreqs, self.Ntimes, self.Njones))
+            for si in range(self.Nspws):
+                temp = -2 * np.pi * np.dot(self.delay_array[:, si, :, :, np.newaxis],
```
adampbeardsley: I'm not sure how universal the sign convention in this exponent is. In other words, does a positive delay imply a certain direction of phase ramp, or is there ambiguity?
Should there be a delay_convention parameter to say whether the time*freq part has a plus or minus? Or an argument to this function which defaults to minus?

zakiali: That's a good point. I'm also not sure how universal it is. I like the second suggestion of an argument that defaults to minus.

bhazelton: That's a very good point. I did this based off of what firstcal does, since I think it's the only thing using the delay type, but I can add a delay convention parameter to the method. Or were you thinking it should be on the object?

adampbeardsley: I first I was thinking something on the object, but it's probably more natural to just be an option to this function.

pyuvdata/uvcal.py
```diff
@@ -432,6 +432,28 @@ def select(self, antenna_nums=None, antenna_names=None,
...
+                                           self.freq_array[si, np.newaxis, :])
+                temp = np.transpose(temp, (0, 3, 1, 2))
+                phase_array[:, si, :, :, :] = temp
+
+            amplitude_array = np.ones_like(phase_array)
```
zakiali: Isn't this redundant?

bhazelton: It probably is. I guess I was trying to be explicit, but I can take that out and note it in a comment.

PR #155, opened 2017-05-17
Add utility functions for calculating local ENU coordinates

bhazelton:
Fixed docstrings for XYZ_from_LatLonAlt and LatLonAlt_from_XYZ which incorrectly stated they were going to/from topocentric coordinates rather than ecef coordinates

Also adapted XYZ_from_LatLonAlt and LatLonAlt_from_XYZ to accept arrays of positions rather than just single positions.

Tests fully cover all new code.


bhazelton:
I think I need to better understand what 'topo' and 'eq' are in the eq2top_m & top2eq_m before I could do that. I'm not perfectly sure if topo corresponds to ENU -- after some googling I couldn't find a consistent definition of what topocentric means precisely, but ENU seemed like the thing we wanted for antenna positions.

bhazelton:
No it does not currently support ECEF rotated so that X intersects the longitude of the array. If you can point me at the correct math to do that, I can add it as a function. I'm uncomfortable trying to write it without a reference or a test.

aelanman:
I think I already have something addressing the aipy-style XYZ coordinates, even if it hasn't been included in the utils already.
Lines 140 through 143 of test_utils.py apply a rotation from local ECEF (ecef-lon) to true ECEF coordinates. I added this because the STABXYZ table in the cotter file headers was in this local XYZ coordinates system.

bhazelton:
I think that your rotation is different than what Danny's talking about, but I'm not sure. I agree that what the uvfits memo describes is something that sounds a bit like what Danny wants, but the cotter code that does the rotation says it puts the z axis through North, whereas I think Danny's coordinate system would have z up. Is that right @dannyjacobs?

PR #156, opened 2017-05-18
Better testing

bhazelton:
Several changes to improve testing and significant coverage improvements.


pyuvdata/miriad.py
```diff
@@ -396,18 +398,50 @@ def read_miriad(self, filepath, correct_lat_lon=True, run_check=True, run_check_
...
+                blt_good = np.where(~np.all(self.flag_array, axis=(1, 2, 3)))
+                if np.isclose(np.mean(np.diff(ra_list[blt_good])), 0.):
+                    self.set_phased()
+                else:
+                    self.set_drift()
+            else:
+                # if there's only one time, checking for consistent RAs doesn't work.
+                # instead check for the presence of an epoch variable, which isn't
+                # really a good option, but at least it prevents crashes.
+                if 'epoch' in uv.vartable.keys():
```
adampbeardsley: Technically if there is only one time, tracking and drift are indistinguishable. However, pyuvdata does assume that "drift" implies drifting while phased at zenith. So perhaps a better check would be to see if the phase center is set to zenith? Otherwise we could be messing up files.

pyuvdata/tests/test_calfits.py
```diff
@@ -52,13 +41,9 @@ def test_readwriteread_delays():
     uv_out = UVCal()
     testfile = os.path.join(DATA_PATH, 'zen.2457698.40355.xx.HH.uvc.fits')
     write_file = os.path.join(DATA_PATH, 'test/outtest_firstcal.fits')
-    read_status = uvtest.checkWarnings(uv_in.read_calfits, [testfile],
-                                       nwarnings=0)
+    uvtest.checkWarnings(uv_in.read_calfits, [testfile], nwarnings=0)
```
adampbeardsley: I thought I had changed this when reviewing calfits, but I guess not.
Do we want to use checkWarnings when nwarnings=0? I suppose it doesn't hurt to make sure we're not getting any, but in other places we only use it when we expect a warning.

pyuvdata/tests/test_miriad.py
```diff
@@ -19,34 +19,163 @@ def test_ReadMiriadWriteUVFits():
...
+    miriad_file = os.path.join(DATA_PATH, 'zen.2456865.60537.xy.uvcRREAA')
+    uvtest.checkWarnings(miriad_uv.read_miriad, [miriad_file], known_warning='miriad')
+    pol_arr = miriad_uv.polarization_array
+
+    miriad = miriad_uv._convert_to_filetype('miriad')
+    miriad.polarization_array = None
+    nt.assert_raises(ValueError, miriad._pol_to_ind, pol_arr[0])
+
+    miriad.polarization_array = [pol_arr[0], pol_arr[0]]
+    nt.assert_raises(ValueError, miriad._pol_to_ind, pol_arr[0])
```
adampbeardsley: Should we also have a quick test for a working case?

pyuvdata/uvfits.py
```diff
@@ -158,15 +158,8 @@ def read_uvfits(self, filename, run_check=True, run_check_acceptability=True):
 
         # get shapes
         self.Nfreqs = hdr.pop('NAXIS4')
-        if self.data_array.shape[2] != self.Nfreqs:
```
adampbeardsley: Why are we getting rid of these warnings?

pyuvdata/miriad.py
bhazelton: That's a good point. The "crashing" I referred to in the comment is that the epoch keyword isn't necessarily present for drift files (although it apparently can be). For phased files it has to be present. Do we have some existing code that tells us what zenith is so that I can compare it to the  RA/Decs? (the term phase center generally means we're talking about phased data)

pyuvdata/tests/test_calfits.py
bhazelton: I thought about removing this entirely, my thinking was similar to yours. I'm happy to do whatever you think is best.

pyuvdata/tests/test_miriad.py
bhazelton: I think that's covered by the tests where actual files are read and processed, I added these tests to cover cases that weren't already covered. But I can add another case if you think I should. Or I can add a docstring or change the name of the test to make it clear I'm testing error handling.

pyuvdata/uvfits.py
bhazelton: Because it's impossible to get them. It turns out that you can't write a fits file that would be broken in these ways.

pyuvdata/miriad.py
adampbeardsley: The more I think about this, the more complicated the various cases can be. One could have an observation which is drifting, but phased off zenith (e.g. low altitude drift scan), and you could store the ra/dec values in non-current epoch. So I think your solution is probably the simplest without really expanding for a lot of crazy fringe cases that probably don't exist.

We really just need to make sure we don't call something drift when it's not phased to zenith. The only case where your solution would break that is if it's scanning off-zenith, and doesn't define the epoch. But then we don't know where's it's looking anyway, so we can't call it phased.

pyuvdata/tests/test_miriad.py
adampbeardsley: If it's covered, it's all good.

pyuvdata/uvfits.py
adampbeardsley: Gotcha. I kind of wondered about that.

pyuvdata/tests/test_calfits.py
adampbeardsley: Ok, I just took out the zero warning checks in the interest of consistency.

PR #158, opened 2017-05-31
Add array for recording array-wide quality values

plaplant:
Adds a new array for tracking array-wide quality values, in addition to per-antenna values. Issuing a pull request to build tests for added code.

### DIFF
--- a/pyuvdata/calfits.py
+++ b/pyuvdata/calfits.py
@@ -368,10 +368,6 @@ class CALFITS(UVCal):
             self.git_hash_cal = hdr['HASHCAL']
         except:
             pass
-        try:
-            self.calfile = hdr['CALFILE']
-        except:
-            pass
 
         # get data.
         if self.cal_type == 'gain':
--- a/pyuvdata/calfits.py
+++ b/pyuvdata/calfits.py
@@ -122,6 +122,8 @@ class CALFITS(UVCal):
             prihdr['ORIGCAL'] = self.git_origin_cal
         if self.git_hash_cal:
             prihdr['HASHCAL'] = self.git_hash_cal
+        if self.calfile:
+            prihdr['CALFILE'] = self.calfile
 
         if self.cal_type == 'unknown':
             raise ValueError("unknown calibration type. Do not know how to"
@@ -350,6 +352,10 @@ class CALFITS(UVCal):
             pass
         try:
             self.git_hash_cal = hdr['HASHCAL']
+        except:
+            pass
+        try:
+            self.calfile = hdr['CALFILE']
         except:
             pass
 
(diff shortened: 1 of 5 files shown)


pyuvdata/uvcal.py
```diff
@@ -200,6 +200,23 @@ def __init__(self):
                                          expected_type=str,
                                          required=False)
 
+        desc = ('Calfile version used in this file.')
+        self._calfile = uvp.UVParameter('calfile', form='str',
```
bhazelton: I'm not sure we want to record the calfile this way, mostly because the concept of cal files is very specific to PAPER & current HERA data and it's going to go away in future HERA processing. pyuvdata is meant to be a general tool for the community, so I'm uncomfortable with having a parameter this specific and transient on the object. I suggest adding a comment to the history parameter that records the cal file. The history parameter is traditionally where code-specific parameters are recorded.

pyuvdata/calfits.py
```diff
@@ -273,8 +285,13 @@ def write_calfits(self, filename, spoof_nonessential=False,
...
         else:
-            prihdu = fits.PrimaryHDU(data=pridata, header=prihdr)
-            hdulist = fits.HDUList([prihdu, ant_hdu])
+            if self.total_quality_array is not None:
+                prihdu = fits.PrimaryHDU(data=pridata, header=prihdr)
+                sechdu = fits.ImageHDU(data=secdata, header=sechdr)
+                hdulist = fits.HDUList([prihdu, ant_hdu, sechdu])
+            else:
+                prihdu = fits.PrimaryHDU(data=pridata, header=prihdr)
+                hdulist = fits.HDUList([prihdu, ant_hdu])
```
bhazelton: It looks to me like you are only handling the total_quality_array if the cal_type is 'gain' and ignoring it if the cal_type is 'delay'. Is this the behavior we want? I think the concept of an overall gain quality should apply equally well for delay type solutions.

bhazelton:
Since you've added new parameters to uvcal, we also need to remake the docs. I can do that once all the code changes are done, or I can walk you through it if you'd like.

pyuvdata/uvcal.py
plaplant: Okay, I've taken out the calfile parameter

pyuvdata/calfits.py
plaplant: I hadn't done it since there already was a secondary HDU in the delay type. For backwards compatibility, I've added the total_quality_array as a tertiary HDU for delay cals, rather than make it the secondary for consistency between types. Is this solution acceptable?

bhazelton: Is it possible to put it in the secondary HDU when we have a delay type (i.e. do the array shapes work for that)? If so, I'd put it there. Otherwise having it in it's own HDU is probably the only option.

plaplant: Unfortunately it's a different number of axes. Seems like it has to be its own HDU.

bhazelton:
This looks good to me now. I'm going to merge it.

PR #161, opened 2017-06-07
fix calfits to conform to FITS standard + minor tweaks

bhazelton:
I just found a problem with calfits that meant that the files didn't conform to the FITS standard. Namely, the CRPIX* values were not being defined. In addition, there was confusion about the meaning of CRVAL* -- in calfits they were being treated as representing the length of the axis (which is really in NAXIS*) rather than the value corresponding to the pixel indicated by CRPIX*.

This pull request fixes that problem (with fixes to the test files) and also cleans up some duplicated code in uvfits and calfits by moving internal modules to utils.py

It also adds the axes description keywords (C*) to the total quality array HDU, cleans up some code duplication and tests that axes that should match between different HUDs actually do match.


pyuvdata/calfits.py
```diff
@@ -127,68 +115,98 @@ def write_calfits(self, filename, run_check=True,
...
+            totqualdata = self.total_quality_array
 
         if self.cal_type == 'delay':
             # Set header variable for delay.
 
             # ANTAXIS axis number differs between delay and gain because there's no frequency axis in delay
             prihdr['CTYPE5'] = ('ANTAXIS', 'See ANTARR in ANTENNA extension for values.')
             prihdr['CUNIT5'] = 'Integer'
-            prihdr['CRVAL5'] = 0
+            prihdr['CRVAL5'] = 1
```
adampbeardsley: Shouldn't this differ between gain and delay? It looks like they're both 1 (previously they were both zero). 
Actually, looking at the structure of the data, it should be 1 for both. But if that's the case, why have this parameter?

pyuvdata/calfits.py
```diff
@@ -257,6 +275,7 @@ def write_calfits(self, filename, run_check=True,
 
         prihdr['CTYPE3'] = ('TIME', 'Time axis.')
         prihdr['CUNIT3'] = ('JD', 'Time in julian date format')
+        prihdr['CRPIX3'] = 1
         prihdr['CRVAL3'] = self.time_array[0]
```
adampbeardsley: Here and elsewhere, is it a requirement of calfits (and uvcal) that the time axis is in order?

pyuvdata/calfits.py
bhazelton: CRVAL is the value of the axis at the pixel given by CRPIX. I think the comment about axis number is saying that for delay it's axis 5 ('CTYPE5') but for gain it's axis 6 ('CTYPE6'). The reason I changed it to 1 is because FITS files are always 1 indexed. But I should have checked that the 1 is subtracted off when we read that array. I'll check and fix that now if it's wrong.

pyuvdata/calfits.py
bhazelton: Oh, yes it should be since it is it's own axis. I don't think that's checked, but I can add a check. Good catch!

pyuvdata/calfits.py
bhazelton: Checking that led me to some other minor fixes, it turns out that calfits wasn't using the axis length parameters at all and instead inserting the length parameters as their own keywords. I fixed that as well.

pyuvdata/calfits.py
bhazelton: It looks like there is a check on the time axis and it would fail if the times weren't in order because it checks that the np.diff of the time array is constant.

plaplant:
These changes look good to me. I hadn't yet created any calfits files with the old format, so there's no backwards compatibility issues.

bhazelton:
Ok, thanks Paul & Adam. I'm going to merge this now in the interest of getting these fixes in as soon as possible.

PR #162, opened 2017-06-09
Calfits backwards compatibility fixes

bhazelton:
This PR adds backwards compatibility for old calfits files, with warnings issued when problems are discovered and with a keyword option (strict_fits) on read_calfits to allow for overriding the compatibility code.

### DIFF
--- a/pyuvdata/uvcal.py
+++ b/pyuvdata/uvcal.py
@@ -511,7 +511,8 @@ class UVCal(UVBase):
             setattr(other_obj, p, param)
         return other_obj
 
-    def read_calfits(self, filename, run_check=True, run_check_acceptability=True):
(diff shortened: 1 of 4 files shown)


bhazelton:
Unfortunately this adds a bunch of untested code,  tests need to be added. I can tackle that next week or someone else can jump in earlier if they'd like.

pyuvdata/uvcal.py
```diff
@@ -511,7 +511,8 @@ def _convert_to_filetype(self, filetype):
             setattr(other_obj, p, param)
         return other_obj
 
-    def read_calfits(self, filename, run_check=True, run_check_acceptability=True):
+    def read_calfits(self, filename, run_check=True, run_check_acceptability=True,
+                     strict_fits=False):
```
adampbeardsley: let's add strict_fits to the args list below

PR #163, opened 2017-06-12
Make change to total_quality_array to reflect actual dimensions from …

plaplant:
…omnical


plaplant:
Omnical returns an overall chisq value per time and per frequency, but not per polarization. The total_quality_array is changed to reflect this. The self-consistency checks are also modified for delay-style calibrations, which will not have a frequency array.

plaplant:
After offline discussion, it was pointed out that this behavior is likely a bug in omnical. The change to the self-consistency check has been kept, since it seems to prevent otherwise valid files from being read in.

PR #165, opened 2017-06-15
Update total quality array self-consistency check

plaplant:
This is a bug-fix to correctly read in delay-type calfits files, since the freq_array only has one entry instead of an array.


bhazelton:
It looks like there's some kind of conflict with merging this branch into master.

PR #170, opened 2017-06-20
Update calfits object and tests so delay-type fits have same rank as …

plaplant:
…gain-type. Previously, gain-type and delay-type objects had a different rank for the solutions, as well as quality objects. Tests and parameter definitions have been changed to make them equal rank, and a warning is raised when reading an "old" delay file.


bhazelton:
This looks pretty good to me and the test coverage is great. I think the one think we need to keep an eye on is updating the calfits memo with all the recent changes.

Issue #136, opened 2017-03-08
Add support for beams


bhazelton:
define formats, conventions.

need a discussion with lots of people to talk about what things need to be accommodated

bhazelton:
The UVBeam object was incorporated in PR #171. There are several requested features that have not yet been implemented, they will be added as issues now.

PR #171, opened 2017-06-21
Add UVBeam object

bhazelton:
Add UVBeam object and support for reading & writing uniformly gridded power and e-field beams.

Addresses issue #136


bhazelton:
This branch now has a reasonably well-developed and tested read/write method for regularly gridded beams, either efield or power beams.

The testing includes reading in and writing out an exported CASA beam, with the caveat that the CASA beam is really in RA/Dec sine projection and the test treats it as if it were in az/za sine projection.

It currently lacks any read/write methods for healpix format beams.
It also currently lacks support for phased_array type antennas.

bhazelton:
This branch now supports reading & writing healpix beam fits files using a custom fits format (rather than the standard healpix fits format which does not support a frequency axis).

bhazelton:
Ok, I think this PR is ready for some external eyes on it. @adampbeardsley @dannyjacobs @jpober 

Some existing limitations that should be addressed in the future:
-- The only data file I have to test with is a CASA power beam from Chris. In order to fully exercise the code, I'm making up dummy test data. It'd be better to base tests on some actual data files. Of course this is limited by the current set of read/write methods (only FITs is supported currently).
-- There is no support yet for phased-array beams in the FITs format
-- There are no import/export methods yet for standard Healpix FITs files (as Danny requested), but the select and add methods they would require do exist.
-- There are no coordinate conversion methods yet to move between different pixel_coordinate_systems (e.g. az_za and healpix).
-- There is no method yet to convert E-field beams to power beams.
-- We need read methods for native model files

I'm not sure how many of these should be addressed before this PR gets merged, I think that should be part of the discussion.

pyuvdata/uvbeam.py
```diff
@@ -0,0 +1,1023 @@
...
+                                             description=desc,
+                                             form=(4, 'Nspws', 'Nfreqs'),
+                                             expected_type=np.float,
+                                             tols=1e-3)
+
+        super(UVBeam, self).__init__()
+
+    def check(self, run_check_acceptability=True):
+        """
+        Add some extra checks on top of checks on UVBase class.
```
adampbeardsley: Are there meant to be more checks other than set_cs_params? If not, we can drop this line of the doc string.

pyuvdata/uvbeam.py
```diff
@@ -0,0 +1,1023 @@
...
+            run_check_acceptability: Option to check acceptable range of the values of
+                required parameters after combining objects. Default is True.
+            inplace: Overwrite self as we go, otherwise create a third object
+                as the sum of the two (default).
+        """
+        if inplace:
+            this = self
+        else:
+            this = copy.deepcopy(self)
+        # Check that both objects are UVData and valid
```
adampbeardsley: UVBeam, not UVData

pyuvdata/beamfits.py
```diff
@@ -0,0 +1,543 @@
...
+            self.data_array = data
+            if primary_header.pop('CTYPE' + str(ax_nums['feed_pol'])).lower().strip() == 'stokes':
+                self.Npols = primary_header.pop('NAXIS' + str(ax_nums['feed_pol']))
+            self.polarization_array = np.int32(uvutils.fits_gethduaxis(primary_hdu,
+                                                                       ax_nums['feed_pol']))
+        elif self.beam_type == 'efield':
+            self.set_efield()
+            if n_dimensions < n_efield_dims:
+                raise (ValueError, 'beam_type is efield and data dimensionality is too low')
+            complex_arrs = np.split(data, 2, axis=0)
```
bhazelton: There aren't N arrays, just two: the real and imaginary values of the data array. I do this because FITs can't store complex values in image arrays. So I'm doubling the number of arrays along a new axis and setting one to be the real and the other to be imaginary. Then I recombine them on read. This is exactly what is done in uvfits (although there they have 3 arrays: the real part, the imaginary part and the weights (which also contains the flagging encoded as the sign of the weights))

pyuvdata/uvbeam.py
```diff
@@ -0,0 +1,1023 @@
...
+                                      'More than one spectral window is not '
+                                      'currently supported.', expected_type=int)
+
+        desc = ('Number of directions in vector_coordinate_system, options '
+                'are 2 or 3 (or 1 if beam_type is "power")')
+        self._Naxes_vec = uvp.UVParameter('Naxes_vec', description=desc,
+                                          expected_type=int, acceptable_vals=[2, 3])
+
+        desc = ('Pixel coordinate system, options are: '
+                ', '.join(self.coordinate_system_dict.keys()))
```
bhazelton: thanks!

pyuvdata/uvbeam.py
```diff
@@ -0,0 +1,1023 @@
...
+                                           form=('Nspws', 'Nfreqs'),
+                                           expected_type=np.float,
+                                           tols=1e-3)  # mHz
+
+        self._spw_array = uvp.UVParameter('spw_array',
+                                          description='Array of spectral window '
+                                          'Numbers, shape (Nspws)', form=('Nspws',),
+                                          expected_type=int)
+
+        desc = 'Normalization standard of data_array, options are: "peak", "solid_angle"'
```
bhazelton: I'm embarrassed to say that I don't remember exactly what it means, it was something Dave, James and I came up with. I think it might be related to the isotropic radiator, I've pinged them to get help remembering.

pyuvdata/uvbeam.py
bhazelton: thanks!

pyuvdata/uvbeam.py
```diff
@@ -0,0 +1,1065 @@
...
+        self._spw_array = uvp.UVParameter('spw_array',
+                                          description='Array of spectral window '
+                                          'Numbers, shape (Nspws)', form=('Nspws',),
+                                          expected_type=int)
+
+        desc = ('Normalization standard of data_array, options are: '
+                '"peak" and "solid_angle". Peak normalized means each frequency '
+                'is separately normalized such that the peak is 1 and the beam '
+                'is dimensionless. Solid angle normalized means the peak '
+                'normalized beam is divided by the integral of the of the beam '
```
adampbeardsley: `of the of the` => `of the`

PR #175, opened 2017-06-23
Catch warnings when reading old data formats during tests

plaplant:
This change wraps the reading of old calfits files with a way to properly catch and check the warnings during testing. These warning messages are no longer displayed when tests are run.


bhazelton:
looks good to me!

PR #178, opened 2017-06-27
Enhance UI

adampbeardsley:
This branch aims to pull some of the user interface functionality of the "DataContainer" object from metrics into pyuvdata.
Essentially it provides some functions to quickly access data, flags, and nsamples by providing an antenna pair and/or polarization, along with a few extra goodies.

### DIFF
--- a/pyuvdata/uvdata.py
+++ b/pyuvdata/uvdata.py
@@ -1452,14 +1452,30 @@ class UVData(UVBase):
         """
         return [self.baseline_to_antnums(bl) for bl in self.get_baseline_nums()]
 
+    def get_pols(self):
+        """
+        Returns list of pols in string format.
+        """
+        return uvutils.polnum2str(self.polarization_array)
+
     def get_antpairpols(self):
         """
         Returns list of unique antpair + pol tuples (ant1, ant2, pol) in data.
         """
         bli = 0
-        pols = uvutils.polnum2str(self.polarization_array)
+        pols = self.get_pols()
         bls = self.get_antpairs()
         return [(bl) + (pol,) for bl in bls for pol in pols]
+
+    def get_feedpols(self):
+        """
+        Returns list of unique antenna feed polarizations (e.g. ['X', 'Y']) in data.
+        NOTE: Will return ValueError if any stokes visibilities are present.
+        """
+        if np.any(self.polarization_array > 0):
+            raise ValueError('Stokes visibilities cannot be interpreted as feed polarizations')
+        else:
+            return list(set(''.join(self.get_pols())))
 
     def antpair2ind(self, ant1, ant2):
         """
--- a/pyuvdata/uvdata.py
+++ b/pyuvdata/uvdata.py
@@ -1433,3 +1433,97 @@ class UVData(UVBase):
         # check if object is self-consistent
         if run_check:
             self.check(run_check_acceptability=run_check_acceptability)
+
+    def get_ants(self):
+        """
+        Returns numpy array of unique antennas in the data.
+        """
+        return np.unique(np.append(self.ant_1_array, self.ant_2_array))
+
+    def get_baseline_nums(self):
+        """
+        Returns numpy array of unique baseline numbers in data.
+        """
+        return np.unique(self.baseline_array)
+
+    def get_bls(self):
+        """
+        Returns list of unique baseline tuples (ant1, ant2) in data.
+        """
+        return [self.baseline_to_antnums(bl) for bl in self.get_baseline_nums()]
+
+    def antpair2ind(self, ant1, ant2):
+        """
+        Get blt indices for given (ordered) antenna pair.
+        """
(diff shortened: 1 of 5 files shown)


pyuvdata/uvdata.py
```diff
@@ -1433,3 +1433,212 @@ def reorder_pols(self, order=None, run_check=True, run_check_acceptability=True)
...
+            Numpy array of nsamples corresponding to key.
+        """
+        ind1, ind2, indp = self._key2inds(key)
+        out = self.nsample_array[:, :, :, indp]
+        out = np.append(out[ind1, :, :, :], np.conj(out[ind2, :, :, :]), axis=0)
+        if squeeze:
+            out = np.squeeze(out)
+        return out
+
+    def __getitem__(self, key):
```
bhazelton: I'm not a fan of this method. It's generally used for very simple objects that contain only an array-like value which is a very poor representation of UVData. I'd be much more comfortable if it was called something else.

dannyjacobs: Also, I agree with Bryna.  Why not just have a .get_data method?

adampbeardsley:
Shoot, I meant to do the tutorials. I'll add them.

pyuvdata/uvdata.py
jsdillon: FWIW, I like the __getitem__ syntax a lot. From my standpoint, getting out a particular piece of data to then do some calculation on it is the first thing I want to do with a pyuvdata object and facilitating that in as simple and obvious a way as possible will go a long way toward getting people up and running quickly on pyuvdata. This is the design of DataContainer and part of what makes it such a breeze to use. 

I like that we have both .get_data() and .get_flags() in addition to __getitem__(). I support Adam's change.

bhazelton: The concern that I have with it is that the implementation in this PR restricts `__getitem__` to returning only the data array, rather than any of the other attributes on a UVData object. We have consistently implemented functions like `__iter__` to work over all the attributes in the class, so this is really inconsistent with the rest of the code. There's a nice discussion about how these things should be related here:
https://docs.python.org/2/reference/datamodel.html#emulating-container-types

jsdillon: Hmmm... Well, what if `__getitem__` is a layer on top of `getattr` when you pass it a string (the name of the attribute) and `.get_data()` only when you pass it three arguments?

Reading over that list, I realized... @adampbeardsley do we want a __setitem__ as well?

adampbeardsley: I think Bryna makes some good points, and I hadn't really thought about the symmetry with the __iter__ function. @jsdillon - I don't think we should have completely different behavior for the getitem depending on the input. That's pretty much the definition of when you should just have a separate function.

jsdillon:
Looks great, Adam!

pyuvdata/uvdata.py
```diff
@@ -1433,3 +1434,225 @@ def reorder_pols(self, order=None, run_check=True, run_check_acceptability=True)
...
+        if type(key) is str:
+            # Single string given, assume it is polarization
+            pol_ind = np.where(self.polarization_array == uvutils.polstr2num(key))[0]
+            if len(pol_ind) == 0:
+                raise KeyError('Polarization {pol} not found in data.'.format(pol=key))
+            blt_ind1 = np.arange(self.Nblts)
+            blt_ind2 = np.array([], dtype=np.int64)
+        elif len(key) == 1:
+            key = key[0]  # For simplicity
+            print(key)
```
bhazelton: did you mean to leave this print statement in?

bhazelton:
@adampbeardsley what's the rational behind accepting either 1-3 values or a tuple of 1-3 values in these methods? Is there a strong reason to support both vs one or the other?

jsdillon:
I asked Adam for that feature because sometimes you're looping through all antenna/pol pairs in the format, e.g. `(81,32,'xy')` as a single key in some dictionary, in which case it's easier to use `uv.get_data(key)` instead of `uv.get_data(key[0],key[1],key[2])` while sometimes you're explicitly handling baselines or polarizations in a loop, in which case it's easier to use `uv.get_data(81,31,'xy')`,

bhazelton:
It seems a little unnecessarily complicated and like it clutters the interface to me (you can just add a second set of parentheses in the second case), but I'll let others weigh in.

pyuvdata/uvdata.py
adampbeardsley: Nope. just pushed its removal.

adampbeardsley:
I don't have a strong preference regarding the 3 arguments or tuple - it was easy enough to implement and I can't think of a clashing situation. I have to admit that I did find the extra parentheses in the earlier commit slightly annoying, so I would lean towards the current version as my choice.
Do @dannyjacobs @AaronParsons or @jpober have an opinion?

PR #179, opened 2017-06-27
Add checking for optional parameters by default in check and eq methods

bhazelton:
This addresses issue 177.

There's also a little bit of cleanup of unused code and comments in ms.py


pyuvdata/tests/test_ms.py
```diff
@@ -135,11 +150,29 @@ def test_readUVFITS_readMS():
         DATA_PATH, 'day2_TDEM0003_10s_norx_1src_1spw.uvfits')
     uvtest.checkWarnings(uvfits_uv.read_uvfits, [uvfits_file],
                          message='Telescope EVLA is not')
-    uvtest.checkWarnings(ms_uv.read_ms, [ms_file, True, True, 'DATA', 'AIPS', True],
+    uvtest.checkWarnings(ms_uv.read_ms, [ms_file, True, True, 'DATA', 'AIPS'],
```
adampbeardsley: This doesn't look like it should be working. The positional arguments for read_ms are:
`filepath, run_check, check_extra, run_check_acceptability, data_column, and pol_order`
So three booleans, but only two are being passed.
Also, the extra arguments are actually not needed if they're being set to the default anyway. It looks like the original line had all these to get to the last keyword.
But....
checkWarnings also takes an argument `func_kwargs`, which is a dictionary to pass keywords to the function in question (read_ms here). So the old line would have looked like:
`uv.test.checkWarnings(ms_uv.read_ms, [ms_file], {test_import_uvfits=True}, ...`
Just an FYI for future use of checkWarnings.

pyuvdata/uvbase.py
```diff
@@ -181,69 +203,75 @@ def __ne__(self, other):
...
-                                                        eshape=eshape))
-                if eshape == ():
-                    # Single element
-                    if not isinstance(param.value, param.expected_type):
-                        raise ValueError('UVParameter ' + p + ' is not the appropriate'
-                                         ' type. Is: ' + str(type(param.value)) +
-                                         '. Should be: ' + str(param.expected_type))
+                # Check parameter shape
+                eshape = param.expected_shape(self)
+                if eshape is None:
```
adampbeardsley: I don't think this is possible. The default of `expected_shape` is to return `()`.

pyuvdata/tests/test_ms.py
```diff
@@ -77,6 +64,24 @@ def test_readMSreadUVFITS():
...
     ms_uv.history = ""
     uvfits_uv.history = ""
+
+    # the objects won't be equal because uvfits adds some optional parameters
+    nt.assert_false(uvfits_uv == ms_uv)
+    # they are equal if only required parameters are checked:
+    nt.assert_true(uvfits_uv.__eq__(ms_uv, check_extra=False))
+
+    # set those parameters to none to check that the rest of the objects match
+    for p in uvfits_uv.extra():
```
bhazelton: We could. I'm agnostic about it but I think this is the first time it's come up. The issue here is comparing a CASA exported uvfits to the matching measurement set and I actually suspect that some of these parameters might be available in measurement sets. I have a question out to Aaron about that (although it might have gotten lost in the noise on the channel. I'll email him about it specifically)

pyuvdata/uvbase.py
bhazelton: Ahh, thanks. I didn't change this (I just indented what was there under an if statement) but I'll take it out.

pyuvdata/ms.py
```diff
@@ -214,7 +214,11 @@ def read_ms(self, filepath, run_check=True, run_check_acceptability=True, data_c
                 ant_names.append(self.antenna_names[antNum])
         self.antenna_names = ant_names
         self.antenna_numbers = self.antenna_numbers[np.invert(antFlags)]
-        self.antenna_positions = self.antenna_positions[np.invert(antFlags), :]
+
+        relative_positions = np.zeros_like(full_antenna_positions)
+        for ind in range(3):
+            relative_positions[:, ind] = full_antenna_positions[:, ind] - self.telescope_location[ind]
```
bhazelton: Thanks!

pyuvdata/tests/test_ms.py
bhazelton: Actually, as I think about it more, I don't think we should, mostly because antenna_positions are technically not required and I think you usually wouldn't want to lose those. So I think it's better to let people clear things as needed.

PR #183, opened 2017-06-29
Fix form selection of total_quality_array

plaplant:
The form of total_quality_array was not being set for different delay types. This change fixes that bug. The intrinsic shape of the total_quality_array has been removed as well, since its shape depends only on `_gain_array.form` or `_delay_array.form` and drops the `Nants_data` dimension.

### DIFF
--- a/pyuvdata/uvcal.py
+++ b/pyuvdata/uvcal.py
@@ -221,8 +221,6 @@ class UVCal(UVBase):
                 'if cal_type is delay, shape is (Nspws, 1, Ntimes, Njones), '
                 'type = float.')
         self._total_quality_array = uvp.UVParameter('total_quality_array', description=desc,
-                                                    form=('Nspws', 'Nfreqs',
-                                                          'Ntimes', 'Njones'),
                                                     expected_type=np.float,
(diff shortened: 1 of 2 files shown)


pyuvdata/uvcal.py
```diff
@@ -221,8 +221,6 @@ def __init__(self):
                 'if cal_type is delay, shape is (Nspws, 1, Ntimes, Njones), '
                 'type = float.')
         self._total_quality_array = uvp.UVParameter('total_quality_array', description=desc,
-                                                    form=('Nspws', 'Nfreqs',
```
bhazelton: These 2 lines should not be removed, they represent the default (just like we have for the quality_array). They can then get updated in the set functions.

plaplant: Okay, I've added the default shape back in. I've also added a test to make sure the array is the right size.

plaplant:
My comment above got swallowed up in the "outdated" section, so I'm adding it here in the main discussion. I added the default size of the array back into the object definition, and added a test to check the size of `total_quality_array` for both gain-style and delay-style files.

Issue #185, opened 2017-07-01
UVData.__add__ operator errors at last check()


nkern:
I'm trying to add two UVData objects via UV1 + UV2 and I am getting an error that seems to originate from the final check() call in the `__add__()` method. I'm not sure if this is a bug or if this is something inherently wrong in my data files. See traceback below:

```
ValueError                                Traceback (most recent call last)
<ipython-input-21-e607b1265f43> in <module>()
----> 1 UV1 + UV2

/home/nkern/anaconda2/lib/python2.7/site-packages/pyuvdata-1.1-py2.7.egg/pyuvdata/uvdata.pyc in __add__(self, other, run_check, check_extra, run_check_acceptability, inplace)
    902         if run_check:
    903             this.check(check_extra=check_extra,
--> 904                        run_check_acceptability=run_check_acceptability)
    905 
    906         if not inplace:

/home/nkern/anaconda2/lib/python2.7/site-packages/pyuvdata-1.1-py2.7.egg/pyuvdata/uvdata.pyc in check(self, check_extra, run_check_acceptability)
    323 
    324         super(UVData, self).check(check_extra=check_extra,
--> 325                                   run_check_acceptability=run_check_acceptability)
    326 
    327         # Check internal consistency of numbers which don't explicitly correspond

/home/nkern/anaconda2/lib/python2.7/site-packages/pyuvdata-1.1-py2.7.egg/pyuvdata/uvbase.pyc in check(self, check_extra, run_check_acceptability)
    244                                          'Parameter shape is {pshape}, expected shape is '
    245                                          '{eshape}.'.format(param=p, pshape=np.shape(param.value),
--> 246                                                             eshape=eshape))
    247                     if eshape == ():
    248                         # Single element

ValueError: UVParameter _zenith_dec is not expected shape. Parameter shape is (10640,), expected shape is (21280,).
```

nkern:
The problem seems to be that the final UVData object expects `_zenith_dec` to have length equal to twice the incoming `_zenith_dec` from each object, but this isn't the case.

adampbeardsley:
This looks like it's my bad. @nkern - what axis/axes are you adding along (e.g. frequency, time, baselines, pols)? Are these data files available I can test against?

nkern:
I'm trying to add data coming originally from miriad files. I wasn't specifying an axis to add over, I guess I assumed the default would be adding over time.

nkern:
[...]

PR #186, opened 2017-07-01
Fix for adding drift UVData objects.

adampbeardsley:
My fault for not checking against drift data before. This should address Issue #185.
@nkern Please check that this fixes your problem.


nkern:
That works for me! Thanks!

PR #200, opened 2017-07-03
Adding smart slicing functionality

jsdillon:
Returns the output of get_data, get_flags, etc. faster by minimizing
unnecessary copying (and as a view rather than copy if possible)


pyuvdata/uvdata.py
```diff
@@ -1583,6 +1583,57 @@ def _key2inds(self, key):
...
+        """
+        if len(ind2)==0:
+            #only unconjugated baselines
+            isRegularlySpaced = len(set(np.ediff1d(ind1))) <= 1
+            if isRegularlySpaced:
+                if len(indp)==1:
+                    out = data[ind1[0]:ind1[-1]+1:ind1[1]-ind1[0], :, :, indp[0]]
+                else: 
+                    out = data[ind1[0]:ind1[-1]+1:ind1[1]-ind1[0], :, :, :]
+                    out = out[:,:,:,indp]
```
adampbeardsley: At this point, if indp is all the polarizations, we can save time (and maybe not have to copy data) by not executing this line.

pyuvdata/uvdata.py
```diff
@@ -1583,6 +1583,57 @@ def _key2inds(self, key):
...
+        Method for quickly picking out the relevant section of data for get_data or get_flags
+
+        Args:
+            data: 4-dimensional array in the format of self.data_array
+            ind1: list with blt indices for antenna pair (e.g. from self._key2inds)
+            ind2: list with blt indices for conjugate antenna pair. (e.g. from self._key2inds)
+            indp: list with polarization indices (e.g. from self._key2inds)
+
+        Returns:
+            out: copy (or if possible, view) of relevant section of data as an numpy array
```
adampbeardsley: I do worry a little about the output sometimes being a view and sometimes being a copy. This makes it not obvious to the user whether changes they make to their output will actually change the uvdata object. Unfortunately I don't know how to get around it.
@bhazelton do you have any thoughts on this?

bhazelton: I agree that this is worrying. I think one thing that would help is to provide an option to force it to be a copy. Then if that option is not set but you have to return a copy you could imagine printing a notification or warning that it was a copy that was being returned.

adampbeardsley: Ah, I like that. I'll try to implement it.

adampbeardsley:
Oh shoot, we need to add some tests for this before we can approve it.

pyuvdata/uvdata.py
jsdillon: Confirmed. That works.

pyuvdata/uvdata.py
jsdillon: Here's a problem: you also copy the data when you conjugate it. I think the typical user wants the data as quickly as possible: a copy if necessary but a view is always preferable since it's much, much faster. On the other hand, I think the user wants pyuvdata to be smart and conjugate the data when the baseline order is backwards. So it's never clear whether the result is going to be a view or a copy.

If the user wants a copy, they can always call `np.array(uv.get_data(...))` which makes a copy. I'd rather clarify the documentation to warn the user against trying to update the underlying data with `get_data`, etc. We can also include a force_copy option if that's helpful.

adampbeardsley: But Josh I feel like you just described exactly why we need a warning - the user wants the data as fast as possible, but sometimes it will be a copy and sometimes not.
One alternative I'm looking into is read-only views. We could return arrays that cannot be changed, unless the user manually changes that setting. thoughts?

jsdillon: That's interesting. Let me play with that.

jsdillon: Yup, that should work. Just pushed. Another neat feature of numpy: views can be made read only even when the underlying array is writable.

jsdillon: Nice idea @adampbeardsley ...I didn't really know about that feature before.

adampbeardsley:
When adding tests, I noticed that the smart slicing function fails if only one blt index is select (e.g. `len(ind1) == 1`. I've been looking for an elegant way of fixing, but so far have only come up with more if statements. I'll get back to it tomorrow, but thought I'd put this out there in case anyone else looks at it between now and then.

jsdillon:
Interesting. Does it fail because of squeezing? Should we make the squeezing default to only operate along the spectral window and polarization dimension? Data container always returns a 2D array, even if one of the dimensions has length 1.

adampbeardsley:
The error I was looking into is the actual indexing, for example `out = data[ind1[0]:ind1[-1] + 1:ind1[1] - ind1[0], :, :, indp[0]]`. But yeah, we could change the squeezing behavior if it's better to retain the time dimension.

adampbeardsley:
This branch should be ready now, including full test coverage. @bhazelton would you mind reviewing because I've gotten too close to be an independent reviewer.

.gitignore
```diff
@@ -65,3 +65,4 @@ target/
 pyuvdata/data/
 pyuvdata/data/test/
 pyuvdata/GIT_INFO
+uvdata/data/test/*
```
bhazelton: Why is this needed? Do tests need to write to a different folder than pyuvdata/data/test/ ?

pyuvdata/uvdata.py
```diff
@@ -1575,72 +1575,168 @@ def _key2inds(self, key):
...
+
+        if squeeze is 'full':
+            out = np.squeeze(out)
+        elif squeeze is 'default':
+            if out.shape[3] is 1:
+                # one polarization dimension
+                out = np.squeeze(out, axis=3)
+            if out.shape[1] is 1:
+                # one spw dimension
+                out = np.squeeze(out, axis=1)
```
bhazelton: By only checking for squeeze == 'full' or 'defualt', you're treating any other value as equivalent to 'none', which is other acceptable value you listed in the docstring. Is this what you want or should it error if you try to pass it something else?

And are you sure that the no squeeze case should be given by the string 'none' vs the value None?

.gitignore
jsdillon: I added it to prevent files from getting automatically added (github desktop tries to automatically track files) added when I run nosetests locally.

pyuvdata/uvdata.py
jsdillon: That's a good point. I'll edit this to raise an error if someone else is passed in.

I prefer the string 'none' to None, since the no squeezing case is not the default. Kind of like `interpolation='none'` in `plt.imshow` (in old versions of matplotlib).

.gitignore
bhazelton: That's strange. The location that test data is written is usually pyuvdata/data/test. If it's getting written somewhere else we should understand why.

jsdillon: Not sure what I was doing before, but I'm not getting anything there anymore. Could be because I was running a subset of the tests. Anyway, I'll remove this line.

jsdillon:
@bhazelton I believe I've addressed your comments. We ready to merge?

Issue #181, opened 2017-06-29
order_pols function is incomplete


adampbeardsley:
This function should reorder flags and nsamples as well. Is probably best to call reorder_pols.
More abstractly, we may want to come up with a way of programmatically finding parameters related to a given axis. For example, `if 'Npols` in p.form`. Something like that might prevent functions from getting bloated with a list of parameters to consider when we act on an axis.

PR #203, opened 2017-07-05
Completing order_pols function

adampbeardsley:
Addressed issue #181


dannyjacobs:
This is a weird one. Adam is catching bugs before they happen.

PR #205, opened 2017-07-11
Add x_orientation to UVData, add support for uvfits and miriad

bhazelton:
Currently there's not support for this in read_fhd or read_ms, waiting on input from @anguta and Ian about whether there is something to read in from those file types or not.


bhazelton:
There is apparently nothing in a measurement set that can be read in to fill in this value, so it will not be set in read_ms.

PR #207, opened 2017-07-12
Add parse_ants function to UVData object

jburba:
Might need to be modified for pep8, but 100% of parse_ants covered by tests.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1,4 +1,4 @@
-polarizationsTutorial
+Tutorial
 ========
 
 UVData: File conversion
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -369,6 +369,7 @@ ________
 
   # Print the number of antenna pairs after the select
   >>> print(len(UV.get_antpairs()))
+  48
 
 2. Individual baseline(s):
 ________
@@ -384,13 +385,13 @@ ________
   >>> UV = UVData()
   >>> filename = 'pyuvdata/data/day2_TDEM0003_10s_norx_1src_1spw.uvfits'
   >>> UV.read_uvfits(filename)
-  # Print the number of antenna pairs in the original file
 
+  # Print the number of antenna pairs in the original file
   >>> print(len(UV.get_antpairs()))
   153
 
   # Apply select to UV object
-  >>> UV.select(ant_str='(1,2)_(3,4)')
+  >>> UV.select(ant_str='(1,2)_(3,6)')
 
   # Print the antennas pairs with data after the select
   >>> print(UV.get_antpairs())
@@ -417,6 +418,7 @@ all antenna pairs kept in the object will retain data for each specified polariz
   >>> UV = UVData()
   >>> filename = 'pyuvdata/data/day2_TDEM0003_10s_norx_1src_1spw.uvfits'
   >>> UV.read_uvfits(filename)
+
   # Print the number of antennas and polarizations with data in the original file
   >>> print(len(UV.get_antpairs()), UV.get_pols())
   (153, ['RR', 'LL', 'RL', 'LR'])
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -348,7 +348,7 @@ d) Select antenna pairs and polarizations using ant_str argument
 Basic options are 'auto', 'cross', or 'all'.  The ant_str can also contain:
 
 1. Individual antenna number(s):
-====
+________
 
 - 1: returns all antenna pairs containing antenna number 1 (including the auto correlation)
 - 1,2: returns all antenna pairs containing antennas 1 and 2
@@ -367,7 +367,7 @@ Basic options are 'auto', 'cross', or 'all'.  The ant_str can also contain:
   print(UV.get_antpairs())
 
 2. Individual baseline(s):
-====
+________
 
 - 1_2: returns only the antenna pair (1,2)
 - 1_2,1_3,1_10: returns antenna pairs (1,2),(1,3),(1,10)
@@ -388,15 +388,14 @@ Basic options are 'auto', 'cross', or 'all'.  The ant_str can also contain:
   print(UV.get_antpairs())
 
 3. Antenna number(s) and polarization(s):
-====
+________
 
 When polarization information is passed with antenna numbers,
 all antenna pairs kept in the object will retain data for each specified polarization
 
 - 1x: returns all antenna pairs containing antenna number 1 and polarizations xx and xy
 - 2x_3y: returns the antenna pair (2,3) and polarization xy
-- 1r_2l,1l_3l,1r_4r: returns antenna pairs (1,2), (1,3), (1,4) and polarizations rr, ll, and rl.
-This yields a complete list of baselines with polarizations of 1r_2l, 1l_2l, 1r_2r, 1r_3l, 1l_3l, 1r_3r, 1r_11l, 1l_11l, and 1r_11r.
+- 1r_2l,1l_3l,1r_4r: returns antenna pairs (1,2), (1,3), (1,4) and polarizations rr, ll, and rl.  This yields a complete list of baselines with polarizations of 1r_2l, 1l_2l, 1r_2r, 1r_3l, 1l_3l, 1r_3r, 1r_11l, 1l_11l, and 1r_11r.
 - (1x,2y)_(3x,4y): returns antenna pairs (1,3),(1,4),(2,3),(2,4) and polarizations xx, yy, xy, and yx
 - 2l_3: returns antenna pair (2,3) and polarizations ll and lr
 - 2r_3: returns antenna pair (2,3) and polarizations rr and rl
@@ -417,7 +416,7 @@ This yields a complete list of baselines with polarizations of 1r_2l, 1l_2l, 1r_
   print(UV.get_antpairs(), UV.get_pols())
 
 4. Stokes parameter(s):
-====
+________
 
 Can be passed lowercase or uppercase
 
@@ -425,7 +424,7 @@ Can be passed lowercase or uppercase
 - q,V: keeps both Stokes Q and V
 
 5. Minus sign(s):
-====
+________
 
 If a minus sign is present in front of an antenna number, it will not be kept in the data
 
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -351,20 +351,24 @@ Basic options are 'auto', 'cross', or 'all'.  The ant_str can also contain:
 ________
 
 - 1: returns all antenna pairs containing antenna number 1 (including the auto correlation)
-- 1,2: returns all antenna pairs containing antennas 1 and 2
+- 1,2: returns all antenna pairs containing antennas 1 and/or 2
 
 ::
 
-  from pyuvdata import UVData
-  UV = UVData()
-  filename = 'pyuvdata/data/day2_TDEM0003_10s_norx_1src_1spw.uvfits'
-  UV.read_uvfits(filename)
-  # Print all the antennas numbers with data in the original file
-  print(UV.get_antpairs())
+  >>> from pyuvdata import UVData
+  >>> UV = UVData()
+  >>> filename = 'pyuvdata/data/day2_TDEM0003_10s_norx_1src_1spw.uvfits'
+  >>> UV.read_uvfits(filename)
+  # Print the number of antenna pairs in the original file
+
+  >>> print(len(UV.get_antpairs()))
+  153
+
   # Apply select to UV object
-  UV.select(ant_str='1,2,3')
-  # Print all the antennas numbers with data after the select
-  print(UV.get_antpairs())
+  >>> UV.select(ant_str='1,2,3')
+
+  # Print the number of antenna pairs after the select
+  >>> print(len(UV.get_antpairs()))
 
 2. Individual baseline(s):
 ________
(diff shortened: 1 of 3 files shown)


pyuvdata/tests/test_uvdata.py
```diff
@@ -1548,3 +1548,160 @@ def test_get_feedpols():
...
+
+    # Multiple baselines as list
+    ant_str = '1_2,1_3,1_10'
+    ant_pairs_nums,polarizations = uv.parse_ants(ant_str)
+    ant_pairs_data = [(1, 2),(1, 3),(1, 10)]
+    nt.assert_items_equal(ant_pairs_nums,ant_pairs_data)
+    nt.assert_is_instance(polarizations,type(None))
+
+    # Multiples baselines with polarizations as list
+    ant_str = '1r_2l,1l_3l,1r_11r'
```
adampbeardsley: We should make it clear in the documentation that if you pass the result from a selection like this, you don't just get the three baselines, but rather you get all the pairs of antennas with all the polarizations. In this case you end up with:
1r_2l, 1l_2l, 1r_2r, 1r_3l, 1l_3l, 1r_3r, 1r_11l, 1l_11l, and 1r_11r.
If the intent is to pass to select, this is the right thing to do, but it should be clear.

pyuvdata/tests/test_uvdata.py
```diff
@@ -1548,3 +1548,160 @@ def test_get_feedpols():
...
+
+    # Antenna numbers with polarizations
+    ant_str = '(1l,2r)_(3l,6r)'
+    ant_pairs_nums,polarizations = uv.parse_ants(ant_str)
+    ant_pairs_data = [(1, 3),(1, 6),(2, 3),(2, 6)]
+    polarizations_data = [-1, -2, -3, -4]
+    nt.assert_items_equal(ant_pairs_nums,ant_pairs_data)
+    nt.assert_items_equal(polarizations,polarizations_data)
+
+    # Antenna numbers with - for avoidance
```
adampbeardsley: Oh I think I see later that it excludes that antenna, rather than include it. Is there a more practical way to test this? The examples given aren't all that useful because e.g. `'1_10'` gives you the same result as `'1_(-3,10)'`, so why would you use the latter? Is the minus useful when otherwise using wildcards? For example, all baselines with antenna 1, excluding baselines with antenna 3?
Hmm.. after playing myself a little, it looks like `'1,-3'` is equivalent to `'1,3'`. Is this correct behavior?

adampbeardsley:
re: @bhazelton I think it would be useful for select to take a string like this, but get_* functions won't really synergize well with this format. The get functions only allow certain slices of the data - single baseline/pol, single baseline with all pols, or single pol with all baselines. It's possible to use the output of this function to sequentially call get_*, but I think that's user level. Might be a good example for the tutorial though.
I also wouldn't want to lose the current select interface. Maybe we can add a new keyword which, if set, just runs select using the output of parse_ants.

jburba:
@adampbeardsley I agree.  Taking a look at the get_* functions, the input syntax is doable from parse_ants output, but I designed the parser with the select function in mind.  Currently working on having parse_ants called from the select function via an antenna string.

pyuvdata/uvdata.py
```diff
@@ -1774,3 +1775,147 @@ def antpairpol_iter(self, squeeze='default'):
...
+    def parse_ants(self,ant_str):
+        """
+        Generates two lists of antenna pair tuples and polarization indices based on
+        parsing of the string ant_str.  If no valid polarizations or antenna numbers
+        are found in ant_str (Stokes params, or combinations of [lr] or [xy]),
+        ant_pairs_nums and polarizations are returned as None.
+
+        Args:
+            ant_str: String containing antenna information to pass to select function.
+                Can be all, auto, cross, or combinations of antenna numers and polarization
```
jburba: Corrected

pyuvdata/uvdata.py
```diff
@@ -1774,3 +1775,147 @@ def antpairpol_iter(self, squeeze='default'):
...
+
+        Output:
+            ant_pairs_nums: List of tuples containing the parsed pairs of antennae numbers.
+                If 'all' is passed as ant_str, returned as None.
+            polarizations: List of desired polarizations.  If no polarizations found in ant_str
+                then returned as None.
+        """
+
+        ant_re = r'(\(((-?\d+[lrxy]?,?)+)\)|-?\d+[lrxy]?)'
+        bl_re = '(^(%s_%s|%s),?)' % (ant_re, ant_re, ant_re)
```
jburba: This came straight from `aipy/scripting.py` but I did add support for circular polarizations.

pyuvdata/uvdata.py
```diff
@@ -1774,3 +1775,147 @@ def antpairpol_iter(self, squeeze='default'):
...
+                        for ant2 in ant_nums:
+                            if not ant1 == ant2:
+                                if ant1 < ant2:
+                                    ant_tuple = tuple((ant1,ant2))
+                                else:
+                                    ant_tuple = tuple((ant2,ant1))
+                                if not ant_tuple in ant_pairs_nums:
+                                    ant_pairs_nums.append(ant_tuple)
+                elif ant_str[str_pos:].upper().startswith('I'):
+                    polarizations.append(1)
```
jburba: Corrected

pyuvdata/uvdata.py
```diff
@@ -1774,3 +1775,147 @@ def antpairpol_iter(self, squeeze='default'):
...
+            m = re.search(bl_re, ant_str[str_pos:])
+            if m is None:
+                if ant_str[str_pos:].startswith('all'):
+                    pass
+                elif ant_str[str_pos:].startswith('auto'):
+                    for ant in ant_nums:
+                        ant_tuple = tuple((ant,ant))
+                        ant_pairs_nums.append(ant_tuple)
+                elif ant_str[str_pos:].startswith('cross'):
+                    for ant1 in ant_nums:
```
jburba: This goes back to the discussion from Slack regarding how much the parser should know about the object.  I previously had implemented a check that an antenna pair would not be output in `ant_pairs_nums` unless it was present in `self.get_antpairs()`.  Should I put this check back in?  Or should that check be left to the select function?

pyuvdata/uvdata.py
```diff
@@ -1774,3 +1775,147 @@ def antpairpol_iter(self, squeeze='default'):
...
+        ant_nums = self.get_ants()
+
+        while str_pos < len(ant_str):
+            m = re.search(bl_re, ant_str[str_pos:])
+            if m is None:
+                if ant_str[str_pos:].startswith('all'):
+                    pass
+                elif ant_str[str_pos:].startswith('auto'):
+                    for ant in ant_nums:
+                        ant_tuple = tuple((ant,ant))
```
jburba: Corrected

pyuvdata/uvdata.py
jburba: Corrected in `a97f1af`

pyuvdata/tests/test_uvdata.py
```diff
@@ -1548,3 +1548,160 @@ def test_get_feedpols():
...
+    ant_pairs_data = [(0, 0),(1, 1),(2, 2),(3, 3),(6, 6),(7, 7),(8, 8),(11, 11),(14, 14),(18, 18),
+                               (19, 19),(20, 20),(21, 21),(22, 22),(23, 23),(24, 24),(26, 26),(27, 27)]
+    nt.assert_items_equal(ant_pairs_nums,ant_pairs_data)
+    nt.assert_is_instance(polarizations,type(None))
+
+    # Cross correlations
+    ant_str = 'cross'
+    ant_pairs_nums,polarizations = uv.parse_ants(ant_str)
+    for ant_pair in ant_pairs_nums:
+        nt.assert_not_equal(ant_pair[0],ant_pair[1])
```
jburba: Corrected in `860ff6e`

pyuvdata/tests/test_uvdata.py
jburba: The behavior of the minus sign now works "correctly."  If you pass '1,-1_3', all baselines including antenna 1 are kept except (1,3).  Similarly, if you pass '1,2,-3' to select via ant_str, all baselines including antennas 1 and 2 are kept except the pairs (1,3) and (2,3).  Corrected in `8d6c986`

docs/tutorial.rst
```diff
@@ -237,7 +237,110 @@ c) Select a few antenna pairs to keep
...
+  # Print all the antennas numbers and polarizations with data after the select
+  print(UV.get_antpairs(), UV.get_pols())
+
+4. Stokes parameter(s):
+________
+
+Can be passed lowercase or uppercase
+
+- i,I: keeps only Stokes I
+- q,V: keeps both Stokes Q and V
```
bhazelton: this needs an example.

pyuvdata/tests/test_uvdata.py
```diff
@@ -1548,3 +1548,366 @@ def test_get_feedpols():
...
+    # All baselines
+    ant_str = 'all'
+    uv2 = uv.select(ant_str=ant_str, inplace=inplace)
+    nt.assert_items_equal(uv2.get_antpairs(), uv.get_antpairs())
+    nt.assert_items_equal(uv2.get_pols(),uv.get_pols())
+
+    # Auto correlations
+    ant_str = 'auto'
+    nt.assert_raises(ValueError, uv.select, ant_str=ant_str, inplace=inplace)
+    # No auto correlations in this data
```
bhazelton: You also need a test with a data set that has auto correlations so you can test that you get the expected number back. We have an autocorrelation test somewhere, so we have at least one test data set with autocorrelations.

pyuvdata/tests/test_uvdata.py
```diff
@@ -1548,3 +1548,366 @@ def test_get_feedpols():
...
+
+    # Cross correlations
+    ant_str = 'cross'
+    uv2 = uv.select(ant_str=ant_str, inplace=inplace)
+    nt.assert_items_equal(uv2.get_antpairs(), uv.get_antpairs())
+    nt.assert_items_equal(uv2.get_pols(), uv.get_pols())
+    # All baselines in data are cross correlations
```
bhazelton: You also need a test with a data set that has auto correlations so you can test that this doesn't just always return everything.

docs/tutorial.rst
```diff
@@ -1,4 +1,4 @@
-Tutorial
+polarizationsTutorial
```
bhazelton: Did you really want to change this?

pyuvdata/tests/test_uvdata.py
```diff
@@ -1548,3 +1548,366 @@ def test_get_feedpols():
...
+    # Pass ant_str with ant_pairs_nums and polarizations to select
+    ant_str = '1l_3r'
+    ant_pairs_nums = [(1, 11)]
+    ant_pairs = [(1, 3), (1, 11)]
+    polarizations = [-1] # RR
+    pols = ['RR', 'LR']
+    uv2 = uv.select(ant_str=ant_str, ant_pairs_nums=ant_pairs_nums,
+                            polarizations=polarizations, inplace=inplace)
+    nt.assert_items_equal(uv2.get_antpairs(), ant_pairs)
+    nt.assert_items_equal(uv2.get_pols(), pols)
```
bhazelton: This last test is great! You should probably also have test for the case where ant_str is set along with antenna_numbers or antenna_names.

jpober:
A couple of general comments/questions for everyone:

(1) It seems like we have some duplicate functionality in that an ant_str can be used to accomplish a lot of the same things that would be accomplished by e.g. antenna_nums or ant_pairs_nums.  I understand not wanting to remove existing functionality, but I think we should consider having select() only accept an ant_str and making sure all the existing selects can be performed with an ant_str (or, maybe, an antpol_str if we want to incorporate polarization functionality completely into the string).  I think it's too confusing to have some things doable only with ant_str and others only doable with other select arguments.

(2) A couple pieces of behavior that don't seem right; some are due to select() and not parse_ants, but the two really should work in tandem.

a) On a data file with only xx polarization visibilities, uv2 = uv.select(ant_str='10x',inplace=False) returns an error that polarization -7 (i.e. XY) isn't in the data.  I think this should be a valid select and return all baselines containing the x polarization of antenna 10.  Not sure whether the right place for a change is in select or parse_ants.

b) parse_ants doesn't handle string arguments ('all', 'cross', 'auto') plus polarizations (or at least I couldn't figure out how to do it.  This would be for a case where e.g. I wanted all autocorrelations in a specific polarization.  I suppose the polarizations argument in select can handle this, but see comment (1) above. 

c) I couldn't find support for getting e.g. all baselines containing antenna 10 in the xx polarization (e.g. '10x_x' or something like that).  

Overall, this is _very_ nice.  Good work!

docs/tutorial.rst
jburba: Is there a file distributed with pyuvdata that contains Stokes params?  The file I used for the other examples only has circular pols.

bhazelton: No, there is not a file that is circulated with Stokes parameters, but for a test you can always change the values in the polarization array after reading in any of the test files and before testing the select.

jburba:
Added some functionality for the plotting script/notebook I am working on now with @jpober as a "replacement" for plot_uv.py from aipy which is why the coverage decreased.  The output formatting for this plotting functionality is not set in stone yet, so I figured there is no immediate reason to test it.  That bit of the code for plotting aside, there's 100% coverage of the code in the parse_ants function written specifically for the select function which is the main contribution of this pull request.

jburba:
Last clarifying comment hopefully.  I moved all of the code I added to parse_ants to another branch parse_ants_plotting to prevent any confusion when looking at the parse_ants function and its tests.  This branch now only contains the pieces of parse_ants that are designed to work with the select function.

bhazelton:
@jburba Can you please rebase your branch on master? There has been a significant change in the way we format the tutorial to support automatic documentation testing via Travis (to ensure that the tutorial does not get out of date). If you'd like help incorporating these changes into your PR after you rebase, I'd be happy to help.

bhazelton:
@jburba this looks like it's really close. It looks to me like there are a few outstanding comments that we should address. Some of them are mine about testing changes -- those should be pretty easy. 

@jpober's comments are more difficult to address. I think some of them are things that need to be discussed and decided by a small group of people while you may be able to address the others. We may need to have a short telecon about this with @adampbeardsley @dannyjacobs in order to make the decisions.

jpober:
I think my comments have largely been addressed, although I haven't looked at the code in a while.  I think the plan was to accept either the existing syntax or the ant_str in select() but to not accept both simultaneously and to provide clear erroring when both are provided.  @jburba can correct me if I'm wrong.

bhazelton:
@jburba I guess not. I was trying to remember where we were on this PR and It looked like there were review comments about testing that hadn't been addressed, but I looked more carefully at the code and it looks like you have actually addressed all of them.

Issue #164, opened 2017-06-13
Add __add__ method to UVCal


bhazelton:
Can closely follow the similar method on UVData

plaplant:
This was addressed by PR #209.

PR #209, opened 2017-07-13
Add __add__ method for UVCal

plaplant:
This PR adds an `__add__` method to the UVCal object and addresses Issue #164. The general structure closely follows the `__add__` method of the UVData object. This method allows the user to combine two different UVCal objects of the same `cal_type` along the antenna, frequency, time, or Jones axis. Special care is taken to appropriately handle the optional parameters `total_quality_array` and `input_flag_array`.

### DIFF
--- a/pyuvdata/uvcal.py
+++ b/pyuvdata/uvcal.py
@@ -947,9 +947,9 @@ class UVCal(UVBase):
                               'it impossible to write this data out to some file types.')
 
         if this.Njones > 2:
-            pol_separation = np.diff(this.polarization_array)
-            if np.min(pol_separation) < np.max(pol_separation):
-                warnings.warn('Combined polarizations are not evenly spaced. This will '
+            jones_separation = np.diff(this.jones_array)
+            if np.min(jones_separation) < np.max(jones_separation):
+                warnings.warn('Combined Jones elements are not evenly spaced. This will '
                               'make it impossible to write this data out to some file types.')
 
         if n_axes > 0:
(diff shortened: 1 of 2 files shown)


pyuvdata/uvcal.py
```diff
@@ -583,3 +583,407 @@ def write_calfits(self, filename, run_check=True, check_extra=True,
...
+        this.check(check_extra=check_extra, run_check_acceptability=run_check_acceptability)
+        if not isinstance(other, this.__class__):
+            raise(ValueError('Only UVCal objects can be added to a UVCal object'))
+        other.check(check_extra=check_extra, run_check_acceptability=run_check_acceptability)
+
+        # Check objects are compatible
+        # Note zenith_ra will not necessarily be the same if times are different.
+        # But phase_center should be the same, even if in drift (empty parameters)
+        compatibility_params = ['_cal_type', '_integration_time', '_channel_width',
+                                '_telescope_name', '_gain_convention', '_x_orientation']
```
plaplant: Good idea. I don't want to be too restrictive, so I think a warning is appropriate.

pyuvdata/uvcal.py
```diff
@@ -583,3 +583,407 @@ def write_calfits(self, filename, run_check=True, check_extra=True,
...
+        # Check we don't have overlapping data
+        both_jones = np.intersect1d(
+            this.jones_array, other.jones_array)
+        both_times = np.intersect1d(
+            this.time_array, other.time_array)
+        if this.cal_type != 'delay':
+            both_freq = np.intersect1d(
+                this.freq_array[0, :], other.freq_array[0, :])
+        else:
+            both_freq = []
```
plaplant: Good catch! I'll fix this and add a test to check for this.

plaplant:
@adampbeardsley I think this is ready for another pass

PR #214, opened 2017-07-19
Miriad extra keywords

jpober:


pyuvdata/tests/test_miriad.py
```diff
@@ -151,6 +167,57 @@ def test_poltoind():
...
+    # check for warnings & errors with extra_keywords that are dicts, lists or arrays
+    uvtest.checkWarnings(uv_in.read_miriad, [miriad_file],
+                         known_warning='miriad')
+    uv_in.extra_keywords['testdict'] = {'testkey': 23}
+    uvtest.checkWarnings(uv_in.check, message=['testdict in extra_keywords is a '
+                                               'list, array or dict'])
+    nt.assert_raises(TypeError, uv_in.write_miriad, testfile, clobber=True,
+                     run_check=False)
+
+    uvtest.checkWarnings(uv_in.read_miriad, [miriad_file],
```
bhazelton: no, the things that gets added previously just needs to be removed. I'll fix that.

pyuvdata/uvdata.py
```diff
@@ -338,6 +338,20 @@ def check(self, check_extra=True, run_check_acceptability=True):
             raise ValueError('Ntimes must be equal to the number of unique '
                              'times in the time_array')
 
+        # issue warning if extra_keywords keys aren't longer than 8 characters
```
bhazelton: good catch!

pyuvdata/miriad.py
```diff
@@ -612,6 +667,60 @@ def write_miriad(self, filepath, run_check=True, check_extra=True,
...
+                       }
+        types = {str: 'a',
+                 int: 'i',
+                 float: 'd',
+                 complex: 'c',
+                 bool: 'j',  # NB: bool will be stored as int
+                 }
+        # make a copy of the extra_keywords dict so that the changes required
+        # to write to miriad don't change the object
+        extra_keywords_use = copy.deepcopy(self.extra_keywords)
```
bhazelton: you're right. fixing it now.

pyuvdata/miriad.py
```diff
@@ -622,6 +677,54 @@ def write_miriad(self, filepath, run_check=True, check_extra=True,
             uv.add_var('diameter', 'd')
             uv['diameter'] = self.antenna_diameters
 
+        # other extra keywords
+        # set up dictionaries to map common python types to miriad types
+        # NB: arrays/lists/dicts could potentially be written as strings or 1D
+        # vectors.  This is not supported at present!
+        # NB: complex numbers *should* be supportable, but are not currently
+        # supported due to unexplained errors
```
adampbeardsley: Could we be a little more specific about the errors? Maybe `due to errors in aipy.miriad and/or its underlying libraries`. Just to clarify we're not being lazy about debugging our own code.

PR #217, opened 2017-07-26
Make spw_array backwards compatible

plaplant:
In the calfits files generated as part of IDR2, the `spw_array` was erroneously 0-indexed (as opposed to 1-indexed), so the conversion from a 1-indexed array as required by the FITS format means `self.spw_array == array([-1])`. A check is performed when reading in files, and 1 is only subtracted if the array is found to be 1-indexed. Though this should not be an issue going forward, it does allow users to read in the calfits files generated as part of IDR2.

### DIFF
--- a/pyuvdata/calfits.py
+++ b/pyuvdata/calfits.py
@@ -438,8 +438,12 @@ class CALFITS(UVCal):
 
             # add this for backwards compatibility when the spw CRVAL wasn't recorded
             try:
-                # subtract 1 to be zero-indexed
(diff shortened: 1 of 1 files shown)


bhazelton:
Actually, @plaplant maybe we should add a comment here indicating that this is a backwards-compatibility thing (so it's clear and so it could potentially be removed at a later date?)

plaplant:
@bhazelton thanks, I've added a comment about it. I also added the check for gain-style calfits too.

PR #224, opened 2017-08-04
Add a test based on an ATCA miriad file, update miriad telescope location handling.

bhazelton:
Before changes to miriad.py, this file resulted in a broken UVData object because the telescope location wasn't set. This codifies the approach to location setting with more warnings. The approach is:
1) if altitude is present, the telescope location is set using the lat/lon/alt (lat & lon are required miriad variables).
2) if the telescope is in the known telescopes, the location from the known telescopes is used (warnings issued if lat & lon aren't close, option `correct_lat_lon` allows for not updating the lat/lon based on the known telescope location). Warning issued.
3) If the mean of the antenna positions is on the surface of the earth and the lat/lon of the mean antenna position is close to the file lat/lon, the mean antenna altitude is used with the file lat/lon to generate the telescope location. Warning issued.
4) If the mean of the antenna positions is on the surface of the earth but the lat/lon of the mean antenna position is not close to the file lat/lon, the telescope location is set using the file lat/lon and an altitude of zero (sea-level). Warning issued.
5) If the mean of the antenna positions is not on the surface of the earth, the telescope location is set using the file lat/lon and an altitude of zero (sea-level) and the antenna positions are assumed to be relative positions rather than absolute positions. Warning issued.


bhazelton:
Thanks for the review @jpober!

I'll work on the test coverage issue. It may require hacking some files by hand.

On the issue of long baselines making a telescope location inside the earth: If this happens it would be caught as a not valid telescope location by the test on line 328. This means it would end up in the 5th category I described above, where we'd use the lat/lon from the file and an altitude of sea-level. I'll try to make a test to cover something like this.

bhazelton:
@jpober I added a bunch of testing (all the new stuff is covered now), which revealed some interesting and wrong behavior, which I've now fixed. Thank you for pushing on that!

To better handle the situation you asked about, where the antenna positions were very far apart, I now check to see both if the antenna positions themselves appear to be on the surface of the earth and whether the mean position is on the surface of the earth. If the antenna positions are not on the surface of the earth, they are treated as relative and I set the telescope location using the lat/lon from the file with a sea level altitude. If the antennas are on the surface of the earth, but the mean position is not on the surface of the earth (because they're so far apart), I treat them as absolute positions, but I still set the telescope location using the lat/lon from the file with a sea level altitude (because I don't have a better way to set the altitude). I have a test to cover this situation.

jpober:
Don't think you're going to work on this anytime soon, but I just had a chance to look it overall and seems like it's ready to go.

PR #225, opened 2017-08-04
Cleanup

bhazelton:
miscellaneous code cleanup


pyuvdata/tests/__init__.py
```diff
@@ -31,6 +31,9 @@ def checkWarnings(func, func_args=[], func_kwargs={},
                   nwarnings=1, message=None, known_warning=None):
     """Function to check expected warnings."""
 
+    if (not isinstance(category, list) or len(category) == 1) and nwarnings > 1:
+        category = [category] * nwarnings
```
adampbeardsley: Can we do the same thing for `message`?

bhazelton: We can. Is it common to want the same message repeated several times?

PR #228, opened 2017-08-16
Change miriad saving of antenna names to be cleaner

bhazelton:
requires new version of aipy, which is in the works.


bhazelton:
This PR is now ready for review. @adampbeardsley @dannyjacobs can one of you look at it? It's a pretty small change.

PR #229, opened 2017-08-22
Add an aipy version requirement

bhazelton:
This addresses the fact that we now require a higher aipy version in order to support the new miriad antenna names encoding.


adampbeardsley:
Sorry @plaplant, I updated the README with this new requirement. Could you review again?

PR #231, opened 2017-08-23
update beam object based on discussions with Eloy and James

bhazelton:
A few small tweaks to the beam definition, with support in beamfits and testing

### DIFF
--- a/pyuvdata/beamfits.py
+++ b/pyuvdata/beamfits.py
@@ -237,11 +237,18 @@ class BeamFITS(UVBeam):
         # check to see if BANDPARM HDU exists and read it out if it does
         if 'BANDPARM' in hdunames:
             bandpass_hdu = F[hdunames['BANDPARM']]
+            bandpass_header = bandpass_hdu.header.copy()
+            self.input_impedence = bandpass_header.pop('inimped', None)
+            self.output_impedence = bandpass_header.pop('outimped', None)
+
             freq_data = bandpass_hdu.data
             columns = [c.name for c in freq_data.columns]
+            self.bandpass_array = freq_data['bandpass']
+            self.bandpass_array = self.bandpass_array[np.newaxis, :]
+
             if 'sys_temp' in columns:
-                self.system_temperature_array = freq_data['sys_temp']
-                self.system_temperature_array = self.system_temperature_array[np.newaxis, :]
+                self.receiver_temperature_array = freq_data['sys_temp']
+                self.receiver_temperature_array = self.receiver_temperature_array[np.newaxis, :]
             if 'loss' in columns:
                 self.loss_array = freq_data['loss']
                 self.loss_array = self.loss_array[np.newaxis, :]
@@ -258,6 +265,9 @@ class BeamFITS(UVBeam):
                 self.s_parameters[1, 0, :] = s12
                 self.s_parameters[2, 0, :] = s21
                 self.s_parameters[3, 0, :] = s22
+        else:
+            # no bandpass information, set it to an array of ones
+            self.bandpass_array = np.zeros((self.Nspws, self.Nfreqs)) + 1.
 
         if run_check:
             self.check(run_check_acceptability=run_check_acceptability)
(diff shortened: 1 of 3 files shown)


bhazelton:
@adampbeardsley can you take a look at this one?

pyuvdata/beamfits.py
```diff
@@ -506,36 +516,41 @@ def write_beamfits(self, filename, run_check=True,
...
-            bandpass_hdu = fits.BinTableHDU.from_columns(coldefs)
-            bandpass_hdu.header['EXTNAME'] = 'BANDPARM'
-            hdulist.append(bandpass_hdu)
+        bandpass_col = fits.Column(name='bandpass', format='D',
+                                   array=self.bandpass_array[0, :])
+        col_list = [bandpass_col]
+
+        if self.receiver_temperature_array is not None:
+            sys_temp_col = fits.Column(name='sys_temp', format='D',
+                                       array=self.receiver_temperature_array[0, :])
```
adampbeardsley: Is there a reason this is called a system temperature in the file, but a receiver temperature in the object? Compliance with other formats?

pyuvdata/uvbeam.py
```diff
@@ -143,11 +143,16 @@ def __init__(self):
...
-                'over the sphere, giving the beam dimensions of 1/stradian.')
+                '"physical", "peak" or "solid_angle". Physical normalization '
+                'means that the frequency dependence of the antenna sensitivity '
+                'is included in the data_array while the frequency dependence '
+                'of the receiving chain is included in the bandpass_array. '
+                'Peak normalized means that for each frequency the data_array'
+                'is separately normalized such that the peak is 1 (so the beam '
+                'is dimensionless) and all frequency dependence is moved to the '
+                'bandpass_array. Solid angle normalized means the peak normalized '
+                'beam is divided by the integral of the beam over the sphere, '
```
adampbeardsley: Is this the integral of the power beam, voltage beam, or does it depend on the beam type?

pyuvdata/uvbeam.py
```diff
@@ -250,9 +264,19 @@ def __init__(self):
                                                description=desc, value={},
                                                spoof_val={}, expected_type=dict)
 
-        desc = 'Array of system temperatures, shape (Nspws, Nfreqs), units K'
-        self._system_temperature_array = \
-            uvp.UVParameter('system_temperature_array', required=False,
+        desc = 'Input impedence of receiving chain, units: Ohms'
+        self._input_impedence = uvp.UVParameter('input_impedence', required=False,
+                                                description=desc,
+                                                expected_type=np.float, tols=1e-3)
```
adampbeardsley: Wait, shouldn't impedences have the option of being frequency dependent?

pyuvdata/beamfits.py
bhazelton: I think that's a mistake, I changed the name from system temperature to receiver temperature to be more accurate and I think I missed that one. Thanks!

pyuvdata/uvbeam.py
bhazelton: It depends on the type. I'll clarify a bit, it really makes the most sense for the power beam, but it can be defined for voltage beam as well.

pyuvdata/uvbeam.py
bhazelton: I have a question about this out to Eloy, I'll ping him again.

adampbeardsley:
Looks good. Should we wait on Eloy's response before merging?

bhazelton:
Yes, let's wait for Eloy. I had forgotten that I was waiting on that before, but I just pinged him, so hopefully I'll hear soon.

bhazelton:
So these impedances are actually just the reference impedances and should not be frequency dependent. The frequency dependence is reflected in the S parameters, which are already on the object. I renamed them to clarify the point and clarified the descriptions of the parameters.

pyuvdata/uvbeam.py
```diff
@@ -150,7 +150,9 @@ def __init__(self):
                 'Peak normalized means that for each frequency the data_array'
                 'is separately normalized such that the peak is 1 (so the beam '
                 'is dimensionless) and all frequency dependence is moved to the '
```
zacharymartinot: It is not true that all frequency dependence is moved to the bandpass. The beam is still frequency dependent in general -- i.e. is not a separable function of space and frequency -- for any normalization.

bhazelton: That's true, but it is common in radio astronomy to split out the direction independent frequency dependence (sorry that's awkward wording) into what is called a bandpass and to therefore have the beam peak normalized at every frequency. That is what I was trying to describe here, I'll see if I can make the language more clear. The main point was that this was different than the physical normalization, where the direction independent frequency dependence is split between the bandpass and the data array based on the source of that frequency dependence.

zacharymartinot: Ya, I think you should just remove the end of that sentence:

 "Peak normalized means that for each frequency the data_array is separately normalized such that the peak is 1 (so the beam is dimensionless). If the beam_type is "efield", then peak normalized means that the absolute value of the peak is 1. Solid angle normalized means the peak normalized beam is divided by the integral of the beam over the sphere, so the beam has dimensions of 1/steradian."

PR #234, opened 2017-08-25
Add class to read CST power beams into UVBeam object, add method to convert to Healpix

bhazelton:
I'm not sure this is quite ready, but I wanted to get more people's eyes on it. If anyone has some example CST files, I'd like to know how this does reading them -- I understand that not all CST files are created equal.


bhazelton:
Ok, I think this is now ready for review. It has been tested by Adam Lanman on the Fagnoni CST files.

bhazelton:
This now supports reading in the Efield beams from the CST files, but converting Efield beams to Healpix is not yet supported (this requires careful work and will be left to a later PR).

pyuvdata/uvbeam.py
```diff
@@ -410,6 +412,71 @@ def set_phased_array(self):
...
+
+        phi_vals, theta_vals = np.meshgrid(self.axis1_array, self.axis2_array)
+
+        min_res = np.min(np.array([np.diff(self.axis1_array)[0], np.diff(self.axis2_array)[0]]))
+        nside_min_res = np.sqrt(3 / np.pi) * np.radians(60.) / min_res
+        nside = int(2**np.floor(np.log2(nside_min_res)))
+        assert(hp.pixelfunc.nside2resol(nside) > min_res)
+
+        npix = hp.nside2npix(nside)
+        hpxidx = np.arange(npix)
```
adampbeardsley: `hpxidx` is never used.

pyuvdata/uvbeam.py
```diff
@@ -1099,3 +1166,41 @@ def write_beamfits(self, filename, run_check=True,
...
         del(beamfits_obj)
+
+    def read_cst_beam(self, filenames, beam_type='power', frequencies=None, telescope_name=None,
+                      feed_name=None, feed_version=None, model_name=None,
+                      model_version=None, history=''):
+        """
+        Read in data from a cst file.
+
+        Args:
+            filename: The cst file or list of files to read from.
```
adampbeardsley: `filenames`

pyuvdata/uvbeam.py
```diff
@@ -1099,3 +1166,41 @@ def write_beamfits(self, filename, run_check=True,
...
+            telescope_name: the name of the telescope corresponding to the filename(s).
+            feed_name: the name of the feed corresponding to the filename(s).
+            feed_version: the version of the feed corresponding to the filename(s).
+            model_name: the name of the model corresponding to the filename(s).
+            model_version: the version of the model corresponding to the filename(s).
+            history: A string detailing the history of the filename(s).
+            run_check: Option to check for the existence and proper shapes of
+                required parameters after reading in the file. Default is True.
+            run_check_acceptability: Option to check acceptable range of the values of
+                required parameters after reading in the file. Default is True.
```
adampbeardsley: The check parameters are not getting in or passed on.

docs/tutorial.rst
```diff
@@ -440,3 +440,101 @@ b) Select 3 antennas to keep using the antenna names, also select 5 frequencies
...
+a) Reading a CST power beam file
+****************
+::
+
+  from pyuvdata import UVBeam
+  import numpy as np
+  import matplotlib.pyplot as plt
+  beam = UVBeam()
+  # you can pass several filenames and the objects from each file will be
+  # combined across the appropriate axis -- in this case frequency
```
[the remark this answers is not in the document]
bhazelton: You're right for CST beams. I was thinking about the `__add__` function, which can go across other axes, and other axes are supported for fits files, but CST files need to be different frequencies (at least for now).

docs/tutorial.rst
```diff
@@ -440,3 +440,101 @@ b) Select 3 antennas to keep using the antenna names, also select 5 frequencies
...
+  filenames = ['pyuvdata/data/HERA_NicCST_150MHz.txt', 'pyuvdata/data/HERA_NicCST_123MHz.txt']
+  # have to specify the telescope_name, feed_name, feed_version, model_name
+  # and model_version because they are not included in the file
+  beam.read_cst_power(filenames, telescope_name='HERA', feed_name='PAPER_dipole',
```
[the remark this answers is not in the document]
bhazelton: yeah, it's fixed in my next PR. I can fix it here, but then there'll be more rebasing issues.

pyuvdata/cst_beam.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+class CSTBeam(UVBeam):
+    """
+    Defines a CST-specific subclass of UVBeam for reading CST text files.
+    This class should not be interacted with directly, instead use the
+    read_cst_beam method on the UVBeam class.
+
+    Assumes the structure in the simulation was symmetric under
+    45 degree rotations about the z-axis.
```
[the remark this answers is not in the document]
bhazelton: Yes, it's inherent to CST because the CST files are so simple that they don't specify all the required metadata. But I'll add it as a parameter that defaults to 'x' but can be specified differently. Actually, it should support passing different files for the different pols... I'll think about that.

pyuvdata/cst_beam.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def read_cst_beam(self, filelist, beam_type='power', frequencies=None, telescope_name=None,
+                      feed_name=None, feed_version=None, model_name=None, model_version=None,
+                      history='', run_check=True, run_check_acceptability=True):
+
+        """
+        Read in data from a cst file.
+
+        Args:
+            filename: The cst file or list of files to read from.
```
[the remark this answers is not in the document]
bhazelton: yes, thanks.

pyuvdata/cst_beam.py
```diff
@@ -0,0 +1,223 @@
...
+                self.data_array[1, 0, 1, freq_i, :, :] = theta_beam2
+
+            self.bandpass_array.append(1.)
+
+        self.freq_array = np.array(self.freq_array)
+        self.bandpass_array = np.array(self.bandpass_array)
+        sort_inds = np.argsort(self.freq_array)
+
+        self.data_array[0, 0, 0, :, :] = self.data_array[0, 0, 0, sort_inds, :]
+        self.data_array[0, 0, 1, :, :] = self.data_array[0, 0, 1, sort_inds, :]
```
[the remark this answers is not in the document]
bhazelton: yes, thanks!

pyuvdata/cst_beam.py
```diff
@@ -0,0 +1,223 @@
...
+        contains a substring with the frequency channel in MHz that the data represents.
+        e.g. "HERA_Sim_120.87MHz.txt" should yield 120.87e6
+
+        Args:
+            fname: filename (string)
+
+        Returns:
+            extracted frequency
+        """
+        fi = fname.find('MHz')
```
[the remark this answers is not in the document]
bhazelton: yes, I'll make it more flexible. I'm sort of hoping we can come up with a better way to pass various metadata from CST simulations in, but for now this is what we have...

pyuvdata/tests/test_cst_beam.py
```diff
@@ -0,0 +1,87 @@
...
+    beam1.read_cst_beam(cst_files, beam_type='power', telescope_name='TEST', feed_name='bob',
+                        feed_version='0.1', model_name='E-field pattern - Rigging height 4.9m',
+                        model_version='1.0')
+
+    beam2.read_cst_beam(cst_files, beam_type='power', frequencies=[150e6, 123e6], telescope_name='TEST',
+                        feed_name='bob', feed_version='0.1',
+                        model_name='E-field pattern - Rigging height 4.9m',
+                        model_version='1.0')
+
+    nt.assert_equal(beam1, beam2)
```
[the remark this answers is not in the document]
bhazelton: yes, good idea.

pyuvdata/uvbeam.py
```diff
@@ -410,6 +412,71 @@ def set_phased_array(self):
...
+        if self.pixel_coordinate_system != 'az_za':
+            raise ValueError('pixel_coordinate_system must be "az_za"')
+        if self.beam_type != 'power':
+            raise ValueError('healpix conversion not yet defined for efield')
+
+        phi_vals, theta_vals = np.meshgrid(self.axis1_array, self.axis2_array)
+
+        min_res = np.min(np.array([np.diff(self.axis1_array)[0], np.diff(self.axis2_array)[0]]))
+        nside_min_res = np.sqrt(3 / np.pi) * np.radians(60.) / min_res
+        nside = int(2**np.floor(np.log2(nside_min_res)))
```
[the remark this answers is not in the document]
bhazelton: I guess it depends on the goal and primary concern about error introduction. I wrote this quickly, but I was concerned about introducing interpolation errors if we oversample. Maybe there needs to be some options controlling this?

pyuvdata/uvbeam.py
bhazelton: thanks!

pyuvdata/uvbeam.py
bhazelton: thanks!

pyuvdata/uvbeam.py
bhazelton: thank you for catching that.

pyuvdata/uvbeam.py
```diff
@@ -410,6 +412,74 @@ def set_phased_array(self):
...
+        self.set_cs_params()
+
+        healpix_data = np.zeros(self._data_array.expected_shape(self), dtype=np.float)
+        data_pixels = hp.ang2pix(nside, theta_vals, phi_vals)
+        pixels = np.arange(npix)
+        hits = np.zeros(self._data_array.expected_shape(self))
+
+        for index in range(self.Naxes1 * self.Naxes2):
+            healpix_data[:, :, :, :, data_pixels[index]] += az_za_data[:, :, :, :, index]
+            hits[:, :, :, :, data_pixels[index]] += 1
```
adampbeardsley: Hmm, now I'm not sure about this interpolation method. When we were undersampling, each covered healpix pixel would get a hit. But now that we're oversampling, we are missing healpix pixels between az_za pixels. For example, about 15% of the pixels get missed when converting Nic's 150MHz beam. If this is desired, that's fine, but it seems to me like we would want to fill those in with interpolation.

bhazelton: yeah, I was worried about that too. Let me take a look at what Ian does in FHD.

adampbeardsley: fwiw, because we require az_za are on a regular grid, you can use some pretty efficient interpolation routines. I have some code here which did this using RectBivariateSpline:
https://github.com/HERA-Team/HERA-Beams/blob/master/NicolasFagnoniBeams/convert_to_healpix.py

bhazelton: So I'm worried about a couple of edge cases. I think actually that the regular grid requirement should not be enforced on the object, but only if it's written to a regularly gridded FITS file. So I think those errors should actually be warnings. Also, it's definitely not required to cover the sphere and if it doesn't, the healpix pixels outside the covered area should just be excluded.

bhazelton: Hmm.. I'm wrong about the regular grid -- that is baked into the object. I was confused because we had talked about doing something else, but we ultimately decided that was a reasonable requirement. I'm still worried about it not covering the sphere though.

PR #240, opened 2017-09-05
Fix backwards compatibility for miriad diameter keyword

plaplant:
When doing PR #239, the `diameter` uv_var should have been removed from the `extra_keywords` dictionary, to prevent it from being propagated. This has been resolved, and a test has been added.


pyuvdata/miriad.py
```diff
@@ -422,6 +422,8 @@ def read_miriad(self, filepath, correct_lat_lon=True, run_check=True,
             # backwards compatibility for when keyword was 'diameter'
             try:
                 self.antenna_diameters = uv['diameter']
+                # if we find it, we need to remove it from extra_keywords to keep from writing it out
+                self.extra_keywords.pop('diameter')
```
adampbeardsley: I'm still a little unsure about how the diameter gets into extra_keywords, but then gets promoted to `self.antenna_diameter`. Should this `pop` exist for the `antdiam` case as well?

Issue #254, opened 2017-10-19
Summing uvfits files doesn't produce expected results


zakiali:
I am trying to comgine uvfits files (omnical mdl files) together by summing the UVData objects together (code below). The result doesn't make sense to me and it looks like only the first file is being preserved. I can however individually read files and concatenate individual baseline data together in a new array to produce correct results.

Code:
`init_mdl = pyuvdata.UVData()`
`init_mdl.read_uvfits(mdl_data_file[0])`
`init_mdl.unphase_to_drift()`
`for mdl_file in mdl_data_file[1:]:`
`      uvmdl = pyuvdata.UVData()`
`      uvmdl.read_uvfits(mdl_file)`
`      uvmdl.unphase_to_drift()`
`      init_mdl += uvmdl`

![image](https://user-images.githubusercontent.com/940272/31796059-fce771de-b4dc-11e7-910a-cbda8d1124ce.png)

adampbeardsley:
What is an mdl file?
Presumably you're plotting an amplitude waterfall? Can you show what each file looks like individually, and the code for producing the plot?
Are these files anywhere that I can access to reproduce the problem?

zakiali:
I meant 'vis.uvfits' files (the model visibilities for a redundant baseline). 
That was actually the phase. You see the same thing in amp too.
You can find the data files at the path in the glob on lustre at nrao. Let me know if you have issues accessing that and I can put them somewhere else. 



```mdl_data_files = glob.glob('/lustre/aoc/projects/hera/zali/HERA2015/2457702/IDR3/zen.2457702.4[5-6]*.HH.uvc.vis.uvfits')
concat_array = []

init_mdl = pyuvdata.UVData()
init_mdl.read_uvfits(mdl_data_files[0])
init_mdl.unphase_to_drift()

concat_array = init_mdl.get_data(9, 20, 'xx')

for mdl_file in mdl_data_files[1:]:
    uvmdl = pyuvdata.UVData()
    uvmdl.read_uvfits(mdl_file)
    uvmdl.unphase_to_drift()
    
    # concatenate
    concat_array = np.concatenate([concat_array, uvmdl.get_data(9, 20, 'xx')])
    init_mdl += uvmdl


plt.figure()
plt.subplot(221)
uvtools.plot.waterfall(init_mdl.get_data(9,20,'xx'), mode='phs', cmap='inferno')
plt.xlabel('freq bin')
plt.ylabel('integration number')
plt.title('phs')
[...]

PR #264, opened 2017-11-04
Fix reordering in add function

adampbeardsley:
If the data weren't already ordered by time, ant2, ant1, adding objects together resulted in a scramble. This addresses issue #254.


adampbeardsley:
Whoops, I thought I added tests but I guess not. I'll write some up quick.

PR #270, opened 2017-11-15
added get_ENU_from_UVData to utils.py w/ tests

nkern:
simple script to return antennas positions in ENU (topocentric) coordinates given a UVData instance.


bhazelton:
This is nice. It seems to me that it should be a method on the UVData object since it provides functionality exclusively for that class. And you might consider adding something to the tutorial :)

nkern:
@bhazelton let me know if you want anything else for this PR

PR #272, opened 2017-11-28
pep8 fixes

bhazelton:
@nkern There's still a bare except in your pyuvdata_inspect script, line 62 that I wasn't sure how to fix. Can you fix it and push it to this branch?

### DIFF
--- a/pyuvdata/tests/test_utils.py
+++ b/pyuvdata/tests/test_utils.py
@@ -182,5 +182,3 @@ def test_jones_num_funcs():
     nt.assert_raises(KeyError, uvutils.jstr2num, 'foo')
     nt.assert_raises(ValueError, uvutils.jstr2num, 1)
     nt.assert_raises(ValueError, uvutils.jnum2str, 7.3)
-
-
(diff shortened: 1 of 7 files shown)


nkern:
@bhazelton its kinda important for the try: except: statement to be there, so I just made a trivial change to make the except hold something, instead of a pass statement. let me know if you were looking for a different kind of fix.

bhazelton:
@nkern I'm sorry, what I wanted you to fix was that the except is bare -- it doesn't specify which error types it is supposed to handle (https://docs.python.org/2/howto/doanddont.html#except). I have no problem with the pass.

nkern:
@bhazelton ah I see. the `except` is bare because each of the `uvd.read_{}` functions raise different kinds of exceptions... so to catch them all I just used a blanket `except` statement

bhazelton:
@nkern You can list multiple exception types in the except statement, it's fine to handle many different ones, but leaving it bare allows it to match exception types you don't expect, as explained in the link I posted. That's why the pep8 standard opposes bare excepts.

I can add in the exception types myself, but I was just looking at the code a bit to figure out what they are, and I am confused by ob_names having 'UVFITS' -- I think it should just have UVData, UVBeam and UVCal, am I missing something?

nkern:
@bhazelton yeah UVFITS can be excluded. ok i updated the except with all of the exceptions the three objects raised when trying to read the wrong files. let me know if you need anything else!

bhazelton:
@nkern Perfect! Thank you!!

bhazelton:
@adampbeardsley can you take a look at this one too? It's just pep8 fixes.

PR #275, opened 2017-12-01
Add a note to the history when converting delays to gains

bhazelton:


bhazelton:
@adampbeardsley can you check this very small PR?

PR #276, opened 2017-12-07
Add pep8 checking to Travis

bhazelton:
To prevent pep8 problems from being missed in PRs.

This ignores the line-length check (code E501).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,11 +27,7 @@ install:
 
   # create environment and install dependencies
   - conda install numpy scipy astropy nose pip
-  - conda install -c conda-forge healpy
-  - conda install -c conda-forge aipy
-  - conda install -c conda-forge python-casacore
-  - conda install -c conda-forge pycodestyle
-  - pip install coveralls
+  - conda install -c conda-forge healpy aipy python-casacore pycodestyle coveralls
   - conda list
 script:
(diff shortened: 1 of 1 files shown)


bhazelton:
@adampbeardsley @dannyjacobs @jpober I'm of two minds about whether to push for this PR to be merged or not. Partly I wanted to figure out if I could do it and I had just spent some time cleaning up some stuff that had been merged in previous PRs. What do you all think?

jpober:
I don't have a strong opinion.  We've decided the code should be pep8 compliant, but maybe it shouldn't come at the expense of an increased bar for participation in the project.  That said, it's not that onerous to get right either.

dannyjacobs:
But how does it actually work? Does the Travis build just fail if all lights are not green? Or is there a service like Coveralls that tracks our pep8-edness?

bhazelton:
Travis fails on any pep8 miss, but if you look at the Travis log it lists all the problematic line numbers. The fact that this PR is passing Travis means we don't have any problems currently.

adampbeardsley:
I can also go either way. Definitely useful, and maybe we need a little extra push to get people to stay compliant. Typically @bhazelton is the hero that goes through and fixes things, so if this makes her life easier, I think we should implement it.

bhazelton:
Ok, I'm going to merge it. If we find it annoying we can always take it out.

Issue #198, opened 2017-07-03
UVBeam test suite should be based on actual data files


bhazelton:
As opposed to fake data. This can be fixed after the read_cst method is implemented.

aelanman:
Based on the Zac's code in polskysim, I've written a script that reads in the Nicolas Fagnoni simulated E-field patterns for HERA and fills a UVBeam object. I'm not sure how read_cst is supposed to work, but I do have a hera beamfits file now. If someone else would be willing to check it, could that serve for tests?

bhazelton:
This is fixed except for e-field healpix beams. Once conversion of e-field beams to healpix works we can finish this off.

PR #277, opened 2017-12-15
base beam testing off actual files to address issue #198

bhazelton:
This doesn't fully do away with faked data because we don't yet have a way to get efield beams into healpix. But it significantly helps to address issue #198


bhazelton:
While working on this branch, I found that the name of my branch (specifically having the word testing in my branch name) caused some tests on the way histories are combined in the various add functions to fail. I fixed that logic and moved it into a utility because it was duplicated across all three classes. So there are some changes to files other than the beam testing files as a result.

PR #285, opened 2018-01-26
Handle new numpy/scipy deprecation warnings properly

bhazelton:


bhazelton:
Some numpy print formatting also changed apparently, which required tutorial changes to pass doctests.

PR #290, opened 2018-01-31
Add parameters to UVCal and calfits to support sky-based calibration

### LINKED ISSUE #289: calfits flexibility for sky-based cal
We need to incorporate some header keyword flexibility that will help as we start including sky-based calibrations (e.g. FHD). Here are some suggestions.

New Required keyword: 
cal_style. Two options I can think of would be (redundant, sky).
Required keywords given the cal_style : 
sky_field (required if cal_style is sky): short string describing field center or dominant source.
sky_catalog (required if cal_style is sky): catalog that was used
ref_antenna_name (required if cal_style is sky): name of antenna that was used as the phase reference antenna
Optional keywords:
n_sources (optional if cal_style is sky): number of sources included in the calibration model
cal_baseline_range (optional if cal_style is sky): range in wavelengths of baselines included in calibration.
diffuse_model (optional if cal_style is sky): short string describing calibration diffuse model used.

bhazelton:
This fixes #289

### DIFF
--- a/pyuvdata/calfits.py
+++ b/pyuvdata/calfits.py
@@ -22,6 +22,13 @@ def _warn_olddelay(filename):
                   'future compatibility.'.format(file=filename))
 
 
+def _warn_oldstyle(filename):
+    warnings.warn('{file} appears to be an old calfits format '
+                  'which has been depricated. '
+                  'Rewrite this file with write_calfits to ensure '
+                  'future compatibility.'.format(file=filename))
+
+
 class CALFITS(UVCal):
     """
     Defines a calfits-specific class for reading and writing calfits files.
@@ -100,6 +107,19 @@ class CALFITS(UVCal):
         prihdr['TELESCOP'] = self.telescope_name
         prihdr['GNCONVEN'] = self.gain_convention
         prihdr['CALTYPE'] = self.cal_type
+        prihdr['CALSTYLE'] = self.cal_style
+        if self.sky_field is not None:
+            prihdr['FIELD'] = self.sky_field
+        if self.sky_catalog is not None:
+            prihdr['CATALOG'] = self.sky_catalog
+        if self.ref_antenna_name is not None:
+            prihdr['REFANT'] = self.ref_antenna_name
+        if self.Nsources is not None:
+            prihdr['NSOURCES'] = self.Nsources
+        if self.baseline_range is not None:
+            prihdr['BL_RANGE'] = '[' + ', '.join(self.baseline_range) + ']'
+        if self.diffuse_model is not None:
+            prihdr['DIFFUSE'] = self.diffuse_model
         prihdr['INTTIME'] = self.integration_time
         prihdr['CHWIDTH'] = self.channel_width
         prihdr['XORIENT'] = self.x_orientation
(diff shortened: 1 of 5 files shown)


pyuvdata/calfits.py
```diff
@@ -109,11 +129,6 @@ def write_calfits(self, filename, run_check=True, check_extra=True,
             prihdr['FRQRANGE'] = ','.join(map(str, self.freq_range))
         prihdr['TMERANGE'] = ','.join(map(str, self.time_range))
 
-        for p in self.extra():
```
plaplant: So does this mean that we're not supporting arbitrary extra parameters anymore? I think that's okay, just wondering if this was added somewhere else and I missed it.

pyuvdata/calfits.py
```diff
@@ -413,7 +440,7 @@ def read_calfits(self, filename, run_check=True, check_extra=True,
             self.Nspws = hdr.pop('NAXIS5')
             # add this for backwards compatibility when the spw CRVAL wasn't recorded
             try:
-                spw_array = uvutils.fits_gethduaxis(F[0], 5, strict_fits=strict_fits) - 1
+                spw_array = uvutils.fits_gethduaxis(F[0], 5, strict_fits=strict_fits)
```
plaplant: Good catch!

pyuvdata/uvcal.py
```diff
@@ -190,6 +196,41 @@ def __init__(self):
                                            description=desc, form=(2,),
                                            expected_type=float, tols=1e-3)
 
+        # --- cal_style parameters ---
+        desc = ('Style of calibration. Values are sky or redundant.')
```
plaplant: This is already defined above. Should it be defined again here?

pyuvdata/uvcal.py
```diff
@@ -250,6 +291,11 @@ def check(self, check_extra=True, run_check_acceptability=True):
             run_check_acceptability: Option to check if values in required parameters
                 are acceptable. Default is True.
         """
+        # if make sure requirements are set properly for cal_style
```
plaplant: Small typo in the comment

pyuvdata/calfits.py
bhazelton: We are, but they go in extra_keywords. This part of the code was apparently to catch actual parameters (meaning ones that are defined explicitly on the object) that were not required to be set and that we didn't want to bother adding to the header explicitly. The problem is that if they need any special handling (like baseline_range) then you have to do them explicitly and then they get caught here as well, which is bad. Also, the keyword in the FITS file gets set by simple truncation which tends to give them somewhat confusing or unclear names and can lead to redundant names. I think this was a bit lazy and caused lots of issues.

pyuvdata/calfits.py
bhazelton: Yeah, I'm not sure how that made it through before, but I had a bug that I was hunting and I found that one (it didn't fix my bug, but I found it somewhere else).

pyuvdata/uvcal.py
bhazelton: oops! thank you! I'll remove one of them

pyuvdata/uvcal.py
bhazelton: thanks!

pyuvdata/uvcal.py
bhazelton: I took out the upper one.

Issue #105, opened 2016-12-14
read FHD cals


dannyjacobs:

bhazelton:
This was done in PR #291

wenyang-li:
If the keyword 'raw' is set to True when reads in FHD cal say files, it does not really read in the raw solution. In fhd_cal.py, res_gain_array_in is not actually used after line 130.

Issue #36, opened 2016-05-26
Improve reading of history from FHD settings files


bhazelton:
It currently just grabs the first 11 lines, should be more programmatic and should get info relevant to visibilities (not images).

bhazelton:
This was fixed in PR #291

PR #291, opened 2018-01-31
Add ability to read FHD calibrations

bhazelton:
This fixes issue #105 
This fixes issue #36 
This fixes issue #292


pyuvdata/fhd_cal.py
```diff
@@ -0,0 +1,195 @@
...
+        self.telescope_name = obs_data['instrument'][0]
+
+        self.Nants_data = int(cal_data['n_tile'][0])
+        self.Nants_telescope = int(cal_data['n_tile'][0])
+        self.antenna_names = cal_data['tile_names'][0].tolist()
+        self.antenna_numbers = np.arange(self.Nants_telescope)
+        self.ant_array = np.arange(self.Nants_data)
+
+        self.set_sky()
+        self.sky_field = 'phase center (RA, Dec): ({ra}, {dec})'.format(
```
plaplant: I'm not familiar with FHD, so forgive the naive question. Are these dictionary keys "standard" keywords that appear in all files? Put another way, is it possible for us to get a KeyError here? Should we bother putting this in a try/except block?

pyuvdata/fhd_cal.py
```diff
@@ -0,0 +1,195 @@
...
+        freq_use = obs_data['baseline_info'][0]['freq_use'][0]
+        # array of used antennas (1: used, 0: flagged)
+        ant_use = obs_data['baseline_info'][0]['tile_use'][0]
+        # array of used times (1: used, 0: flagged)
+        time_use = obs_data['baseline_info'][0]['time_use'][0]
+
+        time_array_use = time_array[np.where(time_use > 0)]
+        self.time_range = [np.min(time_array_use), np.max(time_array_use)]
+
+        # Currently this can't include the times because the flag array
```
plaplant: Should we make the format more flexible to allow for different-sized flag arrays? I'm not sure it's worth the effort since the solution here seems adequate, but if we think it is, we should make an issue.

pyuvdata/tests/test_fhd_cal.py
```diff
@@ -0,0 +1,207 @@
...
+    equality.
+    """
+    fhd_cal = UVCal()
+    calfits_cal = UVCal()
+    if uvtest.pre_1_14_numpy:
+        fhd_cal.read_fhd_cal(cal_testfile, obs_testfile, settings_file=settings_testfile)
+    else:
+        # numpy 1.14 introduced a new deprecation warning
+        n_scipy_warnings, scipy_warn_list, scipy_category_list = \
+            uvtest.get_scipy_warnings(n_scipy_warnings=605)
```
plaplant: Is the quantity `n_scipy_warnings` empirically determined here? If so, it might be useful to note that, in case future versions have a different set of deprecation warnings...

scripts/truncate_fhd_files.pro
```diff
@@ -1,61 +1,114 @@
-pro truncate_fhd_files
```
plaplant: I don't speak IDL, so I'm just assuming this is all okay :)

pyuvdata/fhd.py
```diff
@@ -8,6 +8,39 @@
 import utils as uvutils
 
 
+def get_fhd_history(settings_file, return_user=False):
```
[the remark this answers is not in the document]
bhazelton: yes, it should. Thanks!

pyuvdata/fhd_cal.py
bhazelton: My understanding is that the fields will always be present, but that some of the fields can be Nones or empty strings. I'm iterating with Ian to figure that out and I'll put in the appropriate checks. Thanks!

pyuvdata/fhd_cal.py
bhazelton: I'm of two minds on this one. The issue is that FHD has flag information per integration that we can't include because FHD generates a single calibration across all the times and the flag array is constrained to have the same shape as the calibration array.

I think that constraint makes some sense when you think about applying the calibration to data, but if you're trying to understand how the calibration was calculated you'd need all the flag info. Of course, there can also be visibility level flags that are already being lost and would affect the calibration solution, so maybe we're already in the regime where you don't have all the information.

pyuvdata/tests/test_fhd_cal.py
bhazelton: Yes it's totally empirical :(. It looks to me from Googling that it'll go away with the next scipy version (it's an issue between numpy and scipy). I'll add some comments.

scripts/truncate_fhd_files.pro
bhazelton: I tweaked this until I got the test files I wanted. I included the changes for posterity :)

pyuvdata/fhd_cal.py
bhazelton: Ok, I checked with Ian and it sounds like we're safe here.

PR #295, opened 2018-02-08
Add a Beamfits memo to describe the beam fits files

bhazelton:


docs/references/beamfits_memo.tex
```diff
@@ -0,0 +1,154 @@
...
+\item{\textbf{NORMSTD}: should be set to one of `physical', `peak' or `solid\_angle'. (data\_normalization)}
+\item{\textbf{COORDSYS}: should be set to one of `az\_za', `sin\_zenith' or `healpix'. (pixel\_coordinate\_system)}
+\item{\textbf{TELESCOP}: Telescope name. (telescope\_name)}
+\item{\textbf{FEED}: Feed name (used to distinguish between different physical feed types on the same telescope). (feed\_name)}
+\item{\textbf{FEEDVER}: Feed version (used to distinguish between different versions of the same feed on the same telescope). (feed\_version)}
+\item{\textbf{MODEL}: Model name  (used to distinguish between different models of the same hardware). (model\_name)}
+\item{\textbf{MODELVER}: Model version (used to distinguish between different versions of the same models of the same hardware). (model\_version)}
+\item{\textbf{FEEDLIST}: Only for E-field beams. The list of feeds orientations represented in the data. (feed\_array)}
+\item{\textbf{NSIDE}: Only for HEALPix beams. The NSIDE of the HEALPix map. (nside)}
+\item{\textbf{ORDERING}: Only for HEALPix beams. The ordering parameter of the HEALPix map.(ordering)}
```
isullivan: Since only NESTED or RING are allowed, it might be worthwhile to just list them here.

docs/references/beamfits_memo.tex
```diff
@@ -0,0 +1,154 @@
...
+\section{Primary HDU}
+The primary HDU of \textit{beamfits} files is an Image HDU containing the beam model.
+
+\subsection{Primary Header}
+The following are required keywords in the primary header of a  \textit{beamfits} file. For more detailed explanations 
+of what these keywords mean, see the descriptions on pyuvdata's ReadTheDocs uvbeam\_parameters page. 
+The UVBeam parameter corresponding to each keyword is noted in parentheses.
+As with all FITS files, HISTORY and COMMENT cards are allowed.
+
+\begin{itemize}
```
isullivan: I think it would be helpful to indicate when a field requires a specific type, such as a string or an integer.

docs/references/beamfits_memo.tex
```diff
@@ -0,0 +1,154 @@
...
+\item{\textbf{ORDERING}: Only for HEALPix beams. The ordering parameter of the HEALPix map.(ordering)}
+\end{itemize}
+
+\subsection{Beam Data}
+The axes of the data array in the primary HDU depends on the pixelization scheme.
+
+\subsubsection{HEALPix beam axes}
+For HEALPix beams, the data axes are: 
+\begin{enumerate}
+\item{\textbf{PIX\_IND}: index into the HEALPix pixel array (in the HPX\_INDS extension) This is needed because partial sky maps are supported so pixels may be missing.}
```
isullivan: Is this required for whole sky maps?

docs/references/beamfits_memo.tex
```diff
@@ -0,0 +1,154 @@
...
+\item{\textbf{FREQ}: frequency}
```
isullivan: Is frequency assumed to be Hz?

docs/references/beamfits_memo.tex
```diff
@@ -0,0 +1,154 @@
...
+\item{\textbf{IF}: Spectral window number, currently only one spectral window is supported so this can only have length 1.}
+\item{\textbf{VECIND}: Basis vector index, the index into the dimensions of the basis vector array (stored in the BASISVEC extension). This is usually length 1 for power beams and length 2 for E-field beams.}
+\item{\textbf{COMPLEX}: real and imaginary (axis length 2)}
+\end{enumerate}
+
+\subsubsection{regularly pixelized beam axes}
+For regularly pixelized beams, the data axes are: 
+\begin{enumerate}
+\item{\textbf{First Image Axis (AZ or SIN\_X)}: Coordinates for the first image axis. The axis name is dictated by the COORDSYS keyword.}
+\item{\textbf{Second Image Axis (ZA or SIN\_Y)}: Coordinates for the second image axis. The axis name is dictated by the COORDSYS keyword.}
```
isullivan: Given the similarity between "AZ" and "ZA", it might be helpful to spell out Azimuth and Zenith Angle somewhere.

isullivan: Also, I think you need to have map projection parameters in here somewhere for the regularly gridded case, or at least the pixel scale.

docs/references/beamfits_memo.tex
```diff
@@ -0,0 +1,154 @@
...
+\begin{itemize}
+\item{\textbf{HPX\_INDS}: The HEALPix pixel numbers in the order that they are stored in the beam and basis vector arrays.}
+\end{itemize}
+
+\section{BANDPARM HDU}
+The BANDPARM HDU is an required element of \textit{beamfits} files. It is a Binary Table HDU containing the bandpass and other optional parameters that are functions of frequency only.
+The \textbf{EXTNAME} keyword in the header must be set to BANDPARM. There are no other required keywords in the header, but there are 2 optional header keywords:
+
+\begin{itemize}
+\item{\textbf{REFZIN}: Reference input impedance of the receiving chain (sets the reference for the S parameters), units: Ohms}
```
isullivan: Please provide a note or a link to a description of the S parameters.

pyuvdata/uvbeam.py
```diff
@@ -19,13 +19,15 @@ class UVBeam(UVBase):
...
     coordinate_system_dict = {
-        # uniformly gridded az/za coordinate system, az runs from East to North in radians
-        'az_za': ['az', 'za'],
-        # sine projection at zenith. y points North, x point East
-        'sin_zenith': ['sin_x', 'sin_y'],
-        # HEALPix map with zenith at north pole and
-        # az, za coordinate axes (for efield) where az runs from East to North.
-        'healpix': ['hpx_inds']}
+        'az_za': {'axes': ['az', 'za'],
+                  'description': 'uniformly gridded az/za coordinate system, '
```
isullivan: Please spell out "azimuth" and "zenith angle" in the description.

docs/references/beamfits_memo.tex
bhazelton: yes. I'll make that more explicit.

docs/references/beamfits_memo.tex
bhazelton: The pixel scale is in (i.e. the corresponding CDELTn keyword is set). I'm not super familiar with the map projection parameters, I can see needing them for the sin_zenith projection but do you need them for the azimuth, zenith angle case?

pyuvdata/uvbeam.py
bhazelton: done

docs/references/beamfits_memo.tex
bhazelton: Ok, I've added it to the docs as well as the tex.

docs/references/beamfits_memo.tex
isullivan: I think it's fine as it is, but I believe it is a departure from standard Healpix Fits format. It may seem obvious, but I'd recommend noting explicitly in the comment that it is required even for full sky maps (not just that it's needed for partial sky maps).

bhazelton: ok, will do.

docs/references/beamfits_memo.tex
isullivan: I think I misunderstood what this document would look like when finished. I see `CDELT` in `beamfits.py` but not in this file. Will all of those parameters be automatically imported?

bhazelton: I just went off and read some WCS stuff and I'm still not totally sure I know what to add, beyond the CDELT keywords. I'm still making my way through the 3 papers, but do you have some specific keywords in mind?

bhazelton: I added some language to the tex about all the axes having the standard `CRVALn', `CRPIXn', and `CDELTn' keywords and sometimes having the `CUNITn' keyword for units. I consider that those are required to specify any FITs axis, so I wasn't spelling them out here, I was just explaining what the axes were. Is that too opaque?

bhazelton: I also just realized that the `CUNITn` keyword wasn't being set for the image axes so I added that to the code.

bhazelton:
I just discovered (reading the WCS papers) that angles should always be in degrees for FITs files. I'm going to have to change some code to fix this, we're currently using radians throughout the beamfits code.

bhazelton:
Ok, I've fixed the units to use degrees not radians per FITs standards. I also realized I was erroneously writing empty BASISVEC HDUs for power beams that didn't have basis vector arrays. I fixed that and updated the LaTeX to reflect the fact that that HDU is only required for E-field beams (but allowed for power beams).

bhazelton:
@isullivan Ok, I think I've substantially improved the memo and code thanks to your comments. When you have time, can you give it another read and see if there are any more things to fix? I'll email you new versions of the files because there have been some improvements in this process.

bhazelton:
Thanks @isullivan !

docs/references/beamfits_memo.tex
```diff
@@ -0,0 +1,160 @@
...
+\item{\textbf{ORDERING}: \textit{string} Only for HEALPix beams. The ordering parameter of the HEALPix map, allowed values are `nested' and `ring'. (ordering)}
+\end{itemize}
+
+\subsection{Beam Data}
+The axes of the data array in the primary HDU depends on the pixelization scheme.
+
+\subsubsection{HEALPix beam axes}
+For HEALPix beams, the data axes (values of the CTYPE\textit{m} keywords in order) are: 
+\begin{enumerate}
+\item{\textbf{PIX\_IND}: index into the HEALPix pixel array (in the HPX\_INDS extension) This is needed because partial sky maps are supported so pixels may be missing (full sky maps are not handled differently, they have the same required elements as partial maps). The corresponding CRVAL\textit{m}, CRPIX\textit{m}, and CDELT\textit{m} values should all be 1. (pixel\_array)}
```
adampbeardsley: A question because I'm not sure - does the HEALPix definition of these pixels fully constrain the coordinate system? For example, I know there are ways of going from pixel index to "theta" and "phi" angles, but are those directly related to za and az?

docs/references/beamfits_memo.tex
```diff
@@ -0,0 +1,160 @@
...
+\item{\textbf{COMPLEX}: real and imaginary (axis length 2), the first entry on this axis contains the real part of the complex beam and the second contains the corresponding imaginary component. The corresponding CRVAL\textit{m}, CRPIX\textit{m}, and CDELT\textit{m} keywords should all have the value 1.}
+\end{enumerate}
+
+\subsubsection{regularly pixelized beam axes}
+For regularly pixelized beams, the data axes (values of the CTYPE\textit{m} keywords in order) are: 
+
+\begin{enumerate}
+\item{\textbf{AZIMUTH or ZENX-SIN (First Image Axis)}: The coordinates for the first image axis, either azimuth angle or the East/West axis of a zenith orthoslant projection, should match the coordinate system given by the COORDSYS keyword. The values of the coordinates are set by the CRVAL\textit{m}, CRPIX\textit{m}, and CDELT\textit{m} keywords and the CUNIT\textit{m} keyword gives the units (default is degrees).(axis1\_array)}
```
adampbeardsley: Similar to above - I always need to look it up, but is azimuth strictly defined as toward east from north?

docs/references/beamfits_memo.tex
```diff
@@ -0,0 +1,160 @@
...
+\item{\textbf{IF}: The IF axis enumerates spectral windows (frequency bands). The corresponding CRVAL\textit{m}, CRPIX\textit{m}, and CDELT\textit{m} keywords should all have the value 1. Currently only one spectral window is supported so this can only have length 1. (spw\_array)}
+\item{\textbf{VECIND}: Basis vector index, the index into the dimensions of the basis vector array (stored in the BASISVEC extension). This is usually length 1 for power beams and length 2 for E-field beams. The corresponding CRVAL\textit{m}, CRPIX\textit{m}, and CDELT\textit{m} keywords should all have the value 1.}
+\item{\textbf{COMPLEX}: real and imaginary (axis length 2), the first entry on this axis contains the real part of the complex beam and the second contains the corresponding imaginary component. The corresponding CRVAL\textit{m}, CRPIX\textit{m}, and CDELT\textit{m} keywords should all have the value 1.}
```
adampbeardsley: This corresponds to `data_array` right?

docs/references/beamfits_memo.tex
```diff
@@ -0,0 +1,160 @@
...
+\item{\textbf{STOKES or FEEDIND (Polarization axis)}: STOKES for power beams give the polarization integer convention from uvfits (see AIPS memo 117), FEEDIND for E-field beams give the index into the FEEDLIST. The corresponding CRPIX\textit{m} values should be 1 for both and for FEEDIND the CRVAL\textit{m} and CDELT\textit{m} keywords should 1. (polarization\_array or feed\_array)}
+\item{\textbf{IF}: The IF axis enumerates spectral windows (frequency bands). The corresponding CRVAL\textit{m}, CRPIX\textit{m}, and CDELT\textit{m} keywords should all have the value 1. Currently only one spectral window is supported so this can only have length 1. (spw\_array)}
+\item{\textbf{VECIND}: Basis vector index, the index into the dimensions of the basis vector array (stored in the BASISVEC extension). This is usually length 1 for power beams and length 2 for E-field beams. The corresponding CRVAL\textit{m}, CRPIX\textit{m}, and CDELT\textit{m} keywords should all have the value 1.}
```
adampbeardsley: Does this have a corresponding uvbeam parameter, or is it implied by the data structure when it's in the array?

docs/references/beamfits_memo.tex
bhazelton: Healpix is a mapping of the sphere in a generic theta, phi space. Usually theta, phi map to RA, Dec in sky maps, but for beams the sensible thing is for them to map to az, za. This is specified in the description of the `pixel_coordinate_system` in the docs, but I can add some more language to this memo as well.

docs/references/beamfits_memo.tex
bhazelton: It can be defined differently in different contexts. I was careful to pick a specific definition here so that it was defined. I think this is the most common choice (I don't remember exactly how I came to this choice, but I remember looking at several different references).

docs/references/beamfits_memo.tex
bhazelton: All of these axes are axes in the data array except this one, actually, because the data array is complex but FITs files make you split out the real and imaginary along another axis.

docs/references/beamfits_memo.tex
bhazelton: It's more implied, basically you're specifying 2 (or 3) basis vectors (directions) for which the data array gives beam values. This is just the index along that direction.

docs/references/beamfits_memo.tex
adampbeardsley: Oh, I see. I misunderstood what was actually in here.

docs/references/beamfits_memo.tex
adampbeardsley: Gotcha, that sort of what I thought.

docs/references/beamfits_memo.tex
adampbeardsley: Ah, I was looking at old docs. This and the next question are already addressed.

Issue #296, opened 2018-02-08
imaging files concatenated by `pyuvdata` rotates image?


nkern:
When I image a 10-minute HERA-47 Miriad file, my images appear as I expect. If, however, I concatenate a series of 10-minute Miriad files using `pyuvdata` and image that file, my images appear slightly rotated (see below). This rotation gets worse the more files I concatenate, leading me to think it is perhaps a phasing issue? One thing that happens in the process of getting a Miriad file to a CASA MS is converting to `uvfits` which phases the whole dataset to the first time integration.

As I said, I'm not sure if this is a `pyuvdata` problem or a problem in my imaging scripts, so if someone can try to duplicate this with their own imaging scripts, that would be appreciated.

![gleam02_1file_concat](https://user-images.githubusercontent.com/4307358/35997600-d2aaa554-0cce-11e8-8570-fc1cfc69750d.png)
^ Image of a GLEAM 02h source on a single 10-minute Miriad file

![gleam02_6file_concat](https://user-images.githubusercontent.com/4307358/35997541-a21cd542-0cce-11e8-9631-f7cf3aa4c834.png)
^ Image of the same source having concatenated 6 time-contiguous Miriad files w/ `pyuvdata`

bhazelton:
How did you phase the files?

nkern:
yeah, good point.

This rotation gets worse the more files I concatenate, leading me to think it is perhaps a phasing issue? One thing that happens in the process of getting a Miriad file to a CASA MS is converting to uvfits which phases the whole dataset to the first time integration.

nkern:
I do the Miriad -> uvfits via the `pyuvdata.miriad_to_uvfits.py` script

bhazelton:
yes, but you can phase the files to a specific location if you'd like, rather than just letting it default to the first time sample

nkern:
well, in CASA, I do a `fixvis`, which should phase the data to the location I want. is there some discrepancy between the phasing algorithm in `pyuvdata` and CASA?

bhazelton:
So after reading in the miriad files, you can use the `phase_to_time` or `phase` methods before calling write_uvfits.

It is possible that there is a discrepancy. CASA is known to have some phasing issues. pyuvdata is not perfect either, partly because we have struggled to find reference files to test against, but we believe it is good to better than an arcsecond. See issue #148

nkern:
ok, i'll try that and see if it fixes the problem and let you know

[...]

PR #297, opened 2018-02-08
updated miriad_to_uvfits.py

nkern:
to use argparse instead of optparse, and to optionally phase data to a specified time


nkern:
helps address some concern in #296

PR #300, opened 2018-02-20
Fix labelling of SPW axis in calfits files to be consistent with uvfits format

bhazelton:


bhazelton:
@plaplant can you look at this one? All the tests pass, but I don't want to mess up users of dependent repos like hera_cal.

PR #302, opened 2018-02-23
Add a script to renumber antennas to below 255 in uvfits files

bhazelton:


scripts/renumber_ants.py
```diff
@@ -0,0 +1,51 @@
...
+if os.path.exists(args.file_out) and args.overwrite is False:
+    print("{} exists, not overwriting...".format(args.file_out))
+    continue
+
+uv_obj = UVData()
+uv_obj.read_uvfits(args.file_in)
+
+large_ant_nums = uv_obj.antenna_numbers[np.where(uv_obj.antenna_numbers > 255)[0]]
+
+new_nums = sorted(list(set(range(256)) - set(uv_obj.antenna_numbers)))
```
bnikolic: It needs to be range(255) I think because 0-based antenna 255 is 1-based 256 and that gets turned into 0 in some 8-bit code path in CASA

bhazelton: Ok, I'll decrease the maximum antenna number to 254.

bhazelton:
@adampbeardsley can you look at this? It adds a script to renumber large antenna numbers to make CASA able to ingest the files.

scripts/renumber_ants.py
```diff
@@ -0,0 +1,64 @@
...
+    wh_ant1_arr = np.where(uv_obj.ant_1_array == ant_in)[0]
+    wh_ant2_arr = np.where(uv_obj.ant_2_array == ant_in)[0]
+
+    uv_obj.antenna_numbers[wh_ant_num] = ant_out
+    uv_obj.ant_1_array[wh_ant1_arr] = ant_out
+    uv_obj.ant_2_array[wh_ant2_arr] = ant_out
+
+    blt_inds = np.array(sorted(list(set(wh_ant1_arr.tolist() + wh_ant2_arr.tolist()))))
+    uv_obj.baseline_array[blt_inds] = \
+        uv_obj.antnums_to_baseline(uv_obj.ant_1_array[blt_inds], uv_obj.ant_2_array[blt_inds])
```
adampbeardsley: If you wanted to, you could change lines 58-60 to
`uv_obj.baseline_array = uv_obj.antnums_to_baseline(uv_obj.ant_1_array, uv_obj.ant_2_array)`
outside the for loop. But not necessary.

bhazelton: sure, I'll change it.

PR #303, opened 2018-02-28
Added beam area and beam squared area calculations

acliu:
A fairly simple addition that allows the UVBeam to output the integral of the beam and the integral of the beam squared.

### DIFF
--- a/pyuvdata/uvbeam.py
+++ b/pyuvdata/uvbeam.py
@@ -977,42 +977,75 @@ class UVBeam(UVBase):
         self.__add__(other, inplace=True)
         return self
 
-    def get_beam_area(self,stokes='I'):
+    def get_beam_area(self, stokes='pseudo_I'):
         """
         Computes the integral of the beam, which has units of steradians
+
+        Currently, only the "pseudo Stokes I" beam is supported.
+        See Equations 4 and 5 of Moore et al. (2017) ApJ 836, 154
+        or arxiv:1502.05072 for details.
         """
         if self.beam_type is not 'power':
             raise ValueError('beam_type must be power')
         if self._data_normalization.value != 'peak':
-            print self._data_normalization.value
             raise ValueError('beam must be peak normalized')
+        if self.pixel_coordinate_system is not 'healpix':
+            raise ValueError('Currently only healpix format supported')
+        if self.data_array.shape[0] > 1:
+            raise ValueError('Expect scalar for power beam, found vector')
+
         npix = self.data_array.shape[-1]
-        if stokes == 'I':
-            # Assume A_I = (B_xx + B_yy)/2
-            beam = np.mean(self.data_array[0,0], axis=-3)
+        pol_array = self.polarization_array
+        if stokes == 'pseudo_I':
+            if 1 in pol_array:
+                stokes_I_ind = np.where(np.isin(pol_array, 1))[0][0]
+                beam = self.data_array[0, 0, stokes_I_ind]
+            elif -5 in pol_array and -6 in pol_array:
+                # Assume A_I = (B_xx + B_yy)/2
+                xx_ind = np.where(np.isin(pol_array, -5))[0][0]
+                yy_ind = np.where(np.isin(pol_array, -6))[0][0]
+                beam = 0.5 * (self.data_array[0, 0, xx_ind] + self.data_array[0, 0, yy_ind])
+            else:
+                raise ValueError('Do not have the right polarization information')
         else:
             raise NotImplementedError()
 
         return np.sum(beam, axis=-1) * 4.*np.pi / npix
 
-    def get_beam_sq_area(self,stokes='I'):
+    def get_beam_sq_area(self, stokes='pseudo_I'):
         """
         Computes the integral of the beam**2, which has units of steradians
+
+        Currently, only the "pseudo Stokes I" beam is supported.
+        See Equations 4 and 5 of Moore et al. (2017) ApJ 836, 154
+        or arxiv:1502.05072 for details.
         """
         if self.beam_type is not 'power':
             raise ValueError('beam_type must be power')
         if self._data_normalization.value != 'peak':
-            print self._data_normalization.value
             raise ValueError('beam must be peak normalized')
+        if self.pixel_coordinate_system is not 'healpix':
+            raise ValueError('Currently only healpix format supported')
+        if self.data_array.shape[0] > 1:
+            raise ValueError('Expect scalar for power beam, found vector')
+
         npix = self.data_array.shape[-1]
-        if stokes == 'I':
-            # Assume A_I = (B_xx + B_yy)/2
-            beam = np.mean(self.data_array[0,0], axis=-3)
+        pol_array = self.polarization_array
+        if stokes == 'pseudo_I':
+            if 1 in pol_array:
+                stokes_I_ind = np.where(np.isin(pol_array, 1))[0][0]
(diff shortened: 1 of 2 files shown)


bhazelton:
@acliu It looks like the Travis checks failed because of some pep8 issues. Also, can you add testing to fully cover your functions?

pyuvdata/uvbeam.py
```diff
@@ -978,6 +977,43 @@ def __iadd__(self, other):
...
+        if self.beam_type is not 'power':
+            raise ValueError('beam_type must be power')
+        if self._data_normalization.value != 'peak':
+            print self._data_normalization.value
+            raise ValueError('beam must be peak normalized')
+        npix = self.data_array.shape[-1]
+        if stokes == 'I':
+            # Assume A_I = (B_xx + B_yy)/2
+            beam = np.mean(self.data_array[0,0], axis=-3)
+        else:
```
bhazelton: This worries me for several reasons.

First, the first index is the vector index. Usually it only has length one for power beams, but that's not required. If it is a power beam with Naxes_vec > 1, then it's reporting the power separately in different directions. I suspect that you either want to error on that or sum over that axis. (You can get such a beam for testing by reading in an efield beam and then converting to power using `efield_to_power` and setting `keep_basis_vector=True`).

Even more critically, you need to be careful about the polarization axis. The polarization axis can be in stokes parameters already (values in the pol array between 1-4), or it can be any set of linear pols (xx, yy, xy, yx -> -5:-8) _in any order_. Do you really want to sum over all 4 if they are all present? You can test with a beam with all 4 pols by reading in an efield beam and converting to power with `efield_to_power` -- by default it also calculates the xy & yx power beams.

Finally, I'm not sure I like calling this options 'stokes=I' because summing over the xx & yy polarizations in this fashion is really not what one means by a stokes I beam in general (although I'd need to check details with James, I believe that a stokes I beam has contributions from all 4 pols and they need to be combined in different ways in different parts of the sky -- in fact the conversion should use a different 4x4 matrix for every sky pixel). I'd prefer something that's less overloaded.

pyuvdata/uvbeam.py
```diff
@@ -978,6 +977,43 @@ def __iadd__(self, other):
...
 
+    def get_beam_area(self,stokes='I'):
+        """
+        Computes the integral of the beam, which has units of steradians
+        """
+        if self.beam_type is not 'power':
+            raise ValueError('beam_type must be power')
+        if self._data_normalization.value != 'peak':
+            print self._data_normalization.value
+            raise ValueError('beam must be peak normalized')
```
bhazelton: Note that a 'peak' normalization means that each frequency has been separately peak normalized and the intrinsic variation in beam response with frequency has been pushed into the bandpass variable. Is that what you want? Do you want to take the bandpass into account in any way?

pyuvdata/uvbeam.py
```diff
@@ -978,6 +977,43 @@ def __iadd__(self, other):
...
+        npix = self.data_array.shape[-1]
+        if stokes == 'I':
+            # Assume A_I = (B_xx + B_yy)/2
+            beam = np.mean(self.data_array[0,0], axis=-3)
+        else:
+            raise NotImplementedError()
+
+        return np.sum(beam, axis=-1) * 4.*np.pi / npix
```
bhazelton: I think this is assuming something about how much of the sky the beam model covers and/or the area of the pixels? The beam is not required to cover the full sky.

Also, if this isn't a HEALPix beam I think you need to sum over the last 2 indices.

pyuvdata/uvbeam.py
```diff
@@ -978,6 +977,43 @@ def __iadd__(self, other):
...
+    def get_beam_area(self,stokes='I'):
+        """
+        Computes the integral of the beam, which has units of steradians
+        """
+        if self.beam_type is not 'power':
+            raise ValueError('beam_type must be power')
+        if self._data_normalization.value != 'peak':
+            print self._data_normalization.value
+            raise ValueError('beam must be peak normalized')
+        npix = self.data_array.shape[-1]
```
bhazelton: Is this assuming a Healpix beam? If it's not healpix, there are two pixel axes that should have their length multiplied together to get the number of pixels. I'd suggest using `npix = self.Npixels` for Healpix beams and `npix = self.Naxes1 * self.Naxes2` for regularly gridded beams.

pyuvdata/uvbeam.py
acliu: Definitely! I've changed a bunch of things:
---If Naxes_vec > 1, the code will now error.
---The code now checks to see if a Stokes I beam is available. If it is, it'll use that. Otherwise, it will extract only xx and yy and average those.
---Agreed that it's not really Stokes I. More like what we've sometimes called "pseudo Stokes I". The terminology has changed now, and I've also added a journal reference.

pyuvdata/uvbeam.py
acliu: I think we do want peak normalization. I'll check with some people, but I think it's peak that we want.

pyuvdata/uvbeam.py
acliu: Changed to only allow healpix.

pyuvdata/uvbeam.py
acliu: Changed to only allow healpix.

acliu:
Thanks Bryna! I've made the changes outlined above, written a bunch of tests, and have also checked against pep8 standards.

acliu:
Ok, I just tried to fix the pep8 issues. The weird thing is that the pep8 checker on my computer didn't seem to pick up on them, but when I clicked the link next to the failed build (thanks for the tip!), they showed up.

pyuvdata/uvbeam.py
```diff
@@ -978,6 +978,76 @@ def __iadd__(self, other):
...
+                # Assume A_I = (B_xx + B_yy)/2
+                xx_ind = np.where(np.isin(pol_array, -5))[0][0]
+                yy_ind = np.where(np.isin(pol_array, -6))[0][0]
+                beam = 0.5 * (self.data_array[0, 0, xx_ind] + self.data_array[0, 0, yy_ind])
+            else:
+                raise ValueError('Do not have the right polarization information')
+        else:
+            raise NotImplementedError()
+
+        return np.sum(beam, axis=-1) * 4. * np.pi / npix
```
bhazelton: I'm still worried that this command requires that the beam covers the full sky, and I'm actually not quite sure why you're dividing by npix (this is a sum not an average, right?). But if you want the area covered by the beam, you can use `healpy.pixelfunc.nside2pixarea` to get the pixel area and then multiply by self.Npixels.

acliu: But naively I would've thought that if I'm only accepting Healpix, then the beams are by definition full sky?

I'm multiplying by 4pi/npix because it's meant to approximate an integral, rather than a sum. So the 4pi/npix is meant to be the dOmega differential element.

acliu: Ah, thanks for pointing out this problem! I hadn't realized that you had built support for a non-full sky collection of Healpix pixels.

I've made the changes, working out the dOmega for nside in the object but using the healpy function in the tests to minimize the chances that I messed it up.

bhazelton: That's perfect, thanks! I'm going to add one more little test to cover the case where we have a stokes I polarization for completeness and then I think we can merge it.

PR #306, opened 2018-03-05
updated UVCal.ant_array, antenna_numbers and antenna_names doc

nkern:
Updated documentation on these `UVCal` attributes to more clearly explain their purpose. Please let me know if anyone wants more / less / different wording for the doc on these attributes.


bhazelton:
The doc changes look great, much clearer than before. But it looks like the tests aren't passing due to the change to make antenna_numbers an array. I think an array is fine, but we'll have to change other things in the code to make the tests pass.

nkern:
@bhazelton ok I think this is ready for merge. let me know when you are OK w/ it, and if so feel free to merge it.

Issue #311, opened 2018-03-07
write_calfits introduces small error in time array


adampbeardsley:
We noticed this when creating unittests in hera_cal (see HERA-Team/hera_cal#204). There are two problems that I see with the code below.

https://github.com/HERA-Team/pyuvdata/blob/b4bc89c01860985a571bd6d187be5f5a9943f5f8/pyuvdata/calfits.py#L79-L83

1) Line 79 is checking the closeness of two numbers that are in different units (time_spacing is in days, integration_time is in seconds)
2) The rounded_spacing is not quite achieving the precision needed. While it gives a rounded spacing within the tolerance of the time array, the error compounded over Ntimes is too large. I think the tolerance should be divided by Ntimes, and maybe add an extra digit to be on the safe side. In fact, it's not clear to me why this rounding is happening in the first place.

adampbeardsley:
This was addressed in PR #312

PR #312, opened 2018-03-07
Fixes to write_calfits to ensure units are correct and tols are met

adampbeardsley:
Addresses #311.
I worry that this will affect downstream pipelines, because it looks like there were certain scenarios where the time deltas were being stored in units of seconds, and other instances stored in units of days. So it would be good to have @zakiali @jsdillon and @nkern look at this to see if we need to include backwards compatibility.

### DIFF
--- a/pyuvdata/calfits.py
+++ b/pyuvdata/calfits.py
@@ -76,13 +76,15 @@ class CALFITS(UVCal):
                 raise ValueError('The times are not evenly spaced (probably '
                                  'because of a select operation). The calfits format '
                                  'does not support unevenly spaced times.')
-            if np.isclose(time_spacing[0], self.integration_time):
-                time_spacing = self.integration_time
+            if np.isclose(time_spacing[0], self.integration_time / (24. * 60.**2)):
+                time_spacing = self.integration_time / (24. * 60.**2)
             else:
-                rounded_spacing = np.around(time_spacing, int(np.ceil(np.log10(self._time_array.tols[1]) * -1)))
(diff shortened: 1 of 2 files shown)


bhazelton:
The parameter description of `integration_time` on the uvcal object says that it has units of seconds, but it looks like you're assuming units of days here.

adampbeardsley:
I think it's just the opposite - the old code assumed integration_time and time_array were in the same units. Now I am converting integration_time into days to use as time_spacing, which is used for setting the time axis delta. In other words, CRVAL3 and CDELT3 need to be in the same units, so I'm converting the spacing to match.

bhazelton:
Sorry, I misread it. Thanks for the clarification.

bhazelton:
I approved but didn't merge this in case you wanted to wait for comment from others.

jsdillon:
This doesn't affect `hera_cal.io.load_cal` or `hera_cal.io.update_cal`, but it may affect `hera_cal.io.write_cal`. @nkern what do you think?

nkern:
I've always thought that the `UVCal.integration_time` attribute has units of seconds..

nkern:
@adampbeardsley we can always change `hera_cal.io.write_cal` to match whatever `pyuvdata` wants, so don't worry too much about us

nkern:
@adampbeardsley just be sure to make sure the docs are up-to-date :)

adampbeardsley:
Ok, don't say I didn't warn y'all...

PR #316, opened 2018-03-15
renumber ants for miriad files

nkern:
I'd appreciate having this capability for MIRIAD files, too.

### DIFF
--- a/scripts/renumber_ants.py
+++ b/scripts/renumber_ants.py
@@ -70,4 +70,3 @@ if args.filetype == 'uvfits':
     uv_obj.write_uvfits(args.file_out)
 elif args.filetype == 'miriad':
     uv_obj.write_miriad(args.file_out)
-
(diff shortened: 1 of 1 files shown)


bhazelton:
Out of curiosity, what's the use case for renumbering the antennas for Miriad? It's just a little dangerous because it makes the antenna numbers not match the true numbers for the array...

nkern:
b/c all my personal calibration routines (and `hera_cal` in general) always goes back to the miriad file format before starting the next round of calibration. 

also, i don't see what you mean. in running the script on UVFITS and MIRIAD I get the same behavior for both `antenna_numbers` and `antenna_names`..

bhazelton:
Ok, I'm fine with adding this functionality, I was just trying to understand the use case.

My comment about the antenna numbers not matching applies to both the uvfits and miriad cases -- I just mean that the new (renumbered) antenna numbers don't match the actual HERA numbers for those antennas, so if you see a problem on one of them you have to translate to the original numbers in order to communicate with anyone using the real HERA numbers. Also, as the array changes, the renumbered values can change because they can't clash with other existing numbers. In general I regard this script as a kludge :).

PR #317, opened 2018-03-23
Fix CST beam filename frequency parsing

bhazelton:


pyuvdata/tests/test_cst_beam.py
```diff
@@ -22,6 +22,11 @@ def test_frequencyparse():
     parsed_freqs = [beam1.name2freq(f) for f in test_files]
     nt.assert_equal(parsed_freqs, [150e6, 123e6])
 
+    test_path = 'Simulations/Radiation patterns/E-field pattern - Rigging height 4.9 m/HERA_4.9m_E-pattern_100-200MHz/'
```
adampbeardsley: Could you use `os.path.join`? Also, are spaces robust in paths? I know these are Nic's actual paths, but there's no reason it has to be the same for this test.

bhazelton: yes, I'll fix it. I wrote the test that way to make sure it'd work in my script, but I can take out the spaces.

bhazelton:
@adampbeardsley I think I made the changes you asked for.

PR #324, opened 2018-04-10
allowed vis_units to additionally be mK str

nkern:
it would be nice to allow the `units` of a UVData object to additionally be`mK str`, `mK` or `K`. Specifically, the input to `hera_pspec` will be visibility data in units of `mK` (where we have already divided out by the beam integral to get rid of the `str` units).

### DIFF
--- a/pyuvdata/uvdata.py
+++ b/pyuvdata/uvdata.py
@@ -55,10 +55,10 @@ class UVData(UVBase):
                                                  'Nfreqs', 'Npols'),
                                            expected_type=np.complex)
 
-        desc = 'Visibility units, options are: "uncalib", "Jy" or "K str"'
+        desc = 'Visibility units, options are: "uncalib", "Jy" or "K str", "mK str"'
         self._vis_units = uvp.UVParameter('vis_units', description=desc,
                                           form='str', expected_type=str,
-                                          acceptable_vals=["uncalib", "Jy", "K str"])
+                                          acceptable_vals=["uncalib", "Jy", "K str", "mK str"])
 
         desc = ('Number of data points averaged into each data element, '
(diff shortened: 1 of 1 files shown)


bhazelton:
Just to put the details of a discussion @nkern and I had on Slack here: 

I think the `mK str` unit should be fine (although there needs to be a conversion in uvfits read/write because that unit is not allowed in uvfits files), but I’m worried about how to handle the case where you’ve divided by the integral of the beam. In particular, how to properly communicate that in the various file formats we support. 

I need think about this a little more, I'll try to get to this soon when proposal panic has passed.

bhazelton:
Ok, just getting back to this.

I think this is ok, except that you need to add some code to `read_uvfits` to deal with the fact that the only allowed values for `bunit` (where this is normally stored) are `Jy` or `K str` as `uncalib`. I think the right thing is to write anything where you've divided by the beam as `uncalib`. However, just doing that would make it impossible to round-trip this data through uvfits files, which is something we *really* try to avoid in pyuvdata. So I think we also need to add a new, non-standard header keyword to uvfits to store the actual units. Then the code can check for the presence of that keyword and use it if it exists.

nkern:
So I realized that although we want to change the `vis_units` to `mK`, we don't need to write the `UVData` object to file, so we won't ever run into a situation where we are triggering the `self.run_check` function (which would fail if `vis_units` were set to `mK`). So if its easiest, we actually don't need to change anything in `pyuvdata`, and can just perform the primary beam correction in memory before we run the pspec pipeline..

bhazelton:
Ok, that sounds like a reasonable solution @nkern. Is it ok for me to remove this branch?

nkern:
yup! i can do it...

PR #326, opened 2018-04-14
added XX and YY pol capability to uvbeam.get_beam_areas

nkern:
a small addition to allow for 'XX' and 'YY' beam power area integrals for `UVBeam`, which we need in `hera_pspec`. should be a quick review..

### DIFF
--- a/pyuvdata/uvbeam.py
+++ b/pyuvdata/uvbeam.py
@@ -1019,24 +1019,11 @@ class UVBeam(UVBase):
         self.__add__(other, inplace=True)
         return self
 
-    def get_beam_area(self, stokes='pseudo_I'):
+    def _get_beam(self, stokes):
         """
-        Computes the integral of the beam, which has units of steradians
-
-        Currently, only the "pseudo Stokes I" beam is supported.
-        See Equations 4 and 5 of Moore et al. (2017) ApJ 836, 154
-        or arxiv:1502.05072 for details.
+        Get the healpix beam map corresponding to the specififed stokes polarization,
+        'pseudo_I', 'pseudo_Q', etc., or linear dipole polarization, 'XX', 'YY', etc.
         """
-        if self.beam_type != 'power':
-            raise ValueError('beam_type must be power')
-        if self._data_normalization.value != 'peak':
-            raise ValueError('beam must be peak normalized')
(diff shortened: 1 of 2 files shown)


pyuvdata/uvbeam.py
```diff
@@ -1019,12 +1019,46 @@ def __iadd__(self, other):
         self.__add__(other, inplace=True)
         return self
 
+    def _get_beam(self, stokes):
+        """
+        Get the healpix beam map corresponding to the specififed stokes polarization,
```
bhazelton: It doesn't look to me like this is specific to HEALPix beams. If it is, you should add a check that errors if this is called on a non-HEALPix beam

pyuvdata/uvbeam.py
```diff
@@ -1019,12 +1019,46 @@ def __iadd__(self, other):
...
+                yy_ind = np.where(np.isin(pol_array, -6))[0][0]
+                beam = self.data_array[0, 0, yy_ind]
+            else:
+                raise ValueError('Do not have the right polarization information')
+        else:
+            raise NotImplementedError("Stokes {} not yet implemented...".format(stokes))
+
+        return beam
+
     def get_beam_area(self, stokes='pseudo_I'):
```
bhazelton: I would favor renaming the 'stokes' option to something like 'polarization' or 'polarization_name' because 'xx' and 'yy' are not stokes types, they're polarization types.

bhazelton:
@nkern can you rebase this on master? I think that'll fix the Travis issues.

nkern:
yup! i'll rebase it and address your comments some point this evening

pyuvdata/uvbeam.py
nkern: I agree, and I also think that we should not use 'pseudo_I' or 'pseudo_Q' etc. because it is not consistent with the utils.polstr2num convention we have already adopted within `pyuvdata` and within `hera_cal`. Although I appreciate the philosophy behind making the "pseudo" distinction, I think it makes things unnecessarily complicated and not self-consistent. What are your thoughts?

bhazelton: hmm... I hear what you're saying but I'm very sensitive to the confusion between stokes and pseudo-stokes. Let me talk to a few other people.

PR #331, opened 2018-05-02
change all uses of Stokes to pseudo-Stokes

bhazelton:


pyuvdata/beamfits.py
```diff
@@ -128,8 +128,13 @@ def read_beamfits(self, filename, run_check=True, check_extra=True,
                 self.data_array = np.squeeze(complex_arrs[0] + 1j * complex_arrs[1], axis=0)
             else:
                 self.data_array = data
+
+            # Note: This axis is called STOKES by analogy with the equivilent uvfits axis
+            # However, this confusing because it is NOT a true Stokes axis,
```
adampbeardsley: Couple typos: equivilent -> equivalent; "this confusing" -> "this is confusing"

pyuvdata/beamfits.py
```diff
@@ -449,8 +454,11 @@ def write_beamfits(self, filename, run_check=True, check_extra=True,
             else:
                 pol_spacing = 1
 
+            # Note: This axis is called STOKES by analogy with the equivilent uvfits axis
+            # However, this confusing because it is NOT a true Stokes axis,
```
adampbeardsley: Same typos

pyuvdata/beamfits.py
bhazelton: thanks, fixing them now

pyuvdata/beamfits.py
bhazelton: thanks, fixing them now

bhazelton:
I'm a little worried about the downstream affect on users, but I think that these are only recently being used, so maybe we can make this change quickly and not affect too many people? That's part of why I asked @nkern to look at this too. I'm just very reluctant to be a source of more confusion about this going forward...

nkern:
I think a good compromise would be to have `polnum2str` map from `1 -> pI`, but support compatibility in `polstr2num` (as Adam suggested) to do both `pI -> 1` and `I -> 1`, to support users who have adopted that convention.

bhazelton:
Ok , that sounds reasonable @nkern, I'll make that change

pyuvdata/utils.py
```diff
@@ -315,14 +315,15 @@ def fits_indexhdus(hdulist):
...
+    Use 'pI', 'pQ', 'pU' and 'pV' to make it clear that these are pseudo-Stokes, not true Stokes
 
     Args:
         pol: polarization string
 
     Returns:
         Number corresponding to string
     """
-    poldict = {'I': 1, 'Q': 2, 'U': 3, 'V': 4,
+    poldict = {'PI': 1, 'PQ': 2, 'PU': 3, 'PV': 4,
```
nkern: this is supposed to be lower case on the `p`, no?

nkern: ah, is this b/c you do `pol.upper()` before passing through to the dictionary?

bhazelton: Yes, but the way the `polstr2num` function handles case insensitivity is to make them all upper case. I decided case-insensitivity was a good thing for `polstr2num`, but I used lowercase `p` for `polnum2str`.

nkern: ok makes sense. i'm fine w/ this then.

bhazelton: I can add a comment to that effect...

PR #335, opened 2018-05-05
added data selection for uvdata.read_miriad and miriad.read_miriad

nkern:
Enables partial I/O for `UVData.read_miriad` and `miriad.read_miriad` methods.

For this to work on polarization I had to eliminate the following assertion in `miriad.read_miriad`
```
        if len(self.polarization_array) != self.Npols:
            warnings.warn('npols={npols} but found {n} pols in data file'.format(
                npols=self.Npols, n=len(self.polarization_array)))
```
which also necessitated removing a test in `test_miriad` to test for this warning. I'm not sure what use-case this is trying to account for, but if it is a vital catch then perhaps we can figure out how to put it back in.

### DIFF
--- a/pyuvdata/miriad.py
+++ b/pyuvdata/miriad.py
@@ -52,7 +52,8 @@ class Miriad(UVData):
                 numbers within the tuple does not matter. A single antenna iterable
                 e.g. (1,) is interpreted as all visibilities with that antenna.
             ant_str: A string containing information about what kinds of visibility data
-                to read-in.  Can be 'auto', 'cross', 'all'.
+                to read-in.  Can be 'auto', 'cross', 'all'. Cannot provide ant_str if
+                antenna_nums and/or ant_pairs_nums is not None.
             polarizations: List of polarization integers or strings to read-in.
                 Ex: ['xx', 'yy', ...]
             time_range: len-2 list containing min and max range of times (Julian Date) to read-in.
@@ -204,6 +205,7 @@ class Miriad(UVData):
         if ant_str is not None:
             # type check
             assert isinstance(ant_str, (str, np.str)), "ant_str must be fed as a string"
+            assert antenna_nums is None and ant_pairs_nums is None, "ant_str must be None if antenna_nums or ant_pairs_nums is not None"
             aipy.scripting.uv_selector(uv, ant_str)
         # select on antenna_nums and/or ant_pairs_nums using aipy.scripting.uv_selector
         if antenna_nums is not None or ant_pairs_nums is not None:
--- a/pyuvdata/miriad.py
+++ b/pyuvdata/miriad.py
@@ -48,7 +48,7 @@ class Miriad(UVData):
             antpairs: List of antnum-pair tuples in data to read-in.
                 If a tuple contains a single antnum, read-in all baselines that touch that antnum.
                 In this case, make sure the tuple is an iterable, e.g. (2,) not (2).
-                Ex: [(0, 0), (0, 1), (2,), ...]. 
+                Ex: [(0, 0), (0, 1), (2,), ...].
             pols: List of polarization integers or strings to only read-in.
                 Ex: ['xx', 'yy', ...]
             times: len-2 list containing min and max range of times (Julian Date) to read-in.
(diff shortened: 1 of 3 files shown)


pyuvdata/miriad.py
```diff
@@ -29,7 +30,8 @@ def _pol_to_ind(self, pol):
         return pol_ind
 
     def read_miriad(self, filepath, correct_lat_lon=True, run_check=True,
-                    check_extra=True, run_check_acceptability=True, phase_type=None):
+                    check_extra=True, run_check_acceptability=True, phase_type=None,
+                    antpairs=None, pols=None, times=None):
```
bhazelton: I'd like to see this interface match the select interface, so include these options:
antenna_nums, antenna_names, ant_str, ant_pairs_nums, frequencies, freq_chans, times, polarizations, blt_inds

If some of these can't be supported, we should document why, but we should definitely use the same keywords with the same functionality when possible. (If you think the current select interface should be modified we can discuss that, but I'd really like there to be consistency.)

pyuvdata/miriad.py
```diff
@@ -268,9 +317,7 @@ def read_miriad(self, filepath, correct_lat_lon=True, run_check=True,
             data_accumulator[pol] = np.array(data)
 
         self.polarization_array = np.array(pol_list)
-        if len(self.polarization_array) != self.Npols:
-            warnings.warn('npols={npols} but found {n} pols in data file'.format(
-                npols=self.Npols, n=len(self.polarization_array)))
+        self.Npols = len(pol_list)
```
bhazelton: I'm a little confused about this, is the Miriad `npol` parameter not available unless you read in the whole file?

pyuvdata/tests/test_miriad.py
```diff
@@ -563,9 +559,49 @@ def test_readWriteReadMiriad():
...
 
+    # check partial IO selections
+    full = UVData()
+    full.read_miriad(testfile)
+    uv_in = UVData()
+
+    # test only specified bls were read, and that flipped antpair is loaded too
+    uv_in.read_miriad(testfile, antpairs=[(0, 0), (0, 1), (4, 2)])
+    nt.assert_equal(uv_in.get_antpairs(), [(0, 0), (0, 1), (2, 4)])
+
```
bhazelton: Can you also test that the object overall is equal if you read in the whole file and then do a matching select for all these partial read tests? That's a much more comprehensive test that may catch edge cases. (You may have to have some special handling for the history parameter, you can see how we do that for uvfits if you want.)

pyuvdata/miriad.py
nkern: yup for sure, i'll try to make that happen

pyuvdata/miriad.py
nkern: no it is available, but when you perform a `UV.select`  on the miriad file on polarizations, `npol` doesn't change to reflect the extracted polarizations, it remains what it was for the entire file..

pyuvdata/tests/test_miriad.py
nkern: sure thing

pyuvdata/miriad.py
bhazelton: I see. Maybe we can leave in the check unless we're doing a select along the polarization axis?

nkern: well its not so much that the test breaks things, its that with the new definition of `Npols`, its a trivially-true and non-helpful test.. because  I'm manually setting `self.Npols` to be equal to `len(pol_list)`, which itself has explicitly set to `len(self.polarization_array)`, if you see what I mean.

bhazelton: Yes, I see. I'm just thinking that in the case where there's no down select on the polarization axis, we could still check to see if npol matched the polarization array length. But it's really not critical, so I think it's ok to drop it if it doesn't make sense.

nkern:
@bhazelton so I tried to make the variables match those in `select` as close as possible. I was able to support antenna_nums, ant_str, ant_pairs_nums, time_range, and polarizations. antenna_names is hard b/c I don't think miriad by default stores antenna names. The time_range kwarg is also different than the UVData.select times variable, and I didn't support frequency and blts selection. This is done b/c these are the variables the native `miriad.select` functions supports and they way it allows you to access them. I don't know of an alternative way to pre-select out these variables from miriad objects, but maybe a more advanced miriad user does. But I don't think its that critical: one can always perform a `UVData.select` on the data once loaded in, and the main axes we want to enable partial I/O on are antennas and baselines.

Let me know if you think this is sufficient: it certainly is for what I and the hera_cal / hera_pspec team need in the near term from pyuvdata on this front.

pyuvdata/miriad.py
```diff
@@ -186,6 +200,62 @@ def read_miriad(self, filepath, correct_lat_lon=True, run_check=True,
         for extra_variable in extra_miriad_variables:
             check_variables[extra_variable] = uv[extra_variable]
 
+        # select on ant_str if provided
+        if ant_str is not None:
+            # typc check
```
zakiali: 'typc' -> 'type'

pyuvdata/miriad.py
```diff
@@ -186,6 +200,62 @@ def read_miriad(self, filepath, correct_lat_lon=True, run_check=True,
...
+            assert isinstance(ant_str, (str, np.str)), "ant_str must be fed as a string"
+            uv.select(ant_str, 0, 0)
```
zakiali: I think this line only works if 'auto' is passed. 'cross' and 'all' do not work (at least with my aipy). It might be better to use scripting.uv_selector instead.

nkern:
@bhazelton ready to merge whenever you're ok with it.

PR #338, opened 2018-05-08
Fix bug that causes memory error

### LINKED ISSUE #328: Make check for unflagged nsample=0 visibilities in write_uvfits less memory intensive
This check causes a memory error for large files, presumably because we're making copies of a large part of the data. Seen by Ben McKinley on MWA files.

### LINKED ISSUE #336: MemoryError in write_uvfits
pyuvdata/pyuvdata/uvfits.py, line 621:
wh_nsample0 = np.where(self.nsample_array == 0)[0]
line 622 (np.any(~self.flag_array[wh_nsample0])) throws an MemoryError.
If change line 621 to be: 
wh_nsample0 = np.where(self.nsample_array == 0)
then it will be fine. Not sure if it is a numpy version problem.

bhazelton:
fixes #336 
fixes #328

### DIFF
--- a/pyuvdata/uvfits.py
+++ b/pyuvdata/uvfits.py
@@ -618,7 +618,7 @@ class UVFITS(UVData):
                                          .format(attribute=p))
 
         # check for unflagged data with nsample = 0. Warn if any found
-        wh_nsample0 = np.where(self.nsample_array == 0)[0]
+        wh_nsample0 = np.where(self.nsample_array == 0)
         if np.any(~self.flag_array[wh_nsample0]):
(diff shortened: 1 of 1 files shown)


bhazelton:
@benjyman I think this PR should fix the memory error you were having. It turned out to just be a bug in how I was handling the np.where

bhazelton:
@wenyang-li can you review this one-line PR to fix the bug you identified?

wenyang-li:
Yes, this fixes the issue. Thanks!

wenyang-li (approved):
Yes, this fixes the issue. Thanks!

PR #340, opened 2018-05-10
Changed Stokes 'I' to 'pI' and added option to read (xx,xy,yx,yy) beams

Chuneeta:
changed Stokes 'I' to 'pI' and added the ability to read in (xx,yy,xy,yy) beam and tried to generalize some of the hardcorded part. Work on pQ, pU and pV still in progress.

### DIFF
--- a/pyuvdata/tests/test_uvbeam.py
+++ b/pyuvdata/tests/test_uvbeam.py
@@ -4,8 +4,7 @@ import os
 import numpy as np
 import copy
 import healpy as hp
-#from pyuvdata import UVBeam
-from uvbeam import UVBeam
+from pyuvdata import UVBeam
 import pyuvdata.tests as uvtest
 import pyuvdata.utils as uvutils
 import pyuvdata.version as uvversion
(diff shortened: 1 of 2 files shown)


pyuvdata/tests/test_uvbeam.py
```diff
@@ -1371,3 +1371,20 @@ def test_get_beam_functions():
...
+                             feed_version='0.1',
+                             model_name='E-field pattern - Rigging height 4.9m',
+                             model_version='1.0')
+
+    efield_beam.efield_to_power()
+    power_beam = copy.deepcopy(efield_beam)
+    power_beam.az_za_to_healpix()
+    power_beam.peak_normalize()
+    XY_area = power_beam.get_beam_sq_area("XY")
+    YY_area = power_beam.get_beam_sq_area("YX")
```
bhazelton: I think this should read "YX_area"

Also, there should probably be an assert in this test. Maybe at least test that the XY & YX areas are equal? Also, I think that they should be less than the XX or YY area, right?

PR #343, opened 2018-05-11
uvbeam.get_beam now does polnum (not polstr) check

nkern:
this provides backwards compatibility, and should fix what is causing hera_pspec master to fail currently

### DIFF
--- a/pyuvdata/uvbeam.py
+++ b/pyuvdata/uvbeam.py
@@ -1057,26 +1057,30 @@ class UVBeam(UVBase):
         Currently only 'pI', 'XX' and 'YY' are supported.
 
         Args:
-          pol : polarization string, Ex. a pseudo-stokes pol 'pI', or a linear pol 'XX'
+          pol : polarization string or integer, Ex. a pseudo-stokes pol 'pI', or a linear pol 'XX'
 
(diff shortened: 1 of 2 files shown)


pyuvdata/uvbeam.py
```diff
@@ -1057,36 +1057,40 @@ def _get_beam(self, pol):
...
+                keys = pols_dict[1]
                 if keys[0] in pol_array and keys[1] in pol_array:
                     xx_ind = np.where(np.isin(pol_array, keys[0]))[0][0]
                     yy_ind = np.where(np.isin(pol_array, keys[1]))[0][0]
                     beam = 0.5 * (self.data_array[0, 0, xx_ind] + self.data_array[0, 0, yy_ind])
                 else:
                     raise ValueError('Do not have the right polarization information')
             else:
-                raise NotImplementedError("Polarization {} not yet implemented...".format(pol))
+                raise NotImplementedError("Polarization {} not yet implemented...".format(polstr))
```
bhazelton: I should have caught this in the last PR, but I think this line should be:
`raise ValueError('Do not have the right polarization information')`

You only get here if the pol is allowed but not available in the object.

PR #347, opened 2018-05-15
Implement new UVH5 data format

plaplant:
This PR adds a new "UVH5" data format into pyuvdata. The format is saved as an HDF5 data type, and is a straightforward serialization of a UVData object. Tests show that the UVH5 read and write methods are able to round-trip MIRIAD and UVFITS files, including current HERA data. Future plans include adding support for partially reading files and using the built-in compression filters (in particular, the bitshuffle filter for visibility data), but this PR adds the basic functionality. Any feedback or comments are welcome!

### DIFF
--- a/pyuvdata/uvdata.py
+++ b/pyuvdata/uvdata.py
@@ -1316,7 +1316,7 @@ class UVData(UVBase):
             import uvh5
             other_obj = uvh5.UVH5()
         else:
-            raise ValueError('filetype must be uvfits, miriad, or fhd')
+            raise ValueError('filetype must be uvfits, miriad, fhd, or uvh5')
         for p in self:
             param = getattr(self, p)
             setattr(other_obj, p, param)
@@ -1704,7 +1704,19 @@ class UVData(UVBase):
     def read_uvh5(self, filename, run_check=True, check_extra=True,
                   run_check_acceptability=True):
         """
-        Read uvh5 file
+        Read a UVH5 file.
+
+        Args:
+            filename: The UVH5 file to read.
+            run_check: Option to check for the existence and proper shapes of
+                parameters after reading in the file. Default is True.
+            check_extra: Option to check optional parameters as well as required
+                ones. Default is True.
+            run_check_acceptability: Option to check acceptable range of the values of
+                parameters after reading in the file. Default is True.
+
+        Returns:
+            None
         """
         import uvh5
         if isinstance(filename, (list, tuple)):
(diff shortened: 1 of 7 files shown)


pyuvdata/uvdata.py
```diff
@@ -1698,6 +1701,38 @@ def write_miriad(self, filepath, run_check=True, check_extra=True,
                                 clobber=clobber, no_antnums=no_antnums)
         del(miriad_obj)
 
+    def read_uvh5(self, filename, run_check=True, check_extra=True,
+                  run_check_acceptability=True):
+        """
+        Read uvh5 file
+        """
```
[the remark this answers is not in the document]
plaplant: Good catch, this was just my initial placeholder docstring. I'll make a real one.

pyuvdata/uvh5.py
```diff
@@ -0,0 +1,252 @@
...
+        header['integration_time'] = self.integration_time
+        header['lst_array'] = self.lst_array
+        header['polarization_array'] = self.polarization_array
+        header['spw_array'] = self.spw_array
+        header['ant_1_array'] = self.ant_1_array
+        header['ant_2_array'] = self.ant_2_array
+
+        # treat nsample_array differently, because it is very large
+        # TODO: add filter options
+        header['nsample_array'] = self.nsample_array.astype(np.float32)
```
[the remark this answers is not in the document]
plaplant: That's a great idea, since I'll likely add filter options for it like the data and flags array. I'll reorganize.

bhazelton:
Looking good @plaplant. The travis build got wedged so I restarted it. It looks like you haven't added anything to the tutorial or readme. Are you planning to on this PR or do you not want to advertise the functionality yet?

plaplant:
I think I’d rather wait ~a week to update the docs and tutorial. We might have to update the installation instructions to make bitshuffle available to the users, but I’d like to have this core functionality in place so Jack and I can iterate.

bhazelton:
Ok, then once this is merged, can you make an issue to track that? Also any todos in the code should be issues as well.

plaplant:
Sure, happy to make some issues.

pyuvdata/uvdata.py
```diff
@@ -1698,6 +1701,66 @@ def write_miriad(self, filepath, run_check=True, check_extra=True,
                                 clobber=clobber, no_antnums=no_antnums)
         del(miriad_obj)
 
+    def read_uvh5(self, filename, run_check=True, check_extra=True,
+                  run_check_acceptability=True):
+        """
+        Read a UVH5 file.
```
dannyjacobs: Since this is our first attempt at splitting that previously assumed equality and there is at least one other existing hdf5 format we need to make super clear that UVH5 is "a pyuvdata object stored in an HDF5 container".  That sentence should be plastered everywhere, in all the major functions. That way people trying to figure out why they can't read their random hdf5 file and drop into one of the read functions will get the message.

pyuvdata/uvh5.py
```diff
@@ -0,0 +1,279 @@
...
+            for k in self.extra_keywords.keys():
+                extra_keywords[k] = self.extra_keywords[k]
+
+        # write out history
+        header['history'] = self.history
+
+        # write out data and flags
+        # TODO: add filter options for visdata (bitshuffle), flags, and
+        #     nsample_array (lzf compression)
+        dgrp = f.create_group("Data")
```
dannyjacobs: Jack might get into trouble here. Having to buffer an entire file in memory before writing it might be problematic.  Do you have an idea of how to go about doing streaming writes?

pyuvdata/uvdata.py
plaplant: Okay, I can make this clearer when I update the documentation.

pyuvdata/uvh5.py
plaplant: Yes, I believe I know how to accomplish this, and was going to implement it with the "partial read" functionality. If we need the "partial write" sooner, I can prioritize that.

PR #355, opened 2018-05-18
Added ephem2juldate and test

duncanjmr:
modified:   tests/test_uvdata.py
	modified:   uvdata.py

This pull request adds a function juldate2ephem for switching between julian dates and ephem objects. And tests for both juldate2ephem and ephem2juldate.


bhazelton:
I think @jpober agreed to review this. One comment I have is that I am currently working to replace the phasing code because it doesn't work properly. In that process, I am removing all uses of `ephem` in favor of `astropy` because `ephem` has fundamental problems (e.g. no leap second support) and is no longer supported.

PR #358, opened 2018-05-23
added utils.get_miriad_antpos

nkern:
This function takes a miriad data file and returns *only* the antennas and antenna positions. To achieve this, I moved chunks of `miriad.read_miriad` into separate class methods. This feature will be used in the hera pspec pipeline.

### DIFF
--- a/pyuvdata/miriad.py
+++ b/pyuvdata/miriad.py
@@ -29,7 +29,7 @@ class Miriad(UVData):
                   "polarization_array".format(pol=pol))
         return pol_ind
 
-    def read_miriad(self, filepath, correct_lat_lon=True, run_check=True, read_data=True,
+    def read_miriad(self, filepath, correct_lat_lon=True, run_check=True,
                     check_extra=True, run_check_acceptability=True, phase_type=None,
                     antenna_nums=None, ant_str=None, ant_pairs_nums=None,
                     polarizations=None, time_range=None):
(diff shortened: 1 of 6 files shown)


nkern:
@bhazelton so I originally took that approach but reached a dead-end when I realized that it seems like certain meta-data can only be read-in **with** the data itself, which means one essentially has to loop over the data to get the meta-data. I'm not sure of any other way to get those metadata, but if a more advanced miriad user does, it would be much appreciated. The compromise I took is that there are certain metadata (like the antenna positions) that I don't need to loop over the data for, so I figured I would just build a specific function for getting that information. 

What do you think?

bhazelton:
@nkern I'd suggest that you define the set that can be read in without looping as the lightweight metadata read for miriad (i.e. when you set read_data to False it only reads in the lightweight stuff). That's essentially the approach I took for uvfits -- anything I could read without reading the data were included when read_data is false. Then to get anything that's not defined on the object after such a read, the user has to call a read_data function. The nice thing about breaking it up that way is that you have the same object structure for the metadata and you can use that metadata to inform selections you might like to make on the data (e.g. which antennas exist in the data). Essentially it's the functionality you have now, but attached to the object, not detached.

aelanman:
Does this actually calculate antenna positions from UVWs if the antpos descriptor is missing?

nkern:
@aelanman it only accesses the `uv['antpos']` variable. the `uvw`s are tied with the data in miriad file format, and this is supposed to be a *fast* way to accessing this variable. so currently it just fails or returns None if the `antpos` variable doesn't exist

PR #360, opened 2018-05-24
more generalized beam interpolation functions

bhazelton:
Support interpolating to any frequency and location (not just healpix pixel centers). Infrastructure to allow for different interpolation functions and to keep track of which one was used.


bhazelton:
ping @zacharymartinot @jaguirre. Can you take a look at this and make sure the structure looks appropriate and extensible as needed?

PR #365, opened 2018-05-30
Support pseudo-stokes parameter beam as done in Kohn et al.2018

Chuneeta:
Implemented Figure 1 (equation 8) of Kohn et al. 2018.

### DIFF
--- a/pyuvdata/tests/test_uvbeam.py
+++ b/pyuvdata/tests/test_uvbeam.py
@@ -1334,7 +1334,10 @@ def test_healpix():
                               model_name='E-field pattern - Rigging height 4.9m',
                               model_version='1.0')
 
+    nt.assert_raises(ValueError, efield_beam.efield_to_pstokes, 'pI')
+
     efield_beam.az_za_to_healpix()
+    efield_beam.efield_to_pstokes()
     efield_beam.peak_normalize()
     pI_area = efield_beam.get_beam_sq_area("pI")
     pQ_area = efield_beam.get_beam_sq_area("pQ")
--- a/pyuvdata/tests/test_uvbeam.py
+++ b/pyuvdata/tests/test_uvbeam.py
@@ -1261,8 +1261,8 @@ def test_healpix():
 
     # Check sizes of output
     numfreqs = healpix_norm.freq_array.shape[-1]
-    beam_int = healpix_norm.get_beam_area()
-    beam_sq_int = healpix_norm.get_beam_sq_area()
+    beam_int = healpix_norm.get_beam_area(pol='xx')
+    beam_sq_int = healpix_norm.get_beam_sq_area(pol='xx')
     nt.assert_equal(beam_int.shape[0], numfreqs)
     nt.assert_equal(beam_sq_int.shape[0], numfreqs)
 
@@ -1270,9 +1270,9 @@ def test_healpix():
     dOmega = hp.nside2pixarea(healpix_norm.nside)
     npix = healpix_norm.Npixels
     healpix_norm.data_array = np.ones_like(healpix_norm.data_array)
-    nt.assert_almost_equal(np.sum(healpix_norm.get_beam_area()), numfreqs * npix * dOmega)
+    nt.assert_almost_equal(np.sum(healpix_norm.get_beam_area(pol='xx')), numfreqs * npix * dOmega)
     healpix_norm.data_array = 2. * np.ones_like(healpix_norm.data_array)
-    nt.assert_almost_equal(np.sum(healpix_norm.get_beam_sq_area()), numfreqs * 4. * npix * dOmega)
+    nt.assert_almost_equal(np.sum(healpix_norm.get_beam_sq_area(pol='xx')), numfreqs * 4. * npix * dOmega)
 
     # check XX and YY beam areas work and match to within 5 sigfigs
     XX_area = healpix_norm.get_beam_area('XX')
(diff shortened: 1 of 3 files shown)


pyuvdata/uvbeam.py
```diff
@@ -1084,6 +1084,60 @@ def __iadd__(self, other):
...
         return self
 
+    def _stokes_matrix(self, n):
+        """
+        Calculates Pauli matrices (where indices are reordered from the quantum mechanical convention
+        to an order which gives the ordering of the Stoes vector as [I, Q, U, V]) according to
+        https://arxiv.org/pdf/1401.2095.pdf.
+
+        Args :
+            n : Index for which to calculate the Pauli matrix; n should lie between 0 and 3 for the 4x4
```
bhazelton: We prefer not to use single letter variable names. Please pick something that has more meaning. I don't know reading this what 'n' is supposed to represent: is it a polarization index?

pyuvdata/uvbeam.py
```diff
@@ -1084,6 +1084,60 @@ def __iadd__(self, other):
...
+        Generates Mueller component as done in https://arxiv.org/pdf/1802.04151.pdf
+
+                    Mij = Tr(J sigma_i J sigma_j J^*),
+
+        where sigma_i and sigma_j are Pauli matrices.
+
+        Args :
+            jones : Jones matrices containing the electric field components for the dipole arms or polarizations.
+
+            i : Index at which the first Pauli matrix sigma_i is calculated.
```
bhazelton: Again, use more meaningful variable names.

pyuvdata/uvbeam.py
```diff
@@ -1084,6 +1084,60 @@ def __iadd__(self, other):
...
+
+            j : Index at which the second Pauli matrix sigma_j is calculated.
+
+        Returns :
+            npix numpy array containig the values for the corresponding Mueller component
```
bhazelton: This sounds like it's specific to Healpix. Is there a reason for that?

pyuvdata/uvbeam.py
```diff
@@ -1084,6 +1084,60 @@ def __iadd__(self, other):
...
+        Pj = self._stokes_matrix(j)
+
+        Mij = 0.5 * np.einsum('...ab,...bc,...cd,...ad', Pi, jones, Pj, jones.conj())
+        Mij = np.abs(Mij)
+
+        return Mij
+
     def _get_beam(self, pol):
         """
         Get the healpix beam map corresponding to the specififed polarization,
```
bhazelton: Why is this specific to Healpix?

pyuvdata/uvbeam.py
```diff
@@ -1102,54 +1156,56 @@ def _get_beam(self, pol):
...
-                if keys[0] in pol_array and keys[1] in pol_array:
-                    xx_ind = np.where(np.isin(pol_array, keys[0]))[0][0]
-                    yy_ind = np.where(np.isin(pol_array, keys[1]))[0][0]
-                    beam = 0.5 * (self.data_array[0, 0, xx_ind] + self.data_array[0, 0, yy_ind])
-                else:
-                    raise ValueError('Do not have the right polarization information')
             else:
                 raise ValueError('Do not have the right polarization information')
+        elif pol in [1, 2, 3, 4]:  # pseudo stokes
+            assert self.beam_type == 'efield', "Beam type must be efield for pseudo stokes {}".format(uvutils.polnum2str(pol))
```
bhazelton: The beam can in principal already have been converted to a power beam in pseudo stokes (e.g. using your conversion code)  then you'd have a power beam with the polarization array having values 1:4. In that case, I think you should be able just get the values from the beam.

pyuvdata/uvbeam.py
```diff
@@ -1102,54 +1156,56 @@ def _get_beam(self, pol):
...
 
         Returns:
           omega : float, integral of the beam across the sky [steradians]
         """
-        if self.beam_type != 'power':
-            raise ValueError('beam_type must be power')
+        if isinstance(pol, (str, np.str)):
+            pol = uvutils.polstr2num(pol)
+        if pol in [-5, -6, -7, -8]:
+            if self.beam_type != 'power':
```
bhazelton: I think users will find it confusing that to get 'xx' or 'yy' you need a power beam but to get a pseudo-stokes beam you need an efield beam. I think we should make it consistent and calculate whatever is needed to the extent possible. I think in general it should work for either type unless the information you need isn't present or has been lost.

pyuvdata/tests/test_uvbeam.py
```diff
@@ -1286,30 +1286,19 @@ def test_healpix():
...
-    pI_area = healpix_norm.get_beam_area('pI')
-    area1 = healpix_norm.get_beam_area(1)
-    nt.assert_true(np.allclose(I_area, pI_area))
-    nt.assert_true(np.allclose(I_area, area1))
-
     # Check that if pseudo-Stokes I (pI) is in the beam polarization_array, it just uses it
     healpix_norm.polarization_array = [1, 2]
-    nt.assert_almost_equal(np.sum(healpix_norm.get_beam_area()), 2. * numfreqs * npix * dOmega)
-    nt.assert_almost_equal(np.sum(healpix_norm.get_beam_sq_area()), 4. * numfreqs * npix * dOmega)
+    # nt.assert_almost_equal(np.sum(healpix_norm.get_beam_area()), 2. * numfreqs * npix * dOmega)
```
bhazelton: If these tests aren't needed, just delete them rather than commenting them out. Git will preserve the memory of them if we need.

Chuneeta: Deleted!

pyuvdata/uvbeam.py
Chuneeta: done!

pyuvdata/uvbeam.py
Chuneeta: done!

pyuvdata/uvbeam.py
Chuneeta: Ya right now it's specific to healpix as implemented in Kohn et al 2018, the general implementation still needs to be implemented.

pyuvdata/uvbeam.py
Chuneeta: implemented in Kohn et al 2018, general case implementation n progress.

pyuvdata/uvbeam.py
Chuneeta: I've included a separate function efield_to_pstokes which does that and then call be called in _get_beam.

pyuvdata/uvbeam.py
Chuneeta: True. This is no more the case. You can use efield_to_pseudo function to convert from efield to pstokes parameters/power.

bhazelton:
@Chuneeta I'd like to get this merged in, can you add some examples to the tutorial?

PR #367, opened 2018-06-08
Module to calculate UVWs from antenna locations

rlbyrne:


bhazelton:
@rlbyrne ok, I think this is looking good, but it's failing the Travis check because it has some pep8 formatting issues. Also, we need to add some tests for this function.

pyuvdata/tests/test_uvdata.py
```diff
@@ -2471,3 +2471,42 @@ def test_select_with_ant_str():
     uv2 = uv.select(ant_str=ant_str, inplace=inplace)
     nt.assert_count_equal(uv2.get_antpairs(), ant_pairs)
     nt.assert_count_equal(uv2.get_pols(), uv.get_pols())
+
+def test_juldate2ephem():
```
bhazelton: I think this function somehow got rescued in a rebase. It should not exist anymore, please take it out along with it's tests.

pyuvdata/uvdata.py
```diff
@@ -833,6 +833,50 @@ def phase_to_time(self, time, phase_frame='icrs', use_ant_pos=False):
...
+                raise ValueError('UVW calculation requires unphased data. '
+                                 'Use unphase_to_drift or set '
+                                 'allow_phasing=True.'
+                                 )
+        antenna_locs_ENU = uvutils.ENU_from_ECEF(
+            (self.antenna_positions + self.telescope_location).T,
+            *self.telescope_location_lat_lon_alt
+        ).T
+        uvw_array = np.zeros((self.baseline_array.size, 3))
+        for baseline in list(set(self.baseline_array)):
```
bhazelton: Note that the baselines are repeated often in the baseline array. You might want to only grab the unique ones, something like what we do in the `phase` function.

rlbyrne: "set" should be taking care of that

pyuvdata/tests/test_uvdata.py
rlbyrne: oops you're right

pyuvdata/uvdata.py
bhazelton: yes, you're right. I missed that when I read it the first time.

PR #371, opened 2018-06-15
Remove the external AIPY dependency.

pkgw:
This PR extracts AIPY's code for MIRIAD I/O and removes the AIPY dependency. It uses the latest AIPY codebase which should be compatible with both Python 2 and Python 3, so this should unlock us to start making pyuvdata compatible with both versions as well.


pkgw:
Thanks @plaplant !

@bhazelton care to review? Should other folks take a look at this?

bhazelton:
Yes, I can look at it on Monday.

PR #372, opened 2018-06-19
[DO NOT MERGE] Work on both Python 2 and Python 3

pkgw:
Work-in-progress for simultaneous compatibility with both Python 2 and Python 3.

Current status: installs on both platforms. Test suite passes on Python 2. Various failures on Python 3; from a quick scan, they are numerous but boil down to a much smaller set of actual problems.


pkgw:
Note that the first batch of commits could be separated out into its own PR of "keep the test suite passing on Python 2, but get ready to run on Python 3". Then a separate PR could focus on the changes to get things running on Python 3. Thoughts as to preferred approach?

pkgw:
Actually, yeah, that's what I'll do. Closing for now.

PR #377, opened 2018-06-27
Adds read_miriad_metadata to UVData

nkern:
Adds a read_miriad_metadata function to the UVData and Miriad objects for fast read-in of lightweight attributes associated with a miriad file.


bhazelton:
Thanks @nkern. Is this a replacement for your antpos PR?

nkern:
@bhazelton yes it is. I'll close that one..

pyuvdata/tests/test_miriad.py
```diff
@@ -67,6 +67,8 @@ def test_ReadMiriadWriteUVFits():
     miriad_uv = UVData()
     uvfits_uv = UVData()
     miriad_file = os.path.join(DATA_PATH, 'zen.2456865.60537.xy.uvcRREAA')
+    if not os.path.exists(os.path.join(DATA_PATH, 'test')):
+        os.mkdir(os.path.join(DATA_PATH, 'test'))
```
bhazelton: This is done in the `setup_package` function in `tests/__init__.py` so this shouldn't be necessary here. Was this fixing a particular problem?

pyuvdata/tests/test_miriad.py
```diff
@@ -153,19 +154,23 @@ def test_wronglatlon():
...
     uvtest.checkWarnings(uv_in.read_miriad, [telescopefile], {'run_check': False},
-                         nwarnings=4,
+                         nwarnings=5,
                          message=['Altitude is not present in Miriad file, and telescope',
                                   'Telescope location is set at sealevel at the '
                                   'file lat/lon coordinates. Antenna positions '
                                   'are present, but the mean antenna position',
-                                  telescopefile + ' was written with an old version of pyuvdata',
+                                  'This file was written with an old version of pyuvdata',
+                                  'This file was written with an old version of pyuvdata',
```
bhazelton: why are all these warnings doubled?

pyuvdata/tests/test_miriad.py
```diff
@@ -718,8 +733,9 @@ def test_rwrMiriad_antpos_issues():
     uvtest.checkWarnings(uv_in.read_miriad, [testfile], known_warning='miriad')
     uv_in.antenna_positions = None
     uv_in.write_miriad(write_file, clobber=True)
-    uvtest.checkWarnings(uv_out.read_miriad, [write_file],
-                         message=['Antenna positions are not present in the file.'])
+    uvtest.checkWarnings(uv_out.read_miriad, [write_file], nwarnings=2,
+                         message=['Antenna positions are not present in the file.',
+                                  'Antenna positions are not present in the file.'])
```
bhazelton: Why are the warnings doubled?

pyuvdata/tests/test_utils.py
```diff
@@ -11,6 +11,8 @@
 import numpy as np
 from pyuvdata.data import DATA_PATH
 import pyuvdata.utils as uvutils
+from pyuvdata.miriad import Miriad
```
bhazelton: Why do you need this import? It doesn't look like you added anything that uses it...

pyuvdata/tests/test_miriad.py
nkern: Yeah, this was failing for me locally. I guess for some reason my setup wasn't doing this for me? Anyways, doesn't hurt I guess to have this..

bhazelton: hmm.. I want to understand what's going on. I'll check out the branch and see whether I have the same problem. Were you testing on python 2 or 3 when you had the problem?

pyuvdata/tests/test_miriad.py
nkern: Ah, its because I now call `_load_antpos` twice in a `read_miriad` call: once at the beginning when I'm loading in the metadata, and once at the end when all of the heavy-weight metadata have been loaded in. The first call to `_load_antpos` isn't needed, but its done anyways b/c its lumped in with the `read_miriad_metadata` call (its very fast anyways so speed isn't a concern). But it means that any warning that the function raises is technically raised twice in a `read_miriad` call. We can change this if this is too annoying..

pyuvdata/tests/test_miriad.py
nkern: same here

pyuvdata/tests/test_utils.py
nkern: I may have added that in by accident..

pyuvdata/tests/test_miriad.py
bhazelton: It's not too big a deal. Could just the first one suffice (it makes sense to get it when you only read the metadata), or do you really need the later one?

pyuvdata/tests/test_miriad.py
bhazelton: When I comment out those lines and run `nosetests pyuvdata/tests/test_miriad.py` I don't get an error in either python 2 or 3, but I do get a warning that's not being caught...

pyuvdata/tests/test_miriad.py
nkern: Good question. Not entirely sure, but it looks like the `Nants_telescope` is set using `sorted_unique_ants`, the latter of which is constructed from the `ant_1_array` and `ant_2_arrray` (the heavy-weight metadata arrays). The second call to `_load_antpos` is fed this `sorted_unique_ants`, so it seems like the second call to `_load_antpos` has some added functionality that the first call cannot reproduce...

pyuvdata/tests/test_miriad.py
nkern: This was testing for python2. So  I don't get the error when I also run `nosetest test_miriad.py`, but if I enter an interactive session and try to run the tests via copy-paste (this is what I do to debug when tests fail), then I  get the issue. But I guess that makes sense if you are saying that this is handled in the `tests/__init__.py` file, which I don't call in an interactive session..

pyuvdata/tests/test_miriad.py
bhazelton: Ok, so here's what's going on there. Miriad doesn't understand the concept of antenna numbers -- it essentially requires that the index of the antenna position is the number of that antenna. The problem with that is that sometimes you don't have all the antennas present in a file (like for HERA where we haven't built some of them yet). You could re-number all the antennas to be consecutive, but then a given physical antenna could have a different number for different files which would drive people analyzing the data crazy. So instead we pad the antenna positions with zeros for the indices that do not correspond to existing antenna information. 

It is perfectly reasonable to have antenna position information for antennas that you don't have any data for -- that's why there's a difference between Nants_telescope & Nants_data.  So to calculate Nants_telescope we could just use the non-zero positions, except that some files seem to be missing antenna positions for antennas that *do* appear in the data. So to calculate Nants_telescope we use a union of the antennas that have non-zero positions and the antennas that are present in the data.

So if you just read the antpos variable without reading the data, you can't necessary down-select it to get just the valid antenna positions because you don't know if there are antennas in the data for which the antenna positions are set to zero for some reason.

So now that I understand the complexity, I think you're doing the right thing and we should just live with duplicated warnings (this is assuming that you have to read the data to know what antennas are present in the data, which I believe is true for miriad).

pyuvdata/tests/test_miriad.py
bhazelton: Ok. I'd prefer to remove this from the code, I think that copy and pasting tests into an interactive session is a useful debugging approach but we don't generally keep debugging code (like print statements) in the master branch of the repo.

pyuvdata/tests/test_miriad.py
nkern: Yeah, I agree with that summary.

nkern:
@bhazelton okay coverage added for those lines, and I should be done w/ this PR pending tests pass..

PR #385, opened 2018-07-02
Partial I/O for uvh5 files

### LINKED ISSUE #351: Support partially reading and writing UVH5 objects
We should add partial-reading functionality to UVH5 files. This already exists for UVFITS files, so we should support analogous functions.

plaplant:
This PR adds support for partial reading and partial writing of uvh5 files. The partial reading is very similar to that supported for miriad and uvfits files. The partial writing is new for the HDF5 format, and attempts to write the data to files as efficiently as possible. It also requires first initializing an empty file on disk, and then populating the data in it in pieces.

@bhazelton I'm planning on adding examples of the partial I/O to the tutorial. Let me know if you would like it to be a part of this PR, or I can make an issue and do it at a later date.

Fixes #351.

### DIFF
--- a/pyuvdata/tests/test_uvh5.py
+++ b/pyuvdata/tests/test_uvh5.py
@@ -14,6 +14,7 @@ import pyuvdata.utils as uvutils
 from pyuvdata.data import DATA_PATH
 import pyuvdata.tests as uvtest
 import warnings
+import h5py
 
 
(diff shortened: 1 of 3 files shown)


bhazelton:
@plaplant I approved it, please make the issue to track the tutorial additions.

PR #390, opened 2018-07-04
fix bug in select for single string antenna_name, flatten higher-dimensional arrays in select

bhazelton:
Fixes a bug where if a single antenna_name string was passed it treated the characters in the string as separate antenna names rather than using the whole string as the antenna name.

Also add asserts to prevent problems with nested lists selects.

### DIFF
--- a/pyuvdata/tests/test_uvdata.py
+++ b/pyuvdata/tests/test_uvdata.py
@@ -404,12 +404,20 @@ def test_select_blts():
 
     nt.assert_true(np.all(selected_data == uv_object2.data_array))
 
+    # check that it also works with higher dimension array
+    uv_object2 = copy.deepcopy(uv_object)
+    uv_object2.select(blt_inds=blt_inds[np.newaxis, :])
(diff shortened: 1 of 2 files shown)


adampbeardsley:
I think asserting that ndim=1 is fine. Are there any use cases where people would supply larger dim arrays and we should just flatten?

adampbeardsley:
@bhazelton I went ahead and approved this, but I'll leave it to you to merge pending what you think about my above comment.

adampbeardsley:
Yeah, I'm struggling to think of either a case where flattening is wrong, or an actual use case where higher dim arrays would be input. Either way, we should probably make the assertion error more clear. I think as is it just gives:
`AssertionError:`

bhazelton:
Ok, maybe I'll just change it to flatten the arrays. Then I think I don't need the assert.

bhazelton:
Does the flatten ordering matter?

bhazelton:
@adampbeardsley I changed all the asserts to flattens. Can you take another look when you get a chance?

PR #391, opened 2018-07-04
Allow two objects to be added together if one or both of them is an instance of a subclass of UVData

jsdillon:
I'm currently writing a subclass of UVData and this check is blocking me from loading in data from multiple files using my subclass.


adampbeardsley:
I’m confused by this. Shouldn’t the two objects be your subclass of uvdata? Or are you allowing uvdata objects to be added to your subclass? I think yours is the first case like this, but if we generate more subclasses, should this be the rule or the exception?

jsdillon:
My object is a subclass and I’m adding a uvdata to it. This happens during
list loading, which makes uvdata objects and concatenates them to my
object.

PR #395, opened 2018-07-07
allow read_data = False for single element lists of files in read_uvh5

jsdillon:
Often, it's easier to keep files to load in lists, even if they're length one. This slightly relaxes the requirement that you can't run lists of paths through read_uvh5 with read_data = False without needing to dive into the whole rabbit hole that is adding metadata-only uvdata objects.


jsdillon:
@plaplant done.

Issue #396, opened 2018-07-09
concatenating autocorr-only and crosscorr-only UVData objects fails


nkern:
If I have a `UVData` object (`uvd1`) with only autocorrs and I have another object (`uvd2`) with only cross-corr and I try to concatenate them as `uvd1 + uvd` I get the following error:
```---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-58-0d4b3fe87665> in <module>()
----> 1 uvd1 + uvd2

/Users/nkern/Software/anaconda2/lib/python2.7/site-packages/pyuvdata-1.2.1-py2.7-macosx-10.7-x86_64.egg/pyuvdata/uvdata.pyc in __add__(self, other, run_check, check_extra, run_check_acceptability, inplace)
    977         if run_check:
    978             this.check(check_extra=check_extra,
--> 979                        run_check_acceptability=run_check_acceptability)
    980 
    981         if not inplace:

/Users/nkern/Software/anaconda2/lib/python2.7/site-packages/pyuvdata-1.2.1-py2.7-macosx-10.7-x86_64.egg/pyuvdata/uvdata.pyc in check(self, check_extra, run_check_acceptability)
    357 
    358         super(UVData, self).check(check_extra=check_extra,
--> 359                                   run_check_acceptability=run_check_acceptability)
    360 
    361         # Check internal consistency of numbers which don't explicitly correspond

/Users/nkern/Software/anaconda2/lib/python2.7/site-packages/pyuvdata-1.2.1-py2.7-macosx-10.7-x86_64.egg/pyuvdata/uvbase.pyc in check(self, check_extra, run_check_acceptability)
    288                     if not accept:
    289                         raise ValueError('UVParameter ' + p + ' has unacceptable values. '
--> 290                                          + message)
    291 
    292         return True

ValueError: UVParameter _uvw_array has unacceptable values. Mean of abs values, 4.62852730015, is not in allowed range: (0.0, 0.0)
```

If, on the other hand, I use `uvd2 + uvd1` it works just fine.

This is likely because of lines 345-350 in `uvdata.py`. Can anyone enlighten me as to why this adjustment of the `_uvw_array.acceptable_range` is needed?

PR #397, opened 2018-07-09
changed uvdata._uvw_array.acceptable_range to (0, 1e8)

nkern:
Changes default `uvdata._uvw_array.acceptable_range` to `(0, 1e8)` and removes adjustment to `(0.0, 0.0)` in `uvdata.check()` when only autocorrs are present. This addresses the issue relating to concatenating autocorr-only and crosscorr-only `UVData` objects (#396).

Also adds tests in `UVData.check` to test for non-zero uvw coords of autocorrs, and near-zero uvw magnitudes of cross-corrs. Adds appropriate tests in `test_uvdata`.

### DIFF
--- a/pyuvdata/tests/test_uvdata.py
+++ b/pyuvdata/tests/test_uvdata.py
@@ -1199,6 +1199,7 @@ def test_add():
     uv2 = uv_cross + uv_auto
     nt.assert_equal(uv2.Nbls, uv_auto.Nbls + uv_cross.Nbls)
 
+
 def test_add_drift():
     uv_full = UVData()
     testfile = os.path.join(
--- a/pyuvdata/tests/test_uvdata.py
+++ b/pyuvdata/tests/test_uvdata.py
@@ -217,6 +217,22 @@ class TestUVDataBasicMethods(object):
                                                 == self.uv_object.ant_2_array)[0])
         nt.assert_true(self.uv_object.check())
 
+        # test auto and cross corr uvw_array
+        uvd = UVData()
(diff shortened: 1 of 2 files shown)


bhazelton:
Copying comment from Slack:
I would prefer to add a line to the check function that restores the acceptable range if there are cross-pols present. Note that the way the acceptable values check is implemented may not be intuitive. It’s not a check that every value in the array falls between the acceptable values, it’s a check that the mean of the absolute value falls between the acceptable values. This really should not be allowed to be zero if there are cross-antenna visibilities present, but it should not be allowed to be other than zero if there are only autos present. So I think the fix needs to be entirely in the check function. I can make the change if you want, but I like the new test you added and I agree it should pass.

nkern:
Okay, I'm happy to make the requested changes. Personally, I don't understand why an implicitly auto-corr only object needs to be restricted to being explicitly an only auto-corr object and vice-versa, but seems like the requested changes will fix my problem, so I'm happy.

bhazelton:
Thank you! I think the point is not to try to restrict objects, it's to make sure that the check function catches problems with data sets, which is what it is supposed to do. Your original change does pass all the tests, but it also makes the checking less likely to catch errors.

bhazelton:
If you would rather, I would be completely comfortable with a change that makes the check function not affect the acceptable values at all and adjusts the acceptable values as you did in your initial commit and then adds some extra code in the check function that would catch the errors I'm worried about -- non-zero uvws for autos and zero uvws for crosses.

nkern:
@bhazelton I think that makes more sense, so I'll add that if that's okay.

bhazelton:
Yep, that’s fine. In fact I think it’s better than what we had before.

Issue #398, opened 2018-07-09
get_data, get_flags, and get_nsamples do not properly handle conjugated baseline polarization


jsdillon:
Here's a minimal example:

```
from pyuvdata import UVData
from hera_cal.data import DATA_PATH
four_pol = [os.path.join(DATA_PATH, 'zen.2457698.40355.{}.HH.uvcA'.format(pol)) for pol in ['xx','yy','xy','yx']]
uvd = UVData()
uvd.read_miriad(four_pol)
print np.all(uvd.get_data(80, 81, 'XY') == np.conj(uvd.get_data(81, 80, 'XY')))
# Returns True, should be false
print np.all(uvd.get_data(80, 81, 'XY') == np.conj(uvd.get_data(81, 80, 'YX')))
# Returns False, should be True
```

bhazelton:
@adampbeardsley didn't your PR fix this issue? Should I close it?

adampbeardsley:
Yup, I just forgot to close.

PR #399, opened 2018-07-09
properly conjugate visibilities for cross-corr get_data

adampbeardsley:
Addresses Issue #398

### DIFF
--- a/pyuvdata/uvdata.py
+++ b/pyuvdata/uvdata.py
@@ -2172,8 +2172,8 @@ class UVData(UVBase):
                 blt_ind2 = np.array([], dtype=np.int64)
                 pol_ind2 = np.array([], dtype=np.int64)
             else:
-                pol_ind2 = np.where(self.polarization_array ==
-                                    uvutils.polstr2num(uvutils.conj_pol(key)))[0]
+                pol_ind2 = np.where(self.polarization_array
+                                    == uvutils.polstr2num(uvutils.conj_pol(key)))[0]
                 if len(pol_ind2) == 0:
                     raise KeyError('Polarization {pol} not found in data.'.format(pol=key))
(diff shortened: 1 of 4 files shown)


pyuvdata/uvdata.py
```diff
@@ -2159,28 +2159,48 @@ def _key2inds(self, key):
         Returns:
             blt_ind1: numpy array with blt indices for antenna pair.
             blt_ind2: numpy array with blt indices for conjugate antenna pair.
-            pol_ind: numpy array with polarization indices
+                      Note if a cross-pol baseline is requested, the polarization will
+                      also be reversed. e.g. (1, 2, 'xy') -> conj(2, 1, 'yx')
```
bhazelton: I find this comment confusing. I'm not sure when I'm getting reversed polarizations and whether that means they match what I asked for or not.

adampbeardsley:
@bhazelton I think I've addressed your comments, though I'm not sure how clear my doc string is.. I'm having a hard time articulating it right. Let me know what you think.
I actually changed the behavior for a couple of the places you requests tests for. If the user only supplies a polarization key, I don't think we should be returning the conjugate pol. So now if you ask for 'xy' you only get 'xy', and not conj('yx').

jsdillon:
Isn't there a much simpler way to do this? 

```
from pyuvdata.utils import polstr2num

def conj_pol(pol):
    '''Given V_ij^(pol), return the polarization of V_ji^(conj pol) such
    (V_ji^(conj pol))* = V_ij^(pol). This means xy -> yx and yx -> xy, but
    psuedo-Stokes visibilities are unaffected. Case is left unmodified.'''
    if polstr2num(pol) > 0:  # this includes Stokes and pseudo-Stokes visibilities
        return pol
    else:
        return pol[::-1]
```

PR #404, opened 2018-07-16
Rationalize coords

### LINKED ISSUE #402: Rationalize utility coordinate transform APIs
Currently, about have the transforms expect shapes like (Nlocations, 3) and the other half expect shapes like (3, Nlocations). This is really annoying can lead to confusion.

This will change the API for functions that are used widely by the HERA community, so we will need to prepare them for this change.

bhazelton:
Make all coordinate conversions use same axes order, add future deprecation warnings for old ordering.

fixes #402


bhazelton:
@nkern @philbull This is the PR to rationalize the coordinate transforms, with pending deprecation warnings for calls using the old shapes. Can one or both of you look at it and check that it won't cause you undue problems (other than getting the associated warnings?)

PR #411, opened 2018-07-20
fixes for new org

bhazelton:


bhazelton:
@dannyjacobs can you look at this?

PR #413, opened 2018-07-22
Update tutorial

### LINKED ISSUE #314: Update read_calfits documentation
`read_calfits` supports loading in a list of files (and internally concatenating them), but this is not clear from the documentation.

https://github.com/HERA-Team/pyuvdata/blob/1b666d0996d8a825e799ee2e951c99644e0afb50/pyuvdata/calfits.py#L351

bhazelton:
Adds generic read, more detail & explanation for partial reads of UVData files
Improves UVCal tutorial, adding missing info about adding objects and reading multiple files

fixes #314

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -19,9 +19,13 @@ a) miriad -> uvfits
   >>> from pyuvdata import UVData
   >>> UV = UVData()
 
-  # This miriad file is known to be a drift scan
+  # This miriad file is known to be a drift scan.
+  # Use either the file type specific read_miriad or the generic read function,
+  # optionally specify the file type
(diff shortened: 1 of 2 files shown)


bhazelton:
@plaplant I think that's a good idea. I need to think about it a little bit, because I think it's not restricted to only metadata (and for Miriad, what is read in with `read_data=False` is less than for uvfits or uvh5). So I think there needs to be some thinking about when to set `read_data=False` and when not to.

bhazelton:
@plaplant @nkern I think I'll make an issue to add this to this script. I'd like to not hold up these docs fixes too long and I think it's going to take awhile to work out the right way to handle this in general for the different file types.

PR #414, opened 2018-07-23
updates to miriad_to_uvfits.py given recent changes

nkern:
due to recent changes w/ phasing etc.


scripts/miriad_to_uvfits.py
```diff
@@ -37,16 +38,16 @@
 
     # read in file
```
bhazelton: This assumes the Miriad file is in drift mode, which is a good assumption for HERA but not for other data. The script will just error (not do incorrect things) if it's a phased file, but it'd be nice to check that it's in drift mode before attempting to phase it.

PR #419, opened 2018-07-24
Slim data for PyPI package

bhazelton:
The package is too large for pypi. I found a large ms file that wasn't being used and removed it.
I also fixed a couple of small testing problems that caused warnings and test data to be written to the wrong folder.


bhazelton:
Thanks @jpober. The coveralls report is confusing, there don't appear to be any new uncovered lines when I look at it in detail. I'm re-running the travis build to see if it clears up.

PR #422, opened 2018-07-25
Use the uvw_array tolerances on the extra checks

### LINKED ISSUE #420: Autocorr uvws are not close enough to zero when unphased
https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/cdb3344400ea0a723cf5fe0d436dfa70057674c4/pyuvdata/uvdata.py#L384-L386

These lines are picking up on autocorrelation uvws that are small but not zero to machine precision. When unphasing data, I don't think it's reasonable to require machine precision on uvw coordinates. My experience is that autocorrelation uvws of unphased data tend to be of order tens of microns.

This can be fixed by setting the `atol` (absolute tolerance) keyword on the np.isclose() function.

bhazelton:
fixes #420

### DIFF
--- a/pyuvdata/uvdata.py
+++ b/pyuvdata/uvdata.py
@@ -381,10 +381,14 @@ class UVData(UVBase):
 
         # check auto and cross-corrs have sensible uvws
         autos = np.isclose(self.ant_1_array - self.ant_2_array, 0.0)
-        if not np.all(np.isclose(self.uvw_array[autos], 0.0)):
+        if not np.all(np.isclose(self.uvw_array[autos], 0.0,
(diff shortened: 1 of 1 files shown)


bhazelton:
@aelanman @mwilensky768 can you verify that this fixes the problems you were seeing with autocorrelation uvw lengths?

mwilensky768:
My issue was actually related to cross-correlation uvw magnitudes being nearly zero, which was fixed by upgrading numpy from 1.13.3 to 1.14.5.

aelanman:
That fixes my issue. Thanks!

PR #430, opened 2018-08-03
add backwards compatibility for uvh5 files

bhazelton:


bhazelton:
@plaplant can you look at this and make a test that covers it? (I think that'll require tampering with the file with hd5 tools that you're more familiar with)

PR #432, opened 2018-08-05
enforce boolean flags, adds ordered kwarg to `antpair2ind` method, & integration_time dtype

### LINKED ISSUE #431: bug when getting conjugated bl flags
When trying to get baseline flags that are conjugated, `UVData` returns an integer array, rather than a boolean array

nkern:
to fix #431.

This also updates a `antpair2ind()` method to `UVData` to do unordered antpair slicing along `blts` axis, such that we can easily get any array of len baseline-time. Ex, in lieu of a `get_integrations` function, we can also just do `uvd.integration_time[uvd.antpair2ind(key, ordered=False)]`.

Also enforces the `integration_time` attribute to be `np.float64` upon read-in from every `read_*` function in `UVData`

### DIFF
--- a/pyuvdata/uvdata.py
+++ b/pyuvdata/uvdata.py
@@ -2537,11 +2537,9 @@ class UVData(UVBase):
                 "antpair2ind must be fed an antpair tuple or expand it as arguments"
             ant2 = ant1[1]
             ant1 = ant1[0]
-        if isinstance(ant1, tuple):
-            assert ant2 is None, \
+        else:
+            assert isinstance(ant1, (int, np.integer)), \
                 "antpair2ind must be fed an antpair tuple or expand it as arguments"
-            ant2 = ant1[1]
-            ant1 = ant1[0]
         assert isinstance(ordered, (bool, np.bool)), "ordered must be a boolean"
 
         # get indices
--- a/pyuvdata/uvdata.py
+++ b/pyuvdata/uvdata.py
@@ -2514,16 +2514,16 @@ class UVData(UVBase):
         else:
             return list(set(''.join(self.get_pols())))
 
-    def antpair2ind(self, ant1, ant2=None, pol=None, ordered=True):
+    def antpair2ind(self, ant1, ant2=None, ordered=True):
         """
-        Given an antenna pair (pol) key, return indices along the baseline-time axis.
+        Given an antenna pair key, return indices along the baseline-time axis.
         This will search for either the key as specified, or the key and its
         conjugate.
 
         Args:
-            ant1, ant2, pol:
-                Either an antenna-pair-pol key, or key expanded as arguments.
-                Example: antpair2ind( (10, 20) ) or antpair2ind(10, 20, 'xx')
+            ant1, ant2:
+                Either an antenna-pair key, or key expanded as arguments.
+                Example: antpair2ind( (10, 20) ) or antpair2ind(10, 20)
             ordered : Boolean, if True, search for antpair as provided, else
                 search for it and it conjugate.
 
@@ -2531,15 +2531,15 @@ class UVData(UVBase):
             inds: int-64 ndarray containing indices of the antpair along the
                 baseline-time axis.
         """
-        # check for expanded antpair-pol, or key
+        # check for expanded antpair or key
         if ant2 is None:
             assert isinstance(ant1, tuple), \
-            "antpair2ind must be fed an antpair tuple or expand it as arguments"
+                "antpair2ind must be fed an antpair tuple or expand it as arguments"
             ant2 = ant1[1]
             ant1 = ant1[0]
         if isinstance(ant1, tuple):
-            assert ant2 is None and pol is None, \
-            "antpair2ind must be fed an antpair tuple or expand it as arguments"
+            assert ant2 is None, \
+                "antpair2ind must be fed an antpair tuple or expand it as arguments"
             ant2 = ant1[1]
             ant1 = ant1[0]
         assert isinstance(ordered, (bool, np.bool)), "ordered must be a boolean"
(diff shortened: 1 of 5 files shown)


pyuvdata/uvdata.py
```diff
@@ -2615,9 +2616,17 @@ def _key2inds(self, key):
                 raise KeyError('Polarization {pol} not found in data.'.format(pol=key[2]))
         # Catch autos
         if np.array_equal(blt_ind1, blt_ind2):
-            blt_ind2 = np.array([])
+            blt_ind2 = np.array([], dtype=np.int64)
         return (blt_ind1, blt_ind2, pol_ind)
 
+    def get_blts_inds(self, key):
+        """
+        Given an unordered antpair key, return indices along baseline-time axis.
```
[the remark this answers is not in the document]
bhazelton: So I just discovered that Read The Docs doesn't include underscored methods in the class documentation. So I think that referring to the `_key2inds` documentation is not helpful (it appears that we do that several other places that should be fixed too. That can go in another PR though).

bhazelton:
I think in the interest of keeping our API as stable as possible for users that we should start the deprecation process for `antpair2ind` in favor of `get_blt_inds`. Putting an underscore in front of it breaks the API immediately and I'm not sure I see a need to keep it around long term since I agree that your new function is better. So I'd lean towards issuing a future deprecation warning (and including a note to that effect in the docstring) but leaving the name the same.

bhazelton:
Oh, sorry I missed that. So then I'd advocate for moving the functionality to `_ antpair2ind` but also leaving a `antpair2ind` method that issues a future deprecation warning and then calls `_ antpair2ind`. I just think we need to move towards warning users when the API is changing instead of just changing it.

nkern:
@bryna, okay I think that's also a good idea.

pyuvdata/uvdata.py
```diff
@@ -2615,9 +2616,17 @@ def _key2inds(self, key):
                 raise KeyError('Polarization {pol} not found in data.'.format(pol=key[2]))
         # Catch autos
         if np.array_equal(blt_ind1, blt_ind2):
-            blt_ind2 = np.array([])
+            blt_ind2 = np.array([], dtype=np.int64)
         return (blt_ind1, blt_ind2, pol_ind)
 
+    def get_blts_inds(self, key):
```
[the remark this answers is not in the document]
nkern: I actually entirely agree w/ that suggestion, and am glad you raised it to save me a keystroke!

pyuvdata/uvdata.py
```diff
@@ -2508,12 +2508,19 @@ def get_feedpols(self):
...
-        Get blt indices for given (ordered) antenna pair.
+        Get blt indices for given (ordered) antenna pair. See self.get_blt_inds for
+        indices given an unordered pair.
         """
         return np.where((self.ant_1_array == ant1) & (self.ant_2_array == ant2))[0]
 
+    def antpair2ind(self, *args):
+        """ Backwards Compatability method, but will soon be deprecated """
+        warnings.warn("UVData.antpair2ind will soon be moved to UVData._antpair2ind.\n"
+                      "Try using UVData.get_blt_inds() as a replacement function.")
```
bhazelton: Can you make the type of this warning a `PendingDeprecationWarning`? I think that is a little clearer

adampbeardsley:
I disagree that having a function that does ordered slicing is not user friendly. There are plenty of situations where you want to know the order of the antennas. It seems to me the cleaner thing would be to add an option to `antpair2ind` that allows for non-ordered antenna pairs. That's my 2 cents.

bhazelton:
Sounds like a reasonable compromise with fewer API implications.

nkern:
@adampbeardsley sounds good! I guess I was more complaining about the *lack* of something that does do unordered slicing, not necessarily the availability of an ordered slicer.

nkern:
There were a few things I needed to tidy up, but now this should be good to go!

pyuvdata/uvdata.py
```diff
@@ -2508,11 +2508,46 @@ def get_feedpols(self):
...
+            ordered : Boolean, if True, search for antpair as provided, else
+                search for it and it conjugate.
+
+        Returns:
+            inds: int-64 ndarray containing indices of the antpair along the
+                baseline-time axis.
+        """
+        # check for expanded antpair or key
+        if ant2 is None:
+            assert isinstance(ant1, tuple), \
```
plaplant: Just to add a comment on this latest version: `assert` statements are not as robust as checking conditions with `if` and then raising an error, because one can run python in such a way (with the `-O` flag) that `asssert`s are completely ignored: https://stackoverflow.com/a/2167639.

`assert` could be fine for doing basic self-consistency checks, but in this case we'd rather raise an error if the user doesn't provide the right arguments.

nkern: Hmm.. but i would argue these are self-consistency checks, in that, if I get rid of these `assert` statements and the user knows exactly how to use this function,  then nothing changes. These are only to help the novice user (and occasional advanced-user mistake) by enacting "self-consistency" checks on the data types: i.e. trying to be defensive in my programming.

nkern: another way of saying it: the function will still error if given malformed inputs, but it will give a slightly less informative error. however, core functionality is not taken away if these assert statements vanished

bhazelton: I went off and read a bit about using assertions vs exceptions in python. The consensus largely seems to be that assertions are intended for the developers of the package (indicating that they should chang/fix things), while Exceptions, especially ValueError, are for runtime issues caused by bad inputs by users.

> Assertions are for programmers. Exceptions are for errors that appear on runtime.

plaplant: I agree that nothing changes if the assert statements go away, but my general philosophy is that `assert` should be used to guarantee internal consistency of functions. If you're working directly with user input (which is the case here), then I think it's much safer to raise exceptions rather than waiting for the function to fail.

For instance, if this code were optimized and the user called `uvd.antpair2ind(0)`, instead of your helpful error message, they get `TypeError: 'int' object has no attribute '__getitem__'`, which would be thoroughly cryptic. I think it's better to raise an error in that case.

nkern:
Thanks for the detailed reviews! feel free to merge once this passes!

PR #434, opened 2018-08-07
Check extra metadata in add function

### LINKED ISSUE #433: bug with integration_time when adding polarizations w/ diff integration time
Recently, the `UVData.integration_time` was broadcasted to a vector from a scalar, but that vector only spans the `blts` axis. It should in principle extend to the `polarization` axis as well. One can, for eaxmple, add two `UVData` objects w/ different `integration_time` vectors and different polarizations, and one is `integration_time` is chosen over the other, but the polarizations are added just fine, which is a bug.

plaplant:
This PR ensures that additional metadata is checked if two UVData objects are being added, but not along the blt axis. It was possible for the objects to have different values, and lead to an inconsistent output.

Fixes #433.

### DIFF
--- a/pyuvdata/tests/test_uvdata.py
+++ b/pyuvdata/tests/test_uvdata.py
@@ -1549,6 +1549,7 @@ def test_break_add():
 
     # Wrong class
     uv1 = copy.deepcopy(uv_full)
+    uv1.select(freq_chans=np.arange(0, 32))
     nt.assert_raises(ValueError, uv1.__iadd__, np.zeros(5))
 
     # One phased, one not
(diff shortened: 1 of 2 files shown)


bhazelton:
I'd like to loop @adampbeardsley into this discussion since he wrote this function initially. I have a memory that we had a mechanism to try to handle this kind of thing and I just want to make sure we're being consistent in how we do it for all the nblt length vectors.

adampbeardsley:
This follows the mechanism I originally used. I think I had assumed if the blt arrays are equal, then the integration times would be (which is not necessarily true). I also assumed that if the phasing was set the same, the uvw array should be the same, but it doesn't hurt to check.

PR #438, opened 2018-08-10
Expanded compatibility of read_cst

anguta:


bhazelton:
I'm a little worried that files with 'dir' in their column names and dBi units are reporting directivities, which is different than the other kinds of files I've been working with. I think we need to be careful about this.

anguta:
I distinguish between the 'dir' and 'v' (which needs to be squared) case in lines 216 and 217 of cst_beam.py.

bhazelton:
The files this code was written to read include both efield and power values (in different columns). As we discussed on Slack, directivity files need to have the appropriate normalization set.

anguta:
I've changed the code so that when e-fields and powers specified in directivities, the normalization is set to the appropriate normalization (solid angle -- which integrates to 4\pi steradians).

pyuvdata/cst_beam.py
```diff
@@ -177,6 +176,11 @@ def read_cst_beam(self, filename, beam_type='power', feed_pol='x',
         self.axis2_array = theta_axis
         self.Naxes2 = self.axis2_array.size
 
+        if np.any(np.core.defchararray.find(np.core.defchararray.lower(np.array(column_names)), 'abs(dir.)') != -1):
+            self.data_normalization = 'solid_angle'
+        elif np.any(np.core.defchararray.find(np.core.defchararray.lower(np.array(column_names)), 'abs(v)') != -1):
+            self.data_normalization = 'physical'
+
```
bhazelton: It appears to me that you're doing this search twice, here and below where you define `data_col`. I think you should do it once to figure out whether it's a directivity or not and then use the appropriate string when you define `data_col`. That limits the code replication and ensures that you can't get a weird combination (mixed column types).

pyuvdata/cst_beam.py
```diff
@@ -204,8 +208,17 @@ def read_cst_beam(self, filename, beam_type='power', feed_pol='x',
...
+                                              np.core.defchararray.find(np.core.defchararray.lower(np.array(column_names)), 'abs(dir.)') != -1))[0][0]
+            power_beam1 = data[:, data_col].reshape((theta_axis.size, phi_axis.size), order='F')
+
+            mag_dbi = 'dbi' in np.core.defchararray.lower(units)[data_col]
+
+            if mag_dbi:
+                power_beam1 = 10. ** (power_beam1 / 10.)
+
+            if 'abs(v)' in column_names[data_col]:
+                power_beam1 = power_beam1 ** 2.
```
bhazelton: I think the reason you're not squaring the dBi value is because the factor of 10 (vs 20) in the denominator of the exponent effectively takes care of that for you. Can you make that explicit in a comment for the future reader?

Also this logic is slightly funny in that you could hit both of these if statements, but I don't think that's actually physical -- I don't believe directivities can be express in voltages. Maybe we need some logic to catch a case that doesn't make sense followed by an if/elif that makes it clear that there are only 2 options here.

pyuvdata/cst_beam.py
```diff
@@ -237,6 +255,13 @@ def read_cst_beam(self, filename, beam_type='power', feed_pol='x',
             theta_phase = theta_phase.reshape((theta_axis.size, phi_axis.size), order='F')
             phi_phase = phi_phase.reshape((theta_axis.size, phi_axis.size), order='F')
 
+            if theta_mag_dbi:
+                theta_mag = 10. ** (theta_mag / 20.)
+                theta_phase *= .5
```
bhazelton: Why are you dividing the phase by 2 here?

pyuvdata/cst_beam.py
```diff
@@ -237,6 +255,13 @@ def read_cst_beam(self, filename, beam_type='power', feed_pol='x',
...
+            if phi_mag_dbi:
+                phi_mag = 10. ** (phi_mag / 20.)
+                phi_phase *= .5
```
bhazelton: same question

pyuvdata/tests/test_uvbeam.py
```diff
@@ -146,6 +146,21 @@ def test_errors():
...
+                              feed_version='0.1', feed_pol=['x'],
+                              model_name='E-field pattern - f/D=0.25',
+                              model_version='1.0')
+    power_beam = UVBeam()
+    power_beam.read_cst_beam([os.path.join(DATA_PATH, 'hirax_420_MHz.txt')], beam_type='power', frequency=[420e6],
+                             telescope_name='TEST', feed_name='bob',
+                             feed_version='0.1', feed_pol=['x'],
+                             model_name='E-field pattern - f/D=0.25',
+                             model_version='1.0')
+
```
bhazelton: You should test that when the efield beams are converted to power that they match the power beam read from the file.

pyuvdata/cst_beam.py
bhazelton: Actually, I'm confused by something else here. It looks like you're using the power column names to decide on the normalization whether or not you're reading in a power or an efield beam. Is that the right behavior? Or should the units (dbi) be used as an indicator? What if the power wasn't written to the file along with the efield?

anguta:
efield -> power yields normalization agreement at the 1% level in the main-lobe.

pyuvdata/cst_beam.py
anguta: You're right. I think I accidentally assumed that dBi simply stood for dB but it actually stands for decibal Isotropic which is unambiguously referring to solid angle normalization. I should therefor eliminate this search and normalize the solid angle if units are in "dBi" and physical if units are in "V". I've removed this extraneous search and changed the normalization setting to be dependent on "mag_dbi". I've also added a check that the units are in dBi or V and raise a type error if unexpected units come up so we can explicitly add them as needed.

pyuvdata/cst_beam.py
anguta: I've added a comment and I've moved the squaring into an elif.

pyuvdata/cst_beam.py
anguta: I'm not sure why I did that. I've removed it.

pyuvdata/cst_beam.py
anguta: See above.

bhazelton:
So I had a go at reading in Nic Fagnioni’s directivity file that should line up with one of our existing test files, but I can’t get them to line up.
the first issue is that Nic’s directivity files have no units listed
so we may need to revisit that, but for now I just added the units by hand.
then I read it and the test file in and tried to figure out what to do to get the test file normalized the same way
here’s my code:
    ```# test that directivity file matches other file
    dir_beam = UVBeam()
    dir_beam.read_cst_beam(directivity_file, beam_type='power', frequency=150e6,
                           feed_pol='y', telescope_name='TEST',
                           feed_name='bob', feed_version='0.1',
                           model_name='E-field pattern - Rigging height 4.9m',
                           model_version='1.0')

    phys_beam = UVBeam()
    phys_beam.read_cst_beam(cst_files[0], beam_type='power', frequency=150e6,
                            feed_pol='y', telescope_name='TEST',
                            feed_name='bob', feed_version='0.1',
                            model_name='E-field pattern - Rigging height 4.9m',
                            model_version='1.0')

    peak_beam = copy.deepcopy(phys_beam)
    peak_beam.peak_normalize()
    peak_beam_integral = np.sum(peak_beam.data_array, axis=(4, 5))

    peak_beam_integral = peak_beam_integral[:, :, :, :, np.newaxis, np.newaxis]
    peak_beam_integral = np.tile(peak_beam_integral, (1, 1, 1, 1, peak_beam.Naxes2, peak_beam.Naxes1))

    normalized_data_array = peak_beam.data_array / peak_beam_integral
    print(normalized_data_array[0:10])
    print(dir_beam.data_array[0:10])

    nt.assert_true(np.allclose(normalized_data_array, dir_beam.data_array))```
the `normalized_data_array` and `dir_beam.data_array` are off by 32 orders of magnitude :rolling_on_the_floor_laughing:
so I’m clearly doing it wrong. Do you have the math?

bhazelton:
closing per off-line discussion with @anguta

PR #442, opened 2018-08-22
Moved polarization dicts and lists in utils.py into constants that ca…

AaronParsons:
…n be accessed externally by anyone using pyuvdata for its polarization functionality.

One outstanding issue is that there seems to be an internal inconsistency about whether pseudo-stokes ('pI', 'pU', etc.) have the preceding 'p' capitalized or not.  Could/should we be consistent?

### DIFF
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -34,6 +34,27 @@ else:
     def _bytes_to_str(b):
         return b.decode('utf8')
 
+# polarization constants
+# XXX is the 'p' capitalized or not?
+POLSTR_LIST = ['YX', 'XY', 'YY', 'XX', 'LR', 'RL', 'LL', 'RR', '', 'pI', 'pQ', 'pU', 'pV']
+POLSTR_DICT = {'PI': 1, 'PQ': 2, 'PU': 3, 'PV': 4,
+           'I': 1, 'Q': 2, 'U': 3, 'V': 4,
+           'RR': -1, 'LL': -2, 'RL': -3, 'LR': -4,
+           'XX': -5, 'YY': -6, 'XY': -7, 'YX': -8}
+
+CPOL_DICT = {'XX': 'XX', 'YY': 'YY', 'XY': 'YX', 'YX': 'XY',
+             'JXX': 'jxx', 'JYY': 'jyy', 'JXY': 'jyx', 'JYX': 'jxy',
+             'RR': 'RR', 'LL': 'LL', 'RL': 'LR', 'LR': 'RL',
+             'JRR': 'jrr', 'JLL': 'jll', 'JRL': 'jlr', 'JLR': 'jrl',
+             'I': 'I', 'Q': 'Q', 'U': 'U', 'V': 'V',
+             'PI': 'pI', 'PQ': 'pQ', 'PU': 'pU', 'PV': 'pV'}
+
+JSTR_LIST = ['jyx', 'jxy', 'jyy', 'jxx', 'jlr', 'jrl', 'jll', 'jrr']
+JSTR_DICT = {'jxx': -5, 'jyy': -6, 'jxy': -7, 'jyx': -8,
+         'xx': -5, 'x': -5, 'yy': -6, 'y': -6, 'xy': -7, 'yx': -8,  # Allow shorthand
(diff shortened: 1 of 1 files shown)


pyuvdata/utils.py
```diff
@@ -34,6 +34,27 @@ def _str_to_bytes(s):
     def _bytes_to_str(b):
         return b.decode('utf8')
 
+# polarization constants
+# XXX is the 'p' capitalized or not?
```
bhazelton: I can be persuaded on the capitalization issue, but I lean slightly toward a lower case p. I think it is capitalized in the dict for the purposes of the `polstr2num` function that compares the keys to an input string that is called with `.upper()`, although interestingly `.lower()` is used in the `jstr2num` function. Let's pick a capitalization scheme and then make the functions do the comparisons properly.

pyuvdata/utils.py
```diff
@@ -34,6 +34,27 @@ def _str_to_bytes(s):
...
+POLSTR_LIST = ['YX', 'XY', 'YY', 'XX', 'LR', 'RL', 'LL', 'RR', '', 'pI', 'pQ', 'pU', 'pV']
```
bhazelton: I think we should use reversed dicts rather than these lists. Since the goal is to pull these out to make them useful to other codes, things like the need to add 8 to the value of the list location and the placeholder value in the `POLSTR_LIST` make it hard for other codes to use them correctly.

pyuvdata/utils.py
```diff
@@ -34,6 +34,27 @@ def _str_to_bytes(s):
...
+               'I': 1, 'Q': 2, 'U': 3, 'V': 4,
+               'RR': -1, 'LL': -2, 'RL': -3, 'LR': -4,
+               'XX': -5, 'YY': -6, 'XY': -7, 'YX': -8}
+
+CPOL_DICT = {'XX': 'XX', 'YY': 'YY', 'XY': 'YX', 'YX': 'XY',
+             'JXX': 'jxx', 'JYY': 'jyy', 'JXY': 'jyx', 'JYX': 'jxy',
+             'RR': 'RR', 'LL': 'LL', 'RL': 'LR', 'LR': 'RL',
+             'JRR': 'jrr', 'JLL': 'jll', 'JRL': 'jlr', 'JLR': 'jrl',
+             'I': 'I', 'Q': 'Q', 'U': 'U', 'V': 'V',
+             'PI': 'pI', 'PQ': 'pQ', 'PU': 'pU', 'PV': 'pV'}
```
bhazelton: I wonder whether we really need to have all these duplicates based on case. I think we should be able to fix the functions to deal with case properly.

bhazelton:
We discussed the capitalization convention on our telecon today and decided on lowercase `p` for the pseudo-Stokes polarizations and lower case for the linear and circular polarizations. We decided to use a capital `J` for the Jones matrix elements to help convey that they are matrix elements.

bhazelton:
I have made some changes based on these discussions, I will push them to this branch and you can see if they meet your needs.

AaronParsons:
@bhazelton I'm not sure what broke the unit tests, (it looks like a capitalization difference in the tutorial) but can I ask for your help in shepherding this pull request through to a merge?

Thanks.

bhazelton:
Yes, I'll fix it up.

bhazelton:
@adampbeardsley are you willing to take a look at this PR?

bhazelton:
I'm fixing the issues @adampbeardsley pointed out, should be ready soon.

pyuvdata/tests/test_uvdata.py
```diff
@@ -2170,6 +2170,8 @@ def test_parse_ants():
     ant_pairs_nums, polarizations = uv.parse_ants(ant_str)
     ant_pairs_expected = [(1, 3)]
     pols_expected = [-4]
+    print(ant_pairs_nums)
+    print(polarizations)
```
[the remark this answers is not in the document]
bhazelton: yes, thanks!

bhazelton:
@adampbeardsley Thanks for your review, I made the STR2NUM dicts have integer keys, which is definitely the right thing to do, and removed the print statements.

jsdillon:
Thanks all!

PR #456, opened 2018-09-13
Redundancies

### LINKED ISSUE #417: Convenience method for extracting redundant baseline groups
It would be very useful to have a method on `UVData` objects that can retrieve a list of baselines within a redundant group. We have a few different custom functions to do this already -- one in `hera_cal` and one in `hera_pspec` I believe -- but it would be useful to have this as part of `UVData` to avoid needing to maintain this functionality externally.

For reference, the implementations we currently have select redundant baselines based on antenna separation and angle, within some tolerance.

aelanman:
Adds functions for finding redundant baselines within a specified tolerance. I decided to write a new method instead of adding hera_cal as a dependency, but I've compared the results of this with those of `hera_cal.redcal.get_pos_reds`, and with some careful adjustment they match. There is a key difference between the redundancy conventions that should be considered before merging.

- `get_pos_reds` finds redundant *antenna pairs* under the condition that u>0, v>0 (if u=0), or w>0 (if u=v=0. Thus, none of the redundant groups have a mean `uvw` with negative `u`, etc. However, this means that some of the antenna pairs have `ant_1 < ant_2` and so `uvd.antnums_to_baseline(ant_1, ant_2)` will not be present in the baseline_array.
- `get_baseline_redundancies` finds redundant baselines through their uvw coordinates in the file. This ensures that ant_1> ant_2 for all baselines found, but allows for negative u, v, and w coordinates. It will find more redundant groups but have fewer baselines in each.

I'd argue that for my purposes the second convention is more useful, because I'm interested in simply selecting one baseline from each redundant group, using the uvdata object in a simulation, and then copying over the calculated visibility across redundant baselines. The first method would require keeping track of which baselines have to be flipped within each group... but I'd like to hear what other uses this will have. I could probably revise `get_baseline_redundancies` to optionally return antenna pairs in the same scheme as `get_pos_reds`.

This does not completely fix Issue 417, but it's a start.

### DIFF
--- a/pyuvdata/tests/test_utils.py
+++ b/pyuvdata/tests/test_utils.py
@@ -351,3 +351,27 @@ def test_deprecated_funcs():
                                                      uvversion.version],
                          category=DeprecationWarning,
                          message='The combine_histories function is deprecated')
+
+
+def test_redundancy_finder():
+    """
+        Confirm that get_redundancies returns baselines that have the same length.
+        Should probably check orientation too.
+        Will need a test file that is close to redundant, but not perfectly so.
+    """
+    uvd = pyuvdata.UVData()
+    uvd.read_uvfits(os.path.join(DATA_PATH, 'hera19_8hrs_uncomp_10MHz_000_05.003111-05.033750.uvfits'))
+    uvd.select(times=uvd.time_array[0])
+    uvd.unphase_to_drift()   # uvw_array is now equivalent to baseline
+
+    bl_error_tol = 0.5  # meters
+
+    baseline_groups, vec_bin_centers, lens = uvutils.get_redundancies(uvd.baseline_array, uvd.uvw_array, bl_error_tol=bl_error_tol)
+
+    for gi, gp in enumerate(baseline_groups):
+        for bl in gp:
+            bl_ind = np.where(uvd.baseline_array == bl)
+            bl_vec = uvd.uvw_array[bl_ind]
+
+            print(np.dot(bl_vec, vec_bin_centers[gi]), lens[gi]**2)
+            nt.assert_true(np.isclose(np.sqrt(np.dot(bl_vec, vec_bin_centers[gi])), lens[gi], atol=bl_error_tol))
--- a/pyuvdata/tests/test_utils.py
+++ b/pyuvdata/tests/test_utils.py
@@ -361,7 +361,6 @@ def test_redundancy_finder():
     """
     uvd = pyuvdata.UVData()
     uvd.read_uvfits(os.path.join(DATA_PATH, 'hera19_8hrs_uncomp_10MHz_000_05.003111-05.033750.uvfits'))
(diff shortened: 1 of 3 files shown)


pyuvdata/utils.py
```diff
@@ -682,3 +682,99 @@ def _combine_histories(history1, history2):
...
+        baseline_groups: list of arrays of redundant baseline indices
+        vec_bin_centers: List of vectors describing redundant group centers
+        lengths: List of redundant group baseline lengths in meters
+    """
+    Nants = antenna_numbers.shape[0]
+
+    bl_inds = []
+    bl_vecs = []
+
+    def antnum_to_baseline(ant1, ant2):
```
bhazelton: This is replicating existing code, we should avoid doing that. If the issue is that you want to put this functionality in utils rather than uvdata, make a function in utils that is called by the uvdata method.

pyuvdata/tests/test_utils.py
```diff
@@ -351,3 +351,55 @@ def test_deprecated_funcs():
                                                      uvversion.version],
                          category=DeprecationWarning,
                          message='The combine_histories function is deprecated')
+
+
+def test_redundancy_finder():
+    """
+        Confirm that get_baseline_redundancies returns baselines that have the same length.
+        Should probably check orientation too.
```
bhazelton: Is this comment out of date? It looks like you are checking orientation because you're taking a dot product.

aelanman: Thanks, I missed that. I'll fix it.

bhazelton:
I hate to pick at this, but I think now that you've moved the core of `antnums_to_baseline` to utils that maybe we should also do a similar thing for `baseline_to_antnums`.

aelanman:
No problem. That makes sense.

pyuvdata/uvdata.py
```diff
@@ -474,7 +474,7 @@ def set_telescope_params(self, overwrite=False):
             raise ValueError('Telescope {telescope_name} is not in '
                              'known_telescopes.'.format(telescope_name=self.telescope_name))
 
-    def baseline_to_antnums(self, baseline):
+    def baseline_to_antnums(self, baseline, attempt256=False):
```
bhazelton: one more nitpick: this method doesn't need the `attempt256` keyword.

aelanman: Ah sorry, thought I'd gotten them all. Thanks for reviewing!

aelanman:
I'm not sure why this says coverage decreased... my changes shouldn't have touched any of the lines it's reporting.

bhazelton:
The issue is the (known) coveralls parallel build issue. I fixed it in another PR, when that gets merged it'll be happier. For now, I just looked at the actual coveralls page and it looks fine. All your new code is covered and there weren't any existing covered lines that lost coverage.

PR #458, opened 2018-09-16
Conjugate redundancy

### LINKED ISSUE #457: document the new redundant baseline functionality in the tutorial
Need some examples of how to use it in the tutorial.

@aelanman can you take this on?

aelanman:
Adds the option `with_conjugates` to to `utils.get_baseline_redundancies` and `utils.get_antenna_redundancies`. This option flips baselines to have the same orientation before finding redundant groups, and returns a list of which baselines were flipped.

The result is the same redundant groups returned by `hera_cal.get_pos_reds`.

Also includes a description of the redundancy methods in the tutorial.

Fixes #457


aelanman:
An interesting bug came up while working on this... Apparently there are different floating point errors between python3 and python2. Since the test file is unphased before going through the redundancy finder, there are errors of order 10^-6 in the baseline positions. Baselines get flipped when u<0 or , if u==0, when v<0 (and if u==v==0, then if w< 0 it's flipped). But in this case, baselines with u very close to zero were ending up with some small and positive and some small and negative, which split a redundant group in half. This only happened in python3.

I've put in a unit test to check for this case, but it might be worth thinking a little more about how we're handling floating point errors from phasing/unphasing.

bhazelton:
That fits with the behavior we saw in pyuvsim where we needed to unphase using the antenna locations in order to get the simulations done in python 3 to match the values in the saved file (presumably done in python 2). Note that we also saw something similar between linux & mac machines in python 2, might be worth checking that out as well.

I'm not entirely sure what to do about this problem globally...

docs/tutorial.rst
```diff
@@ -777,6 +777,49 @@ done after the read, which does not save memory.
...
+    # The with_conjugates option includes baselines that are redundant when reversed.
+    # If used, the conjugates list will contain a list of indices of baselines that must be flipped to be redundant.
+    >>> baseline_groups, vec_bin_centers, lengths, conjugates = uvutils.get_baseline_redundancies(uvd.baseline_array, uvd.uvw_array, tol=tol, with_conjugates=True)
+    >>> print(len(baseline_groups))
+    31
+
+    # Alternatively, antenna positions can be used.
+    >>> baseline_groups, vec_bin_centers, lengths, conjugates = uvutils.get_antenna_redundancies(uvd.antenna_numbers, uvd.antenna_positions, tol=tol, include_autos=True)
+    >>> print(len(baseline_groups))
+    52
```
bhazelton: The fact that this method seems to find the same number of groups as the baseline method _without_ conjugates confuses me -- it's not clear what ordering is considered to be unconjugated vs conjugated. We should at least clarify that point in the tutorial text.

pyuvdata/utils.py
```diff
@@ -816,6 +837,7 @@ def get_antenna_redundancies(antenna_numbers, antenna_positions, tol=1.0, includ
         antenna_positions:  Antenna position vectors in meters. Cartesian frame  (float), shape (Nants, 3)
         tol = (float)   Redundancy tolerance in meters. (float)
         include_autos =  Include autocorrelations (bool). Default is false
+        with_conjugates = Include baselines that are redundant if flipped. Default is false.
```
bhazelton: This is confusing because it's unclear what the conjugated or unconjugated means for antennas. I'd probably lean toward always getting all possible pairs, but at a minimum we need to document what direction is assumed (i.e. is it the vector from the smaller to larger antenna number or the other way around?).

pyuvdata/utils.py
```diff
@@ -746,26 +747,46 @@ def antnums_to_baseline(ant1, ant2, Nants_telescope, attempt256=False):
...
+        baseline_ind_conj: List of baselines that are redundant when reversed
 
     """
     Nbls = baseline_inds.shape[0]
 
     if not baseline_vecs.shape == (Nbls, 3):
         raise ValueError("Baseline vectors must be shape (Nbls, 3)")
 
+    baseline_vecs = copy.copy(baseline_vecs)              # Protect the vectors passed in.
+    baseline_vecs = np.around(baseline_vecs, decimals=4)  # Round to nearest 0.1mm, to avoid floating errors from unphasing.
```
bhazelton: I'm a little uncomfortable with hardcoding this, but I understand the problem you were trying to fix. I guess the good news is it's smaller than the tolerance on the uvws (which I believe is 1mm).

bhazelton:
The "solution" in the pyuvsim context was to unphase using the antenna locations. Does that fix the problem you were seeing with the test file? Not that we don't want it to work without that fix, but it would just be good to know how it behaves.

pyuvdata/utils.py
aelanman: That seems to work. It looks like unphasing with `use_ant_pos` assigns uvws as pos(ant2) - pos(ant1), while `get_antenna_redundancies` uses the opposite convention. I can switch that around, but then I'll get different uvws from what's already in the file. I'll keep at it.

pyuvdata/utils.py
aelanman: The antenna method just makes baseline vectors from antenna positions and numbers and then uses `get_baseline_redundancies`. It defines baselines such that the larger antenna number is first, instead of making a condition on the baseline orientation, which is why it has the same result as `get_baseline_redundancies` when run without the `with_conjugate` option. I guess that behavior is a bit unclear... it might be more useful to have it return antenna pair tuples equivalent to the `with_conjugates` mode.

aelanman:
I agree that this definitely needs better documentation, I'm just not really sure how best to phrase it (and keep it concise enough for a docstring). Do you think it'd be worth renaming the `with_conjugates` keyword to better clarify? I started naming things thinking about what would have to happen to visibilities within redundant groups, but nothing gets complex-conjugated in these functions. 

The redundancy finder without the `with_conjugates` option just splits up the list of baseline indices according to how the corresponding `baseline_vecs` cluster. It only considers uvw coordinates that are present in the file. Under this condition, baselines that are equal in length and opposite in direction are not counted as redundant.
The `with_conjugates` option enforces that baselines are oriented with u>0, which I think is the convention used by `hera_cal`. To make that work, it has to flip baselines around before binning. Since each redundant group is just a list of baseline indices, the `conjugates` list holds onto which baselines were flipped in order to fit into their groups.

bhazelton:
I guess I think that the `with_conjugates` option makes total sense for the baseline-based function, but I'm not sure it should exist at all for the antenna-based function, because there is no strong reason for ordering them in either direction (there is for the baseline one, because there's an ordering intrinsic to the visibility in the file). So I'd be inclined to have the antenna-based function get all possible redundant pairs (and maybe you could return the pairs in the order that makes them redundant).

I'm ok with keeping the option in the antenna-based function if you think it's really useful, but in that case we need to be super clear in the docs what is being returned if `with_conjugates` is turned off. And I'd probably lean towards defaulting it to True. I think the real issue is: what is the use case for this option being turned off?

aelanman:
I just pushed some changes that default to this behavior: `get_antenna_redundancies` enforces that all baselines are in the positive-u direction and returns baselines under that condition. So its default behavior is equivalent to `get_baseline_redundancies` with the `with_conjugates` option, except that it doesn't return a `conjugates` list because none of the baselines in the returned groups need to be flipped.

The key difference now: `get_baseline_redundancies` will only ever return baseline indices that are given to it, while `get_antenna_redundancies` returns, essentially, antenna pairs that are redundant (in the form of baseline indices).

Both return antenna pairs as baseline indices, so it might be worth adding an option to return antenna number tuples instead...

bhazelton:
Ok, thanks. I think an option to return antenna tuples might be good, and we should ask some users, but it might be natural to default to returning antenna tuples from the antenna-based method and baselines from the baseline-based method.

pyuvdata/utils.py
```diff
@@ -818,7 +838,7 @@ def get_antenna_redundancies(antenna_numbers, antenna_positions, tol=1.0, includ
         include_autos =  Include autocorrelations (bool). Default is false
 
     Returns:
-        baseline_groups: list of arrays of redundant baseline indices
+        baseline_groups: list of lists of redundant baseline indices
```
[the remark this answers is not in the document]
aelanman: Both functions return a list of lists, not arrays. I've updated the other docstring.

aelanman:
Apparently changing a couple comments caused a segfault?

bhazelton:
I think it's probably that something else changed under the hood, unrelated to your change. I just pushed a branch with no changes relative to master to see if it errors.

bhazelton:
It does error, so it's nothing to do with your changes. I'm working on figuring out the problem now.

bhazelton:
ok, after some digging it's related to `import casacore.tables`. Peter Williams says there was a recent update to the conda `casacore` package but not the `python-casacore` package, which may have caused the problem. He will update the `python-casacore` package when he can (he's away from his computer this weekend).

bhazelton:
@aelanman the casacore issue has been resolved and the tests are passing now. There is one new line in utils that is not covered, can you add a test to cover it?

aelanman:
I'm hope this change improves coverage... The uncovered line is used when `u==0` and `v != 0`, which previously wasn't present in the test file. I shifted an antenna over by a little so that this condition comes up.

bhazelton:
Hmm, it looks like line 781 in utils.py is still not covered.

PR #459, opened 2018-09-24
Beam interpolation reuse

aelanman:
Currently, UVBeam.interp reruns the interpolation function each time it is called. While this can quickly calculate a bunch of interpolated beam values at once (ie., if given an array of az/za/freq values), it's inefficient for codes like pyuvsim which make many calls to `interp` for single groups of az/za/freq.

This PR provides an option to store the spline fits as an array of callables on the `UVBeam` object, which can then be called again without rerunning `scipy.interpolate.RectBivariateSpline`. Some profiling indicated at least a factor of 3 improvement in speed when run from within interp, but using the interpolated functions separately showed orders of magnitude improvement in speed. Further testing is needed to see if the array of callables can be used more cleanly and efficiently.

Some lines got changed by `autopep8 --ignore=W503,E501`. I'll reverse those changes if necessary. I've also included a unit test that makes sure the reused functions return the same values as the initial spline interpolation.

### DIFF
--- a/pyuvdata/uvbeam.py
+++ b/pyuvdata/uvbeam.py
@@ -676,17 +676,17 @@ class UVBeam(UVBase):
                     for comp_i in range(2):
                         power_data[0, :, pol_i] += \
                             ((efield_data[0, :, pair[0]]
-                             * np.conj(efield_data[0, :, pair[1]]))
+                              * np.conj(efield_data[0, :, pair[1]]))
                              * beam_object.basis_vector_array[0, comp_i]**2
                              + (efield_data[1, :, pair[0]]
-                             * np.conj(efield_data[1, :, pair[1]]))
+                                * np.conj(efield_data[1, :, pair[1]]))
                              * beam_object.basis_vector_array[1, comp_i]**2
                              + (efield_data[0, :, pair[0]]
-                             * np.conj(efield_data[1, :, pair[1]])
-                             + efield_data[1, :, pair[0]]
-                             * np.conj(efield_data[0, :, pair[1]]))
+                                * np.conj(efield_data[1, :, pair[1]])
+                                + efield_data[1, :, pair[0]]
+                                * np.conj(efield_data[0, :, pair[1]]))
                              * (beam_object.basis_vector_array[0, comp_i]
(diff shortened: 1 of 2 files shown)


pyuvdata/uvbeam.py
```diff
@@ -828,49 +829,56 @@ def _interp_az_za_rect_spline(self, az_array, za_array, freq_array):
...
+            Npol_feeds = self.Npols
+        else:
+            Npol_feeds = self.Nfeeds
+        splines_saved = hasattr(self, 'saved_interp_functions')
+
+        def get_lambda(real_lut, imag_lut=None):
+            if imag_lut is None:
+                return lambda za, az: real_lut(za, az, grid=False)
+            else:
+                return lambda za, az: (real_lut(za, az, grid=False) + 1j * imag_lut(za, az, grid=False))
```
bhazelton: These spines need to be calculated and saved separately for the different vector axes (Naxes_vec) and spws (Nspws). I'm a little worried that it's passing the tests using the same spline for the different vector axes, that suggests inadequate testing.

aelanman: This function just returns the lambda functions, which get stored in an array with the full shape `(Naxes_vec, Nspws, Npol_feeds, Nfreqs)`. The resulting array does have separate splines saved along the different axes.
Right now the reuse test just checks for a power beam, but I'll add one for an E-field beam as well.

bhazelton: ok, but I'm worried that they were pulled out of the for loops over those axes -- I think we needed those for loops to make sure the splines did not interact between those axes.

aelanman: It should just be this small function definition that's outside of the for loops. The splines are still made, stored, and used within those loops.

bhazelton: Oh, I see. Sorry, I missed that.

aelanman:
I realized earlier today that this doesn't fully work as advertised, since it doesn't take into account cases when the new frequencies are requested of `UVBeam.interp`. It will take some further work to find a solution that makes sense, so I'm going to close this for now.

Issue #341, opened 2018-05-10
Improve beam interpolation to healpix


bhazelton:
Currently, interpolating an Efield beam and then converting to power gives a different result than converting to power and then interpolating at about a 5% level.

PR #476, opened 2018-10-25
Enable efield_to_pstokes when in az_za coordinates

aelanman:
Some minor changes that enable efield beams to be converted to pstokes while still in az/za coordinates. I'm leaving this as "work in progress" because my unit test shows some substantial discrepancy between two different paths:
az_za efield -> pstokes -> healpix
az_za_efield -> healpix -> pstokes

The healpix interpolation is done in power for the first path and in e-field components for the second, so it's not all that surprising that they're different. If I peak_normalize the results of both, I can get them to match at a tolerance of 0.01.

### DIFF
--- a/pyuvdata/uvbeam.py
+++ b/pyuvdata/uvbeam.py
@@ -569,7 +569,6 @@ class UVBeam(UVBase):
 
         efield_data = beam_object.data_array
         _sh = beam_object.data_array.shape
-        efield_data = beam_object.data_array
         Nfreqs = beam_object.Nfreqs
 
         if self.pixel_coordinate_system != 'healpix':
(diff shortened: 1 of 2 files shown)


pyuvdata/uvbeam.py
```diff
@@ -567,14 +567,19 @@ def efield_to_pstokes(self, run_check=True, check_extra=True, run_check_acceptab
         if beam_object.beam_type != 'efield':
             raise ValueError('beam_type must be efield.')
 
-        if self.pixel_coordinate_system != 'healpix':
-            raise ValueError('Currently only healpix format is supported')
-
-        # construct jones matrix containing the electric field
+        efield_data = beam_object.data_array
```
bhazelton: this line appears to be a duplicate with line 572. Do we need both?

bhazelton:
Interpolation is already known to only be good to ~5% (see #341), so I wouldn't worry about a 1% difference in your 2 paths

aelanman:
@bhazelton This should be ready for your review, Bryna.

PR #477, opened 2018-10-25
Make string handling more precise for uvh5 files

plaplant:
This PR fixes the way strings are handled in I/O for uvh5 files. See http://docs.h5py.org/en/latest/strings.html for full details. The summary is that when writing to HDF5, strings are explicitly cast to `np.string_` objects, and when reading back again, the `pyuvdata.utils._bytes_to_str()` function is called. New tests pass locally, ~~and the change seems to be backwards compatible with old uvh5 files (written with python2, I'm not sure if the python version makes a difference).~~

The changes were no longer backwards compatible with old files, but explicit backwards compatibility has been added, with a warning to the user to rewrite the file if the deprecated style is detected.

### DIFF
--- a/pyuvdata/uvh5.py
+++ b/pyuvdata/uvh5.py
@@ -40,7 +40,11 @@ def _read_uvh5_string(dataset, filename):
     if dataset.dtype.type is np.object_:
         warnings.warn("Strings in metadata of {file} are not the correct type; rewrite with "
                       "write_uvh5 to ensure future compatibility".format(file=filename))
-        return uvutils._bytes_to_str(dataset.value)
+        try:
+            return uvutils._bytes_to_str(dataset.value)
+        except AttributeError:
+            # dataset.value is already <str> type, and doesn't need to be decoded
+            return dataset.value
     else:
         return uvutils._bytes_to_str(dataset.value.tostring())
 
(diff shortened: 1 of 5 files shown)


pyuvdata/uvh5.py
```diff
@@ -102,7 +126,7 @@ def _read_header(self, header, filename):
         self.Nants_telescope = int(header['Nants_telescope'].value)
         self.ant_1_array = header['ant_1_array'].value
         self.ant_2_array = header['ant_2_array'].value
-        self.antenna_names = [uvutils._bytes_to_str(n) for n in header['antenna_names'].value]
+        self.antenna_names = [uvutils._bytes_to_str(n.tostring()) for n in header['antenna_names'].value]
```
bhazelton: This line looks different from the others (it's not using `_read_uvh5_string`). Is that what you want?

plaplant:
@bhazelton one special case that isn't handled correctly is arrays of strings that show up in `extra_keywords`... should I explicitly cover that? Or just make a comment in the forthcoming docs that it's not supported?

bhazelton:
We already do not support arrays, lists or dicts in extra_keywords for uvfits or miriad formats, so I think it's ok not to support them with uvh5.

bhazelton:
@plaplant we have a warning about that in uvdata.py, line 403. You could add uvh5 to the warning there if you wanted to.

PR #482, opened 2018-11-02
Fix pol conjugation again

### LINKED ISSUE #478: Reorder polarization axis in _key2inds for conjugate data when cross pols are present
When using functions like `get_data` that call `_key2inds`, requested baselines whose conjugation is present in the data are properly conjugated. (E.g., if the file contains baseline (0, 1), but the user requests (1, 0), then the visibility data for (0, 1) is returned with a conjugation.) However, when cross-polarizations are present, the polarization should also be reordered (e.g., xy and yx should be swapped), so that the returned array's polarization axis has the same convention as the `polarization_array` contained in the UVData object. The calculation of the polarization array for conjugated data in `_key2inds` should be changed (e.g., here: https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/master/pyuvdata/uvdata.py#L2616), which should make everything downstream "just work".

adampbeardsley:
Fixes #478 
We realized conjugated visibilities need their pols conjugated as well back in PR #399. But we neglected to reorder data when the user does not specify a polarization. The resulting data should be in the order of the polarization_axis, which means we need to reorder the data before conjugating it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 ### Fixed
-- _key2inds now properly reorders polarization axis for conjugated visibilities. This also effects the get_data function.
+- `_key2inds` now properly reorders polarization axis for conjugated visibilities. This also effects the `get_data` function.
 - long strings are saved correctly in miriad files from python3
 
 ## [1.3.3] - 2018-11-01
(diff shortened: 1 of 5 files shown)


plaplant:
This looks good to me too. @jack-h found the initial bug, but he said in his testing that swapping the `xy` and `yx` polarization axes for conjugated baselines resolved his issue. So I think this solves the problem.

PR #484, opened 2018-11-05
 Add option to select only metadata

bhazelton:
helpful for pyuvsim

Also improved error messages in utils polarization functions


pyuvdata/uvdata.py
```diff
@@ -1492,6 +1495,12 @@ def select(self, antenna_nums=None, antenna_names=None, ant_str=None,
         # do select operations on everything except data_array, flag_array and nsample_array
         uv_object._select_metadata(blt_inds, freq_inds, pol_inds, history_update_string)
 
+        if metadata_only is True:
+            if not inplace:
+                return uv_object
+            else:
+                return
```
adampbeardsley: I worry about these returns in the case where the uv object already has data but the user does a metadata select. In that case the data no longer match the metadata. A simple option would be to check if the data_array is not None, and if so a) throw a warning and b) make it (and nsample, and flags) None.

bhazelton: I think you're right to be concerned. Do you think that approach (setting the arrays to None) is better than just erroring (presumably before doing any selection)?

PR #485, opened 2018-11-05
filter useless numpy warnings

bhazelton:
fix broken master build

### DIFF
--- a/pyuvdata/__init__.py
+++ b/pyuvdata/__init__.py
@@ -13,4 +13,9 @@ from .uvcal import *
 from .uvbeam import *
 from . import version
 
+# Filter annoying Cython warnings that serve no good purpose. see numpy#432
+import warnings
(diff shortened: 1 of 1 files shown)


bhazelton:
@adampbeardsley I need help. We have some numpy/Cython warnings that are messing up our warning catching (the master branch is failing on Travis). After doing some reading, it looks like these warnings should always just be ignored. I tried to add a filter for them in the `__init__.py`, but that didn't work as I'd hoped...

adampbeardsley:
Hmm, I have a couple of interesting data points here:
- I'm not able to recreate the test failure locally, but I do extra warnings when I do `from pyuvdata import UVData`.
- I tried putting your filter above the other imports in the init, and that fixed the warnings when I import UVData.

Another thing that seems odd is that only the one test failed. It makes me wonder if we are treating the FHD object slightly differently in terms of the structure of its imports.

PR #488, opened 2018-11-08
Save lat/lon in degrees in uvh5 files instead of radians

plaplant:
Because the lat/lon values are internally stored in radians in UVData objects, uvh5 inherited this behavior. It makes more sense for the values to be written in degrees, to line up with "real world" surveyed values more closely. This is changed for files going forward, and backwards compatibility with previously written files is accomplished by checking for "small" values (absolute value less than pi). If they are, the values are interpreted as radians. (Geographically, these latitudes and longitudes correspond to the middle of the Atlantic Ocean, where no current radio telescopes are known to exist...)

This PR also runs the `self.set_telescope_params()` function on uvh5 objects as they are being read in, which is done for uvfits files. This helps ensure the metadata is self-consistent for known telescope objects, like HERA.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 ### Added
 - support for python3.7
 
+### Changed
+- latitude and longitude in uvh5 files are written in degrees instead of radians.
+
 ### Fixed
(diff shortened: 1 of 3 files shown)


plaplant:
Thanks for the suggestion @bhazelton! I've made it a `DeprecationWarning`.

PR #489, opened 2018-11-12
read uvh5 files with integer datatypes

plaplant:
For a variety of reasons, uvh5 files written by the HERA correlator will have their visibility data saved as integer types. This PR adds support for reading and writing such files. It also adds a (highly down-selected) uvh5 file as a testfile, to ensure we can read and write these types of files.


pyuvdata/uvdata.py
```diff
@@ -1996,14 +2001,15 @@ def read_uvh5(self, filename, antenna_nums=None, antenna_names=None,
...
+                               run_check_acceptability=run_check_acceptability,
+                               data_array_dtype=data_array_dtype)
             self._convert_from_filetype(uvh5_obj)
             del(uvh5_obj)
 
     def write_uvh5(self, filename, run_check=True, check_extra=True,
                    run_check_acceptability=True, clobber=False,
                    data_compression=None, flags_compression="lzf",
-                   nsample_compression="lzf"):
+                   nsample_compression="lzf", data_write_dtype='c8'):
```
bhazelton: Is this the right defaulting? I'd probably lean toward 'c16' or something determined by the type of the data_array.

plaplant: That's a good point, I'll change it so if the user doesn't define their own datatype, I'll use the same precision as data_array. Thanks!

plaplant:
@bhazelton I've updated the defaults to higher precision/the existing precision of the data_array. Thanks for the suggestion.

PR #491, opened 2018-11-14
Fixes a bug related to boolean arrays/lists in the get_baseline_reduncies method

aelanman:
I was getting a warning about using boolean lists as index arrays in numpy, and it turns out that numpy was interpreting [True, False, True, True....] as [1,0,1,1,...], and making a new array with repetitions of the first two elements of the original.

This fix tells numpy that the index array is a boolean array, and now it shows the correct behavior. I'm not sure if this is a result of the version of numpy I'm using (1.15?).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 - latitude and longitude in uvh5 files are written in degrees instead of radians.
+- Fixes a bug in redundancy methods for when there are no redundant baselines.
 
 ### Fixed
 - `_key2inds` now properly reorders polarization axis for conjugated visibilities. This also effects the `get_data` function.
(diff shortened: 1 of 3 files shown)


aelanman:
The latest commit adds a unit test for conjugate baselines. Also fixes a bug when get_baseline_redundancies() is run for a non-redundant array.

adampbeardsley:
whoops, can I take my approval back? Looks like py3's `map` doesn't return a list as expected, resulting in errors.

aelanman:
Uh oh... thanks for catching that. I'll see if I can fix it.

On an unrelated note -- is there a way to get the UVParameter class's `__eq__` method to simply return `False` if comparison fails, instead of printing all the mismatched parts?

aelanman:
@adampbeardsley I've fixed the python3 problem, so it should be ready for review again.

PR #493, opened 2018-11-18
Compress on redundancies

aelanman:
Five new methods in UVData:

(For convenience) `UVData.get_antenna_redundancies` and `UVData.get_baseline_redundancies` return redundancies based on the UVData's antenna_positions and uvw_array parameters, respectively. Comparable to the functions of the same name in `utils.py`.

`UVData._set_u_positive` goes through the baseline array and flips baselines / antenna numbers to follow the rule that `u>0` or `v>0 if u==0` or `w>0 if u == v == 0`. This is required for inflation, because the inflate method uses `get_antenna_redundancies` which only returns baselines within this convention. I can potentially add another method that switches to the `ant_1  > ant_2` convention instead.

`UVData.compress_by_redundancy` does the redundancy selection as discussed, up to a given tolerance. Should I store this tolerance in metadata?

`UVData.inflate_by_redundancy` takes a UVData object with missing data (presumably after a compression, but doesn't require that) and fills out the full data array according to redundancy. Will skip over any missing data.

Note that inflation ***does not*** preserve baseline/time ordering. That is:
```
uv_backup = copy.deepcopy(uv)
uv.compress_by_redundancy()
uv.inflate_by_redundancy()
uv != uv_backup
```
This is because the inflation method is agnostic about any earlier compression operations, and baseline ordering is necessarily lost in compression. The data_array does line up correctly with the baseline_array and uvw_array entries, though.

Still a work in progress for now, but I want to see if this is what people have in mind and get some feedback.

### DIFF
--- a/pyuvdata/uvdata.py
+++ b/pyuvdata/uvdata.py
@@ -3150,3 +3150,84 @@ class UVData(UVBase):
             is in units of days, and integration_time has units of seconds, so we need to convert.
         """
         return np.diff(np.sort(list(set(self.time_array))))[0] * 86400
+
+    def get_redundant_baselines(self, tol=1.0):
+        """
+        Convenience method for getting redundancies to a given tolerance.
+
+        Returns the output of utils.get_baseline_redundancies.
+
+        """
+        baseline_vecs = self.uvw_array[:self.Nbls]
+        baseline_inds = self.baseline_array[:self.Nbls]
+
+        return uvutils.get_baseline_redundancies(baseline_inds, baseline_vecs, tol=tol, with_conjugates=True)
+
+
+    def compress_by_redundancy(self, tol=1.0, inplace=True):
+        """
+        Uses utility functions to find redundant baselines to the given tolerance, then select on those.
+
+        Args:
+            tol: Redundancy tolerance in meters (default 1m)
+            inplace: Do selection on current object.
+        Returns:
+            UVData: if not inplace, returns the compressed UVData object
+        """
+
+        red_gps, centers, lengths, conjugates = self.get_redundant_baselines(tol)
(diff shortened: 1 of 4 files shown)


pyuvdata/uvdata.py
```diff
@@ -3150,3 +3150,122 @@ def _calc_single_integration_time(self):
...
+        Returns the output of utils.get_antenna_redundancies.
+        """
+        antpos, numbers = self.get_ENU_antpos(center=False)
+        return uvutils.get_antenna_redundancies(numbers, antpos, tol=tol, include_autos=include_autos)
+
+    def get_baseline_redundancies(self, tol=1.0):
+        """
+        Convenience method for getting baseline redundancies to a given tolerance.
+
+        Returns the output of utils.get_baseline_redundancies.
```
bhazelton: Let's make this docstring more explicit so that users don't have to look at another piece of code to figure out what this function returns (and for the previous function as well).

pyuvdata/uvdata.py
```diff
@@ -3150,3 +3150,122 @@ def _calc_single_integration_time(self):
...
         """
         return np.diff(np.sort(list(set(self.time_array))))[0] * 86400
+
+    def get_antenna_redundancies(self, tol=1.0, include_autos=True):
+        """
+        Convenience method for getting baseline redundancies to a given tolerance
+        from antenna positions.
+
+        Note that the baseline numbers returned will only correspond with those in the file under the
+        u-positive convention.
```
bhazelton: Maybe mention here that there's a function to reorder to get all of them? Or add an option that will call that function before calling the `get_antenna_redundancies` function.

aelanman:
@bhazelton 

I updated the docstrings and it's passing all tests. Thanks for your comments!

aelanman:
Thanks for reviewing Bryna!

PR #496, opened 2018-11-25
Add uvh5 memo

### LINKED ISSUE #348: Write memo describing UVH5 format
We need to precisely define the UVH5 format, at the level that a tool outside of pyuvdata can make a valid object.

plaplant:
This PR adds a memo describing the uvh5 format.

Fixes #348.


docs/references/uvh5_memo.tex
```diff
@@ -0,0 +1,509 @@
...
+various groups. We note in parenthesis the corresponding attribute of a UVData
+object. Note that in nearly all cases, the names are coincident, to make things
+as transparent as possible to the user.
+
+\section{Header}
+\label{sec:header}
+The \verb+Header+ group of the file contains the metadata necessary to interpret
+the data. We begin with the required parameters, then continue to optional
+ones. Unless otherwise noted, all datasets are scalars (i.e., not arrays). The
+preceision of the data type is also not specified as part of the format, because
```
bhazelton: preceision -> precision

docs/references/uvh5_memo.tex
```diff
@@ -0,0 +1,509 @@
...
+  Nants\_telescope. Note there must be one entry for every unique antenna in
+  ant\_1\_array and ant\_2\_array, but there may be additional
+  entries. (\textit{antenna\_names})
+\item \textbf{baseline\_array}: \textit{int} An array of baseline indices
+  corresponding to the data. This is a one-dimensional array of size Nblts. The
+  baseline index is calculated as:
+  \[
+    \mathtt{baseline} = 2048*(\mathtt{ant2}+1)+(\mathtt{ant1}+1) + 2^{16}.
+  \]
+  For current implementations, 32-bit integers (an \verb+int+ type in C) are
```
bhazelton: should "type in C" be "type of C"?

Also, reading through this I realized that the baseline_array and ant_1_array + ant_2_array is really providing redundant info. That's perfectly fine, but maybe the baseline_array should be optional (in the file format, not on the object)?

docs/references/uvh5_memo.tex
```diff
@@ -0,0 +1,509 @@
...
+  reference date. (\textit{gst0})
+\item \textbf{rdate}: \textit{string} Date for which GST0 (or whichever time
+  saved in that field) applies. (\textit{rdate})
+\item \textbf{timesys}: \textit{string} Time system. pyuvdata currently only
+  supports UTC. (\textit{timesys})
+\item \textbf{x\_orientation}: \textit{string} The orientation of the x-arm of a
+  dipole antenna. It is assumed to be the same for all antennas in the
+  dataset. For instance, ``E'' or ``East'' may be
+  used. (\textit{x\_orientation}).
+\item \textbf{telescope\_name}: \textit{string} The name of the telescope used
```
bhazelton: Interesting that this is optional. I think that's because you've made "instrument" required and I think that makes sense for telescopes where those two field are the same (e.g. MWA, PAPER, HERA), but at least in uvfits and miriad, the telescope name is in the `telescop` field and then `instrume` describes the receiver or backend.

docs/references/uvh5_memo.tex
```diff
@@ -0,0 +1,509 @@
...
+  to take the data. The value is used to check that metadata is self-consistent
+  for known telescopes in pyuvdata. (\textit{telescope\_name})
+\item \textbf{antenna\_positions}: \textit{float} An array of the antenna
```
bhazelton: I'd push for this to be required. pyuvdata's support for objects without antenna_positions is being deprecated.

docs/references/uvh5_memo.tex
```diff
@@ -0,0 +1,509 @@
...
+amount of time that went into a visibility. The best practice is for the
+nsamples dataset to track flagging within an integration time (leading to a
+decrease of the nsamples array value to be less than 1) and LST averaging
+(leading to an increase in the nsamples array value). Datasets that have not
+been LST averaged should have values in nsamples that are less than or equal to
+1. Although this convention is not adhered to by all data formats serviced by
+\verb+pyuvdata+, it is recommended to follow it as closely as possible in UVH5
+files. What \textit{should} be true is the product of the integration\_time
+array and nsamples array corresponding to the total amount of time included in a
+visibility.
```
bhazelton: I really like this paragraph! Very clear.

docs/references/uvh5_memo.tex
plaplant: thanks!

docs/references/uvh5_memo.tex
plaplant: good point re: the baseline array. I've noticed that whenever we're doing our antenna selection with, e.g., `get_data`, the ant_1 and ant_2 arrays are what's used anyway, so making this array optional is not a bad idea.

docs/references/uvh5_memo.tex
plaplant: ah, fair point. I think I'll make it a required parameter, especially because we have those self-consistency checks conditioned on the telescope name.

docs/references/uvh5_memo.tex
plaplant: good point, I'll make the change

plaplant:
@bhazelton I think this is ready for another look when you get a chance.

PR #498, opened 2018-12-05
remove support for astropy <1.3, fix travis 3.7 builds

bhazelton:
In writing fits files, we were still testing for astropy <1.3, but our stated dependency is astropy >= 2.0

this PR removes the handling for old astropy, cleaning up the code and removing untested lines.

It also fixes a Travis issue where python 3.7 was being downgraded in the conda install step

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ install:
   - conda info -a
 
   # install dependencies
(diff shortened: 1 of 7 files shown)


plaplant:
A note for posterity: we noticed that the travis builds for the python3.7 version were actually being downgraded to python3.6 as part of the conda package installation. We've resolved that issue, and added a check to the install script that ensures we're using the right version in the future.

PR #499, opened 2018-12-07
Improve test coverage

bhazelton:
add test coverage, fix some newly exposed bugs, remove some unreachable code.


bhazelton:
this will need to be rebased once PR #500 is merged to get tests to pass

pyuvdata/parameter.py
```diff
@@ -312,7 +313,7 @@ class LocationParameter(UVParameter):
     def __init__(self, name, required=True, value=None, spoof_val=None, description='',
                  acceptable_range=(6.35e6, 6.39e6), tols=1e-3):
         super(LocationParameter, self).__init__(name, required=required, value=value,
-                                                spoof_val=spoof_val, form=(3,),
+                                                spoof_val=spoof_val, form=3,
```
plaplant: Why does `form` go from a tuple to a flat value? Is this just to make it so the user doesn't have to specify things as a tuple? (Happens a few other places, this is just the first one I saw.)

bhazelton: So we should get @adampbeardsley to weigh in here. My read of the code is that it was always supposed to support a single value as well as a tuple, but there was no testing of the lines that used the single value, so I changed this to a single value to test those lines of code.

bhazelton: Actually, I should have said that I added testing for that case. I changed it here because it seemed simpler. (Sorry, got confused about what file we were talking about)

pyuvdata/cst_beam.py
```diff
@@ -197,11 +197,8 @@ def read_cst_beam(self, filename, beam_type='power', feed_pol='x',
             if not np.allclose(roll_rot_phi, phi_data):
                 raise ValueError('Rotating by pi/2 failed')
 
-            # theta should not be affected by the rotation
+            # theta is not affected by the rotation
             rot_theta = theta_data
-            roll_rot_theta = np.roll(rot_theta, int((np.pi / 2) / delta_phi), axis=1)
-            if not np.allclose(roll_rot_theta, theta_data):
-                raise ValueError('Rotating by pi/2 failed')
```
bhazelton: I made this change because I found that it was impossible to trigger this error, by construction.

pyuvdata/uvfits.py
```diff
@@ -44,14 +44,6 @@ def _get_parameter_data(self, vis_hdu):
...
         # using data.par('date') is general -- it will add them together if there are 2
         self.time_array = vis_hdu.data.par('date')
-        if np.finfo(self.time_array[0]).precision < 5:
-            raise ValueError('JDs in this file are not precise to '
-                             'better than a second.')
-        if (np.finfo(self.time_array[0]).precision > 5
-                and np.finfo(self.time_array[0]).precision < 8):
-            warnings.warn('The JDs in this file have sub-second '
-                          'precision, but not sub-millisecond. '
-                          'Use with caution.')
```
bhazelton: It turns out that this code is unreachable and I understand why after looking into it more. The only options for precision in FITS files are all higher than we're testing for here. I think this is ancient cruft.

pyuvdata/tests/test_parameter.py
```diff
@@ -141,3 +153,30 @@ def __init__(self):
     obj.p3 = np.array([4, 9])
     nt.assert_raises(ValueError, obj.check)
     nt.assert_equal(obj._p3.expected_shape(obj), (2,))
+
+
+def test_angle_set_degree_none():
+    param1 = uvp.AngleParameter(name='p2', value=1)
+    param1.set_degrees(None)
```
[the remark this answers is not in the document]
bhazelton: I think it's not strictly needed but it avoids a potential problem that a user could cause.

pyuvdata/parameter.py
bhazelton: I can take it out if we want. It's not super documented and if we want tuples I'm ok with that, I was just trying to cover the code that existed.

pyuvdata/tests/test_parameter.py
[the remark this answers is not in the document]
bhazelton: I'm not sure I want to document it carefully, all it does is make sure that if you do something like `UVData.phase_center_ra_degree = None` it does the same thing as `UVData.phase_center_ra = None`

PR #500, opened 2018-12-07
fix astropy deprecation warnings

bhazelton:
fix new astropy deprecation warnings (from astropy 3.1) that are breaking tests.


bhazelton:
Arg, it broke for python 2. It seems that astropy 2 & 3 conflict on this, trying to figure out how to do it in a way that works for both

plaplant:
Maybe a conditional like `if astropy.__version__ < 3.1:`, like we used to have for the old fits keywords?

bhazelton:
I can do that if I have to. First I'm trying to figure out if I can avoid using that keyword.

bhazelton:
Ok, I got rid of the ones I could and then cased the ones I had to use by python version using six.PY2

PR #505, opened 2018-12-18
Change the defaulting for get_ENU_antpos

bhazelton:
To not remove the median antenna location

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 - support for python3.7
 
 ### Changed
+- UVdata.get_ENU_antpos() now defaults to using the telescope_location as the center rather than the mean antenna position.
 - UVBeam.efield_to_pstokes() no longer restricted to healpix coordinates
 - latitude and longitude in uvh5 files are written in degrees instead of radians.
 - Fixes a bug in redundancy methods for when there are no redundant baselines.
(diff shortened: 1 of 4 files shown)


pyuvdata/uvdata.py
```diff
@@ -2495,6 +2495,13 @@ def get_ENU_antpos(self, center=True, pick_data_ants=False):
         antpos : ndarray, antenna positions in TOPO frame and units of meters, shape=(Nants, 3)
         ants : ndarray, antenna numbers matching ordering of antpos, shape=(Nants,)
         """
+        if center is None:
+            center = False
+            warnings.warn('The default for the `center` keyword has changed. '
+                          'Previously it defaulted to True, using the mean '
```
plaplant: s/mean/median/

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 - support for python3.7
 
 ### Changed
+- UVdata.get_ENU_antpos() now defaults to using the telescope_location as the center rather than the mean antenna position.
```
plaplant: I think we were using the median before, not the mean.

pyuvdata/uvdata.py
bhazelton: thanks!

bhazelton:
@plaplant I fixed the language. Thanks!

PR #510, opened 2018-12-20
Stop Travis from downloading the iers table

### LINKED ISSUE #508: Fix tests to not require the IERS table to be downloaded
To prevent stalling on outages.

bhazelton:
fixes #508

### DIFF
--- a/pyuvdata/tests/__init__.py
+++ b/pyuvdata/tests/__init__.py
@@ -30,15 +30,15 @@ def setup_package():
         os.mkdir(testdir)
 
     # try to download the iers table. If it fails, turn off auto downloading for the tests
-    # and turn it back on in teardown_package
+    # and turn it back on in teardown_package (done by extending auto_max_age)
     try:
         iers_a = iers.IERS_A.open(iers.IERS_A_URL)
     except(urllib.error.URLError):
-        iers.conf.auto_download = False
+        iers.conf.auto_max_age = None
 
 
 def teardown_package():
-    iers.conf.auto_download = True
+    iers.conf.auto_max_age = 30
 
 
 # Functions that are useful for testing:
--- a/pyuvdata/tests/__init__.py
+++ b/pyuvdata/tests/__init__.py
@@ -94,6 +94,25 @@ def checkWarnings(func, func_args=[], func_kwargs={},
         warnings.simplefilter("always")  # All warnings triggered
         warnings.filterwarnings("ignore", message="numpy.dtype size changed")
         warnings.filterwarnings("ignore", message="numpy.ufunc size changed")
+
+        # filter iers warnings if iers.conf.auto_max_age is set to None, as we do in testing if the iers url is down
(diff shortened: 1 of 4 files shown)


bhazelton:
Unfortunately, these tests passed when the IERS server was online, so the new functionality wasn't tested.

bhazelton:
I just tested it locally and it fails.

bhazelton:
But it fails for a new reason: unexpected warnings breaking our warnings catcher

plaplant:
Could we treat those warnings as "ignorable" like the numpy dtype ones?

bhazelton:
I think so maybe. I'm looking at it

bhazelton:
So the problem is that what worked for the numpy warning was to ignore it in the pyuvdata `__init__.py` before all other imports. I tried putting in the ignore for the astropy warning in the test setup_package function and it didn't work. I'm not sure that we want to ignore them outside of the test framework...

bhazelton:
@plaplant Ok, I have this working now on my machine when I turn off wifi and delete the IERS cache file. It involves some ugly handling and filtering of errors and warnings that we'd normally want to let happen but we need to catch in the case of no IERS info. I'd be happy to entertain suggestions of how to do that part better.

pyuvdata/tests/__init__.py
```diff
@@ -26,6 +29,17 @@ def setup_package():
...
+    # try to download the iers table. If it fails, turn off auto downloading for the tests
+    # and turn it back on in teardown_package (done by extending auto_max_age)
+    try:
+        iers_a = iers.IERS_A.open(iers.IERS_A_URL)
+    except(urllib.error.URLError):
+        iers.conf.auto_max_age = None
+
+
+def teardown_package():
+    iers.conf.auto_max_age = 30
```
plaplant: I know this is the default value, and few people have reason to change it, but it might be nice to save and restore the `iers.conf.auto_max_age` value if we need to reset it to `None`. I think this can be done if we switch to unittest-style `setUpPackage` and `tearDownPackage` methods. Let me know if you'd like me to implement that.

bhazelton: You're right, that would be better. If you want to tackle it, go for it!

bhazelton:
The only thing that worries me slightly is that some of the filtering and error handling could mask an error that would show up if the iers url was live. But I think that's a pretty rare edge case that would be caught fairly quickly once iers came back up (or by running the tests on a machine that has a cached iers file), so it's probably ok.

pyuvdata/tests/__init__.py
plaplant: So, it seems that the value of `auto_max_age` is reset between python sessions, so it shouldn't strictly be necessary to reset it after we're done. I don't think it hurts to be thorough, so we should leave it, but there's no danger of it overwriting the user's default.

bhazelton:
I agree. I just wanted to air the concern :)

PR #511, opened 2018-12-22
Use actual antenna numbers in get_antenna_redundancies

aelanman:
Fixes a bug I just found. get_antenna_redundancies wasn't actually using the antenna numbers passed into it to get baseline indices. This should correct it.


bhazelton:
@plaplant It looks like there's a new h5py deprecation warning that might be breaking the Travis builds on this PR:
H5pyDeprecationWarning('dataset.value has been deprecated. Use dataset[()] instead.',)

aelanman:
Thanks @plaplant for fixing the h5py bug. This is passing all checks now @bhazelton .

PR #513, opened 2019-01-02
added run_check_accept to uvh5 read_header

nkern:
This extends the `run_acceptability_check` kwarg for the UVH5 `_read_header` and `_check_header`
for doing certain acceptability checks on metadata, specifically, the checking of proper LST array values. This is important because it is actually a bottleneck when performing partial IO on a large file. My benchmarking shows that it takes 13 seconds just to do the `get_lst_for_time` (without loading any actual data in) for a file that extends across a large LST range

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,13 @@
 All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
+### Changed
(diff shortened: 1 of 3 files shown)


nkern:
@plaplant okay I think this does what you requested. lmk if you see anything else! thanks

PR #525, opened 2019-01-20
Add option for discarding unused antenna metadata after a selection

plaplant:
This PR adds a keyword `keep_missing_antennas` (a name I don't love, so I'm open to other suggestions...) for allowing the user to discard metadata associated with antennas no longer in a UVData object after a selection operation. All metadata that has size `Nants_telescope` are adjusted. By default, this downselection is not invoked, so this does not break backwards compatibility.

### DIFF
--- a/pyuvdata/uvdata.py
+++ b/pyuvdata/uvdata.py
@@ -1407,10 +1407,18 @@ class UVData(UVBase):
 
         return blt_inds, freq_inds, pol_inds, history_update_string
 
-    def _select_metadata(self, blt_inds, freq_inds, pol_inds, history_update_string):
+    def _select_metadata(self, blt_inds, freq_inds, pol_inds, history_update_string,
+                         keep_missing_antennas=True):
         """
-        Internal function to perform select on everything except the data_array,
-        flag_array and nsample_array to allow for re-use in uvfits reading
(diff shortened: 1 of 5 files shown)


pyuvdata/uvdata.py
```diff
@@ -1407,10 +1407,18 @@ def _select_preprocess(self, antenna_nums, antenna_names, ant_str, bls,
...
-        Internal function to perform select on everything except the data_array,
-        flag_array and nsample_array to allow for re-use in uvfits reading
+        Internal function to perform select on everything except the data-sized arrays.
+
+        Args:
+            blt_inds: list of baseline-time indices to keep. Can be None (to keep everything).
+            freq_inds: list of frequency indices to keep. Can be None (to keep everything).
+            pol_inds: list of polarization indices to keep. Can be None (to keep everything).
+            history_update_string: string to append to the end of the history.
+            keep_antennas: whether to keep metadata for antennas that are no longer in the dataset.
```
bhazelton: The docstring keyword name should match the real keyword name. I think I'd slightly prefer something with metadata in the name, maybe `keep_all_metadata`?

plaplant: Good catch! I'll make sure it matches everywhere. And I like `keep_all_metdata`, I'll make that change too.

PR #527, opened 2019-01-29
Redundancy unique axis

### LINKED ISSUE #526: Redundancy finder breaks when array is only redundant baselines
When an array is 100% redundant, like PAPER, the built in redundancy finder will return that each baseline is a unique redundant group. 

This is most likely caused by the `np.unique(bl_grps).tolist()` in `utils.get_baseline_redundancies()`. np.unique by default will flatten input arrays, which is not an issue when different groups have diffrent baselines, but when all baselines fit in one group it return an array of shape (nbls), not (1, nbls).

Unfortunately `np.unique` will fail if you try to call it with the `axis` keyword but all the sub-lists are potentially different sizes.

mkolopanis:
## Description

Redundancy code returns only one group is all baselines are redundant. Previously returned Nbls groups.
## Motivation and Context

 Closes #526 
## Types of changes

## Checklist:


mkolopanis:
hmm I think this padding and un-padding actually negates the call here: 

https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/ebe8cfd567928eea9d2ae5b8c576f73b6372f206/pyuvdata/utils.py#L848-L849

if this type of solution is something that we want to keep they could be removed. the axis keyword in unique will not turn the single element lists back into ints like it used to

pyuvdata/utils.py
```diff
@@ -841,8 +841,15 @@ def get_baseline_redundancies(baseline_inds, baseline_vecs, tol=1.0, with_conjug
         group.sort()
         bl_gps.append(group)
 
+    # Groups can be different lengths, but we need to take a unique over an axis
+    # to properly identity unique groups
```
aelanman: typo -- identify

mkolopanis:
the unique doesn't fail, I think it's just the expected behavior if all the entries are the same (like have all redundant baselines) the flattening of the array during unique also squeezes out the information about the group.

PR #536, opened 2019-02-09
Fix tests in python 2.7 when IERS url is down

### LINKED ISSUE #535: Tests still fail in python 2 if IERS table is not available

bhazelton:
## Description
Reduce the tolerance on an LST calculation in test_miriad.py in python 2.7 when the IERS url is down. 

## Motivation and Context
prevent PRs from failing because IERS isn't available
Fixes #535  

## Types of changes

## Checklist:

- [ ] My change requires an update to the [CHANGELOG](CHANGELOG.md).
  - [ ] I have updated the [CHANGELOG](CHANGELOG.md) accordingly.

### DIFF
--- a/pyuvdata/tests/test_miriad.py
+++ b/pyuvdata/tests/test_miriad.py
@@ -10,10 +10,12 @@ from __future__ import absolute_import, division, print_function
 import os
 import shutil
 import copy
+import six
 import numpy as np
 import nose.tools as nt
 from astropy.time import Time, TimeDelta
 from astropy import constants as const
+from astropy.utils import iers
 
 from pyuvdata import UVData
 from pyuvdata.miriad import Miriad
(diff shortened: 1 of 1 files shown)


pyuvdata/tests/test_miriad.py
```diff
@@ -923,13 +925,19 @@ def test_readMiriadwriteMiraid_check_time_format():
```
plaplant: I've tried to test this with my internet connection off, and unfortunately the tests fail for me... On the line where `delta_lst` is calculated, I get an `IERSRangeError`. I've tried to explicitly set `iers.conf.auto_max_age = None` in the course of the test, but that still raises the error. Does this happen for you locally?

bhazelton: No, it doesn't. What version of astropy do you have?

plaplant: I'm also on astropy v2.0.11. The tests pass for me if I've previously downloaded the IERS table and then have no internet connection, or if I have an active internet connection. (Interestingly, in both cases I get a warning about the time being outside of the IERS range...) But if I clear the download cache and try without internet, I still get the `IERSRangeError`.

I'm inclined to just approve this anyway though, because the tests on travis do not raise the warning I'm seeing locally, and so my local install might be misbehaving in some way. It might be nice to see if tests for a third user pass or fail, but honestly I'm also okay approving now.

bhazelton: Did you use an existing environment or did you make a new one?

PR #540, opened 2019-02-18
Makes _set_u_positive work in the ENU frame.

aelanman:
## Description

Changes `_set_u_positive` to flip baselines according to their vectors in the ENU frame. Previously, it simply used the existing uvws.

## Motivation and Context

If the data are phased, the `_set_u_positive` method worked with phased uvw coordinates. This method was written to ensure that the baselines found by `get_antenna_redundancies` were present in UVData object, since that method returns baselines within the convention of u>=0 in the ENU frame. Using phased UVWs meant that, for some phased data, the wrong baselines were flipped, and `inflate_by_redundancy` failed.

## Types of changes

## Checklist:


pyuvdata/tests/test_uvdata.py
```diff
@@ -2762,7 +2762,7 @@ def test_redundancy_contract_expand():
     # and restored to its original form.
 
     uv0 = UVData()
-    uv0.read_uvfits(os.path.join(DATA_PATH, 'hera19_8hrs_uncomp_10MHz_000_05.003111-05.033750.uvfits'))
+    uv0.read_uvh5(os.path.join(DATA_PATH, 'fewant_randsrc_airybeam_Nsrc100_10MHz.uvh5'))
```
bhazelton: We're fighting package size limits on pyuvdata. I can live with a new 1.05 MB file in the test data if it really needs to be that big, but smaller would be better. Or is the old file you were using (hera19_8hrs_uncomp_10MHz_000_05.003111-05.033750.uvfits) not needed in the test data anymore and we could remove it?

aelanman: I've managed to replace `hera19_8hrs_uncomp_10MHz_...` with a reduced version of the new file (only about 260 KB) for most tests.

Fixing this revealed a potential complication in the inflate method. With the way the redundancies work, it is possible for a single baseline to be redundant with two other baselines that are not redundant with each other. When the tolerance was set to 1cm in `test_redundancy_contract_expand`, this happened, and broke the inflate method. I think it may be because the new test file has actual antenna positions, and those might not be redundant to 1cm precision.

`inflate_by_redundancy`  should only work if there is a unique redundant group for each baseline, so that a single visibility may be assigned to each. It makes sense that this would break the test, but I'm not sure how best to handle this case in the future, so I just raised the tolerance to 2cm for now.

bhazelton:
Looks good to me too.

PR #545, opened 2019-03-04
Add support for a yaml settings file for CST beams

bhazelton:
Add support for a yaml settings file to specify information about CST beams. This allows for metadata to be specified and pass to the UVBeam object and also prevents having to parse frequencies from filenames.

## Description
Developed a new yaml settings file format that specifies metadata for CST beams. This was done in close consulation with Nicolas Fagnioni who will provide these files for his beams going forward.

## Motivation and Context
A lot of critical metadata is missing from CST text files with no good way to add it back in. These settings files allow the metadata to be specified once by an expert and can be kept with the text files.

Talking with Nicolas, I've changed some of the optional attributes on UVBeam to better reflect what information the simulations use and provide. While these are technically breaking changes, I do not think they are in use by anyone, so I do not expect it to impact anyone.

## Types of changes

## Checklist:

- [x] My change introduces new functionality that should be highlighted in the tutorial.
  - [x] I have updated the tutorial accordingly.


pyuvdata/cst_beam.py
```diff
@@ -201,7 +209,7 @@ def read_cst_beam(self, filename, beam_type='power', feed_pol='x',
             rot_theta = theta_data
 
         # get beam
-        if self.beam_type is 'power':
+        if self.beam_type == 'power':
             data_col = np.where(np.array(column_names) == 'abs(v)')[0][0]
```
nkern: So this was one of the lines that got tripped up using Nic's new CST `.txt` files. In this new files, the appropriate column is labeled `E-Field`, rather than `abs(v)`. Nic confirmed with me that in the new files this is indeed the appropriate column to use, but it means that our way of finding this column here will fail, which is why in the `HERA-Beams` PR I added a function for actually editing the CST files to replace `E-Field` with `V` so that the correct data column is caught.

Not sure what the best way to handle this is. My short-term solution was to include some utility functions in `convert_to_uvbeam.py` that actually re-writes the CST files with headers compliant with this standard.

bhazelton: Thanks for this, Nick. I haven't really tested against the new files, but I will do that now and include needed changes in this PR. I was thinking about that as a separate step, but I agree those changes make sense in this PR.

Note that the new settings files actually list the filenames and the frequencies (in two ordered lists) so we no longer need to derive the frequency from the filename.

bhazelton: @nkern I'm not seeing the problem you report with the column name. Which specific directory were you reading from?

bhazelton:
@nkern I think this should now just work on the new files. I included one of them along with the yaml for them that Nicolas gave me and added some tests using them.

pyuvdata/tests/__init__.py
```diff
@@ -199,3 +199,13 @@ def skipIf_no_h5py(test_func):
     except(ImportError):
         return skip(reason)(test_func)
     return test_func
+
+
+def skipIf_no_yaml(test_func):
+    """defines a decorator to skip tests that require h5py."""
```
bhazelton: h5py -> yaml

pyuvdata/uvbeam.py
```diff
@@ -1817,34 +1815,72 @@ def write_beamfits(self, filename, run_check=True, check_extra=True,
...
+                    |  telescope_name (str)
+                    |  feed_name (str)
+                    |  feed_version (str)
+                    |  model_name (str)
+                    |  model_version (str)
+                    |  history (str)
+                    |  frequencies (list(float))
+                    |  cst text filenames (list(str)) -- path relative to yaml file location
+                    |  feed_pol (str) or (list(str))
+                and it may include the following optional keywords:
```
bhazelton: it -> they

pyuvdata/uvbeam.py
```diff
@@ -1817,34 +1815,72 @@ def write_beamfits(self, filename, run_check=True, check_extra=True,
...
+                Specifying any of the associated keywords to this function will
+                override the values in the settings file.
+            beam_type (str): what beam_type to read in ('power' or 'efield'). Defaults to 'power'.
+            feed_pol (str): the feed or polarization or list of feeds or polarizations the files correspond to.
                 Defaults to 'x' (meaning x for efield or xx for power beams).
-            rotate_pol: If True, assume the structure in the simulation is symmetric under
+            rotate_pol (bool): If True, assume the structure in the simulation is symmetric under
                 90 degree rotations about the z-axis (so that the y polarization can be
                 constructed by rotating the x polarization or vice versa).
                 Default: True if feed_pol is a single value, False if it is a list.
```
bhazelton: this is wrong, fix it.

pyuvdata/uvbeam.py
```diff
@@ -1817,34 +1815,73 @@ def write_beamfits(self, filename, run_check=True, check_extra=True,
...
+                    |  model_name (str)
+                    |  model_version (str)
+                    |  history (str)
+                    |  frequencies (list(float))
+                    |  cst text filenames (list(str)) -- path relative to yaml file location
+                    |  feed_pol (str) or (list(str))
+                and they may include the following optional keywords:
+                    | ref_imp (float): beam model reference impedance
+                    | all other fields will go into the extra_keywords attribute
+                Specifying any of the associated keywords to this function will
```
plaplant: I think this is a good practice, but do we want to warn/alert the users if an override happens? I'm not familiar with the day-to-day use of the beams, so I'm not sure how common it is. If users frequently override what's in the file, then it'd probably be annoying to get lots of warnings.

pyuvdata/uvbeam.py
```diff
@@ -1817,34 +1815,73 @@ def write_beamfits(self, filename, run_check=True, check_extra=True,
...
         Read in data from a cst file.
 
         Args:
-            filename: The cst file or list of files to read from. If a list is passed,
+            filename (str): Either a settings yaml file or a cst text file or
+                list of cst text files to read from. If a list is passed,
                 the files are combined along the appropriate axes.
-            beam_type: what beam_type to read in ('power' or 'efield'). Defaults to 'power'.
-            feed_pol: the feed or polarization or list of feeds or polarizations the files correspond to.
+                Settings yaml files must include the following keywords:
```
plaplant: This is great documentation, but it might be a good idea to have a separate document/memo that defines exactly what goes in these files. Pulling this information out of the docstring/code itself might be too high a barrier to entry for people making the yaml files on the CST simulation side...

pyuvdata/uvbeam.py
bhazelton: I think it'd be pretty unusual to want to override this values. I can add warnings.

pyuvdata/uvbeam.py
bhazelton: I agree. I'll work on putting one together. I developed this format in close collaboration with Nicolas Fagnioni, so he's on board, but there should be better documentation.

bhazelton:
On backwards compatibility: This only adds functionality, it doesn't change the existing functionality at all.

The one small caveat is the changes to some of the optional parameters, but that metadata information isn't in the cst txt files (which is all we've had in the past), so those parameters were not being used except in internal testing. One of the value adds to going to these yaml files is being able to specify that (and other) metadata.

pyuvdata/uvbeam.py
bhazelton: done.

pyuvdata/uvbeam.py
bhazelton: I added a page in the docs giving more details about the settings files. You can see it here: https://pyuvdata.readthedocs.io/en/cst_yaml/cst_settings_yaml.html

When I was checking that it worked on RTD, I discovered that our builds were failing there (I need to figure out how to get pinged when that happens). It was a unicode error that was fixed by having RTD build in python 3.6, so those changes are also in this PR now.

plaplant: Thanks! The new documentation is excellent. Thanks for getting RTD going again too.

PR #547, opened 2019-03-12
Added UVBeam healpix interpolation, updated freq interpolation

nkern:
healpix interpolation and general interpolation functionality enhancements

## Description

- Added simple healpix interpolation `UVBeam` via `healpy` routines.
- Updated `_interp_freq` to allow for fast and efficient return of `data_array` slice if requested frequencies are within some distance tolerance of nearest neighbors. 
- Updated `interp` functions to take `polarizations` as an optional kwarg

## Motivation and Context

This provides increased functionality for healpix beams, which are used widely in our visibility simulators. One necessity was for fast and efficient return of an interpolated beam without performing interpolation across the frequency axis, which wasn't possible before.

## Types of changes

## Checklist:

### DIFF
--- a/pyuvdata/tests/test_uvbeam.py
+++ b/pyuvdata/tests/test_uvbeam.py
@@ -534,10 +534,8 @@ def test_healpix_interpolation():
     interp_data_array, interp_basis_vector = power_beam.interp(az_array=az_orig_vals,
                                                                za_array=za_orig_vals,
                                                                freq_array=freq_orig_vals)
-
     data_array_compare = power_beam.data_array
     interp_data_array = interp_data_array.reshape(data_array_compare.shape, order='F')
-
     nt.assert_true(np.allclose(data_array_compare, interp_data_array))
 
     # basis_vector exception
@@ -549,10 +547,8 @@ def test_healpix_interpolation():
     interp_data_array, interp_basis_vector = power_beam.interp(az_array=az_orig_vals,
                                                                za_array=za_orig_vals,
                                                                freq_array=freq_orig_vals)
-
     data_array_compare = power_beam.data_array
     interp_data_array = interp_data_array.reshape(data_array_compare.shape, order='F')
-
     nt.assert_true(np.allclose(data_array_compare, interp_data_array))
 
     # assert not feeding frequencies gives same answer
@@ -567,6 +563,15 @@ def test_healpix_interpolation():
     interp_data_array2, interp_basis_vector2 = power_beam.interp(az_array=az_orig_vals, za_array=za_orig_vals,
                                                                  polarizations=['xx'])
     nt.assert_true(np.allclose(interp_data_array[:, :, :1], interp_data_array2[:, :, :1]))
+
+    # change complex data_array to real data_array and test again
+    power_beam.data_array = np.abs(power_beam.data_array)
+    interp_data_array, interp_basis_vector = power_beam.interp(az_array=az_orig_vals,
+                                                               za_array=za_orig_vals,
+                                                               freq_array=freq_orig_vals)
+    data_array_compare = power_beam.data_array
+    interp_data_array = interp_data_array.reshape(data_array_compare.shape, order='F')
+    nt.assert_true(np.allclose(data_array_compare, interp_data_array))
 
     # test no inputs equals same answer
     interp_data_array2, interp_basis_vector2 = power_beam.interp()
--- a/pyuvdata/tests/test_uvbeam.py
+++ b/pyuvdata/tests/test_uvbeam.py
@@ -410,7 +410,6 @@ def test_freq_interpolation():
     except ValueError:
         raise AssertionError("UVBeam._interp_freq didn't return an array slice as expected")
 
-
     # test errors if one frequency
     power_beam_singlef = power_beam.select(freq_chans=[0], inplace=False)
     nt.assert_raises(ValueError, power_beam_singlef._interp_freq, np.array([150e6]))
--- a/pyuvdata/tests/test_uvbeam.py
+++ b/pyuvdata/tests/test_uvbeam.py
@@ -395,6 +395,17 @@ def test_interpolation():
 
     nt.assert_true(np.allclose(data_array_compare, interp_data_array))
 
+    # test only a single polarization
+    interp_data_array, interp_basis_vector = power_beam.interp(az_array=az_orig_vals,
+                                                               za_array=za_orig_vals,
+                                                               freq_array=freq_orig_vals,
+                                                               polarizations=['xx'])
+
+    data_array_compare = power_beam.data_array[:, :, :1]
+    interp_data_array = interp_data_array.reshape(data_array_compare.shape, order='F')
+
+    nt.assert_true(np.allclose(data_array_compare, interp_data_array))
+
     # test no errors using different points
     az_interp_vals = np.array(np.arange(0, 2 * np.pi, np.pi / 9.0).tolist()
                               + np.arange(0, 2 * np.pi, np.pi / 9.0).tolist())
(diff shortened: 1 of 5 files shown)


pyuvdata/tests/test_uvbeam.py
```diff
@@ -403,7 +414,15 @@ def test_interpolation():
...
-                                                               freq_array=freq_interp_vals)
+                                                               freq_array=freq_interp_vals)    
+
+    # test if fed original frequencies that not interp_bool is triggered
+    # by using only one freq chan in object, which should trigger a ValueError if interp_bool is True
+    _pb = power_beam.select(frequencies=power_beam.freq_array[0, :1], inplace=False)
+    try:
+        interp_data_array, interp_bandp = _pb._interp_freq(_pb.freq_array[0], kind='linear')
+    except:
+        raise AssertionError("UVBeam._interp_freq didn't return array slice as expected...")
```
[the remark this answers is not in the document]
nkern: The point is, if with the inputs given, `UVBeam._interp_freq` should error if it interpolate the beam, rather than slices it so the test is actually supposed to test that it *doesn't* raise an error in this case, and this is the clearest way to do that in my mind. I could just make the function call and not check any of the outputs, but I've gotten pushback on that from other people before, so the `try/except` with an error message is in my mind is more of a clear way to do it. Lmk if you have any suggestions.

pyuvdata/tests/test_uvbeam.py
```diff
@@ -485,6 +504,32 @@ def test_interpolation():
     nt.assert_raises(ValueError, power_beam.interp, az_array=az_interp_vals,
                      za_array=za_interp_vals + np.pi / 2)
 
+    # test healpix interpolation
```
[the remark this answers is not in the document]
nkern: sure thing

pyuvdata/uvbeam.py
```diff
@@ -40,7 +40,8 @@ class UVBeam(UVBase):
                                    'az, za coordinate axes (for the basis_vector_array) '
                                    'where az runs from East to North'}}
 
-    interpolation_function_dict = {'az_za_simple': '_interp_az_za_rect_spline'}
+    interpolation_function_dict = {'az_za_simple': '_interp_az_za_rect_spline',
+                                   'healpix_simple': '_interp_healpix_bilinear'}
```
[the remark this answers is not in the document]
nkern: sure thing

nkern:
@bhazelton there is a line in `_interp_az_za_rect_spline` that reads
```
        if az_array is None:
            return input_data_array, self.basis_vector_array
```
and I'm wondering if we should actually have this read
```
        if az_array is None or za_array is None:
            return input_data_array, self.basis_vector_array
```

pyuvdata/tests/test_uvbeam.py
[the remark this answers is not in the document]
bhazelton: I guess I would just do the call and then assert that the returned array matched the data_array on the object. But I'm ok with this now that you've added the comments to explain what you're doing.

pyuvdata/uvbeam.py
```diff
@@ -294,6 +298,12 @@ def __init__(self):
                                                        form='str', expected_type=str,
                                                        description=desc,
                                                        acceptable_vals=list(self.interpolation_function_dict.keys()))
+        desc = ('String indicating frequency interpolation kind. '
+                'See scipy.interpolate.interp1d for details. Default is linear.')
+        self._freq_interpolation_kind = uvp.UVParameter("freq_interpolation_kind",
+                                                        required=False, form='str',
+                                                        expected_type=str, description=desc)
+        self.freq_interpolation_kind = 'linear'
```
bhazelton: This is great, thank you.

I think we need essentially the same thing for the frequency interpolation tolerance.

pyuvdata/uvbeam.py
```diff
@@ -792,13 +816,14 @@ def _interp_az_za_rect_spline(self, az_array, za_array, freq_array, reuse_spline
...
-            input_data_array = self._interp_freq(freq_array)
+            input_data_array, _ = self._interp_freq(freq_array, kind=freq_interp_kind, tol=freq_interp_tol)
             input_nfreqs = freq_array.size
         else:
             input_data_array = self.data_array
             input_nfreqs = self.Nfreqs
             freq_array = self.freq_array[0]
-        if az_array is None:
+
+        if az_array is None or za_array is None:
```
bhazelton: thanks for catching this!

pyuvdata/uvbeam.py
```diff
@@ -900,14 +936,126 @@ def get_lambda(real_lut, imag_lut=None):
...
+    def _interp_healpix_bilinear(self, az_array, za_array, freq_array, freq_interp_kind='linear',
+                                 polarizations=None):
+        """
+        Simple bi-linear interpolation wrapper for healpix.
+
+        Args:
+            az_array: azimuth angles to interpolate to [radians]
+            za_array: zenith angles to interpolate to [radians]
+            freq_array: frequency values to interpolate to [Hz]
+            freq_interp_kind: str, interpolation method across frequency. See scipy.interpolate.interp1d for details.
```
bhazelton: shouldn't the `freq_interp_tol` show up here in the same way it does in `_interp_az_za_rect_spline`?

pyuvdata/uvbeam.py
```diff
@@ -919,9 +1067,13 @@ def interp(self, az_array=None, za_array=None, freq_array=None, reuse_spline=Fal
...
         if self.interpolation_function is None:
             raise ValueError('interpolation_function must be set on object first')
+        if self.freq_interpolation_kind is None:
+            raise ValueError('freq_interpolation_kind must be set on object first')
 
-        interp_func = self.interpolation_function_dict[self.interpolation_function]
-        return getattr(self, interp_func)(az_array, za_array, freq_array, reuse_spline)
+        interp_func = self.interpolation_function_dict[self.interpolation_function]['func']
+        return getattr(self, interp_func)(az_array, za_array, freq_array,
+                                          freq_interp_kind=self.freq_interpolation_kind,
```
bhazelton: I think the `freq_interp_tol` should show up here too.

nkern:
@bhazelton I don't think the freq_interp_tol needs to be propagated as a UVBeam attribute. I think probably what should happen is that if the tolerance is not exceeded, then I need to change the freq_interp_kind to 'nearest' and then just return the beam array slice, otherwise if it is exceeded then I keep freq_interp_kind as is and return The interpolated beam. What do you think?

bhazelton:
@nkern The issue is that the `to_healpix` method uses the interpolation functions and makes a new UVData object using the interpolated data (maybe you can just use that rather than what you're talking about?). We need to preserve the information about how that interpolation is done on the new object. The frequency tolerance absolutely affects how that new beam is made. I'm open to achieving this information by setting the `freq_interp_kind` to `nearest` if the tolerance is reached, but we need some way to know what was actually done.

nkern:
@bhazelton okay, then I'll propagate that to the `to_healpix` function and add more tests too!

nkern:
@bhazelton but I see now that `to_healpix` doesn't do any frequency interpolation, so I think its fine actually.

pyuvdata/uvbeam.py
```diff
@@ -900,14 +960,134 @@ def get_lambda(real_lut, imag_lut=None):
...
             az_array: az values to interpolate to (same length as za_array)
             za_array: za values to interpolate to (same length as az_array)
             freq_array: frequency values to interpolate to
+            freq_interp_tol: float, frequency distance tolerance [Hz] of nearest neighbors.
+                If *all* elements in freq_array have nearest neighbor distances within
+                the specified tolerance then return the beam at each nearest neighbor,
+                otherwise interpolate the beam.
+            polarizations: list of str, polarizations to interpolate if beam_type is 'power'.
+                Default is all polarizations in self.polarization_array.
+            kwargs: dictionary of keyword arguments to pass to interpolation function. See it for details.
```
bhazelton: So the problem with this docstring, is that the interpolation functions are private functions (starting with the underscore character), which means that their docstrings are not part of the public API and are not printed in the ReadTheDocs auto documentation. We made those private because we didn't want users to call those functions directly, we wanted them to call this `interp` method. I think therefore that we need to have the keywords that only apply to specific interpolation methods documented here. I think I'd really prefer to pass those with their proper names and raise an error or warning if they do not apply to the chosen interpolation function. That approach is much more transparent to the user than passing them via `kwargs`. I'm happy to help with this if you like. It looks to me like the only keywords that this applies to currently are `reuse_spline` and `new_object`, am I missing something?

nkern: Sure thing!

nkern:
@bryna okay updated the doc string as requested. Let me know if there is anything else you want from this PR! thanks

pyuvdata/uvbeam.py
```diff
@@ -1106,7 +1107,9 @@ def interp(self, az_array=None, za_array=None, freq_array=None, freq_interp_tol=
         return getattr(self, interp_func)(az_array, za_array, freq_array,
                                           freq_interp_kind=self.freq_interp_kind,
                                           freq_interp_tol=freq_interp_tol,
-                                          polarizations=polarizations, **kwargs)
+                                          polarizations=polarizations,
+                                          freq_interp_kind=freq_interp_kind,
```
bhazelton: why is `freq_interp_kind` coming in here twice?

bhazelton:
@nkern thank you! I just have one more question (above)

pyuvdata/uvbeam.py
nkern: that was a typo that I fixed and is now outdated, I believe..

PR #551, opened 2019-03-22
Fix overlapping data addition bug

### LINKED ISSUE #455: overlapping data bug when reading in a list of files
We've seen some strange issues when passing in a list of miriad files to read, which triggers the in-place-add feature to combine them into a single UVData object. In some cases, the error `ValueError: These objects have overlapping data and cannot be combined` is raised, even though there should not be overlapping data. This was seen by @ianmalcolm to happen for some IDR2 data (when combining all 4 polarizations from a single time) and by me with some old HERA-19 data (when combining data from adjacent times). I'll try to come up with a minimal working example to reproduce the bug and follow up here.

plaplant:
## Description
This PR allows for combining two UVData objects that ostensibly overlap along the baseline-time, frequency, and polarization axes, but one contains "invalid" data (defined to have all zeros for the data array and be totally flagged). In such a case, the invalid data is overwritten by the valid data in the second object.

## Motivation and Context
When a UVData object is expanded to fill out "dummy indices" that are not part of the combined object but required to create a valid object, the dummy data are zero-valued and completely flagged. This PR makes changes such that if two UVData objects to be combined overlap, and the overlapping data is dummy-like, then it combines the objects anyway.

One caveat is that, due to the way that the objects are combined, the addition operation does _not_ commute in all cases anymore. (Under the hood, the overlapping bits of the first object are overwritten by the second one. If the second object contains more data than the first one, that data is not propagated to the combined object.) To protect against this edge case, an error is raised instructing the user to reorder the addition operation and try again. (As a possible solution, I tried to swap the "this" and "other" objects, which worked for combining not-in-place, but failed for in-place addition. Rather than make lots of complicated if/else blocks for what I assume will be an edge case, I opted for throwing an error instead.)

Fixes #455. 

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file.
 ### Added
 - Support for a yaml settings file to collect and propagate metadata for CST beam files.
 
+### Fixed
+- Combining UVData objects to overwrite invalid "dummy data" no longer produces errors.
+
 ## [1.3.6] - 2019-02-15
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 - Support for a yaml settings file to collect and propagate metadata for CST beam files.
 
 ### Fixed
-- Combining UVData objects to overwrite invalid "dummy data" no longer produces errors.
+- Combining overlapping data along multiple axes (most common when reading in multiple files) no longer errors.
 
 ## [1.3.6] - 2019-02-15
 
--- a/pyuvdata/tests/test_uvdata.py
+++ b/pyuvdata/tests/test_uvdata.py
@@ -2956,10 +2956,10 @@ def test_overlapping_data_add():
     uvfull = uv1 + uv2
     uvfull += uv3
     uvfull += uv4
-    extra_history = ("  Downselected to specific baseline-times, polarizations using pyuvdata. "
+    extra_history = ("Downselected to specific baseline-times, polarizations using pyuvdata. "
                      "Combined data along polarization axis using pyuvdata. Combined data along "
                      "baseline-time axis using pyuvdata. Overwrote invalid data using pyuvdata.")
-    nt.assert_equal(uvfull.history, uv.history + extra_history)
+    nt.assert_true(uvutils._check_histories(uvfull.history, uv.history + extra_history))
     uvfull.history = uv.history  # make histories match
     nt.assert_equal(uv, uvfull)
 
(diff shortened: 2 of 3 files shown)


pyuvdata/tests/test_uvdata.py
```diff
@@ -2936,3 +2936,45 @@ def test_redundancy_finder_when_nblts_not_nbls_times_ntimes():
...
+    nt.assert_equal(uvfull.history, uv.history + extra_history)
+    uvfull.history = uv.history  # make histories match
+    nt.assert_equal(uv, uvfull)
+
+    # check combination not-in-place
+    uvfull = uv1 + uv2
+    uvfull += uv3
+    uvfull = uvfull + uv4
+    uvfull.history = uv.history  # make histories match
+    nt.assert_equal(uv, uvfull)
```
bhazelton: I like these tests, but I also want to be sure that if the are written out as separate files and all 4 filenames are passed to the generic `read` method that there are no errors and it works as expected. I think that's the most common way this code will actually be tripped.

It's a little interesting that in that case I think there will still be something in the history about overwriting invalid data, even though there was no invalid data in any of the individual files. I'm not sure whether I'm worried about that or not.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file.
 ### Added
 - Support for a yaml settings file to collect and propagate metadata for CST beam files.
 
+### Fixed
+- Combining UVData objects to overwrite invalid "dummy data" no longer produces errors.
```
bhazelton: maybe expand this to say something more about the case where people are likely to have run into this, where they have input files that don't have any dummy data (so they might not realize this fixes a bug they care about)

bhazelton: maybe "fixes a bug that occurred when combining data files across multiple axes"?

pyuvdata/tests/test_uvdata.py
plaplant: Good thought! I've added that test.

Yes, it does still add the "invalid data" string to the history. I suppose one option could be to inspect the stack and see if the parent function is a `read*` method, and not add the string in that case. Otherwise I'm not sure how we can tell whether it's worth adding or not.

CHANGELOG.md
plaplant: Updated.

pyuvdata/tests/test_uvdata.py
bhazelton: Yeah, I think we leave it. It just struck me. Thanks for adding the test.

PR #552, opened 2019-03-29
added add_to_history kwarg to UVH5.write_uvh5_part

nkern:
A very small PR that adds ability to append to history when performing UVH5 partial writing.
Also changes a ValueError to a simple print statement when UVH5 writes to a file that already exists (i.e `clobber=False`).
This was very annoying when looping over file writes and I think was not supposed to be the intended behavior.

## Description

A very small PR that adds ability to append to history when performing UVH5 partial writing.
Also changes a ValueError to a simple print statement when UVH5 writes to a file that already exists (i.e `clobber=False`).
This was very annoying when looping over file writes and I think was not supposed to be the intended behavior.

## Types of changes

## Checklist:

- [] All new and existing tests pass in both python 2 and python 3.

### DIFF
--- a/pyuvdata/uvh5.py
+++ b/pyuvdata/uvh5.py
@@ -594,7 +594,7 @@ class UVH5(UVData):
     def write_uvh5(self, filename, run_check=True, check_extra=True,
                    run_check_acceptability=True, clobber=False,
                    data_compression=None, flags_compression="lzf", nsample_compression="lzf",
-                   data_write_dtype=None):
+                   data_write_dtype=None, add_to_history=None):
         """
         Write an in-memory UVData object to a UVH5 file.
 
@@ -642,7 +642,7 @@ class UVH5(UVData):
             if clobber:
                 print("File exists; clobbering")
             else:
-                raise ValueError("File exists; skipping")
+                print("File exists; skipping")
 
         # open file for writing
         with h5py.File(filename, 'w') as f:
@@ -722,7 +722,7 @@ class UVH5(UVData):
             if clobber:
                 print("File exists; clobbering")
             else:
-                raise ValueError("File exists; skipping")
+                print("File exists; skipping")
 
         # write header and empty arrays to file
         with h5py.File(filename, 'w') as f:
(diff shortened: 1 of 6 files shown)


nkern:
@bhazelton looks like this is ready. lmk if you have any requests on this, or if you'd like someone else to review. thanks!

pyuvdata/uvh5.py
```diff
@@ -642,7 +642,7 @@ def write_uvh5(self, filename, run_check=True, check_extra=True,
             if clobber:
                 print("File exists; clobbering")
             else:
-                raise ValueError("File exists; skipping")
+                print("File exists; skipping")
```
bhazelton: Why did you replace this and the next error with a print? I think it is an error if you're trying to write a file and it doesn't get written.

bhazelton:
I asked Paul to review it because he's more familiar with this code.

pyuvdata/uvh5.py
nkern: I guess I've just never really encountered that in `clobber` statements: and I don't think it was part of `pyuvdata` until recently. It sets up somewhat of a headache when iterating over multiple files, some of which might trigger this which we would want to skip but not fail. Of course we could just catch the exception and pass, but this just seems like a silly barrier, and might let other ValueErrors from an actual code failure slip through.

Said another way, you could reason that this is not at all a code failure: the `clobber` kwarg is set to handle this, so not writing the file is indeed exactly what you want it to do, and __should not__ trigger an exception.

nkern:
@plaplant let me know what you think about changing all of the `clobber == False` ValueError to just print statements (see my comment above). I think __not__ raising an exception is the appropriate way to handle it given the logic. If you agree, I can change this in all of the objects in `pyuvdata` to just return a print statement rather than raising a `ValueError`

nkern:
@plaplant I think I disagree with the logic here. The entire notion of a `clobber` or `overwrite` kwarg is to handle the issue of file overwrite for you. If you are trying to overwrite a file and you purposely set `clobber = False`, then that means that you know as the user that there is a chance that the file already exists, and in the case that it does, you want it to __not__ overwrite the file. That is the __expected__ behavior and should not raise an exception.

__However__, I see that `astropy.io.fits` does the same thing (raising an `OSError` if `overwrite=False` but the file already exists), so I'm willing to concede that I may be of the minority opinion.

I think what I'd like to do is keep the error but have it raise an `IOError` instead, which is more apt that `ValueError`. This way, in my looping example, I can catch `IOError` instead of `ValueError`, the concern being that `ValueError` is so widely used that catching it might actually allow other errors to pass silently.

LMK what you think about that. I'll also add the other additions you suggest, thanks!

nkern:
Great, if there aren't further comments then I'll merge this! Thanks for the feedback Paul!

PR #558, opened 2019-04-04
Tweak ci settings

bhazelton:
## Description

- Break circleci tests into 2 workflows, one for building & testing pyuvdata and another for external tests.
- Prevent codecov from waiting on Travis before reporting coverage.
- Add badges for circleci and codecov

## Motivation and Context

Splitting the circleci workflows will make it easier to rerun things separately (hera_cal has a test or two that fail intermittently and sometimes need to be rerun. I've pinged Josh multiple times about this but he doesn't see the error and hasn't been able to track it down).

codecov automatically detects from GitHub when all the status checks are done and doesn't report before then in case more coverage reports are coming in. This is a nice feature in general, but Travis is slow and isn't currently reporting to codecov anyway (it reports to coveralls currently) so it just delays the codecov report unnecessarily.

## Types of changes

## Checklist:

- [ ] My change requires an update to the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/master/CHANGELOG.md).
  - [ ] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/master/CHANGELOG.md) accordingly.


bhazelton:
No coveralls only pulls from Travis at the moment, so it doesn't need to be required. My plan was to pick one of coveralls and codecov going forward and to have only circleci report to it. Maybe we should have a vote on the next telecon to decide which one to keep.

plaplant:
Sounds like a plan, this is good to go in my book.

bhazelton:
hmm. looks like codecov is still waiting on Travis. I suppose it's possible that it's because it's not merged into master yet, but don't think that's how it's supposed to work

PR #559, opened 2019-04-05
combine order_pols and reorder_pols into one method

bhazelton:
## Description

Combine `order_pols` and `reorder_pols` into one method that keeps the functionality of both.

## Motivation and Context

We should not have two methods to do this.

## Types of changes

## Checklist:

- [x] If this is a breaking change, it includes backwards compatibility and deprecation warnings (if possible)

### DIFF
--- a/pyuvdata/uvdata.py
+++ b/pyuvdata/uvdata.py
@@ -936,6 +936,17 @@ class UVData(UVBase):
             self.check(check_extra=check_extra,
                        run_check_acceptability=run_check_acceptability)
 
+    def order_pols(self, order='AIPS'):
+        """
+        Soon to be deprecated method, now just calls reorder_pols.
+
+        Args:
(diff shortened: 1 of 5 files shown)


pyuvdata/uvdata.py
```diff
@@ -910,6 +886,67 @@ def set_uvws_from_antenna_positions(self, allow_phasing=False,
             self.phase(phase_center_ra, phase_center_dec, phase_center_epoch,
                        phase_frame=output_phase_frame)
 
+    def reorder_pols(self, order='AIPS', desired_order=None, run_check=True,
```
adampbeardsley: It doesn't look like keyword `desired_order` is used (or documented).

bhazelton: oops, thanks for catching that. It's cruft.

Issue #548, opened 2019-03-15
Move UVFlag object into pyuvdata


bhazelton:
First step: copy over from hera_qm
second step: base it on UVBase
third step: win!

plaplant:
Steps 2 and 3 are their own issues. Step 1 done in #560.

PR #560, opened 2019-04-09
Move preliminary uvflag feature from hera_qm to pyuvdata

adampbeardsley:
## Description

Moved UVFlag class from hera_qm to pyuvdata. UVFlag is currently _not_ a UVBase object, but we intend to eventually upgrade to such. We would like to move it to pyuvdata "as-is" as a non-documented feature to help development of other codes that use it but do not want to require hera_qm.
We also moved relevant utility functions, and all the associated tests (for utils and UVFlag).

## Motivation and Context

Provides an object for handling flags, removes hera_qm dependency from other code bases.

Addresses first point of issue #548.

## Types of changes

## Checklist:


pyuvdata/tests/__init__.py
```diff
@@ -201,6 +201,13 @@ def skipIf_no_h5py(test_func):
     return test_func
 
 
+def skipIf_no_hera_cal(test_func):
+    """defines a decorator to skip tests that require h5py."""
```
bhazelton: `h5py` -> `hera_cal`

pyuvdata/utils.py
```diff
@@ -959,3 +959,123 @@ def _reraise_context(fmt, *args):
...
+    Args:
+        a (array): Input array to process.
+        alg (str): Algorithm to use. Must be defined in this function with
+            corresponding subfunction below.
+        weights (array, optional): weights for collapse operation (e.g. weighted mean).
+            NOTE: Some subfunctions do not use the weights. See corresponding doc strings.
+        axis (int, tuple, optional): Axis or axes to collapse. Default is all.
+        returned (Bool): Whether to return sum of weights. Default is False.
+    '''
+    if alg == 'mean':
```
bhazelton: Might be cleaner to define a dict with keys equal to the algorithm names and values equal to the functions. Then you can use `getattr` to call the appropriate function. For a similar example, see the `interp` method on UVBeam.

pyuvdata/utils.py
```diff
@@ -959,3 +959,123 @@ def _reraise_context(fmt, *args):
         ex.args = (cstr, ) + ex.args[1:]
 
     raise
+
+
+def collapse(a, alg, weights=None, axis=None, returned=False):
+    ''' Parent function to collapse an array with a given algorithm.
+    Args:
+        a (array): Input array to process.
```
bhazelton: I'm slightly opposed to single letter variable names. Why not just use `array`?

pyuvdata/utils.py
```diff
@@ -959,3 +959,123 @@ def _reraise_context(fmt, *args):
...
+        alg (str): Algorithm to use. Must be defined in this function with
+            corresponding subfunction below.
+        weights (array, optional): weights for collapse operation (e.g. weighted mean).
+            NOTE: Some subfunctions do not use the weights. See corresponding doc strings.
+        axis (int, tuple, optional): Axis or axes to collapse. Default is all.
+        returned (Bool): Whether to return sum of weights. Default is False.
```
bhazelton: this is not a super intuitive keyword name for me. Maybe `return_weights`?

pyuvdata/uvflag.py
```diff
@@ -0,0 +1,832 @@
...
+        self.weights_array = np.swapaxes(self.weights_array, 0, 1)[np.newaxis, np.newaxis,
+                                                                   :, :, :]
+        self.weights_array = self.weights_array.repeat(len(uv.ant_array), axis=0)
+        self.ant_array = uv.ant_array
+        self.type = 'antenna'
+        self.history += 'Broadcast to type "antenna" with ' + self.pyuvdata_version_str
+
+    def to_flag(self):
+        '''Convert to flag mode. NOT SMART. Simply removes metric_array and initializes
+        flag_array with Falses.
```
bhazelton: Is there a method to threshold a metric to turn it into a flag array?

pyuvdata/tests/test_uvflag.py
```diff
@@ -0,0 +1,1280 @@
...
+@uvtest.skipIf_no_hera_cal
+def test_init_HERAData():
+    from hera_cal.io import HERAData
+    uv = UVData()
+    uv.read_miriad(test_d_file)
+    uvf1 = UVFlag(uv)
+    hd = HERAData(test_d_file, filetype='miriad')
+    hd.read()
+    uvf2 = UVFlag(hd)
+    nt.assert_equal(uvf1, uvf2)
```
bhazelton: What utility are you testing here that requires hera_cal? It seems like all of this can be done with just pyuvdata...

mwilensky768:
Another note is that, as is, install requires h5py.

bhazelton:
I think it's ok to add this "as-is" as a non-documented feature, but we should be sure to make all the relevant issues to track the needed improvements and documentation in conjunction with merging this PR.

pyuvdata/uvflag.py
```diff
@@ -0,0 +1,832 @@
...
+            force_pol: If True, will use 1 pol to broadcast to any other pol.
+                       Otherwise, will require polarizations match.
+        '''
+        if self.type == 'baseline':
+            return
+        if not (issubclass(uv.__class__, UVData) or (isinstance(uv, UVFlag) and uv.type == 'baseline')):
+            raise ValueError('Must pass in UVData object or UVFlag object of type '
+                             '"baseline" to match.')
+        if self.type != 'waterfall':
+            raise ValueError('Cannot convert from type "' + self.type + '" to "baseline".')
```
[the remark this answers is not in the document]
adampbeardsley: We made an issue (#565 ) to address this after we merge this PR.

pyuvdata/uvflag.py
adampbeardsley: Currently, this is done in `xrfi`, and the function here is just for converting the UVFlag mode. We could implement a very simple version here (a `threshold` keyword?), but I don't think we want to get into the business of flagging algorithms. What do you think?

pyuvdata/uvflag.py
```diff
@@ -0,0 +1,832 @@
...
+        if isinstance(input, (list, tuple)):
+            self.__init__(input[0], mode=mode, copy_flags=copy_flags, waterfall=waterfall, history=history)
+            if len(input) > 1:
+                for i in input[1:]:
+                    fobj = UVFlag(i, mode=mode, copy_flags=copy_flags, waterfall=waterfall, history=history)
+                    self += fobj
+                del(fobj)
+        elif isinstance(input, UVFlag):
+            self.type = input.type
+            self.history = input.history + 'Copied by ' + self.pyuvdata_version_str + history
```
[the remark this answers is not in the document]
adampbeardsley: If you're ok with it, we'd like to punt on the version string (made issue #568). We got around the second comment by simply removing the option to instantiate a UVFlag with another UVFlag (functionality remains with uvflag.copy)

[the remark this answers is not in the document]
bhazelton: yep, sounds good. Thanks for making the issue.

pyuvdata/uvflag.py
bhazelton: yeah, this probably belongs in analysis code. If someone wants it in the future it can be added then.

bhazelton: A very simple "threshold" version would also be totally reasonable.

pyuvdata/tests/test_uvflag.py
adampbeardsley: This will be moved back to hera_qm.

adampbeardsley:
Ok @bhazelton I think we've addressed your comments (either changes to the code, or punting with associated issues). Tests are passing, and we updated the CHANGELOG, so I think we're ready for a real review.

pyuvdata/uvflag.py
```diff
@@ -0,0 +1,832 @@
...
+from __future__ import print_function, division, absolute_import
+import numpy as np
+import os
+from . import UVData
+from . import UVCal
+from . import utils as uvutils
+from . import version as uvversion
+from . import telescopes as uvtel
+import warnings
+import h5py
```
bhazelton: I think we agreed to move these imports inside the function where they're needed and to decorate tests that require h5py with the `skipIf_no_healpy` decorator.

PR #581, opened 2019-04-25
UVBeam interpolation spline reuse fixes

aelanman:
The way UVBeam was storing interpolation splines for reuse had several flaws that came up when trying to use `reuse_splines` while calling `interp` with specified polarizations.

## Description

The `saved_interp_functions` attribute is now simply a dictionary of functions, where each key specifies a single polarization (or feed), frequency, efield component, and spectral window.
I've also split the spatial interpolation test function into two (one for power, one for efield).

## Motivation and Context

Originally, the `saved_interp_functions` dictionary had frequencies as keys, and each entry had a shape `(Naxes_vec, Nspws, Npols or Nfeeds`. This was fine if you ran `interp` without passing in a list of `polarizations`. However, if you passed in a specific polarization and then, on a subsequent call, requested a different polarization at the same frequency, the saved_interp_function entry for that frequency wouldn't have an entry for that polarization. The function would not carry out interpolation for the new polarization/feed.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- Changed the way interpolation splines are saved in UVBeam to fix errors related to polarization selections.
+
 ## [1.3.8] - 2019-05-01
 
 ### Added
@@ -18,7 +21,6 @@ All notable changes to this project will be documented in this file.
 - `UVData.order_pols` method in favor of `UVData.reorder_pols`.
 
 ### Fixed
-- Changed the way interpolation splines are saved in UVBeam to fix errors related to polarization selections.
 - Building pyuvdata on macOS now targets minimum macOS 10.9 if run on macOS 10.9 or above
 - Possible bug where `check_variables` dictionary can change size during `read_miriad` call
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.
 - `UVData.order_pols` method in favor of `UVData.reorder_pols`.
 
 ### Fixed
+- Changed the way interpolation splines are saved in UVBeam to fix errors related to polarization selections.
 - Building pyuvdata on macOS now targets minimum macOS 10.9 if run on macOS 10.9 or above
 - Possible bug where `check_variables` dictionary can change size during `read_miriad` call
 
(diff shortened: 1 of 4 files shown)


bhazelton:
There's a redundancy test failing in python 2.7 that I don't really understand...

aelanman:
That's a test that takes perfectly redundant baselines, shifts them randomly up to a maximum displacement, then checks that the same redundancies are recovered using that length as a tolerance. Since it's doing something random, it's possible it only fails some of the time. I don't think this is an issue with the redundancy finder, but with the test design.

What was happening was in very rare cases the random shifts brought two baselines from different groups to within the tolerance, changing the redundant groups found. The problem might be that the redundancies in the test file are not perfect. I've made two changes --- It now runs the "baseline shift and redundancy check" for many trials, so it's more capable of finding those failure cases. I've also slightly reduced the maximum distance of those shifts. With these changes, it passed when I raised the number of trials to 500. I lowered it to 50 to keep things quick.

bhazelton:
Ok. Are we sure we want a random number generator used in a test? We've gone away from that over time because of these kinds of issues.

aelanman:
This is really a check of the "tolerance" setting on the redundancy finder, confirming that the tolerance reflects maximum deviation baselines can have from redundancy. The best way I an think of to check that is to move the baseline vectors at random up to that tolerance and confirm that the same redundant groups are found.

I suppose an alternative might be to define fixed shifts at displacement near the tolerance? Possibly check different tolerances?

bhazelton:
I'd prefer having well chosen fixed shifts at displacements of interest. It guarantees that we are testing the thing we want to test.

aelanman:
That makes sense. I've changed the test to define a set of 5 shift vectors in a circle with radius equal to the tolerance, then loop over the baseline array and shift each by each vector one at a time. Each time it does this, it will check that you get the same redundancies as before.

It's a little slower as a result, but accomplishes the same thing.

bhazelton:
@aelanman if you rebase this, the hera_cal test will pass.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.
 - `UVData.order_pols` method in favor of `UVData.reorder_pols`.
 
 ### Fixed
+- Changed the way interpolation splines are saved in UVBeam to fix errors related to polarization selections.
```
bhazelton: This should go under 'unrealeased' rather than under 1.3.8 which was released Wednesday.

aelanman: Whoops... should be fixed now.

PR #582, opened 2019-04-25
Save lst_array in uvfits files

### LINKED ISSUE #567: write lst_array to uvfits file
to avoid recalculating on every read, which is very slow for files with a lot of times (e.g. pyuvsim reference sim 1.2)

bhazelton:
## Description

Add lst_array to uvfits metadata. This allows for it to not be recomputed on read (if run_check_acceptability is False).

## Motivation and Context

close #567 

## Types of changes

## Checklist:

### DIFF
--- a/pyuvdata/tests/test_uvfits.py
+++ b/pyuvdata/tests/test_uvfits.py
@@ -259,7 +259,13 @@ def test_readwriteread():
         hdulist.writeto(write_file, overwrite=True)
 
     uvtest.checkWarnings(nt.assert_raises, [ValueError, uv_out.read, write_file],
-                         message=['Telescope EVLA is not'])
+                         message=['Telescope EVLA is not',
+                                  'ERFA function "utcut1" yielded 1 of "dubious year (Note 3)"',
+                                  'ERFA function "utctai" yielded 1 of "dubious year (Note 3)"',
+                                  'LST values stored in this file are not self-consistent'],
+                         nwarnings=4,
+                         category=[UserWarning, astropy._erfa.core.ErfaWarning,
+                                   astropy._erfa.core.ErfaWarning, UserWarning])
 
     # check that unflagged data with nsample = 0 will cause warnings
     uv_in.nsample_array[list(range(11, 22))] = 0
--- a/pyuvdata/tests/test_uvfits.py
+++ b/pyuvdata/tests/test_uvfits.py
@@ -91,9 +91,23 @@ def test_source_group_params():
 
         par_names = vis_hdu.data.parnames
         group_parameter_list = []
-        # need to account for PZERO values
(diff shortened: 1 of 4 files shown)


bhazelton:
I just realized that all angle parameters in uvfits files are saved in degrees, so I'm going to change to save the lst_array in degrees in the uvfits files to be consistent.

pyuvdata/tests/test_uvfits.py
```diff
@@ -129,9 +143,23 @@ def test_multisource_error():
...
-        # need to account for PZERO values
+
+        lst_ind = 0
         for index, name in enumerate(par_names):
-            group_parameter_list.append(vis_hdu.data.par(name) - vis_hdr['PZERO' + str(index + 1)])
+            par_value = vis_hdu.data.par(name)
+            # lst_array needs to be split in 2 parts to get high enough accuracy
+            if name.lower() == 'lst':
+                if lst_ind == 0:
+                    # first lst entry, par_value has full lst value (astropy adds the 2 values)
```
plaplant: From the comment here, it sounds like astropy handles splitting the array into two entries. Have we checked that the CASA uvfits reader can handle this? Because the uvfits reader is one of the only ways to import pyuvdata objects into CASA, I'd like to check it works before merging this in.

bhazelton: We should check. It is the standard way to handle the time array, although I have seen files that don't do this. Also, I'm not sure if CASA will even read this because it's not one of the standard arrays. Do you know who is working with CASA and could do this test for us?

plaplant: I have read uvfits into CASA before, but don't have experience with ensuring the equality of two different measurement sets. @mkolopanis or @nkern may be able to offer some insight.

mkolopanis: It's a little tough to see, and when I say a little I mean almost impossible but CASA keeps crashing on me. I read the file `day2_TDEM0003_10s_norx_1src_1spw.uvfits` and wrote out in this branch. Here is a plot the amplitudes of the first file and the file written by this branch. The blue dots are the before, the red diamonds are the file written by this branch, if you zoom they mach up to the eye's precision at least. 

It doesn't _seem_ to cause an issue with casa's reader, but I bet there is a more rigorous test we could make too.

image

bhazelton: @plaplant does this meet your needs or do you want something more extensive?

plaplant: Thanks @mkolopanis. Yes, I think this is sufficient. I was mostly concerned about CASA choking entirely and not importing the data, so I'm satisfied.

PR #595, opened 2019-05-14
Nose to pytest conversion

### LINKED ISSUE #507: Move from using nosetests to pytest
Nose is no longer supported.

### LINKED ISSUE #594: A UVcal test asserts nothing
Found this test while going through the nose to pytest conversion. 

https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/1207782698bed06dc319193ba0ced06c38250907/pyuvdata/tests/test_uvcal.py#L147-L151
the first call to `nt.assert_true` does not actually test anything. assert on a function will just return True. 

Didn't want to forget about this as the conversion continues, but correcting the line to call the check function will fail the test.

```(py3) matthew@bahamut:~/src/pyuvdata/pyuvdata (nost_to_pytest)$ python -m pytest -x tests/test_uvcal.py::TestUVCalBasicMethods -s
=============================================================================================== test session starts ================================================================================================
platform linux -- Python 3.6.7, pytest-4.4.0, py-1.8.0, pluggy-0.9.0
rootdir: /home/matthew/work/src/pyuvdata
plugins: remotedata-0.3.1, openfiles-0.3.1, doctestplus-0.1.3, cov-2.7.1, arraydiff-0.3, nbval-0.9.1
collected 8 items                                                                                                                                                                                                  

tests/test_uvcal.py ....18
F

===================================================================================================== FAILURES =====================================================================================================
_________________________________________________________________________________ TestUVCalBasicMethods.test_nants_data_telescope __________________________________________________________________________________

self = <pyuvdata.tests.test_uvcal.TestUVCalBasicMethods testMethod=test_nants_data_telescope>

    def test_nants_data_telescope(self):
        self.gain_object.Nants_data = self.gain_object.Nants_telescope - 1
        print(self.gain_object.Nants_data)
>       assert self.gain_object.check()

tests/test_uvcal.py:147: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
uvcal.py:322: in check
    run_check_acceptability=run_check_acceptability)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pyuvdata.uvcal.UVCal object at 0x7f09e42c10b8>, check_extra = True, run_check_acceptability = True

    def check(self, check_extra=True, run_check_acceptability=True):
        """
        Check that required parameters exist. Check that parameters have
        appropriate shapes and optionally that the values are acceptable.
    
        Args:
            check_extra: If true, check shapes and values on all parameters,
        """
        if check_extra:
            p_check = [p for p in self.required()] + [p for p in self.extra()]
        else:
            p_check = [p for p in self.required()]
    
        for p in p_check:
            param = getattr(self, p)
            # Check required parameter exists
            if param.value is None:
                if param.required is True:
                    raise ValueError('Required UVParameter ' + p
                                     + ' has not been set.')
            else:
                # Check parameter shape
                eshape = param.expected_shape(self)
                # default value of eshape is ()
                if eshape == 'str' or (eshape == () and param.expected_type == 'str'):
                    # Check that it's a string
                    if not isinstance(param.value, str):
                        raise ValueError('UVParameter ' + p + ' expected to be '
                                         'string, but is not')
                else:
                    # Check the shape of the parameter value. Note that np.shape
                    # returns an empty tuple for single numbers. eshape should do the same.
                    if not np.shape(param.value) == eshape:
                        raise ValueError('UVParameter {param} is not expected shape. '
                                         'Parameter shape is {pshape}, expected shape is '
                                         '{eshape}.'.format(param=p, pshape=np.shape(param.value),
>                                                           eshape=eshape))
E                       ValueError: UVParameter _ant_array is not expected shape. Parameter shape is (19,), expected shape is (18,).

uvbase.py:264: ValueError
======================================================================================== 1 failed, 4 passed in 0.87 seconds ========================================================================================
```
 and I'm not sure about the actual behavior of this object or what is expected/not for this.

mkolopanis:
Converts nose style test to the pytest framework.
## Description

Overhauls test framework to rely on pytest instead of nose. Redefines the setup and teardown to use a pytest.fixture in `conftest.py`. makes Test.. Classes subclasses of unittest.TestCase so setUp and teardown are called.
## Motivation and Context

This conversion will keep the testing framework up to date since `nose` is no longer maintained. 

Discovered an issue with a UVCal test which does nothing, but turning the test on causes a failure. See #594 . I'm not personally familiar with the expectation from this class and wasn't sure the best course to solve that issue so it is not part of this PR.

 Also now closes closes #594. thanks @plaplant 

Closes #507 

Retained the `skip` function and added a deprecation warning when it is called.

Other `skipIF_` functions have been replaced with the pytest style skips but retain the old name. I've briefly used github's search feature and no other repos use these specific functions however they can be restored and the new pytest functions re-named if desired too.

## Types of changes

- [ ] Version change

## Checklist:

Breaking change checklist:  


README.md
```diff
@@ -142,8 +142,8 @@ To install via conda,  run
 
 
 ## Tests
-Requires installation of `nose` package.
-From the source pyuvdata directory run ```nosetests pyuvdata```.
+Uses the `pytest` package to execute test suite.
+From the source pyuvdata directory run ```pytest``` or ```python -n pytest```.
```
plaplant: `python -m pytest`?

pyuvdata/tests/__init__.py
```diff
@@ -143,13 +169,18 @@ def skip(reason):
...
     skip wrappers to skip tests that require optional dependencies.
-
     This is needed because nose doesn't respect unittest skip_if decorators.
     Based on: https://stackoverflow.com/questions/21936292/conditional-skip-testcase-decorator-in-nosetests
     Args:
         reason: String describing the reason for skipping a test.
     """
+    warnings.simplefilter('always', DeprecationWarning)  # turn off filter
+    warnings.warn("The skip function was built to work with the `nose` test "
+                  "suite and is deprecated. It will be released in a future version.",
```
plaplant: "removed in a future version"?

README.md
mkolopanis: yeah should be -m

pyuvdata/tests/__init__.py
mkolopanis: I could have sworn I wrote "removed in version 1.5". I blame the headache of the decorator that I originally tried to use

pyuvdata/tests/conftest.py
```diff
@@ -0,0 +1,41 @@
+# -*- mode: python; coding: utf-8 -*
+# Copyright (c) 2018 Radio Astronomy Software Group
```
plaplant: Copyright date should be 2019.

pyuvdata/tests/conftest.py
```diff
@@ -0,0 +1,41 @@
...
+
+import functools
+import types
+import six
+import six.moves.urllib as urllib
+
+from astropy.utils import iers
+
+from pyuvdata.data import DATA_PATH
+import pyuvdata.utils as uvutils
```
plaplant: I don't think this import is actually used.

mkolopanis:
In the current state I left in all the testObjects which use objects to setUp and teardown features of unittest. These could be converted to use fixtures and remove the objects/unittest entirely too. Thoughts?

pyuvdata/tests/test_utils.py
```diff
@@ -804,20 +803,20 @@ def test_mean_infs():
...
     ans[0] = 0
-    nt.assert_true(np.all(wo == ans))
+    assert np.all(wo == ans)
     print(data)
     out, wo = uvutils.mean_collapse(data, axis=1, return_weights=True)
     ans = np.mean(np.arange(data.shape[1])[1:]) * np.ones(data.shape[0])
     ans[0] = np.inf
     print(out)
     print(ans)
-    nt.assert_true(np.all(out == ans))
```
bhazelton: looks like there are some stray print statements above this line, can we remove them?

pyuvdata/tests/test_uvcal.py
```diff
@@ -138,105 +135,110 @@ def teardown(self):
...
-        self.gain_object.Nants_data = self.gain_object.Nants_telescope - 1
-        nt.assert_true(self.gain_object.check)
+        # make sure it's okay for Nants_telescope to be strictly greater than Nants_data
+        self.gain_object.Nants_telescope += 1
+        # add dummy information for "new antenna" to pass object check
+        self.gain_object.antenna_names = np.concatenate((self.gain_object.antenna_names, ["dummy_ant"]))
+        self.gain_object.antenna_numbers = np.concatenate((self.gain_object.antenna_numbers, [20]))
+        assert self.gain_object.check()
+        # make sure an error is raised if Nants_data is strictly greater than Nants_telescope
         self.gain_object.Nants_data = self.gain_object.Nants_telescope + 1
```
bhazelton: I'm concerned that this might be raising the wrong error. Changing Nants_data could cause errors because Nants_data is just wrong, not because it's smaller than Nants_telescope.

pyuvdata/tests/test_uvdata.py
```diff
@@ -234,48 +227,46 @@ def teardown(self):
 
     def test_baseline_to_antnums(self):
         """Test baseline to antnum conversion for 256 & larger conventions."""
-        nt.assert_equal(self.uv_object.baseline_to_antnums(67585), (0, 0))
-        nt.assert_raises(
-            Exception, self.uv_object2.baseline_to_antnums, 67585)
+        assert self.uv_object.baseline_to_antnums(67585) == (0, 0)
+        pytest.raises(Exception, self.uv_object2.baseline_to_antnums, 67585)
```
bhazelton: Should this be a more specific Exception type?

pyuvdata/tests/test_uvdata.py
```diff
@@ -234,48 +227,46 @@ def teardown(self):
...
-        nt.assert_equal(self.uv_object.antnums_to_baseline(
-            0, 0, attempt256=True), 257)
-        nt.assert_equal(self.uv_object.antnums_to_baseline(257, 256), 594177)
+        assert self.uv_object.antnums_to_baseline(0, 0, attempt256=True) == 257
+        assert self.uv_object.antnums_to_baseline(257, 256) == 594177
         uvtest.checkWarnings(self.uv_object.antnums_to_baseline, [257, 256],
                              {'attempt256': True}, message='found > 256 antennas')
-        nt.assert_raises(
-            Exception, self.uv_object2.antnums_to_baseline, 0, 0)
+        pytest.raises(Exception, self.uv_object2.antnums_to_baseline, 0, 0)
```
bhazelton: same question about the Exception type

ci/pyuvdata_min_deps_tests.yml
```diff
@@ -4,8 +4,8 @@ channels:
   - conda-forge
 dependencies:
   - astropy
-  - nose
   - numpy
   - scipy
   - six
   - coverage
+  - pytest-cov
```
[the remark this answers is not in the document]
mkolopanis: To my knowledge the `pytest-cov` is a plugin which uses `coverage` and formats the outputs differently. We don't need to specifically declare `coverage` since conda will grab it anyway. It can be removed if you want.

pyuvdata/tests/__init__.py
```diff
@@ -96,7 +119,6 @@ def checkWarnings(func, func_args=[], func_kwargs={},
         warnings.filterwarnings("ignore", message="numpy.ufunc size changed")
 
         # filter iers warnings if iers.conf.auto_max_age is set to None, as we do in testing if the iers url is down
-        from astropy.utils import iers
```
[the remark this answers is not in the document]
mkolopanis: it is imported in the file header too

pyuvdata/tests/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+    # try to download the iers table. If it fails, turn off auto downloading for the tests
+    # and turn it back on in teardown_package (done by extending auto_max_age)
+    try:
+        iers_a = iers.IERS_A.open(iers.IERS_A_URL)
+    except(urllib.error.URLError):
+        iers.conf.auto_max_age = None
+
+    yield
+
+    iers.conf.auto_max_age = 30
```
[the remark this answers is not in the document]
mkolopanis: It does. the `yield`  allows all the tests to run, then executes line 33 after completion. It might be possible to do something like 
```
def fin():
   do finishing things
```
 instead of a yield

pyuvdata/tests/test_uvdata.py
mkolopanis: It should be, I'll dig in and check the exact Exception raised

pyuvdata/tests/__init__.py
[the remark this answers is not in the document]
bhazelton: ah thanks. Sorry I missed that.

pyuvdata/tests/conftest.py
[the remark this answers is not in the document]
bhazelton: ok, that's fine. Thanks!

pyuvdata/tests/test_uvcal.py
mkolopanis: I think this part of the test existed even before the additions @plaplant made :thinking: I think it may just be throwing a ValueError because some attributes have shapes `(Nants_data, blah)`. Is it worth trying to rework this test, or just changing the caption to be explicit about the fact we're basically testing `check`?

pyuvdata/tests/test_uvdata.py
mkolopanis: this tests the following lines https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/63d6f7f7bb30ddea01f917b9a031a03ff52f20a5/pyuvdata/utils.py#L936-L939

pyuvdata/tests/test_uvdata.py
mkolopanis: this tests teh following https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/63d6f7f7bb30ddea01f917b9a031a03ff52f20a5/pyuvdata/utils.py#L936-L939

pyuvdata/tests/test_uvdata.py
mkolopanis: These lines also raise just an `Exception`
https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/63d6f7f7bb30ddea01f917b9a031a03ff52f20a5/pyuvdata/utils.py#L908-L910

pyuvdata/tests/test_uvcal.py
plaplant: I just inferred the reason for the test from its general structure. If indeed the test is raising an error because things like the `gain_array` are the wrong size, we should modify dimensions of data in the test so that the _only_ reason for failing the test is the incompatibility between `Nants_data` and `Nants_telescope`.

ci/pyuvdata_min_deps_tests.yml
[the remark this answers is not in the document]
bhazelton: ok, I'm fine with leaving it if it doesn't add something we don't need.

pyuvdata/tests/test_uvdata.py
bhazelton: hmm, interesting. Thanks for digging in!

pyuvdata/tests/test_uvdata.py
bhazelton: thanks!

pyuvdata/tests/test_uvcal.py
bhazelton: yep, I just checked and it's not raising the advertised error (another reason to move toward always specifying the expected error message). I'll push a fix soon.

bhazelton: Interestingly, when I tried to fix it it didn't error! So I added something to the check method to catch this problem, both in uvcal and in uvdata (and added tests for them as well)

mkolopanis:
I think this is ready for another look now

mkolopanis:
Check the contributing guide too

scripts/test_coverage.sh
```diff
@@ -6,4 +6,6 @@ cd $DIR/..
 python setup.py install
 
 cd pyuvdata/tests
-nosetests --with-coverage --cover-erase --cover-package=pyuvdata --cover-html "$@"
+python -m pytest --cov=pyuvdata --cov-config=../../.coveragerc\
+       --cov-report term --cov-report html:cover \
+       "$@"
```
adampbeardsley: Will this still work the same way? For example, if I just wanted to run `test_utils.py`, I would do:
`./scripts/test_coverage.sh test_utils.py`
or to do a specific test:
`./scripts/test_coverage.sh test_utils.py:test_XYZ_from_LatLonAlt`

mkolopanis: to run a specific file you can still call 
`./scripts/test_coverage.sh test_utils.py`
however to run a specific test you need two `:` e.g.
`./scripts/test_coverage.sh test_utils.py::test_XYZ_from_LatLonAlt`

mkolopanis: I'll leave reading all the options up to users, but two common ones `-x` is the same (exit after one failure or error) and the no output capture: `-s`.

PR #600, opened 2019-05-21
uvh5 antenna names 

mkolopanis:
Users casting `antenna_name` to an numpy array will break the uvh5 saving file.
## Description

added test to see if  `antenna_name` can be cast to an array before writing to uvh5.

## Motivation and Context

This test seems a little silly at first, since `antenna_names` is designed to be a list of string (or byte) of names. However I don't think it is inconceivable that a user could cast the names to an array or during an operation.  The uvh5 writer will write out the `antenna_names` as a single instance of a `bytes` object. It is very interesting resulting in something akin to a silent failure on write which cannot be read in.

It is easy to be robust against something like this by adding the `keepdims` keyword to the `np.string_` call while writing uvh5.

This is exclusively a python 3 issue.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 - Fixed a bug in UVData when reading in an FHD file with a single time integration.
 - Fixed a bug in how the longitudinal branch cut was handled in beam interpolation
 - Changed the way interpolation splines are saved in UVBeam to fix errors related to polarization selections.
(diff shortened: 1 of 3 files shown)


mkolopanis:
I have a fix for this too just was not sure if it is really a bug :thinking:

mkolopanis:
The fix looked just straight forward enough that I was concerned I was missing some minute hdf5 detail in it and I figured if anyone would know it would be Paul.

PR #623, opened 2019-06-14
Fixing version construction

### LINKED ISSUE #590: installation, version info and dependencies
I will try to summarize the issue. To be fair, this works fine in pyuvdata, at least with the conda installation. But several other packages which use the same infrastructure do *not* work (eg. ``hera_cal``). 

Basically, the package is structured like:

```
project/
   - setup.py
   - package/
      - __init__.py
      - version.py
```

Now, the ``version.py`` does some logic to get the version of the package, and also construct really cool versioning info (such as git hashes etc.). Typically, the ``__init__.py`` will look something like

```python
from . import module1
from . import module2
...

from .version import version
```

The problem comes in the ``setup.py``, which looks something like:

```python
from package import version
...

setup(
    install_requires = [...]
)
```

The problem that occurs is that the ``from package import version`` implicitly imports the ``package.__init__.py``, which itself tries to import all the modules in the package (which may or may not be justified in its own right). Typically those modules then try to import all the dependencies of the package, but those dependencies may not yet be installed, because the installer hasn't gotten to the actual ``setup`` function yet. 

This can be avoided merely by doing a ``sys.path.append("package")`` before the import (in ``setup.py``, and then doing ``import version`` instead of ``from package import version``. 
This will only import the ``version.py`` module, which has no other dependencies. 

The meta-problem is I guess how do you enforce this on other packages that want to use this functionality?

bhazelton:
## Description

Move version construction into setup.py, fix numpy.get_include issue.

## Motivation and Context

closes  #590 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/__init__.py
+++ b/pyuvdata/__init__.py
@@ -7,11 +7,19 @@
 """
 from __future__ import absolute_import, division, print_function
 
+import os
+
 # Filter annoying Cython warnings that serve no good purpose. see numpy#432
 # needs to be done before the imports to work properly
 import warnings
 warnings.filterwarnings("ignore", message="numpy.dtype size changed")
 warnings.filterwarnings("ignore", message="numpy.ufunc size changed")
+
+version_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'VERSION')
+with open(version_file) as f:
+    version = f.read().strip()
+
+__version__ = version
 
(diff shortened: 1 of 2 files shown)


bhazelton:
abandoning this because it looks like astropy has released new conda packages to fix the problem.

PR #631, opened 2019-07-04
Add option to return a new beam option from `interp`

### LINKED ISSUE #622: uv beam frequency interpolation with new object returns tuple
calling _interp_freq with new_object=true gives a tuple with a uv beam object as the zeroth entry and 'None' as the first entry. Is this the desired behavior?

bhazelton:
## Description

Add option to return a new beam object from `UVBeam.interp`. Remove corresponding functionality from the private `_interp_freq` method.

## Motivation and Context

closes #622 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

Breaking change checklist:  

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,9 +19,15 @@ All notable changes to this project will be documented in this file.
 - `UVCal.get_*` methods for accessing data arrays with antenna-polarization keys
 - `utils.uvcalibrate` for automated calibration of `UVData` by `UVCal`
 
+### Added
+- option for `UVBeam.interp` to return a new beam object.
+
 ### Fixed
 - Fixed a bug in select that caused bls and antenna_names/numbers to be or'ed rather than and'ed together.
 - Fixed a bug where `baseline_to_antnums` could accept a numpy array as input but not other array_like objects.
+
+### Changed
+- removed `new_object` keyword from `UVBeam._interp_freq` in favor of new functionality in `UVBeam.interp`
 
 ## [1.4.0] - 2019-05-23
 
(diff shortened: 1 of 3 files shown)


jpober:
Sam is on vacation until next week.  We'll take a look then and see if this is a clean fix for our use case.  (PS: Why can't I tag her here?  She opened an issue, but maybe you can only tag people that are part of the team?)

mkolopanis:
It's possible you might be able to tag but it won't autocomplete for you? let's try: @samanthamcgraw

bhazelton:
Great. @aelanman can you check to make sure this doesn't break anything in healvis or pyuvsim?

aelanman:
@bhazelton It does seem to have broken some pyuvsim tests, but this might just be issues that can be addressed in pyuvsim. One is because `UVBeam.interp` no longer accepts the keyword `freq_interp_kind`. I've fixed that by setting the attribute on each UVBeam object.

There's another that's a little harder to track down. One test is bringing up an error from scipy.interp1d:
`ValueError: The number of derivatives at boundaries does not match: expected 2, got 0+0`

This is coming from the frequency interpolation step.

`healvis` on the other hand needs to be updated to call the new functions, instead of the private `_interp_freq` method.

aelanman:
Update -- This last error is caused by using cubic frequency interpolation instead of linear. Curious that this test passed before...

I realized that although the antenna class functions accept `freq_interp_kind` as a keyword, there is no way for the user to set interpolation methods on the UVBeams. I'll make an issue on pyuvsim for that.

bhazelton:
@jpober @samanthamcgraw I think I fixed the bug you told me about on this branch. Sorry for the delay! Let me know if you still have trouble.

jpober:
This is working for us, thanks @bhazelton.  I'm happy to approve the PR, but @aelanman can you confirm that nothing will break for you if this change gets into master?

PR #632, opened 2019-07-06
Option in UVBase.check() to ignore unset required parameters

aelanman:
## Description

Adds a new option to the UVBase.check(), `only_check_set`. This ignores cases when required parameters aren't set, and only checks non-None parameters for acceptability.

## Motivation and Context

In pyuvsim, it happened that we weren't setting up certain parameters correctly at the start (specifically, `integration_time`). Though the initial UVData object does not have all the required parameters set, those that are set need to be checked for acceptability. I'm not aware of any other way to do this.

## Types of changes

- [ ] Version change

## Checklist:

- [ ] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/master/.github/CONTRIBUTING.md).

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 ## [1.4.0] - 2019-05-23
 
 ### Added
+- Option in UVBase to check acceptability of set parameters, ignoring whether all required parameters are set.
 - made an option to not save the `lst_array` to uvfits files.
 - `conjugate_bls` option to `UVData.get_antenna_redundancies`
 - `UVData.conjugate_bls` method to conjugate baselines to get the desired baseline directions.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ All notable changes to this project will be documented in this file.
 ## [1.4.0] - 2019-05-23
 
 ### Added
-- Option in UVBase to check acceptability of set parameters, ignoring whether all required parameters are set.
+- Option in UVBase.check() to ignore whether required parameters are set.
 - made an option to not save the `lst_array` to uvfits files.
 - `conjugate_bls` option to `UVData.get_antenna_redundancies`
 - `UVData.conjugate_bls` method to conjugate baselines to get the desired baseline directions.
--- a/pyuvdata/tests/test_uvbase.py
+++ b/pyuvdata/tests/test_uvbase.py
@@ -235,8 +235,8 @@ def test_warning():
     assert output, "hello world\n"
 
 
-def test_check_acceptability_only():
+def test_check_ignore_unset():
     test_obj = UVTest()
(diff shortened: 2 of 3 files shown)


pyuvdata/uvbase.py
```diff
@@ -232,6 +240,9 @@ def check(self, check_extra=True, run_check_acceptability=True):
                 otherwise only check required parameters.
             run_check_acceptability: Option to check if values in required parameters
                 are acceptable. Default is True.
+            only_check_set: Option to check only that values are acceptable for parameters
+                that have been set. This is equivalent to run_check_acceptability, but will
```
bhazelton: I don't think this docstring is quite right. `run_check_acceptability` checks whether the parameters have acceptable values, the basic check is what checks that the shapes are correct (and that the values aren't None). This code changes the basic check, not the acceptability check. Feel free to improve the docstring for `run_check_acceptability` if it's not clear enough.

pyuvdata/tests/test_uvbase.py
```diff
@@ -230,3 +233,10 @@ def test_location_degree_none():
 def test_warning():
     output = _warning("hello world")
     assert output, "hello world\n"
+
+
+def test_check_acceptability_only():
```
bhazelton: This test name confuses me, it's testing the `only_check_set` option, not the `run_check_acceptability` option.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 ## [1.4.0] - 2019-05-23
 
 ### Added
+- Option in UVBase to check acceptability of set parameters, ignoring whether all required parameters are set.
```
bhazelton: Again, I think this is confused between the acceptability check that the basic check.

pyuvdata/uvbase.py
aelanman: I've been having trouble determining how best to phrase this. It's mostly useful to the acceptability check, since it lets the user check that parameters are acceptable even when the UVData is incomplete. But it is a change to the basic check. In a way, it's separating the acceptability check from the requirements check.

I've renamed the keyword to `ignore_requirements` and tried to better explain the use for it in the docstring.

bhazelton: But if you *don't* set `run_check_acceptability`, it still lets you check that the shapes are right, which you can't without this option.

aelanman: Oh, I think I follow. So the acceptability check just checks that the values make sense, but the basic check also looks at shapes?

bhazelton: Yes. The basic check checks shapes and data types. The acceptibility check checks that the values are within specified ranges.

bhazelton: Apparently the docstring isn't clear enough, can you update it to make it clearer?

Issue #637, opened 2019-07-12
iterator for "data-like" attributes


steven-murray:
It would be good to have some kind of iterator over the "data-like" attributes of the `UVData` class, which would then also lend itself to quickly determining whether the current instance was a "metadata-only" object.

jpober:
Alternatively (additionally?), we could just have a parameter on UVData that identifies the object as metadata only and then relevant methods check for this parameter and behave differently.

PR #640, opened 2019-07-12
add metadata_only property, allow metadata only reads of multiple files

### LINKED ISSUE #621: Add uv objects which are metadata only
Should enable setting read_data=False when reading _list of files_.  This solution will be similar to what was worked out for select() which has a metadata_only option.

This is critically needed to load large datasets.

### LINKED ISSUE #636: Have select on partial objects use ignore_requirements
Currently we turn off checks for selecting on partial uvdata objects.  We can now use the ignore_requirements option in check to check those that are defined.

bhazelton:
## Description

Add a metadata_only property to UVData objects that adjusts whether the data_array, flag_array and nsample_array are required. This allows for checking metadata only objects and makes other metadata only operations easier.

Add support for reading in multiple files with metadata only (setting read_data = False).

## Motivation and Context

closes #621 
closes #636 

partially addresses #637, but we could still add an iterator over the "data_like" elements.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
+
+### Added
+- `metadata_only` property on `UVData` to automatically detect if data-like arrays are present
+- support for combining metadata only objects and reading in multiple files as metadata only
 - `UVCal.get_*` methods for accessing data arrays with antenna-polarization keys
 - `utils.uvcalibrate` for automated calibration of `UVData` by `UVCal`
 
--- a/pyuvdata/tests/test_miriad.py
+++ b/pyuvdata/tests/test_miriad.py
@@ -569,7 +569,6 @@ def test_readWriteReadMiriad():
     uv_out = UVData()
     testfile = os.path.join(DATA_PATH, 'zen.2456865.60537.xy.uvcRREAA')
     write_file = os.path.join(DATA_PATH, 'test/outtest_miriad.uv')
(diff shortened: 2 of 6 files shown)


bhazelton:
I took out the `metadata_only` keyword out of a couple of methods, maybe we should add them back for backwards compatibility (maybe with a deprecation period?)

steven-murray:
@bhazelton doesn't having the keywords potentially break consistency? As in, if the object has no data, and the user specifies `metadata_only=False`, that's broken, right? 

I guess you could just add a deprecation warning if it's set by the user at all, otherwise it just gets set to the class-level attribute. If the user sets it in an inconsistent way, raise an error.

plaplant:
@bhazelton I think we should keep the `metadata_only` in, and add a deprecation warning if there's a desire to remove them in a future version. I believe the hera_cal tests are failing right now for a genuine reason (i.e., this changeset breaks the repo's tests), not the IERS issues we've been seeing. I'm a little out of the loop as far as the issues that this PR is addressing so I need to catch up, but I'd like to at least mention that I think we should keep that keyword in for now.

bhazelton:
Ok, I'll add the keywords back with suitable deprecation warnings and errors if they're in conflict with the property.

bhazelton:
@dannyjacobs can you look at this PR? It's to address your issue about reading in multiple files with metadata only. We rethought how metadata only objects were handled throughout the code, cleaning it up and making more development along those lines easier.

PR #643, opened 2019-07-17
add metadata_only property, allow metadata only reads of multiple files

### LINKED ISSUE #621: Add uv objects which are metadata only

### LINKED ISSUE #636: Have select on partial objects use ignore_requirements

### LINKED ISSUE #637: iterator for "data-like" attributes

bhazelton:
## Description

Add a metadata_only property to UVData objects that adjusts whether the data_array, flag_array and nsample_array are required. This allows for checking metadata only objects and makes other metadata only operations easier.

Add support for reading in multiple files with metadata only (setting read_data = False).

## Motivation and Context

closes #621
closes #636
closes #637

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
+
+### Added
+- `metadata_only` property on `UVData` to automatically detect if data-like arrays are present
+- support for combining metadata only objects and reading in multiple files as metadata only
 - `utils.uvcalibrate` flag propagation bug fix
 - `UVCal.ant2ind` indexing bug fix
 - `UVCal.get_*` methods for accessing data arrays with antenna-polarization keys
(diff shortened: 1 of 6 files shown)


bhazelton:
@dannyjacobs that is correct, but it does some other things as well. 

It changes the `required` attribute of the three data-like arrays to be `False` which allows us to run `check` on a metadata only object without errors, so that we can check that everything else makes sense. 

It also refactors the `metadata_only` option on `select` and `compress_by_redundancy` to use the new property rather than requiring user input.

And it does some minor fixes to the read methods to permit reading metadata only from multiple files.

PR #656, opened 2019-08-15
fix type error raising bug

bhazelton:
## Description

Fix the call raising a TypeError in `phase_to_time`, also fix test that wasn't catching the bug.

## Motivation and Context

Calling `phase_to_time` with a time that wasn't an Astropy Time object generated an incorrect error message because the TypeError wasn't being raised properly.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ All notable changes to this project will be documented in this file.
 ### Changed
 - `UVFlag` inherits from `UVBase` object.
 
+### Fixed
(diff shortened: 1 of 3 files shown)


bhazelton:
We could do that. I think we were not wanting to handle all the possible input formats, so just required a Time object to make the user figure out their format. But given that we already use jds for the time array, allowing a jd as well as a Time object might be a good choice.

PR #658, opened 2019-08-15
Redundancy branch cut fix

### LINKED ISSUE #650: get_antenna_redundancies fails to properly group North-South oriented baselines.
This example demonstrates the issue:

```python
import numpy as np
import pyuvdata
import matplotlib.pyplot as plt

eps = 1e-10 # the size of the error induced by pyuvdata ENU <-> ECEF transformations

# antenna positions at corners of a square with two small perturbations
ant_pos = np.array([
    [-eps,1.,0.],
    [1.,1.,0.],
    [eps,0.,0.],
    [1.,0.,0.]
])

ant_nums = np.arange(4)

red_grps, _, _ = pyuvdata.utils.get_antenna_redundancies(ant_nums, ant_pos, tol=0.1)

for grp in red_grps:
    bl_num = grp[0] # get one baseline from each redundant group
    ai, aj = pyuvdata.utils.baseline_to_antnums(bl_num, 4)
    bij = ant_pos[aj] - ant_pos[ai]
    
    plt.plot(bij[0], bij[1], '.k', ms=10)
plt.ylim(-1.1,1.1)
plt.xlim(-1.1,1.1)
plt.show()
```
produces this figure of baseline positions for this demo array

redundancy_group_failure

`red_grps` has 5 groups when there are really only 4. The points at (0, 1) and (0, -1) should be in the same redundant group. This looks like a problem with the initial attempt in `get_antenna_redundancies` to select only half the points.

aelanman:
## Description

Fixes a bug in `utils.get_antenna_redundancies` for North-South baselines. This method flipped baselines according to the u-positive convention. Under this convention, North-South baselines (u ~ 0) would be flipped only if u is exactly zero and v is negative. This means that a baseline with u = -ε and v < 0 would be flipped, but another with u = +ε w and v<0 would not. Redundancies among North-South baselines to the given tolerance are not identified.

This fixes that defining ` u = 0` as `np.isclose(u, 0, atol=tol)` (ie., the specified redundancy tolerance, and `u<0` as `u < -tol` (likewise for v and w). This means that very nearly north-south baselines will be flipped if v<0 (ie., pointing south), ensuring that North-south redundancies are identified.

The fix is tested using the example code from issue #650 .

## Motivation and Context

This fixes a significant problem for the redundancy finder when handling realistic baselines.

closes #650 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/tests/test_uvdata.py
+++ b/pyuvdata/tests/test_uvdata.py
@@ -3499,21 +3499,23 @@ def test_deprecated_redundancy_funcs():
                                                           category=DeprecationWarning,
                                                           message='UVData.get_baseline_redundancies has been replaced')
 
-    red_gps_new, _, _, _ = uv0.get_redundancies(include_autos=False, use_antpos=True)
+    red_gps_new, _, _, = uv0.get_redundancies(include_autos=False, use_antpos=True)
     assert red_gps_new == redant_gps
 
 
 def test_get_antenna_redundancies():
     uv0 = UVData()
     uv0.read_uvfits(os.path.join(DATA_PATH, 'fewant_randsrc_airybeam_Nsrc100_10MHz.uvfits'))
-    uv0.conjugate_bls('u>0')
-    red_gps, centers, lengths, _ = uv0.get_redundancies(use_antpos=True, include_autos=False)
+
+    old_bl_array = np.copy(uv0.baseline_array)
+    red_gps, centers, lengths = uv0.get_redundancies(use_antpos=True, include_autos=False, conjugate_bls=True)
+    # new and old baseline Numbers are not the same (different conjugation)
+    assert not np.allclose(uv0.baseline_array, old_bl_array)
 
     # assert all baselines are in the data (because it's conjugated to match)
     for i, gp in enumerate(red_gps):
         for bl in gp:
             assert bl in uv0.baseline_array
-    old_bl_array = np.copy(uv0.baseline_array)
 
     # conjugate data differently
     uv0.conjugate_bls(convention='ant1<ant2')
@@ -3523,8 +3525,9 @@ def test_get_antenna_redundancies():
 
     assert conjs is None
 
-    # new and old baseline Numbers are not the same (different conjugation)
-    assert not np.allclose(uv0.baseline_array, old_bl_array)
+    apos, anums = uv0.get_ENU_antpos()
+    new_red_gps, new_centers, new_lengths = uvutils.get_antenna_redundancies(
+        anums, apos, include_autos=False)
 
     # all redundancy info is the same
     assert red_gps == new_red_gps
@@ -3539,11 +3542,10 @@ def test_redundancy_contract_expand():
     uv0 = UVData()
     uv0.read_uvfits(os.path.join(DATA_PATH, 'fewant_randsrc_airybeam_Nsrc100_10MHz.uvfits'))
 
-    tol = 0.02   # Fails at lower precision because some baselines falling into multiple redundant groups
+    tol = 0.02   # Fails at lower precision because some baselines fall into multiple redundant groups
 
     # Assign identical data to each redundant group:
-    uv0.conjugate_bls('u>0')
-    red_gps, centers, lengths, _ = uv0.get_redundancies(tol=tol, use_antpos=True)
+    red_gps, centers, lengths = uv0.get_redundancies(tol=tol, use_antpos=True, conjugate_bls=True)
     for i, gp in enumerate(red_gps):
         for bl in gp:
             inds = np.where(bl == uv0.baseline_array)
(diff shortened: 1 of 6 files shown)


aelanman:
@zacharymartinot I'm having trouble tagging you as a reviewer, but could you take a look at this change?

aelanman:
@bhazelton I think I know why the remaining test is failing... The baseline conjugation done by `get_antenna_redundancies` no longer matches what's written in `conjugate_bls`. Fixing it may require changing the conjugate_bls function to use a tolerance when setting the index arrays.

bhazelton:
@aelanman `conjugate_bls` can do any arbitrary conjugation if you pass it a list of bls to conjugate. Does that help?

aelanman:
That might...  there's a lot more that needs to be fixed too, it seems. I'll keep at it.

aelanman:
@bhazelton This is passing all the tests now.

I've had to make a pretty significant change. The u>0 convention required by redundancy methods is different from the definition in `UVData.conjugate_bls()`, so I've changed the redundancy finding methods in utils.py to return baseline numbers in the convention `ant1<ant2`.

This should provide more predictable behavior while fixing the bug with north-south baselines.

pyuvdata/tests/test_uvdata.py
```diff
@@ -3518,31 +3523,31 @@ def test_redundancy_contract_expand():
...
     uv2.history = uv0.history
     # Inflation changes the baseline ordering into the order of the redundant groups.
     # reorder bls for comparison
-    uv0.reorder_blts()
-    uv2.reorder_blts()
+    uv0.reorder_blts(conj_convention='u>0')
+    uv2.reorder_blts(conj_convention='u>0')
     uv2._uvw_array.tols = [0, tol]
-
+#    import IPython; IPython.embed()
```
bhazelton: Should this line be removed?

pyuvdata/tests/test_uvdata.py
```diff
@@ -3518,31 +3523,31 @@ def test_redundancy_contract_expand():
...
     uvtest.checkWarnings(
         uv3.inflate_by_redundancy,
         [tol],
-        nwarnings=2,
-        category=[DeprecationWarning, UserWarning],
-        message=['The default for the `center` keyword has changed.',
-                 'Missing some redundant groups. Filling in available data.']
+        nwarnings=1,
+        category=[UserWarning],
+        message=[  # 'The default for the `center` keyword has changed.',
```
bhazelton: Just take this line out, no need to leave a commented line there.

pyuvdata/uvdata.py
```diff
@@ -4457,11 +4457,25 @@ def get_antenna_redundancies(self, tol=1.0, include_autos=True,
         baselines in this object, this method will conjugate baselines on this
         object to the 'u>0' convention unless `no_conjugate` is set to True.
```
bhazelton: This part of the docstring is out of date. There is no `no_conjugate` parameter, and I'm not sure the quoted convention is correct.

pyuvdata/utils.py
```diff
@@ -1184,8 +1195,13 @@ def get_antenna_redundancies(antenna_numbers, antenna_positions, tol=1.0, includ
         List of vectors describing redundant group centers
     lengths : list of float
         List of redundant group baseline lengths in meters
+
+
+    Notes
+    -----
+    Baseline numbers are returned in the convention ant1<ant2.
```
bhazelton: I'm confused about this convention. Don't they need to be in particular orientations to be redundant? I see no reason to always return in a specific convention, just stick with the orientations that make them redundant. Then you don't need the `return_conj` option, right?

aelanman: I could have it flip the baseline numbers when `return_conj` is false?

I need the `get_antenna_redundancies` method to be able to return baselines with the `ant1<ant2` convention for `inflate_by_redundancy` to work. `inflate_by_redundancy` first puts the UVData object in `ant1<ant2` form, then uses the conjugate ist from `utils.get_antenna_redundancies` to flip baselines. [Really, this step is done in the UVData.get_antenna_redundancies method]. This ensures that the mapping from compressed -> full doesn't need to conjugate any data along the way.

bhazelton: I don't think this is a general solution. It would not work if the baseline you had was one from the group that needed to be oriented the other direction to be redundant. I think the only reason it works in the tests is because you select the first baseline from the group in `compress_by_redundancy`.

aelanman: I'm sorry, but I don't follow... prior to inflating, all the baselines in the object are oriented such that nothing needs to be conjugated or flipped for the redundancy to hold within each group. _This_ is not in the `ant1<ant2` convention. The baseline numbers in the list of groups are also set this way. It's similar to u>0, but using the tolerance.

PR #673, opened 2019-08-27
Add BDA resampling methods

### LINKED ISSUE #428: convenience method for BDA handling: resampling averaged baselines

plaplant:
## Description
This PR adds methods on `UVData` objects for resampling datasets to either a uniform shorter integration time (`bda_upsample`) or longer time (`bda_downsample`). There are other related smaller changes, such as adding a `copy` method to the `UVData` object as well as support for phasing on metadata-only objects.

## Motivation and Context
With the advent of BDA datasets, resampling to a common integration time is important to test resampling datasets and understanding how code will be affected. It is also useful for reimplementing existing applications after a dataset has been resampled.

Closes #428.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:


jsdillon:
Sorry to take a while to look at this @plaplant (I've been traveling) but this looks pretty good from a high-level interface perspective. I think this will do what we need on the `hera_cal` side, but I won't be 100% sure until I go ahead and implement BDA support.

Two minor comments:
- It would be nice to have a `bda_resample()` function that dynamically upsamples or downsamples baselines as appropriate.
- The documentation for ordering parameters should refer to the documentation on `reorder_blts`

bhazelton:
Might need an option to do integrating vs averaging -- to match uncalibrated HERA data.

bhazelton:
@adampbeardsley @mkolopanis I think this branch might be ready for another look. I think we've addressed all the issues you raised before and the others that came up since. The only open thing I'm aware of is whether we want to rename the `integrate` keyword.

bhazelton:
Need to add something to the history to mark that resampling happened.

bhazelton:
change `integrate` to `summing_correlator_mode`

bhazelton:
change `bda_*` to `*_in_time`

pyuvdata/uvdata.py
```diff
@@ -4628,3 +4654,497 @@ def inflate_by_redundancy(self, tol=1.0, blt_order='time', blt_minor_order=None)
...
+                        # Then we use a test against 0 to turn it into a Boolean
+                        temp_flag[temp_idx] = np.sum(~self.flag_array[averaging_idx], axis=0) == 0
+
+                        mask = self.flag_array[averaging_idx]
+                        # need to update mask if a downsampled visibility will be flagged
+                        # so that we don't set it to zero
+                        if (temp_flag[temp_idx]).any():
+                            ax1_inds, ax2_inds, ax3_inds = np.nonzero(temp_flag[temp_idx])
+                            for ii in range(ax1_inds.size):
+                                mask[:, ax1_inds[ii], ax2_inds[ii], ax3_inds[ii]] = False
```
mkolopanis: I'm a little confused at the need to iterate through the arrays here. Is it not able to just be indexed as `mask[:, ax1_inds, ax2_inds, ax3_inds]`?  We're saying here that all indices for the 0th axis should be flagged right?

pyuvdata/uvdata.py
```diff
@@ -506,6 +507,28 @@ def check(self, check_extra=True, run_check_acceptability=True):
...
+
+        Returns
+        -------
+        uv : UVData
+            Copy of self.
+        """
+        uv = UVData()
+        for param in self:
+            # parameter names have a leading underscore we want to ignore
+            if metadata_only and param[1:] in self._data_params:
```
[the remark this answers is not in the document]
mkolopanis: also worth noting this would remove all leading `_` characters.

pyuvdata/uvdata.py
```diff
@@ -4628,3 +4654,497 @@ def inflate_by_redundancy(self, tol=1.0, blt_order='time', blt_minor_order=None)
...
+                                  "time gaps.".format(bl=self.baseline_to_antnums(bl)))
+                elif not np.isclose(dtime[0], int_times[0]):
+                    warnings.warn("The time difference between integrations is "
+                                  "not the same as the integration time for "
+                                  "baseline {bl}. The output may average across "
+                                  "longer time intervals than "
+                                  "expected".format(bl=self.baseline_to_antnums(bl)))
+
+            else:
+                # varying integration times for this baseline, need to be more careful
```
adampbeardsley: I _think_ one could simply run this else block instead of having the if statement at all.

pyuvdata/uvdata.py
adampbeardsley: I could be wrong, but I don't think that's quite the same. For example, if `ax1_inds = [0, 1, 2]`, `ax2_inds = [0, 1, 2]`, and `ax3_inds = [0, 1, 2]`, the code as written will set the diagonal to `False`, whereas your code would set the whole thing to `False`. In other words, code as written allows for fine grained selection of each individual pixel, but yours will only unflag slices.

pyuvdata/uvdata.py
```diff
@@ -4628,3 +4654,497 @@ def inflate_by_redundancy(self, tol=1.0, blt_order='time', blt_minor_order=None)
...
+                # time, or we've hit the end of the time samples for this baseline.
+                if over_min_int_time or last_sample:
+                    if last_sample and not (over_min_int_time or keep_ragged):
+                        # don't do anything -- implicitly drop these integrations
+                        continue
+                    # sum together that number of samples
+                    temp_baseline[temp_idx] = bl
+                    # this might be wrong if some of the constituent times are *totally* flagged
+                    averaging_idx = bl_inds[summing_idx:summing_idx + n_sum]
+                    temp_time[temp_idx] = np.average(self.time_array[averaging_idx])
```
adampbeardsley: I think there's an implicit assumption here (and line 5006) that the integration times are uniform within this baseline. Here I think it can be fixed with something like:
```
temp_time[temp_idx] = (np.sum(self.time_array[averaging_idx] * self.integration_integration_time[averaging_idx])
                       / np.sum(self.integration_integration_time[averaging_idx])
```
And a similar weighted mean would make sense below.

pyuvdata/uvdata.py
```diff
@@ -4628,3 +4654,497 @@ def inflate_by_redundancy(self, tol=1.0, blt_order='time', blt_minor_order=None)
...
+        # make sure we've populated the right number of baseline-times
+        assert temp_idx == temp_Nblts, ("Wrong number of baselines. Got {:d}, "
+                                        "expected {:d}. This is a bug, please "
+                                        "make an issue at https://github.com/"
+                                        "RadioAstronomySoftwareGroup/pyuvdata/"
+                                        "issues".format(temp_idx, temp_Nblts))
+
+        # drop data where we downsampled
+        inds_to_keep = np.nonzero(self.integration_time >= min_int_time)
+        self.baseline_array = self.baseline_array[inds_to_keep]
```
adampbeardsley: Basically from here to the end of the function is repeated between up/down sampling. Would it make sense to abstract this out to a function so we don't have so much repeated code?

pyuvdata/uvdata.py
mkolopanis: we should test this. If we don't need a for loop i would prefer to avoid it.

pyuvdata/uvdata.py
```diff
@@ -4628,3 +4654,497 @@ def inflate_by_redundancy(self, tol=1.0, blt_order='time', blt_minor_order=None)
...
+            self.unphase_to_drift()
+
+        # reorganize along blt axis
+        self.reorder_blts(order=blt_order, minor_order=minor_order)
+
+        # check the resulting object
+        self.check()
+
+        # add to the history
+        history_update_string = (" Upsampled data to {:f} second integration time "
```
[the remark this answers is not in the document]
plaplant: Fixed, thanks!

pyuvdata/uvdata.py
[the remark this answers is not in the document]
plaplant: Good suggestion, I used `lstrip`.

pyuvdata/uvdata.py
```diff
@@ -4628,3 +4654,497 @@ def inflate_by_redundancy(self, tol=1.0, blt_order='time', blt_minor_order=None)
...
+            # compute the new times of the upsampled array
+            t0 = self.time_array[ind]
+            dt = self.integration_time[ind] / n_new_samples[i]
+
+            offset = 0.5 + 0.5 * (n_new_samples[i] % 2)
+            n2 = n_new_samples[i] // 2
+            for ii, idx in enumerate(range(i0, i1)):
+                idx2 = ii + offset + n2 - n_new_samples[i]
+                nt = ((t0 * units.day) + (dt * idx2 * units.s)).to(units.day).value
+                temp_time[idx] = nt
```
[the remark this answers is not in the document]
plaplant: I've added some comments. Hopefully it's clearer now.

pyuvdata/uvdata.py
plaplant: I've tested this empirically. Dropping in the `ax?_inds` seems to Do The Right Thing and just zip them together (rather than take the outer product). I've opted for @mkolopanis's approach to eliminate the for loop.

pyuvdata/uvdata.py
```diff
@@ -4628,3 +4654,497 @@ def inflate_by_redundancy(self, tol=1.0, blt_order='time', blt_minor_order=None)
...
+        Resample to a shorter integration time.
+
+        This method will resample a UVData object such that all data samples have
+        an integration time less than or equal to the `max_int_time`. The new
+        samples are copied from the original samples (not interpolated).
+
+        Parameters
+        ----------
+        max_int_time : float
+            Maximum integration time to upsample the to in seconds.
```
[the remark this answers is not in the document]
plaplant: Thanks!

pyuvdata/uvdata.py
plaplant: I think you're right, but I've kept it as is because I'm lazy ;-)

pyuvdata/uvdata.py
plaplant: Good point! I've updated it as indicated here and in the actual data averaging below (which was slightly trickier to get right because of the mis-match in array dimensions between the data and `integration_time`).

pyuvdata/uvdata.py
plaplant: Yeah, probably better to have less repeated code. I've pulled out most of the repeated bits into a common function.

plaplant:
@adampbeardsley @mkolopanis thanks for the (multiple) thorough reviews! I think I've addressed all of the comments, but let me know if there is anything else that needs attention.

pyuvdata/tests/test_uvdata.py
```diff
@@ -3909,3 +3914,1014 @@ def test_lsts_from_time_with_only_unique():
...
+    out_wf = uv_object.get_data(0, 1)
+    # we need a "large" tolerance given the "large" data
+    new_tol = 1e-2 * np.amax(np.abs(uv_object.data_array))
+    assert np.isclose(init_wf[0, 0, 0], out_wf[0, 0, 0], atol=new_tol)
+
+    # this should be true because there are no flags
+    out_ns = uv_object.get_nsamples(0, 1)
+    assert np.isclose(init_ns[0, 0, 0], out_ns[0, 0, 0])
+
+    # try again with allow_drift=False
```
mkolopanis: might even break up the with and without `allow_drift` case

pyuvdata/tests/test_uvdata.py
```diff
@@ -3909,3 +3914,1014 @@ def test_lsts_from_time_with_only_unique():
...
+    # raise an error for a too-large integration time
+    max_integration_time = 1e3 * np.amax(uv_object.integration_time)
+    with pytest.raises(ValueError) as cm:
+        uv_object.downsample_in_time(max_integration_time)
+    assert str(cm.value).startswith("Increasing the integration time by more than")
+
+    # make copies for later
+    uv_object2 = uv_object.copy()
+    uv_object3 = uv_object.copy()
+    uv_object4 = uv_object.copy()
```
mkolopanis: again these sound like multiple tests

pyuvdata/tests/test_uvdata.py
```diff
@@ -3909,3 +3914,1014 @@ def test_lsts_from_time_with_only_unique():
...
+    assert out_data_136_137.size / 2 == init_data_136_137.size
+
+    # check some values
+    assert np.isclose(np.mean(init_data_1_136[0:4, 0, 0]), out_data_1_136[0, 0, 0])
+    assert np.isclose(np.mean(init_data_1_137[0:2, 0, 0]), out_data_1_137[0, 0, 0])
+    assert np.isclose(init_data_1_138[0, 0, 0], out_data_1_138[0, 0, 0])
+    assert np.isclose(init_data_136_137[0, 0, 0], out_data_136_137[0, 0, 0])
+
+    # again, with only_downsample set
+    uv_object2.resample_in_time(8, only_downsample=True)
```
mkolopanis: The only this branch or only that branch of the `resample_in_time` tests could probably be their own tests too.

pyuvdata/tests/test_uvdata.py
```diff
@@ -3495,17 +3500,17 @@ def test_set_uvws_from_antenna_pos():
...
         [ValueError, uv_object.set_uvws_from_antenna_positions, True, 'gcrs', 'xyz'],
-        message='Warning: Data will be unphased'
+        message='Data will be unphased'
     )
     uvtest.checkWarnings(
         uv_object.set_uvws_from_antenna_positions,
         [True, 'gcrs', 'gcrs'],
-        message='Warning: Data will be unphased'
+        message='Data will be unphased'
     )
```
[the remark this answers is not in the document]
plaplant: Thanks for catching this, I've explicitly tested the error messages are what we expect.

pyuvdata/tests/test_uvdata.py
```diff
@@ -3909,3 +3914,1014 @@ def test_lsts_from_time_with_only_unique():
...
+    assert np.allclose(uv_object_copy2.integration_time, max_integration_time * 0.5 / 0.75)
+    # we should double the size of the data arrays
+    assert uv_object_copy2.data_array.size == 2 * init_data_size
+    # output data should be different by a factor of 2
+    out_wf = uv_object_copy2.get_data(0, 1)
+    assert np.isclose(init_wf[0, 0, 0], out_wf[0, 0, 0])
+
+    # this should be true because there are no flags
+    out_ns = uv_object_copy2.get_nsamples(0, 1)
+    assert np.isclose(init_ns[0, 0, 0], out_ns[0, 0, 0])
```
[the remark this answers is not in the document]
plaplant: Done, these are now separate tests.

pyuvdata/tests/test_uvdata.py
```diff
@@ -3909,3 +3914,1014 @@ def test_lsts_from_time_with_only_unique():
...
+    assert uv_object_copy2.data_array.size == 2 * init_data_size
+    # output data should be half the input
+    out_wf = uv_object_copy2.get_data(0, 1)
+    assert np.isclose(init_wf[0, 0, 0] / 2, out_wf[0, 0, 0])
+
+    # this should be true because there are no flags
+    out_ns = uv_object_copy2.get_nsamples(0, 1)
+    assert np.isclose(init_ns[0, 0, 0], out_ns[0, 0, 0])
+
+    return
```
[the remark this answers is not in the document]
plaplant: Done.

pyuvdata/tests/test_uvdata.py
```diff
@@ -3909,3 +3914,1014 @@ def test_lsts_from_time_with_only_unique():
...
+    assert str(cm.value).startswith("Decreasing the integration time by more than")
+
+    # catch a warning for doing no work
+    uv_object2 = uv_object.copy()
+    max_integration_time = 2 * np.amax(uv_object.integration_time)
+    uvtest.checkWarnings(uv_object.upsample_in_time, [max_integration_time],
+                         message="All values in integration_time array are already shorter")
+    assert uv_object == uv_object2
+
+    return
```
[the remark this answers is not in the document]
plaplant: Done.

pyuvdata/tests/test_uvdata.py
```diff
@@ -3909,3 +3914,1014 @@ def test_lsts_from_time_with_only_unique():
...
+
+
+@uvtest.skipIf_no_h5py
+@pytest.mark.filterwarnings("ignore:The xyz array in ENU_from_ECEF")
+@pytest.mark.filterwarnings("ignore:The enu array in ECEF_from_ENU")
+def test_upsample_in_time():
+    """Test the upsample_in_time method"""
+    uv_object = UVData()
+    testfile = os.path.join(DATA_PATH, 'zen.2458661.23480.HH.uvh5')
+    uv_object.read(testfile)
```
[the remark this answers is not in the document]
plaplant: Good point, I've added a fixture for this test file (and the BDA one, because I broke the test that needed it up into separate tests.

pyuvdata/tests/test_uvdata.py
```diff
@@ -3909,3 +3914,1014 @@ def test_lsts_from_time_with_only_unique():
...
+    uv_object.phase_to_time(Time(uv_object.time_array[0], format="jd"))
+    # reorder to make sure we get the right value later
+    uv_object.reorder_blts(order="baseline", minor_order="time")
+
+    # make a copy for later
+    uv_object_copy = uv_object.copy()
+    uv_object_copy2 = uv_object.copy()
+    uv_object_copy3 = uv_object.copy()
+    uv_object_copy4 = uv_object.copy()
```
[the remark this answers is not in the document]
plaplant: Done.

pyuvdata/tests/test_uvdata.py
```diff
@@ -3909,3 +3914,1014 @@ def test_lsts_from_time_with_only_unique():
...
```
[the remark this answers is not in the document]
plaplant: Done.

pyuvdata/tests/test_uvdata.py
```diff
@@ -3909,3 +3914,1014 @@ def test_lsts_from_time_with_only_unique():
...
+    assert np.isclose((init_wf[0, 0, 0] + init_wf[1, 0, 0]) / 2., out_wf[0, 0, 0])
+
+    # this should be true because there are no flags
+    out_ns = uv_object.get_nsamples(0, 1)
+    assert np.isclose((init_ns[0, 0, 0] + init_ns[1, 0, 0]) / 2., out_ns[0, 0, 0])
+
+    # check that there are no flags
+    assert np.nonzero(uv_object.flag_array)[0].size == 0
+
+    # try again with allow_drift=False
```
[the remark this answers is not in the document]
plaplant: Done.

pyuvdata/tests/test_uvdata.py
plaplant: Done.

pyuvdata/tests/test_uvdata.py
plaplant: Indeed, these are now separate.

pyuvdata/tests/test_uvdata.py
plaplant: Done.

PR #676, opened 2019-08-30
Adjacency list calculation speedup

aelanman:
## Description

The redundancy finder was significantly slowed down by two nested for loops over the baselines, calculating distances between baselines for each pair. This changes to use the function `scipy.spatial.distance.pdist` to compute pairwise distances and build the adjacency list.

## Motivation and Context

This is much faster. Previously, trying to solve for the redundant baselines of HERA 350 took over five hours and still didn't finish. With this change, the calculation ran in minutes.

However -- this does increase the memory usage. The `pdist` function returns an array of size `Nbls * (Nbls - 1)/ 2`. This is then expanded to an `Nbls x Nbls` boolean array. For HERA350, this ultimately required around 15G of memory to handle. There may be a way to reduce this.

## Types of changes

- [ ] Version change

New feature checklist:
   (The existing redundancy tests are sufficient to check this).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
 - `UVFlag` information on Read The Docs
 
 ### Changed
+- utils.get_baseline_redundancies uses scipy pdist functions instead of for loops (faster)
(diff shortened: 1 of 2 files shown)


bhazelton (approved):
Looks good, thanks!

Issue #681, opened 2019-09-10
modernize test files


adampbeardsley:

bhazelton:
Some progress along this line, there are some recentish HERA data and cal files in the test data called:
zen.2458098.45361.HH.uvh5_downselected
zen.2458098.45361.HH.omni.calfits_downselected

PR #674, opened 2019-08-27
Warning cleanup

bhazelton:
## Description
Catch warnings in tests that weren't being caught. Move towards better warning & error handling in some tests:
- Use pytest's warning filtering mechanism for warnings unrelated to tests  (as opposed to checkWarning that really should be used for warnings that are related to the tests).
- Check that error messages are as expected (not just that errors are raised).

## Motivation and Context
Several uncaught warnings had crept into the tests. The logs shown on CircleCI were cut off before the warning section, so it was hard to tell from them that warnings weren't being caught. I removed the `-v` argument to pytest on CircleCI to shorten the logs.

While tackling the uncaught warnings, I went ahead and improved the warning and error handling in `test_miriad`, `test_uvfits` and some of the tests in `test_uvdata`. And I fixed one small bug that was exposed in this process (a warning was raised inappropriately). I don't really thing it merits a line in the changelog.

I also streamlined the circleci config to make sure there weren't differences in how the minimum dependencies tests were run because we were having weird errors with them.

## Types of changes
None of these types really apply. I guess bug fix is closest.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation change (documentation changes only)
- [ ] Version change
- [x] Build or continuous integration change

## Checklist:
- [x] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/master/.github/CONTRIBUTING.md).
- [x] My code follows the code style of this project.

Bug fix checklist:
- [x] My fix includes a new test that breaks as a result of the bug (if possible).
- [x] All new and existing tests pass in both python 2 and python 3.
- [ ] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/master/CHANGELOG.md).

PR #690, opened 2019-09-17
Make h5py a required package

### LINKED ISSUE #687: make h5py a primary dependency
change setup.py, readme (also list six in readme!)

remove all test skipping related to h5py & min dependency CI environment

plaplant:
## Description
This PR moves `h5py` from an optional dependency to a required one. Tests and package installation have been updated accordingly.

## Motivation and Context
Due to how integral `h5py` has become to multiple components of the repo (uvh5 files, UVFlag files, and the MWA beam), we have decided to move `h5py` to a required package. This change makes the test files simpler, and ensures that the user can access this functionality without additional installation requirements.

Closes #687.

## Types of changes
- [ ] Version change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file.
 - `UVFlag` information on Read The Docs
 
 ### Changed
+- `h5py` is now a required package instead of an optional one.
 - Phasing now supports metadata only `UVData` objects
 - utils.get_baseline_redundancies uses scipy pdist functions instead of for loops (faster)
 - UVData.get_antenna_redundancies will no longer automatically conjugate baselines.
--- a/README.md
+++ b/README.md
@@ -146,10 +146,11 @@ Required:
 * numpy >= 1.15
 * scipy
 * astropy >= 2.0
+* h5py
+* six
 
 Optional:
 
-* h5py (for reading and writing uvh5 format and uvflag save files)
 * python-casacore (for working with CASA measurement sets)
 * healpy (for working with beams in HEALPix formats)
 * pyyaml (for working with settings files for CST beam files)
--- a/ci/pyuvdata_min_deps_tests.yml
+++ b/ci/pyuvdata_min_deps_tests.yml
@@ -7,6 +7,7 @@ dependencies:
   - numpy
   - scipy
   - six
+  - h5py
   - coverage
   - pytest-cov
   - setuptools_scm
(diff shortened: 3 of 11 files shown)


plaplant:
It looks like the CI is failing on the same test (though at different points...) on python3.7 for linux and mac. It's in the `test_readWriteReadMiriad_partial` test, and in both cases the error is a mysterious `Runtime error in MIRIAD`.

In both cases though, the stderr shows the same error of `### Fatal Error:  Invalid preamble "time/baseline"`. It looks like the preamble checking is being done here:

https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/fce2f76c0905e8cf7d954ac569a9df1fa0c52478/pyuvdata/src/uvio.c#L2675

It's not clear to me why the preamble has an invalid value here, because in the wrapper into miriad, we're always hard-coding the preamble type to be the same thing:

https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/fce2f76c0905e8cf7d954ac569a9df1fa0c52478/pyuvdata/src/miriad_wrap.cpp#L66

And, as in so many of these ephemeral miriad errors, I cannot reproduce the issue locally. I'm not sure why these errors keep popping up, especially because as of #674, we delete the miriad objects in between each segment of the test (which is what I thought was causing these errors before). I'm wondering if there's some kind of weird race condition between different segments within the same test...? Honestly clutching at straws here.

Maybe this issue will go away when we address #681, but I don't know, because we can't seem to reproduce this reliably. I know @mkolopanis was digging into this issue a while back... Maybe there's been additional insight since then?

mkolopanis:
I got about as far as you did here. Tracked the issue down to those two lines. I was able to reproduce it once in a docker image with a fresh environment but only in python 3.7. I also came the same conclusion about deleting the these uvdata object. That usually would fix it and it would stay gone until I took out the delete and re-installed.

bhazelton:
I'm seeing the same error on my healpix branch, but only on linux python 3.7, not the mac one. I've rerun it several times and it keeps happening, so it seems to be getting a little bit more repeatable.

bhazelton:
I just noticed that we're not seeing this on the bda_resample PR, which is a bit out of date with master. Maybe we can change something back (even if it's not super obvious why it's causing this?)

mkolopanis:
I think you can re-base this and cross fingers for some relief from segfault/runtime issues.

plaplant:
@mkolopanis I agree that rebasing on master would probably make the errors go away, but I'm waiting for the BDA PR to go through so I can rebase off that, because that adds lots of testing skip decorators that I'm going to have to take out anyway.

PR #700, opened 2019-10-10
Uvbeam freq interp real interp bug fix

nkern:
Appropriately passes `freq_interp_kind` kwarg when beam is purely real.

## Description

The `freq_interp_kind`  wasn't passed to `scipy.interpolate.interp1d` when the beam was purely real, and would therefore default to simply linear: just a simple oversight, probably on my part. Anyways this fixes that and adds a test to make sure real and complex interpolation match when using the same beam for higher-than-linear interpolation.

This fixes the high delay interpolation artifacts we were seeing before. @aelanman @zacharymartinot @AaronParsons @bhazelton @nithyanandan 

beam_interp
^Frequency FT of beam averaged across the sky with uniform and beam weighting. As suspected, the beam-weighted average has less structure, meaning that not surprisingly the large zenith angles have more spectral structure. The beam is inherently sampled at 1 MHz meaning it achieves a maximum delay of 500 ns.

beam_linear_interp
^This is with linear interpolation onto a full 1024 channelization between 100-200 MHz, which is what @aelanman was showing before b/c the interpolation code was defaulting to linear by mistake.

beam_quad_interp
^This is with quadratic interpolation, which has lower sub-nyquist sidelobes, as suspected

beam_cubic_interp
^This is with cubic interpolation, which has even lower.

@aelanman I haven't tried with the GPR yet, but I think that the cubic interpolation does a good enough job, getting you 60-70 dB worth of suppression at sub-nyquist scales that it should be okay given what you want to do I think.

## Types of changes

- [ ] Version change

## Checklist:

- [ ] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/master/.github/CONTRIBUTING.md).

Bug fix checklist:

New feature checklist:

Breaking change checklist:  

Version change checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ All notable changes to this project will be documented in this file.
 - `UVFlag` objects can now be initialized without inputs
 
 ### Fixed
+- A bug in UVBeam._interp_freq where kind parameter was not passed for real-only beams
 - A bug in get_antenna_redundancies for nearly N-S baselines.
 - A bug where `conj_pol` could not handle cardinal direction polarizations.
 - A bug that gave the wrong error message when calling `UVData.phase_to_time` without an Astropy Time object.
--- a/pyuvdata/tests/test_uvbeam.py
+++ b/pyuvdata/tests/test_uvbeam.py
@@ -484,6 +484,36 @@ def test_freq_interpolation():
                   freq_array=freq_orig_vals, polarizations=['xx'])
 
 
+def test_freq_interp_real_and_complex():
+    # test interpolation of real and complex data are the same
+    power_beam = UVBeam()
+    power_beam.read_cst_beam(cst_files, beam_type='power', frequency=[150e6, 123e6],
+                             telescope_name='TEST', feed_name='bob',
+                             feed_version='0.1', feed_pol=['x'],
+                             model_name='E-field pattern - Rigging height 4.9m',
+                             model_version='1.0')
+    power_beam.interpolation_function = 'az_za_simple'
+
+    # make a new object with more frequencies
+    freqs = np.linspace(123e6, 150e6, 4)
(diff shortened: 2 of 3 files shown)


aelanman:
This is a good find, but I'm not sure it's the cause of the issue. For the beamfits file that I have, the `data_array` is of complex data type when loaded, which passes the `np.iscomplexobj` check and so is evaluated by the other part (which was passing along the freq_interp_kind).

I'm running some GSM simulations with different UVBeam interpolations now. I'll compare their delay spectra once they're done.

I'm not sure how to interpret your plots... it seems like you're just comparing the integral of A to the integral of A^2?

nkern:
@aelanman Okay, yeah I'd try running some quick, single spectrum, single baseline simulations with this pyuvdata branch to see if things are fixed. I think this these plots are telling us clearly that if the spline interpolation is quadratic or higher, you shouldn't be seeing those spikes, so hopefully things can get resolved.

aelanman:
healvis converts the efield beam to power using `UVBeam.efield_to_power`, but this doesn't change the data type of the object. That said, in more recent versions the UVBeam's data_array is put into a shared memory array (`mparray`), which forces it to float. However, the frequency interpolation is done prior to the simulation starting by calling `UVBeam._interp_freq` directly, and it explicitly passes along the interpolation kind.

freq_interp_compare_pspecs

^ These are power spectra for a single baseline from healvis simulations with different interpolation schemes. (But _not_ with this branch). GPR refers to a sim with gpr smoothing on a scale of 1MHz. "near" is nearest-neighbor, "five" is with a fifth-order spline, etc. It looks like the fifth-order defaulted to linear for some reason, since those lines overlap perfectly, but the others are different from each other. So I don't think the issue was a result of this bug.

PR #704, opened 2019-10-12
adds utils.apply_uvflag for applying uvflag to uvdata

nkern:
Adds uvflag.apply_uvflag and fixes a bug in UVFlag.to_baseline(..., force_pol=True)

## Description

1) adds a utils.apply_uvflag function for applying UVFlag objects to UVData objects
2) fixes a bug in UVFlag.to_baseline where the force_pol method did not work as advertised in the regime where the input UVData.Npols > 1

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

New feature checklist:

Breaking change checklist:  

Version change checklist:


nkern:
@adampbeardsley bump on this. lmk if someone else is more apt for a review, thanks!

pyuvdata/tests/test_uvflag.py
```diff
@@ -1300,6 +1300,24 @@ def test_to_baseline_force_pol():
...
+    uvf.flag_array[0, 10, 0] = True  # Flag time0, chan10
+    uvf.flag_array[1, 15, 0] = True  # Flag time1, chan15
+
+    uv2 = copy.deepcopy(uv)
+    uv2.polarization_array[0] = -6
+    uv += uv2
+    uvf.to_baseline(uv, force_pol=True)
+    assert np.all(uvf.baseline_array == uv.baseline_array)
+    assert np.all(uvf.time_array == uv.time_array)
+    assert np.array_equal(uvf.polarization_array, uv.polarization_array)
```
adampbeardsley: It looks like the bug that was fixed was that `Npol` wasn't updated. Could you add a check for it here so you're testing the bug?

pyuvdata/utils.py
```diff
@@ -492,6 +492,83 @@ def uvcalibrate(uvdata, uvcal, inplace=True, prop_flags=True, flag_missing=True,
...
+        UVData object to add flags to.
+    uvf : UVFlag object
+        A UVFlag object in flag mode.
+    inplace : bool
+        If True overwrite flags in uvd, otherwise return new object
+    force_pol : bool
+        If True, broadcast flags to all polarizations if they do not match.
+        Only works if uvf.Npols == 1.
+    unflag_first : bool
+        If True, completely unflag the UVData before applying flags.
```
adampbeardsley: Please clarify the behavior if False (ie, `OR` the flags)

pyuvdata/utils.py
```diff
@@ -492,6 +492,83 @@ def uvcalibrate(uvdata, uvcal, inplace=True, prop_flags=True, flag_missing=True,
...
+                raise ValueError("Input uvf and uvd polarizations do not match")
+
+        # make sure polarization ordering is correct: also edits inplace
+        uvf.polarization_array = uvf.polarization_array[[uvd_pols.index(pol) for pol in uvf_pols]]
+
+    # unflag if desired
+    if unflag_first:
+        uvd.flag_array[:] = False
+
+    # iterate over antpairs and apply flags: TODO need to be able to handle conjugated antpairs
```
adampbeardsley: You can use `ordered=False` in the call to `antpair2ind` to handle the conjugates. This should be done for this PR.

pyuvdata/utils.py
```diff
@@ -492,6 +492,83 @@ def uvcalibrate(uvdata, uvcal, inplace=True, prop_flags=True, flag_missing=True,
...
+                uvf.polarization_array = np.repeat(uvf.polarization_array, uvd.Npols, axis=-1)
+                uvf.Npols = len(uvf.polarization_array)
+                uvf_pols = uvf.polarization_array.tolist()
+
+            else:
+                raise ValueError("Input uvf and uvd polarizations do not match")
+
+        # make sure polarization ordering is correct: also edits inplace
+        uvf.polarization_array = uvf.polarization_array[[uvd_pols.index(pol) for pol in uvf_pols]]
+
```
adampbeardsley: There should be additional checks to be sure the time and frequency axes are the same. Ideally if they don't match, you can still flag what overlap there is, and have default behavior for anything else. But I think it's reasonable for a first attempt to simply check that the time array and freq arrays are equivalent and throw an error if they aren't.

pyuvdata/tests/test_uvflag.py
nkern: Sure thing. I did end up covering that b/c the failure would appear any time a `uvf.check()` was called, which I have covered, but I can just add an explicit check

pyuvdata/utils.py
nkern: `antpair2ind` doesn't have this feature in `UVFlag`

bhazelton:
When this is rebased, the changelog needs to be carefully examined to make sure that the changes from this PR are under the Unreleased section.

pyuvdata/tests/test_utils.py
```diff
@@ -1007,6 +1007,85 @@ def test_uvcalibrate_flag_propagation():
...
+    uvf2 = pyuvdata.UVFlag(uvd)
+    uvf2.to_flag()
+    uvd2.polarization_array[0] = -8
+    pytest.raises(ValueError, uvutils.apply_uvflag, uvd2, uvf2)
+
+    # test time and frequency mismatch exceptions
+    uvf2 = uvf.select(frequencies=uvf.freq_array[:, :2], inplace=False)
+    pytest.raises(ValueError, uvutils.apply_uvflag, uvd, uvf2)
+    uvf2 = uvf.select(times=np.unique(uvf.time_array)[:2], inplace=False)
+    pytest.raises(ValueError, uvutils.apply_uvflag, uvd, uvf2)
```
mkolopanis: There are a number of `pytest.raises` calls in this block. Can you please update them to explicitly check the error message to verify the correct error is raised.

pyuvdata/utils.py
```diff
@@ -492,6 +492,91 @@ def uvcalibrate(uvdata, uvcal, inplace=True, prop_flags=True, flag_missing=True,
...
+        # check time and freq shapes match: if Ntimes or Nfreqs is 1, allow implicit broadcasting
+        if uvf.Ntimes != uvd.Ntimes and uvf.Ntimes > 1:
+            raise ValueError("UVFlag and UVData have mismatched Ntimes")
+        if uvf.Nfreqs != uvd.Nfreqs and uvf.Nfreqs > 1:
+            raise ValueError("UVFlag and UVData have mismatched Nfreqs")
```
mkolopanis: I think we should be checking for overlapping times and frequencies, not that they are the same size. With this check we could have the same length of observations with a time translation and the flags would still be applied.

nkern: I don't actually think that's necessary. I do agree that it is true one can do that, but I don't think we should be in a position to tell the user that its the wrong thing to do: I can think of some reasons why one might want to knowingly do something like that. The check we do have (above) is more of a practical one: when Ntimes and Nfreqs don't match, you simply cannot add two arrays together, so its more of an elegant failure, rather than telling the user they can't do something.

pyuvdata/tests/test_utils.py
nkern: sure

pyuvdata/utils.py
adampbeardsley: Gotcha. Nevertheless, this should be addressed in some way for this PR. I think we could add a check within the `if ap not in uvf_antpairs` block to check for the conjugate.
At the least the TODO in comment here should be promoted to an issue on the repo.

pyuvdata/tests/test_utils.py
nkern: @mkolopanis what's the model for this for pytest? none of the other tests in `uvflag` have this feature.

pyuvdata/utils.py
mkolopanis: I can see the advantage to that, maybe there can be a flag to force time/frequency mismatches? It feels like the base use for this kind of function would be: turn metrics from a observation into flags, apply those flags to the same data the metrics/etc came from. From that perspective it feels like a check should be made that they match.  I think this would more elegantly do what @adampbeardsley suggested. Check for overlap, if no overlap and flag_anyway and size are same do it.

adampbeardsley: In addition to what Matt wrote, the broadcasting should be tested (if we keep it). It's not clear to me line 574 will work if one of the objects only has one time.

mkolopanis: or just if flag anyway and size match do it

pyuvdata/tests/test_utils.py
nkern: oh nvm, i see it now

pyuvdata/utils.py
```diff
@@ -492,6 +492,91 @@ def uvcalibrate(uvdata, uvcal, inplace=True, prop_flags=True, flag_missing=True,
...
+    uvf : UVFlag object
+        A UVFlag object in flag mode.
+    inplace : bool
+        If True overwrite flags in uvd, otherwise return new object
+    force_pol : bool
+        If True, broadcast flags to all polarizations if they do not match.
+        Only works if uvf.Npols == 1.
+    unflag_first : bool
+        If True, completely unflag the UVData before applying flags.
+        Else, OR the inheret uvd flags with uvf flags.
```
adampbeardsley: typo "inheret"

pyuvdata/utils.py
adampbeardsley: Ack, I was in the middle of a review when the other comments appeared, so this is a little out of order.
But I absolutely agree with Matt that the default should be that data with a time (or frequency) label should only be applied to data with the same label. Keywords to override this is fine, but the general user would expect labels to match.

nkern: Okay, so to compromise, what I'll do is just check the arrays match within the tolerance in the `UVFlag` in the case that `uvf.Nfreqs == uvd.Nfreqs` and raise an error if they don't, so that one can still flag in this scenario but they would need to manually raise `uvf._freq_array.tols` (so its kind of an advanced user kind of thing).

and the implicit broadcasting is covered: last test in the test function

pyuvdata/utils.py
nkern: I think that's should really use deeper infrastructure that still needs to be incorporated into `UVFlag`. I'll make a comment on this on the "Graduating UVFlag" issue if that's okay

adampbeardsley: Can you make a separate issue? Graduating UVFlag is supposed to coordinate things, but it'll get hard to track if we add individual issues.

adampbeardsley:
Just thinking about this.. we should also do something with the history in this function.

pyuvdata/utils.py
mkolopanis: Okay that will work for me for now. :+1:

nkern:
@adampbeardsley okay I think I've addressed your and Matt's comments, lmk if you want to see anything else with this PR! thanks

nkern:
@adampbeardsley just rebased this, feel free to approve and merge whenever you get a chance, thanks!

PR #705, opened 2019-10-15
Add rephasing

### LINKED ISSUE #644: Option to phase multiple files on read, enabling add on files with different phase centers
If you have a lot of files with different phase centers but want to create a single uvdata object from them with a single phase center (or a drift scan object), there isn't a graceful way to do it with a single read.  You need to read each one individually, then phase it, then add them.  Would be nice to just do it with a single call to read and a keyword that lets you define the new phase center or drift scan mode.

### LINKED ISSUE #654: Should UVData.read take arrays of filenames?
@kbharatgehlot was trying to read a set of files, and happened to have them in a numpy array (output of np.sort). I don't see a reason we shouldn't support that as an input to read.

### LINKED ISSUE #694: Should the loop over multiple files happen in the generic read or the file-type specific reads?
Right now the generic read doesn't loop and each file type specific read has similar code to do the looping. Is that the pattern we want or should we move (with a deprecation period) to doing the loop in the generic read?

### LINKED ISSUE #695: Calling self.phase on a phased object errors -- should it?
If we as to phase a phased object, should we allow it to phase to the new phase center?  It will probably be via unphasing to drift and then rephasing but that's what the user would have to do by hand anyway.

bhazelton:
## Description

Add better support for rephrasing phased data so that when reading in multiple phased files they can be rephrased and combined. Also deprecates calling file-specific read methods (e.g. `read_uvfits`) with multiple filenames (or sets of filenames for FHD or mwa_corr_fits). Adds support for arrays of filenames (not just lists).

## Motivation and Context

fixes #644 
fixes #695 
fixes #694
fixes #654 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

Breaking change checklist:  

### DIFF
--- a/pyuvdata/miriad.py
+++ b/pyuvdata/miriad.py
@@ -52,15 +52,15 @@ class Miriad(UVData):
         Parameters
         ----------
         filepath : str
-            The miriad file directoryto read from.
+            The miriad root directory to read from.
             Support for a list/array of file directories will be
             deprecated in version 2.0 in favor of a call to the generic
             `read` method.
         antenna_nums : array_like of int, optional
             The antennas numbers to read into the object.
         bls : list of tuple, optional
-            A list of antenna number tuples (e.g. [(0,1), (3,2)]) or a list of
-            baseline 3-tuples (e.g. [(0,1,'xx'), (2,3,'yy')]) specifying baselines
+            A list of antenna number tuples (e.g. [(0, 1), (3, 2)]) or a list of
+            baseline 3-tuples (e.g. [(0, 1, 'xx'), (2, 3, 'yy')]) specifying baselines
             to include when reading data into the object. For length-2 tuples,
             the ordering of the numbers within the tuple does not matter. For
             length-3 tuples, the polarization string is in the order of the two
@@ -73,7 +73,7 @@ class Miriad(UVData):
             and polarizations (e.g. '1', '1_2', '1x_2y').  See tutorial for more
             examples of valid strings and the behavior of different forms for ant_str.
             If '1x_2y,2y_3y' is passed, both polarizations 'xy' and 'yy' will
-            be kept for both baselines (1,2) and (2,3) to return a valid
+            be kept for both baselines (1, 2) and (2, 3) to return a valid
             pyuvdata object.
             An ant_str cannot be passed in addition to any of `antenna_nums`,
             `bls` or `polarizations` parameters, if it is a ValueError will be raised.
@@ -83,8 +83,11 @@ class Miriad(UVData):
             len-2 list containing min and max range of times in Julian Date to
             include when reading data into the object. e.g. [2458115.20, 2458115.40]
         read_data : bool
-            Read in the visibility and flag data. If set to false,
-            only the metadata will be read in. Setting read_data to False
+            Read in the uvws, times, visibility and flag data. If set to False,
+            only the metadata that can be read quickly (without reading the data)
+            will be read in. For Miriad, some of the normally required metadata
+            are not fast to read in (e.g. uvws, times) so will not be read in
+            if this keyword is False. Therefore, setting read_data to False
             results in an incompletely defined object (check will not pass).
         phase_type : str, optional
             Option to specify the phasing status of the data. Options are 'drift',
@@ -106,6 +109,18 @@ class Miriad(UVData):
             Option to check acceptable range of the values of parameters after
             reading in the file (the default is True, meaning the acceptable
             range check will be done). Ignored if read_data is False.
+
+        Raises
+        ------
+        IOError
+            If root file directory doesn't exist.
+        ValueError
+            If incompatible select keywords are set (e.g. `ant_str` with other
+            antenna selectors, `times` and `time_range`) or select keywords
+            exclude all data or if keywords are set to the wrong type.
+            If the data are multi source or have multiple
+            spectral windows.
+            If the metadata are internally consistent.
 
         """
         if not os.path.exists(filepath):
@@ -584,7 +599,7 @@ class Miriad(UVData):
         Parameters
         ----------
         filename : str
-            The miriad file directory to write to.
+            The miriad root directory to write to.
         run_check : bool
             Option to check for the existence and proper shapes of parameters
             after before writing the file (the default is True,
@@ -601,6 +616,15 @@ class Miriad(UVData):
         no_antnums : bool
             Option to not write the antnums variable to the file.
             Should only be used for testing purposes.
+
+        Raises
+        ------
+        ValueError
+            If the frequencies are not evenly spaced or are separated by more
+            than their channel width.
+            The `phase_type` of the object is "unknown".
+        TypeError
+            If any entry in extra_keywords is not a single string or number.
 
         """
         # change time_array and lst_array to mark beginning of integration, per Miriad format
@@ -920,8 +944,8 @@ class Miriad(UVData):
 
         Parameters
         ----------
-        filename : str
-            The miriad file to read.
+        uv : aipy.miriad.UV
+            aipy object to load metadata from.
         correct_lat_lon : bool
             Option to update the latitude and longitude from the known_telescopes
             list if the altitude is missing.
--- a/pyuvdata/miriad.py
+++ b/pyuvdata/miriad.py
@@ -84,7 +84,7 @@ class Miriad(UVData):
 
         # load metadata
         (default_miriad_variables, other_miriad_variables, extra_miriad_variables,
-         check_variables) = self.read_miriad_metadata(uv, correct_lat_lon=correct_lat_lon)
+         check_variables) = self._read_miriad_metadata(uv, correct_lat_lon=correct_lat_lon)
 
         if not read_data:
             # don't read in the data. This means the object is incomplete,
@@ -835,7 +835,10 @@ class Miriad(UVData):
 
     def read_miriad_metadata(self, filename, correct_lat_lon=True):
         """
-        Read in metadata (parameter info) but not data from a miriad file.
+        Deprecated: Read in metadata (parameter info) but not data from a miriad file.
+
+        Deprecated in favor of `_read_miriad_metadata` because it is not an
+        API level method (it's only called internally.)
 
         Args:
             filename : The miriad file to read
@@ -847,7 +850,12 @@ class Miriad(UVData):
             other_miriad_variables: list of other miriad variables
             extra_miriad_variables: list of extra, non-standard variables
             check_variables: dict of extra miriad variables
+
         """
+        warnings.warn('The read_miriad_metadata method is deprecated in favor of '
+                      '_read_miriad_metadata because it is not API level code. This '
+                      'function will be removed in version 1.6', DeprecationWarning)
+
         # check for data array
         if self.data_array is not None:
             raise ValueError('data_array is already defined, cannot read metadata')
@@ -857,7 +865,25 @@ class Miriad(UVData):
             uv = aipy_extracts.UV(filename)
         elif isinstance(filename, aipy_extracts.UV):
             uv = filename
+        self._read_miriad_metadata(uv)
 
+
+    def _read_miriad_metadata(self, uv, correct_lat_lon=True):
+        """
+        Read in metadata (parameter info) but not data from a miriad file.
+
+        Args:
+            uv : aipy_extracts.UV object
+            correct_lat_lon: flag -- that only matters if altitude is missing --
+                to update the latitude and longitude from the known_telescopes list
+
+        Returns:
+            default_miriad_variables: list of default miriad variables
+            other_miriad_variables: list of other miriad variables
+            extra_miriad_variables: list of extra, non-standard variables
+            check_variables: dict of extra miriad variables
+
+        """
         # load miriad variables
         (default_miriad_variables, other_miriad_variables,
          extra_miriad_variables) = self._load_miriad_variables(uv)
--- a/pyuvdata/miriad.py
+++ b/pyuvdata/miriad.py
@@ -120,7 +120,7 @@ class Miriad(UVData):
             exclude all data or if keywords are set to the wrong type.
             If the data are multi source or have multiple
             spectral windows.
-            If the metadata are internally consistent.
+            If the metadata are not internally consistent.
 
         """
         if not os.path.exists(filepath):
--- a/pyuvdata/miriad.py
+++ b/pyuvdata/miriad.py
@@ -26,8 +26,10 @@ from . import aipy_extracts
 class Miriad(UVData):
     """
     Defines a Miriad-specific subclass of UVData for reading and writing Miriad files.
-    This class should not be interacted with directly, instead use the read_miriad
+
+    This class should not be interacted with directly, instead use the read
     and write_miriad methods on the UVData class.
+
     """
 
     def _pol_to_ind(self, pol):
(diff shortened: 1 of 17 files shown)


mkolopanis:
The `hera_cal` and `hera_qm` failures here are a little annoying. In `hera_cal` the HERAData object which is used for processing overloads the `read` function. As a result when pyuvdata tries to call `self.read` while looping over a list of files it instead calls the subclass's `read` function.

We may be okay with these updates causing breaking changes on that object and the downstream repos updating to adjust. If we want to try to play nice with it we could do something like the following whenever we call `self.read` for the looping over lists.
```
            read_func = self.read
            if issubclass(self.__class__, UVData):
                read_func = super(self.__class__, self).read
            read_func
```

bhazelton:
This seems like something that the downstream repos should fix.

mkolopanis:
Interesting the tests I added failing on Mac OSX, uvw arrays seem to be the same to within one part in 10^5 but the data arrays are failing the == check.

bhazelton:
@mkolopanis there's a known error introduced by phasing & unphasing at something like a part in 10^5 (see the phasing memo), so in the existing phasing tests we increase the tolerance on the uvw_array to account for it. Not ideal, but the best we can currently do. I think increasing the tolerance for this test is entirely reasonable and in keeping with the other tests. Maybe add a comment that references the phasing memo.

bhazelton:
Oh, wait. The tolerances are already only 1e-3. Now I understand the confusion. I'll investigate on my mac.

bhazelton:
This is crazy. I'm seeing differences in the data array of order 1.

bhazelton:
does this also close #654?

pyuvdata/miriad.py
```diff
@@ -842,7 +850,12 @@ def read_miriad_metadata(self, filename, correct_lat_lon=True):
             other_miriad_variables: list of other miriad variables
             extra_miriad_variables: list of extra, non-standard variables
             check_variables: dict of extra miriad variables
+
         """
+        warnings.warn('The read_miriad_metadata method is deprecated in favor of '
+                      '_read_miriad_metadata because it is not API level code. This '
+                      'function will be removed in version 1.6', DeprecationWarning)
```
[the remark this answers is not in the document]
bhazelton: probably just 2.0, I copied this from another message. I can change it.

pyuvdata/tests/test_uvdata.py
```diff
@@ -453,20 +450,16 @@ def test_generic_read():
...
+    uv_raw = UVData()
+    uv_raw.read_uvh5(testfile)
+    # uvws in the file are wrong. reset them.
+    uv_raw.set_uvws_from_antenna_positions()
 
-    yield UV_phase, UV_raw
+    uv_phase = uv_raw.copy()
 
-    del UV_phase, UV_raw
+    yield uv_phase, uv_raw
```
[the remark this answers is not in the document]
bhazelton: sure.

pyuvdata/tests/test_uvdata.py
```diff
@@ -593,8 +586,9 @@ def test_unknown_phase_unphaseHERA_errors(
     UV_raw.set_unknown_phase_type()
     # if this is phase_to_time, use this index set in the dictionary and
     # assign the value of the time_array associated with that index
-    # this is a little hacky, but we cannot acces UV_raw.time_array in the parametrize
-    if phase_func == "phsae_to_time":
+    # this is a little hacky, but we cannot acces UV_raw.time_array in the
+    # parametrize
```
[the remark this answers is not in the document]
bhazelton: @mkolopanis can you help here?

pyuvdata/tests/test_uvdata.py
```diff
@@ -619,8 +613,9 @@ def test_phase_rephaseHERA_errors(
     UV_raw.phase(0., 0., epoch="J2000")
     # if this is phase_to_time, use this index set in the dictionary and
     # assign the value of the time_array associated with that index
-    # this is a little hacky, but we cannot acces UV_raw.time_array in the parametrize
-    if phase_func == "phsae_to_time":
+    # this is a little hacky, but we cannot acces UV_raw.time_array in the
+    # parametrize
+    if phase_func == "phase_to_time":
```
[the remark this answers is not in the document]
bhazelton: @mkolopanis and here?

pyuvdata/tests/test_uvdata.py
```diff
@@ -2103,6 +2103,235 @@ def test_break_add():
     pytest.raises(ValueError, uv1.__iadd__, uv2)
 
 
+@pytest.mark.filterwarnings("ignore:Telescope EVLA is not")
+@pytest.mark.parametrize("test_func,extra_kwargs",
+                         [("__add__", {}),
```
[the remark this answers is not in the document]
bhazelton: I think it's wrong, not sure why flake8 doesn't pick it up. I'll fix it.

pyuvdata/tests/test_uvfits.py
```diff
@@ -25,39 +25,46 @@
...
     """Test reading in a CASA tutorial uvfits file."""
     UV = UVData()
-    testfile = os.path.join(DATA_PATH, 'day2_TDEM0003_10s_norx_1src_1spw.uvfits')
+    testfile = os.path.join(DATA_PATH,
+                            'day2_TDEM0003_10s_norx_1src_1spw.uvfits')
     expected_extra_keywords = ['OBSERVER', 'SORTORD', 'SPECSYS',
                                'RESTFREQ', 'ORIGIN']
-    uvtest.checkWarnings(UV.read, [testfile], message='Telescope EVLA is not')
-    assert expected_extra_keywords.sort() == list(UV.extra_keywords.keys()).sort()
+    uvtest.checkWarnings(UV.read_uvfits, func_args=[testfile],
```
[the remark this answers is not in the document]
bhazelton: I used a marker everywhere else, I left it in this test to test the warning.

pyuvdata/uvfits.py
```diff
@@ -500,101 +495,6 @@ def read_uvfits(self, filename, antenna_nums=None, antenna_names=None,
                            blt_inds, False, run_check, check_extra, run_check_acceptability,
                            keep_all_metadata)
 
-    def read_uvfits_metadata(self, filename, run_check_acceptability=True):
```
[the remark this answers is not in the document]
bhazelton: yes.

pyuvdata/uvdata.py
```diff
@@ -70,6 +70,7 @@ def __init__(self):
         self._data_array = uvp.UVParameter('data_array', description=desc,
                                            form=('Nblts', 'Nspws',
                                                  'Nfreqs', 'Npols'),
+                                           tols=(0, 1e-08),
```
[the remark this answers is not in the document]
bhazelton: I just used the default numpy.allclose absolute tolerance. I don't really have a reason other than that, but all existing tests pass with this set. I'm willing to change it to 1e-7 if you think that's better.

pyuvdata/uvdata.py
```diff
@@ -3444,150 +3506,237 @@ def read(self, filename, axis=None, file_type=None,
...
-                             read_data=read_data, read_metadata=read_metadata,
-                             run_check=run_check, check_extra=check_extra,
-                             run_check_acceptability=run_check_acceptability,
-                             keep_all_metadata=keep_all_metadata, axis=axis)
+            raise ValueError('Only one of antenna_nums and antenna_names can '
+                             'be provided.')
+
+        if multi:
+            if file_type == 'uvfits':
+                if not read_data and not read_metadata:
```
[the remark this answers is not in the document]
bhazelton: I don't know. It has existed for awhile, so we at least have to support it with deprecation. I'm willing to deprecate it if we think it's not actually useful.

bhazelton:
@PyxieLouStar I did some work integrating `read_mwa_corr_fits` in this PR to make it symmetric with other read methods. Can you please take a look and make sure it makes sense?

pyuvdata/uvdata.py
```diff
@@ -3726,6 +3778,38 @@ def read(self, filename, axis=None, file_type=None, allow_rephase=True,
...
+                else:
+                    select = False
+
+                self.read_mwa_corr_fits(
+                    filename, run_check=run_check,
+                    use_cotter_flags=use_cotter_flags,
+                    correct_cable_len=correct_cable_len,
+                    phase_data=phase_to_pointing_center,
+                    check_extra=check_extra,
+                    run_check_acceptability=run_check_acceptability)
```
PyxieLouStar: I think that either this should contain a `phase_center` keyword option, or we should remove that keyword from `read_mwa_corr_fits`. Maybe it makes sense to give the reader only a `phase_to_pointing_center` option. Or, maybe it makes sense to move phasing outside of the reader altogether.

bhazelton: So the phase_center business is taken care of below, where the read method applies the requested phasing. We could remove it from the `read_mwa_corr_fits` altogether, but I think we need to keep the pointing center information because it is only known to that method. Let me think a little more about whether there's a clean way to get that information back to the read method so the phasing can be done in one place.

pyuvdata/uvdata.py
```diff
@@ -3514,158 +3672,305 @@ def read(self, filename, axis=None, file_type=None,
...
+                    bls=select_bls,
+                    frequencies=select_frequencies,
+                    freq_chans=select_freq_chans,
+                    times=select_times,
+                    polarizations=select_polarizations,
+                    blt_inds=select_blt_inds,
+                    run_check=run_check,
+                    check_extra=check_extra,
+                    run_check_acceptability=run_check_acceptability,
+                    keep_all_metadata=keep_all_metadata)
```
mkolopanis: Just popping in after looking at the tests. With the exception of `select_times` it looks like all of these select keywords only get defined if the file type is one of `fhd`, `ms`, or `mwa_corr_fits`. I am betting this is because other files types support select on read but we _also_ want to support the selection with the `time_range` keyword. Is it crazy to have two selects? One inside the if file type check for all the `select_meta_data` parameters, and one outside for the select times? Sounds like it might slow things down a little for large datasets though.

bhazelton: In the other file types, if `time_range` is set then `select_times` is set and this select is run. There used to be separate selects for each file type but I consolidated them here.

bhazelton: see e.g. line 3781

mkolopanis:
I realized there very simply could just be an else statement to set the
select_antenna_nums and others that are specific to those there to None
also. I must have been tired this morning. I just noticed that it's trying
to select on say a uvh5 read and erroring because a few of the select
arguments are not set.

bhazelton:
Oh, maybe I screwed up the refactor, I'll take a look at the errors.

pyuvdata/uvdata.py
bhazelton: just fixed this.

pyuvdata/uvdata.py
bhazelton: I did what we discussed on slack -- I deprecated the `phase_center` keyword and changed (with deprecation warnings) the `phase_data` keyword to `phase_to_pointing_center`.

PyxieLouStar: Great, thanks!

pyuvdata/tests/test_uvdata.py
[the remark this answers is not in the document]
bhazelton: @mkolopanis ping on this

pyuvdata/uvdata.py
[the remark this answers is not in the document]
bhazelton: I will deprecate it.

pyuvdata/uvfits.py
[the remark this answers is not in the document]
bhazelton: I put them back and deprecated them properly.

pyuvdata/tests/test_uvdata.py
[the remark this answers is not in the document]
mkolopanis: Oh yeah this is to check the error given by these parameters, without looking at the actual values in the UVData object I didn't know what would be a reasonable JD to give phase to time. So i added the If statement after this comment to just extract the JD when `phase_to_time` is given as the phase function.
```
("phase_to_time", {"time": 0, "allow_rephase": False},
 "The phasing type of the data is unknown. Set the phase_type")
```

bhazelton:
@plaplant we can talk about this more in an hour, but I think it's ready for another pass when you have time.

pyuvdata/uvdata.py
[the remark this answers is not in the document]
bhazelton: we'll go back to the numpy defaults.

pyuvdata/tests/test_mwa_corr_fits.py
```diff
@@ -32,17 +32,54 @@ def test_ReadMWAWriteUVFits():
...
+
+
+def test_select_on_read():
+    mwa_uv = UVData()
+    mwa_uv2 = UVData()
+    mwa_uv.read_mwa_corr_fits(filelist[0:2], correct_cable_len=True)
+    unique_times = np.unique(mwa_uv.time_array)
+    select_times = unique_times[np.where((unique_times >= np.min(mwa_uv.time_array))
+                                         & (unique_times <= np.mean(mwa_uv.time_array)))]
+    print(unique_times)
```
plaplant: We should get rid of these print statements.

pyuvdata/tests/test_uvfits.py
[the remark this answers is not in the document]
plaplant: Got it, thanks!

pyuvdata/tests/test_uvdata.py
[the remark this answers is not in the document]
plaplant: Got it, thanks.

pyuvdata/tests/test_uvdata.py
```diff
@@ -655,38 +650,45 @@ def test_phasing():
...
+                              orig_phase_frame='gcrs',
                               phase_frame='gcrs',
                               use_ant_pos=True)
 
-    # the tolerances here are empirical -- based on what was seen in the external
-    # phasing test. See the phasing memo in docs/references for details
+    # the tolerances here are empirical -- based on what was seen in the
+    # external phasing test. See the phasing memo in docs/references for
+    # details.
+    print(np.max(np.abs(uvd1.uvw_array - uvd2_rephase.uvw_array)))
```
plaplant: We don't need this print anymore.

pyuvdata/tests/test_uvdata.py
```diff
@@ -2142,6 +2145,235 @@ def test_break_add():
...
+    with pytest.raises(ValueError) as cm:
+        getattr(uv_full, test_func)(uv_full, phase_center_radec=(0, 45),
+                                    unphase_to_drift=True,
+                                    **extra_kwargs
+                                    )
+    assert str(cm.value).startswith('phase_center_radec cannot be set if '
+                                    'unphase_to_drift is True.')
+
+
+@pytest.fixture()
```
plaplant: I have a slight preference for putting all fixture definitions at the top of the test file so they're in one place, but we can keep it here too.

pyuvdata/uvfits.py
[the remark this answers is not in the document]
plaplant: Thanks!

pyuvdata/uvdata.py
```diff
@@ -1550,19 +1596,67 @@ def __add__(self, other, run_check=True, check_extra=True,
         ValueError
```
plaplant: I'm not sure what the "right" way to do this is, but you mention that a `ValueError ` is raised in the description of the `phase_center_radec` parameter. I'm not sure if it should also go in this section, or the description above is sufficient. Honestly either is fine with me, more just thinking out loud.

plaplant: Though we should mention that a `ValueError` is also raised if `phase_center_radec` is not the right size.

pyuvdata/uvdata.py
```diff
@@ -1550,19 +1596,67 @@ def __add__(self, other, run_check=True, check_extra=True,
...
+            if (other.phase_type != 'drift'):
+                warnings.warn("Unphasing other UVData object to drift")
+                other.unphase_to_drift(phase_frame=orig_phase_frame,
+                                       use_ant_pos=use_ant_pos)
+
+        if phase_center_radec is not None:
+            if np.array(phase_center_radec).size != 2:
+                raise ValueError('phase_center_radec should have length 2.')
+
+            if (this.phase_type == 'drift'
```
plaplant: Thanks very much for changing these comparisons to use `np.isclose`! While I think that's the correct approach, it now makes the `if` clauses hard to parse. Maybe we can add an in-line comment here and above the other clauses explaining that we're checking that the phase centers are the same?

pyuvdata/uvdata.py
```diff
@@ -2914,8 +3043,11 @@ def read_miriad(self, filepath, axis=None, antenna_nums=None, ant_str=None,
 
         Parameters
         ----------
-        filepath : str or list of str
-            The miriad file directory or list of directories to read from.
+        filepath : str
+            The miriad file directoryto read from.
```
plaplant: I think it reads better as `The miriad root directory to read from.`

pyuvdata/uvdata.py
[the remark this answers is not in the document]
plaplant: Thanks!

pyuvdata/uvdata.py
```diff
@@ -3614,158 +3791,320 @@ def read(self, filename, axis=None, file_type=None,
...
 
             if select:
-                unique_times = np.unique(self.time_array)
-                times_to_keep = unique_times[np.where((unique_times >= np.min(time_range))
-                                                      & (unique_times <= np.max(time_range)))]
-                self.select(times=times_to_keep, run_check=run_check, check_extra=check_extra,
-                            run_check_acceptability=run_check_acceptability,
-                            keep_all_metadata=keep_all_metadata)
+                if file_type in ['fhd', 'ms', 'mwa_corr_fits']:
+                    # these file types do not have select on read, so set all
```
plaplant: Very clear demarcation! Love the detailed comments :D

pyuvdata/uvdata.py
```diff
@@ -3575,17 +3738,31 @@ def read(self, filename, axis=None, file_type=None,
             Option to check acceptable range of the values of parameters after
             reading in the file (the default is True, meaning the acceptable
             range check will be done). Ignored if read_data is False.
+
```
plaplant: I think we should add a `Raises` section. From a quick glance, it looks like we raise `ValueError` if we can't infer the file type, or if the user specifies both `antenna_nums` and `antenna_names`. (There may be more, I didn't do an exhaustive search.)

pyuvdata/miriad.py
```diff
@@ -47,44 +49,77 @@ def read_miriad(self, filepath, antenna_nums=None, ant_str=None, bls=None,
...
+        ant_str : str, optional
+            A string containing information about what antenna numbers
+            and polarizations to include when reading data into the object.
+            Can be 'auto', 'cross', 'all', or combinations of antenna numbers
+            and polarizations (e.g. '1', '1_2', '1x_2y').  See tutorial for more
+            examples of valid strings and the behavior of different forms for ant_str.
+            If '1x_2y,2y_3y' is passed, both polarizations 'xy' and 'yy' will
+            be kept for both baselines (1,2) and (2,3) to return a valid
+            pyuvdata object.
+            An ant_str cannot be passed in addition to any of `antenna_nums`,
```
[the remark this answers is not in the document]
bhazelton: yes, I'll add one. Turns out there are lots of reasons a ValueError can be raised.

pyuvdata/miriad.py
```diff
@@ -47,44 +49,77 @@ def read_miriad(self, filepath, antenna_nums=None, ant_str=None, bls=None,
...
+            `bls` or `polarizations` parameters, if it is a ValueError will be raised.
+        polarizations : array_like of int or str, optional
+            List of polarization integers or strings to read-in. e.g. ['xx', 'yy', ...]
+        time_range : list of float, optional
+            len-2 list containing min and max range of times in Julian Date to
+            include when reading data into the object. e.g. [2458115.20, 2458115.40]
+        read_data : bool
+            Read in the visibility and flag data. If set to false,
+            only the metadata will be read in. Setting read_data to False
```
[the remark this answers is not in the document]
bhazelton: It is true for Miriad because the metadata that can be read from Miriad files without reading the data (so the fast part) doesn't include all the metadata we require for a metadata only object. I can add more explanation.

pyuvdata/miriad.py
```diff
@@ -852,7 +912,32 @@ def read_miriad_metadata(self, filename, correct_lat_lon=True):
...
             uv = filename
+        self._read_miriad_metadata(uv)
+
+    def _read_miriad_metadata(self, uv, correct_lat_lon=True):
+        """
+        Read in metadata (parameter info) but not data from a miriad file.
+
+        Parameters
+        ----------
+        filename : str
```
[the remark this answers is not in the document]
bhazelton: I think it's an aipy object, thanks for catching that!

pyuvdata/tests/test_fhd.py
```diff
@@ -244,60 +253,60 @@ def test_breakReadFHD():
...
-    uvtest.checkWarnings(fhd_uv.read_fhd, [bad_filelist], {'run_check': False},
+    uvtest.checkWarnings(fhd_uv.read, [bad_filelist], {'run_check': False},
                          nwarnings=5, message=warn_messages)
 
     # bad flag file
     broken_flag_file = testdir + testfile_prefix + 'broken_flags.sav'
     bad_filelist = testfiles[1:] + [broken_flag_file]
     fhd_uv = UVData()
-    pytest.raises(ValueError, fhd_uv.read_fhd, bad_filelist)
+    pytest.raises(ValueError, fhd_uv.read, bad_filelist)
```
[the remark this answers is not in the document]
bhazelton: yep, I'll do that. I believe in them.

pyuvdata/tests/test_mwa_corr_fits.py
bhazelton: oops, thanks!

pyuvdata/tests/test_uvdata.py
bhazelton: thanks!

pyuvdata/tests/test_uvdata.py
bhazelton: There were several scattered throughout, I moved them all to the top.

pyuvdata/uvdata.py
bhazelton: sure, I added some text, you can see if it works for you.

pyuvdata/uvdata.py
bhazelton: sure!

pyuvdata/uvdata.py
bhazelton: I added a bunch of raises sections for the file type specific reads and writes throughout.

pyuvdata/miriad.py
```diff
@@ -47,44 +49,92 @@ def read_miriad(self, filepath, antenna_nums=None, ant_str=None, bls=None,
...
+        ------
+        IOError
+            If root file directory doesn't exist.
+        ValueError
+            If incompatible select keywords are set (e.g. `ant_str` with other
+            antenna selectors, `times` and `time_range`) or select keywords
+            exclude all data or if keywords are set to the wrong type.
+            If the data are multi source or have multiple
+            spectral windows.
+            If the metadata are internally consistent.
```
plaplant: Should this be "not internally consistent"?

pyuvdata/tests/test_uvdata.py
```diff
@@ -130,8 +128,91 @@ def bda_test_file():
     return
 
 
+@pytest.fixture(scope='function')
```
plaplant: Thanks for grouping these all together!

pyuvdata/uvfits.py
```diff
@@ -250,66 +245,98 @@ def read_uvfits(self, filename, antenna_nums=None, antenna_names=None,
...
+        ------
+        IOError
+            If filename doesn't exist.
+        ValueError
+            If incompatible select keywords are set (e.g. `ant_str` with other
+            antenna selectors, `times` and `time_range`) or select keywords
+            exclude all data or if keywords are set to the wrong type.
+            If the data are multi source or have multiple
+            spectral windows.
+            If the metadata are internally consistent or missing.
```
plaplant: Should this be "not internally consistent" or "internally inconsistent"?

pyuvdata/uvdata.py
```diff
@@ -2973,51 +3168,45 @@ def read_miriad(self, filepath, axis=None, antenna_nums=None, ant_str=None,
...
+        ------
+        IOError
+            If root file directory doesn't exist.
+        ValueError
+            If incompatible select keywords are set (e.g. `ant_str` with other
+            antenna selectors, `times` and `time_range`) or select keywords
+            exclude all data or if keywords are set to the wrong type.
+            If the data are multi source or have multiple
+            spectral windows.
+            If the metadata are internally consistent.
```
plaplant: "not internally consistent"?

pyuvdata/miriad.py
bhazelton: yes, thanks!

PR #712, opened 2019-11-07
Fix a bug in fhd_cal

### LINKED ISSUE #490: UVCal.read_fhd_cal returns gains ~2
The UVCal.read_fhd_cal function returns gains that are exactly twice those used internally in FHD. Is there a reason for this? If not we should correct it.

bhazelton:
## Description

Fix a copy/paste bug in fhd_cal that may explain an open bug report.

## Motivation and Context

I was doing some code cleanup with flake8, which pointed out that there was an unused variable. Examining the code, I think it was the result of a copy/paste error and I think it may fix issue #490 where FHD calibration solutions were getting doubled.

I need some testing to verify it fixes that bug.
edit: Ruby confirmed it fixed the bug.
fixes #490

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:


bhazelton:
@rlbyrne could you test this branch to see if it fixes the issue? If it does, we need to add a unit test that fails on master as a result of this bug and passes on this branch.

bhazelton:
And this is erroring because of a problem with the test file that was exposed by this fix. In the test file we have a different number of antennas and frequencies in the fit gain array than in the residual gain array.

rlbyrne:
I've tested this PR and confirmed that it resolves issue #490. Thank you!

adampbeardsley:
Just FYI - it looks like when you wrote "may fix #490", github would have closed the issue when we merge this. It's fine since Ruby confirmed, but may want to be careful in the future.

bhazelton:
Oops! I'll be more careful next time. And I'll also update the top message to say it does since Ruby confirmed.

bhazelton:
@adampbeardsley I had to rebase because of another PR merge, can you re-approve?

PR #727, opened 2019-11-26
Select time range

### LINKED ISSUE #627: add uvd.select(times=[time_min, time_max]
Add a select function that allows a time range rather than explicit time samples.  This should make it much easier to do the select once we have BDA (which will have different time samples for different baseline lengths).  Potentially it could also speed the selection by not requiring explicit matching of every time in the list, but just time_min <= t < time_max.

bhazelton:
## Description

Adds a `time_range` option to `select` and the various `read` methods for select on read.

## Motivation and Context

fixes #627 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/tests/test_uvdata.py
+++ b/pyuvdata/tests/test_uvdata.py
@@ -1118,7 +1118,7 @@ def test_select_time_range():
     Nblts_selected = np.nonzero((uv_object.time_array <= time_range[1])
                                 & (uv_object.time_array >= time_range[0]))[0].size
 
-    uv_object2 = copy.deepcopy(uv_object)
+    uv_object2 = uv_object.copy()
     uv_object2.select(time_range=time_range)
 
     assert times_to_keep.size == uv_object2.Ntimes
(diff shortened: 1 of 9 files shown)


pyuvdata/tests/test_uvdata.py
```diff
@@ -1102,6 +1102,54 @@ def test_select_times():
...
+    # check for error setting times and time_range
+    times_to_keep = unique_times[[0, 3, 5, 6, 7, 10, 14]]
+    with pytest.raises(ValueError) as cm:
+        uv_object.select(time_range=time_range, times=times_to_keep)
+    assert str(cm.value).startswith('Only one of "times" and "time_range" can be set')
+
+    # check for errors time_range not length 2
+    with pytest.raises(ValueError) as cm:
+        uv_object.select(time_range=time_range[0])
+    assert str(cm.value).startswith('time_range must be length 2')
```
mkolopanis: I know they use the same setup as the initial selection test, but can we please break these error checks into separate tests.

pyuvdata/uvdata.py
```diff
@@ -2415,6 +2419,25 @@ def _select_preprocess(self, antenna_nums, antenna_names, ant_str, bls,
                     raise ValueError(
                         'Time {t} is not present in the time_array'.format(t=jd))
 
+        if time_range is not None:
+            if np.array(time_range).size != 2:
```
mkolopanis: I don't know if it is anymore efficient that this, probably honestly it is the same operation, but numpy does offer a `np.size` function which just returns the size instead of us having to cast as an array then take the size. Not sure it matters in the end, but if you are interested. I've attached a rather bad benchmark, they looks to take about the same mean time but `np.size` has a lower variance at least in python 3.8 with numpy 1.17.4

```
In [4]: %timeit np.size([l for l in range(int(1e5))])                                                                                                                                                               
9.49 ms ± 8.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [5]: %timeit np.array([l for l in range(int(1e5))]).size                                                                                                                                                         
9.57 ms ± 34.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [6]: print(np.__version__)                                                                            
1.17.4
```

pyuvdata/uvh5.py
```diff
@@ -853,71 +860,95 @@ def _check_header(self, filename, run_check_acceptability=True):
...
-                An ant_str cannot be passed in addition to any of the above antenna
-                args or the polarizations arg.
-            frequencies: The frequencies to include when writing data to the file.
-            freq_chans: The frequency channel numbers to include when writing data to the file.
-            times: The times to include when writing data to the file.
-            polarizations: The polarizations to include when writing data to the file.
-            blt_inds: The baseline-time indices to include when writing data to the file.
-                This is not commonly used.
-            add_to_history: String to append to history before write out. Default is no appending.
+        Parameters
```
mkolopanis: Thanks for the docstring conversion here.

pyuvdata/tests/test_uvdata.py
bhazelton: sure

PR #732, opened 2019-12-12
Error if baselines appear in multiple redundant groups

### LINKED ISSUE #615: Make sure find redundancy methods only put each baseline in one redundant group
Should be asserts to enforce that if it's already the case.

aelanman:
## Description

Checks that the total number of baselines in redundant groups does not exceed the total number of baselines given to the finder. This ensures that baselines are only assigned to one group each.

## Motivation and Context

Redundancy is defined to within a tolerance. If this tolerance is too high, baselines can appear in multiple "redundant" groups, but this is physically meaningless. The redundancy finder should error instead.

closes #615 
## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -1390,6 +1390,10 @@ def get_baseline_redundancies(baselines, baseline_vecs, tol=1.0, with_conjugates
 
     lens = np.sqrt(np.sum(vec_bin_centers**2, axis=1))
 
+    if np.sum([len(bg) for bg in bl_gps]) > Nbls:
+        raise ValueError("Tolerance is too high. Some baselines are falling into multiple"
+                         " redundant groups.")
+
     return bl_gps, vec_bin_centers, lens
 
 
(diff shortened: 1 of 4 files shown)


pyuvdata/utils.py
```diff
@@ -1389,6 +1389,9 @@ def get_baseline_redundancies(baselines, baseline_vecs, tol=1.0, with_conjugates
         vec_bin_centers[gi] = np.mean(baseline_vecs[inds, :], axis=0)
 
     lens = np.sqrt(np.sum(vec_bin_centers**2, axis=1))
+    if np.sum([len(bg) for bg in bl_gps]) > Nbls:
+        raise ValueError("Some baselines are falling into multiple"
+                         " redundant groups. Redundancy tolerance is causing ambiguity.")
```
bhazelton: Maybe suggest decreasing the tolerance? It should be clear that's the problem from what you've got but a specific direction about how to fix it would be nice.

bhazelton:
can you rebase this? The coverage report is crazy and I think it's because it's not based off the current master.

PR #733, opened 2019-12-16
Remove all references to version.py in favour of SCM

steven-murray:
## Description

This PR removes the `version.py` module in favor of using `setuptools_scm` and `pygitversion`. 
The version string will now follow the basic tenets of `setuptools_scm`, except that (due to `pygitversion`), if it is not a pure version, the hash *and* branch name will be displayed. 

I have removed the separate-line `uvdata_version_str`, and just subbed-in the new single-line version, which has this info in it. 

## Motivation and Context

The main purpose of this PR is to test the workings of the new `pygitversion` repo on a real-world repo, before advertising it more broadly. The motivation is to have a single extricated module/repo that every repo can refer to, rather than copy-pasting `version.py` everywhere. `setuptools_scm` does most of the work that we required, and `pygitversion` adds the ability to note the branch name in the version string. 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation change (documentation changes only)
- [ ] Version change
- [x] Build or continuous integration change

## Checklist:

- [x] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/master/.github/CONTRIBUTING.md).
- [x] My code follows the code style of this project.

Build or continuous integration change checklist:
- [x] If required or optional dependencies have changed (including version numbers), I have updated the readme to reflect this.
- [ ] If this is a new CI setup, I have added the associated badge to the readme and to references/make_index.py (if appropriate).

PR #740, opened 2019-12-23
Scm version

steven-murray:
## Description

Use `setuptools_scm` along with `pygitversion` to handle the versioning (see #733). This has been closed and re-opened to be merged to v2.0.x because it doesn't support py2.

## Types of changes

- [ ] Version change


steven-murray:
Looks like everything is "working". Decreased relative coverage due to a couple of lines I don't know how to test (for them to trigger, the package has to not be installed...). Otherwise it's just downstream stuff that's breaking.

I'll let either @bhazelton or @mkolopanis review when you're ready.

setup.py
```diff
@@ -103,11 +100,13 @@ def is_platform_mac():
...
-    'version': version.version,
+    'scripts': [fl for fl in glob.glob('scripts/*') if not os.path.isdir(fl)],
+    'use_scm_version': {
+        'local_scheme': branch_scheme
+    },
     'include_package_data': True,
-    'setup_requires': ['pytest-runner', 'numpy>=1.15'],
-    'install_requires': ['numpy>=1.15', 'scipy', 'astropy>=2.0', 'h5py'],
+    # 'setup_requires': ['pytest-runner', 'numpy>=1.15'],
+    'install_requires': ['numpy>=1.15', 'six>=1.10', 'scipy', 'astropy>=2.0', 'h5py'],
```
mkolopanis: Why did `six` get re-introduced here?

steven-murray: almost certainly because I botched the rebase/merge

pyuvdata/VERSION
```diff
@@ -1 +0,0 @@
-1.4.2
```
mkolopanis: Why do we delete this file again? Isn't this where we define our current version?

steven-murray: Current version is gotten in `__init__` using magic.

mkolopanis: MAGIC!! On a serious note though, how do we describe version numbers with scm, I've never used it before. I see where `__version__` is defined in the `__init__` though but not sure where it pulls the context.

steven-murray: I don't actually understand the magic in the `__init__` very well. But basically you issue a normal git tag to change the version, and that's it. SCM will automatically get the version based on the git tag -- adding relevant info like whether the repo is dirty and how many commits there have been since the last tag, and what the current hash is if so. When creating a distribution for upload to pypi etc., it will make sure that it is a clean version, and save that information in the upload, so that if a user installs from pypi, they also get the version (even though they don't have the repo).

mkolopanis: very cool. Thank you.

setup.py
```diff
@@ -103,10 +100,12 @@ def is_platform_mac():
...
     ],
-    'scripts': glob.glob('scripts/*'),
-    'version': version.version,
+    'scripts': [fl for fl in glob.glob('scripts/*') if not os.path.isdir(fl)],
+    'use_scm_version': {
+        'local_scheme': branch_scheme
+    },
     'include_package_data': True,
-    'setup_requires': ['pytest-runner', 'numpy>=1.15'],
+    # 'setup_requires': ['pytest-runner', 'numpy>=1.15'],
```
mkolopanis: since this is moving to the `pyproject.toml` why not delete this line entirely?

pyuvdata/uvbase.py
```diff
@@ -62,12 +62,14 @@ def __init__(self):
...
-                                          + '.  Git branch: ' + uvversion.git_branch
-                                          + '.  Git description: ' + uvversion.git_description + '.')
+                                     + __version__ + '.')
+
+        # remove the following because __version__ has this info in it if available.
+        # if uvversion.git_hash != '':
+        #     self.pyuvdata_version_str += ('  Git origin: ' + uvversion.git_origin
+        #                                   + '.  Git hash: ' + uvversion.git_hash
+        #                                   + '.  Git branch: ' + uvversion.git_branch
+        #                                   + '.  Git description: ' + uvversion.git_description + '.')
```
mkolopanis: would also say the commented lines here should just be removed.

setup.py
```diff
@@ -39,9 +36,9 @@ def run(self):
         build_ext.run(self)
 
 
-data = [version.git_origin, version.git_hash, version.git_description, version.git_branch]
-with open(os.path.join('pyuvdata', 'GIT_INFO'), 'w') as outfile:
-    json.dump(data, outfile)
+# data = [version.git_origin, version.git_hash, version.git_description, version.git_branch]
+# with open(os.path.join('pyuvdata', 'GIT_INFO'), 'w') as outfile:
+#     json.dump(data, outfile)
```
mkolopanis: comment removal

.github/workflows/macosx_ci.yaml
```diff
@@ -34,13 +34,17 @@ jobs:
         run: |
           export PATH="$HOME/miniconda/bin:$PATH"
           source activate ${ENV_NAME}
-          python setup.py build_ext --force --inplace
+          pip install .
+#          pip install --global-option=build_ext --global-option="--inplace" --global-option="--force" .
```
mkolopanis: this one too.

pyuvdata/__init__.py
```diff
@@ -6,28 +6,32 @@
...
 from __future__ import absolute_import, division, print_function
+import warnings
+from pkg_resources import get_distribution, DistributionNotFound
+
+# Set the version automatically from the package details.
+try:
+    __version__ = get_distribution(__name__).version
+except DistributionNotFound:
+    # package is not installed
+    pass
```
mkolopanis: these are the lines that require the package not to be installed right? Could noqa them since I can't think of a reasonable way to even test it :thinking:

steven-murray: Good idea.

.github/workflows/macosx_ci.yaml
steven-murray: Done.

setup.py
steven-murray: Done

pyuvdata/uvbase.py
steven-murray: Done

setup.py
steven-murray: Yup.

bhazelton:
Looks good to me. If those lines are untestable, they can be marked with `# pragma: no cover` and they will be ignored in the coverage report. If we think they should be testable but haven't figured out how, then leaving them without the mark will help us get back to them later.

steven-murray:
@bhazelton I switched the `noqa` to `pragma: nocover` (woops).

@mkolopanis I dunno if keeping the `version.py` is that useful. I'll let @bhazelton weigh in on that.

steven-murray:
@mkolopanis @bhazelton  is this good to rebase/merge?

mkolopanis:
Yeah I think it is fine. Bryna doesn't seem to have any opinions on the version part so let's do it.

PR #749, opened 2020-01-16
Add routine mwa flagging for edge channels, beginning/end of obs, center channels

mwilensky768:
## Description
This adds the option to flag on read when reading in gpubox files using both generic `read` and `read_mwa_corr_fits`. How much of each category to flag (in title) is specified by a float input to a keyword, which needs to be an integer multiple of the corresponding resolution for the desired quanitity e.g. edge_width=80e3 will flag two fine channels per edge per coarse channel if the channel_width is exactly 40 kHz. Flagging the dc_offset is just a bool - default True. Routine flagging is default _off_ (perhaps routine is a bit of a misnomer if that is the default). There is some defensive error messaging for faulty inputs. 

## Motivation and Context
This change helps make the mwa_corr_fits reader emulate cotter slightly more by doing some flagging that cotter would also do when handling box files. 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

New feature checklist:

Breaking change checklist:  

Version change checklist:

### DIFF
--- a/pyuvdata/mwa_corr_fits.py
+++ b/pyuvdata/mwa_corr_fits.py
@@ -139,7 +139,7 @@ class MWACorrFITS(UVData):
     def read_mwa_corr_fits(self, filelist, use_cotter_flags=False, correct_cable_len=False,
                            phase_to_pointing_center=False, run_check=True, check_extra=True,
                            run_check_acceptability=True, flag_init=False,
-                           edge_width=80e3, start_flag=4.0, end_flag=6.0,
+                           edge_width=80e3, start_flag=2.0, end_flag=2.0,
                            flag_dc_offset=True):
         """
         Read in MWA correlator gpu box files.
--- a/pyuvdata/mwa_corr_fits.py
+++ b/pyuvdata/mwa_corr_fits.py
@@ -78,11 +78,14 @@ class MWACorrFITS(UVData):
         Parameters
         ----------
         edge_width: float
-            The width to flag on the edge of each coarse channel, in hz.
+            The width to flag on the edge of each coarse channel, in hz. Set to
+            0 for no edge flagging.
         start_flag: float
             The number of seconds to flag at the beginning of the observation.
+            Set to 0 for no flagging.
         end_flag: floats
-            The number of seconds to flag at the end of the observation.
+            The number of seconds to flag at the end of the observation. Set to
+            0 for no flagging.
         flag_dc_offset: bool
             Set to True to flag the center fine channel of each coarse channel.
 
@@ -182,14 +185,14 @@ class MWACorrFITS(UVData):
             channel. See associated keywords.
         edge_width: float
             Only used if flag_init is True. The width to flag on the edge of
-            each coarse channel, in hz. Errors if less than the channel_width of
-            the observation.
+            each coarse channel, in hz. Errors if not equal to integer multiple
+            of channel_width. Set to 0 for no edge flagging.
         start_flag: float
             Only used if flag_init is True. The number of seconds to flag at the
-            beginning of the observation.
+            beginning of the observation. Set to 0 for no flagging.
         end_flag: floats
             Only used if flag_init is True. The number of seconds to flag at the
-            end of the observation.
+            end of the observation. Set to 0 for no flagging.
         flag_dc_offset: bool
             Only used if flag_init is True. Set to True to flag the center fine
             channel of each coarse channel.
--- a/pyuvdata/mwa_corr_fits.py
+++ b/pyuvdata/mwa_corr_fits.py
@@ -69,9 +69,81 @@ class MWACorrFITS(UVData):
         self.data_array *= np.exp(-1j * 2 * np.pi * cable_len_diffs / const.c.to('m/s').value
                                   * self.freq_array.reshape(1, self.Nfreqs))[:, :, None]
 
+    def flag_init(self, edge_width=80e3, start_flag=4.0, end_flag=6.0,
+                  flag_dc_offset=True):
+        """
+        Do routine flagging of the edges, beginning and end of obs, as well as
+        the center fine channel of each coarse channel.
+
+        Parameters
+        ----------
+        edge_width: float
+            The width to flag on the edge of each coarse channel, in hz.
+        start_flag: float
+            The number of seconds to flag at the beginning of the observation.
+        end_flag: floats
+            The number of seconds to flag at the end of the observation.
+        flag_dc_offset: bool
+            Set to True to flag the center fine channel of each coarse channel.
+
+        Raises
+        ------
+        ValueError
+            If edge_width is not an integer multiple of the channel_width of the data (0 also acceptable).
+            If start_flag is not an integer multiple of the integration time (0 also acceptable).
+            If end_flag is not an integer multiple of the integration time (0 also acceptable).
+        """
+        if (edge_width % self.channel_width) > 0:
+            raise ValueError("The edge_width must be an integer multiple of the"
+                             "channel_width of the data or zero.")
+        if (start_flag % self.integration_time) > 0:
+            raise ValueError("The start_flag must be an integer multiple of the"
+                             "integration_time of the data or zero.")
+        if (end_flag % self.integration_time) > 0:
+            raise ValueError("The end_flag must be an integer multiple of the"
+                             "integration_time of the data or zero.")
+
+        num_ch_flag = int(edge_width / self.channel_width)
+        num_start_flag = int(start_flag / self.integration_time)
+        num_end_flag = int(end_flag / self.integration_time)
+
+        # Constant from the DSP architecture
+        num_coarse_chan = 24
+        num_fine_chan = len(self.Nfreqs) / num_coarse_chan
+
+        if num_ch_flag > 0:
+            edge_inds = []
+            for ch_count in range(num_ch_flag):
+                # count up from the left
+                left_chans = list(range(ch_count, self.Nfreqs, num_fine_chan))
+                # count down from the right
+                right_chans = list(range(self.Nfreqs - 1 - ch_count, 0, -num_fine_chans))
+                edge_inds = edge_inds + left_chans + right_chans
+
+            self.flag_array[:, :, edge_inds, :] = True
+
+        if flag_dc_offset:
+            center_inds = list(range(num_fine_chan / 2, self.Nfreqs, num_fine_chan))
+
+            self.flag_array[:, :, center_inds, :] = True
+
+        if (num_start_flag > 0) or (num_end_flag > 0):
+            shape = self.flag_array.shape
+            reshape = [self.Ntimes, self.Nbls, self.Nspws, self.Nfreqs, self.Npols]
+            self.flag_array = np.reshape(self.flag_array, reshape)
+            if num_start_flag > 0:
+                self.flag_array[:num_start_flag, :, :, :, :] = True
+            if num_end_flag > 0:
+                self.flag_array[-num_end_flag:, :, :, :, :] = True
+            self.flag_array = np.reshape(self.flag_array, shape)
+
+
+
     def read_mwa_corr_fits(self, filelist, use_cotter_flags=False, correct_cable_len=False,
                            phase_to_pointing_center=False, run_check=True, check_extra=True,
-                           run_check_acceptability=True):
+                           run_check_acceptability=True, flag_init=False,
+                           edge_width=80e3, start_flag=4.0, end_flag=6.0,
+                           flag_dc_offset=True):
         """
         Read in MWA correlator gpu box files.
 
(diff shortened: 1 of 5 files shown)


mwilensky768:
@bhazelton @PyxieLouStar Have not put this through the testing battery yet - there are probably bugs abound. Just thought I would grab y'all's attention while developing this in case there are some new directions required.

pyuvdata/mwa_corr_fits.py
```diff
@@ -69,9 +69,82 @@ def correct_cable_length(self, cable_lens):
...
+            raise ValueError("The end_flag must be an integer multiple of the"
+                             "integration_time of the data or zero.")
+
+        num_ch_flag = int(edge_width / self.channel_width)
+        num_start_flag = int(start_flag / self.integration_time)
+        num_end_flag = int(end_flag / self.integration_time)
+
+        # Constant from the DSP architecture
+        num_coarse_chan = 24
+        num_fine_chan = len(self.Nfreqs) / num_coarse_chan
```
bhazelton: Need to be careful here, we can pass only some of the coarse band files to read_mwa_corr_fits and we still want this to work.

bhazelton: I'd suggest using the fact that we know the bandwidth of the coarse and fine channels.

mwilensky768: It turns out this number is computed during read_mwa_corr_fits using the header information in the box file, so I have opted to just pass that in from there in the most recent push.

pyuvdata/tests/test_mwa_corr_fits.py
```diff
@@ -294,3 +294,64 @@ def test_diff_obs():
...
+        mini6[3].header['TIME'] = mini6[1].header['TIME'] + 1
+        mini6.writeto(spoof_file6)
+
+    flag_testfiles = [spoof_file1, spoof_file6, filelist[0]]
+
+    uv = UVData()
+    uv.read(flag_testfiles, flag_init=True, start_flag=0, end_flag=0)
+    freq_inds = [0, 1, 4, 6, 7, 8, 9, 12, 14, 15]
+
+    assert np.all(uv.flag_array[:, :, freq_inds, :]), "Not all of edge and center channels are flagged!"
```
bhazelton: This ensures that the frequencies that should be flagged are flagged, but I don't see a test that ensures that frequencies that shouldn't be flagged are not flagged.

pyuvdata/tests/test_mwa_corr_fits.py
```diff
@@ -294,3 +294,64 @@ def test_diff_obs():
...
+
+    uv.read(flag_testfiles, flag_init=True, start_flag=1.0, end_flag=1.0,
+            edge_width=0, flag_dc_offset=False)
+    reshape = [uv.Ntimes, uv.Nbls, uv.Nspws, uv.Nfreqs, uv.Npols]
+    time_inds = [0, 1, -1, -2]
+    assert np.all(uv.flag_array.reshape(reshape)[time_inds, :, :, :, :]), "Not all of start and end times are flagged."
```
bhazelton: Similar comment to above -- we should check that only the expected times are flagged and others are not.

pyuvdata/mwa_corr_fits.py
```diff
@@ -69,9 +69,78 @@ def correct_cable_length(self, cable_lens):
...
+
+        if flag_dc_offset:
+            center_inds = list(range(num_fine_chan // 2, self.Nfreqs, num_fine_chan))
+
+            self.flag_array[:, :, center_inds, :] = True
+
+        if (num_start_flag > 0) or (num_end_flag > 0):
+            shape = self.flag_array.shape
+            reshape = [self.Ntimes, self.Nbls, self.Nspws, self.Nfreqs, self.Npols]
+            self.flag_array = np.reshape(self.flag_array, reshape)
```
bhazelton: This is really requiring a particular ordering of the baseline-time axis, which is not in general guaranteed for UVData objects. I guess it's ok if this is only ever called when the data are initially read in and that reading guarantees an ordering (need to check that with @PyxieLouStar). It can be written more generally by making a list of the time stamps, picking out the ones you want to flag and then finding their position in the `time_array`.

pyuvdata/mwa_corr_fits.py
```diff
@@ -69,9 +69,78 @@ def correct_cable_length(self, cable_lens):
...
+            if num_start_flag > 0:
+                self.flag_array[:num_start_flag, :, :, :, :] = True
+            if num_end_flag > 0:
+                self.flag_array[-num_end_flag:, :, :, :, :] = True
+            self.flag_array = np.reshape(self.flag_array, shape)
+
     def read_mwa_corr_fits(self, filelist, use_cotter_flags=False, correct_cable_len=False,
                            phase_to_pointing_center=False, run_check=True, check_extra=True,
-                           run_check_acceptability=True):
+                           run_check_acceptability=True, flag_init=False,
```
bhazelton: Do we really want the default to be False? I'd lean the other way but we should hear from @PyxieLouStar.

pyuvdata/uvdata.py
```diff
@@ -3817,6 +3826,27 @@ def read(self, filename, axis=None, file_type=None, allow_rephase=True,
...
+            Only used if flag_init is True. The number of seconds to flag at the
+            beginning of the observation. Set to 0 for no flagging. Only used if
+            file_type is 'mwa_corr_fits'.
+        end_flag: floats
+            Only used if flag_init is True. The number of seconds to flag at the
+            end of the observation. Set to 0 for no flagging. Only used if
+            file_type is 'mwa_corr_fits'.
+        flag_dc_offset: bool
+            Only used if flag_init is True. Set to True to flag the center fine
+            channel of each coarse channel. Only used if file_type is 'mwa_corr_fits'.
```
bhazelton: I'm not sure these descriptions are sufficient for non MWA users. Maybe adding something to the first sentence saying they're specific to the MWA? Or maybe changing the parameter name in the `read` method to make it clear they are MWA specific @mkolopanis @plaplant would love your input here.

pyuvdata/tests/test_mwa_corr_fits.py
mwilensky768: Unfortunately this is difficult because the two test box files being read in are not contiguous in time. This means that the data_array gets filled with zeros and the flags get filled with ones where there are not actually data.

pyuvdata/tests/test_mwa_corr_fits.py
mwilensky768: This section is also complicated for similar reasons as in the frequency flagging case. In this case, I tried to read only one box file in, so that there would not be discontiguous integrations but it seems there are still flags in the midsection, perhaps due to a bad antenna. Maybe @PyxieLouStar can chime in on the bad antenna hypothesis.

pyuvdata/mwa_corr_fits.py
mwilensky768: I believe this shape is required on read - conferred with Pyxie on this one. You can see it is also used on lines 565-567. I see you've tagged her so hopefully she'll chime in here as well.

pyuvdata/mwa_corr_fits.py
mwilensky768: _We_ might not want the default to be False but I'm not sure what other groups will prefer outside of EoR, etc.

pyuvdata/uvdata.py
mwilensky768: What if I switch the first and last sentences?

pyuvdata/tests/test_mwa_corr_fits.py
PyxieLouStar: Yes, there are bad antennas which get flagged.

pyuvdata/mwa_corr_fits.py
PyxieLouStar: I forced time as the primary axis and baseline as the secondary axis when the time and baseline axes get combined after reading in the data for no real reason except that ordering made more sense to me.

pyuvdata/mwa_corr_fits.py
PyxieLouStar: I think the fact that there are so many knobs to turn with this option is a good reason to make it opt-in.

pyuvdata/mwa_corr_fits.py
PyxieLouStar: Actually, I set it up that way because that's how a cotter file is organized when pyuvdata reads it in.

pyuvdata/mwa_corr_fits.py
```diff
@@ -69,9 +69,78 @@ def correct_cable_length(self, cable_lens):
...
+                           edge_width=80e3, start_flag=4.0, end_flag=6.0,
```
PyxieLouStar: Should this be start_flag=2.0, end_flag=2.0?

pyuvdata/tests/test_mwa_corr_fits.py
```diff
@@ -294,3 +294,64 @@ def test_diff_obs():
...
+    # spoof box files of the appropriate size
+    with fits.open(filelist[1]) as mini1:
+        mini1[1].data = np.repeat(mini1[1].data, 8, axis=0)
+        extra_dat = np.copy(mini1[1].data)
+        for app_ind in range(2):
+            mini1.append(fits.ImageHDU(extra_dat))
+        mini1[2].header['MILLITIM'] = 500
+        mini1[2].header['TIME'] = mini1[1].header['TIME'] + 0.5
+        mini1[3].header['MILLITIM'] = 1000
+        mini1[3].header['TIME'] = mini1[1].header['TIME'] + 1
```
PyxieLouStar: header['TIME'] should be an integer. header['MILLITIM'] should be 0 or 500. Removing lines 315-316 (and 327-328) will give the desired result.

pyuvdata/uvdata.py
mkolopanis: I was thinking yesterday it would be nice if we could partition the parameters sections by like `generic`, `uvfits`, `miriad`, etc specific parameter sections. Numpy docstyle does not seem to sepcifically allow that, or disallow that though. There is only `parameters` and `other parameters`. The latter reserved for "lesser used parameters". Does not seem to fit the need.

pyuvdata/tests/test_mwa_corr_fits.py
PyxieLouStar: Would it be sufficient to check that no other frequency is flagged in entirety?

bhazelton: yes, I think that would address my concern.

pyuvdata/mwa_corr_fits.py
bhazelton: ok. And the idea is that this function would only ever get called on read?

pyuvdata/mwa_corr_fits.py
bhazelton: ok. I can't really imagine that standard users would opt out of this flagging, but setting one option is not onerous.

pyuvdata/uvdata.py
bhazelton: I think the order of the parameters can be re-arranged and that might help. I'm ok with adding section headings too, we'd need to do some testing to make sure they render properly with sphinx.

pyuvdata/mwa_corr_fits.py
mwilensky768: This particular function would only get called on read, although now that I think about it, it might have been nice to write the function in utils so it could be called whenever. A complication that comes to mind in that scenario is that if one wants to call it after averaging then they will need to be careful with the flag_dc_offset keyword.

pyuvdata/mwa_corr_fits.py
mwilensky768: Yes, thanks!

pyuvdata/tests/test_mwa_corr_fits.py
mwilensky768: I think when I excluded 315-316 and 327-328 there was a problem on read when line 492 of mwa_corr_fits.py is executed. It seems like that line expects every hdu to have a TIME line in the header.

PyxieLouStar: Oh, I see. The lines should be `mini[2].header['TIME'] = mini[1].header['TIME']` and `mini[3].header['MILLITIM'] = 0` respectively. TIME advances by whole seconds every other hdu, and nonzero MILLITIM accounts for the half seconds in between.

bhazelton: Is this robust against the case where we average to 2s inside the correlator? We do that sometimes, I think 2014 data might be taken that way.

pyuvdata/mwa_corr_fits.py
mwilensky768: I think that since using the flag_array is itself optional within a pyuvdata object in the sense that one does not _have_ to apply the flag_array to the data during whatever processing, I will just let the default be True. I think that is more in line with how COTTER is deployed anyway.

pyuvdata/tests/test_mwa_corr_fits.py
mwilensky768: I've added a check that no other frequency is flagged in its entirety.

pyuvdata/tests/test_mwa_corr_fits.py
mwilensky768: Similar to the frequency axis, I've added a check to see that no intermediate time is flagged in its entirety.

pyuvdata/tests/test_mwa_corr_fits.py
mwilensky768: I removed ` + 0.5` on lines 315, 327 and replaced 1000 with 0 on lines 316, 328

pyuvdata/uvdata.py
mwilensky768: I have combined the first and last sentence, putting the last sentence first so that it is clear that the keyword is MWA-specific.

bhazelton:
This is looking very good. I think the only remaining items are updating the changelog and the tutorial.

PR #755, opened 2020-01-22
Add an extras_require section to setup.py

### LINKED ISSUE #675: make an `extras_require` section setup.py
To have a place to put all the optional dependencies so that people can install if needed.

### LINKED ISSUE #717: astropy-healpix dep
It seems like when installing `pyuvdata` with pip, `astropy-healpix` needs to be installed manually (at least, in `hera_sim` we need to, for tests). This should be fixed.

bhazelton:
## Description

Add an `extras_require` section to setup.py to make it easier for users to get all the required packages based on their use cases.

## Motivation and Context

closes #675 
closes #717

## Types of changes

- [ ] Version change

## Checklist:

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -1,7 +1,7 @@
 name: pyuvdata
 channels:
-  - conda-forge
   - defaults
+  - conda-forge
 dependencies:
   - astropy
   - astropy-healpix
--- a/environment.yaml
+++ b/environment.yaml
@@ -1,0 +1,22 @@
+name: pyuvdata
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - astropy
+  - astropy-healpix
+  - h5py
+  - flake8
+  - numpy>=1.15
+  - pypandoc
+  - python-casacore
+  - pyyaml
+  - scipy
+  - six
+  - sphinx
+  - coverage
+  - pytest-cov
+  - pip
+  - pip:
+      - flake8-pytest
+      - pytest-cases
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,14 @@ global_c_macros = [
     ('NPY_NO_DEPRECATED_API', 'NPY_1_7_API_VERSION'),
 ]
 
+casa_reqs = ['python-casacore']
+healpix_reqs = ['astropy_healpix']
+cst_reqs = ['pyyaml']
+test_reqs = (casa_reqs + healpix_reqs + cst_reqs
+             + ['pytest', 'pytest-cases', 'pytest-cov', 'coverage', 'flake8',
+                'flake8-pytest'])
+doc_reqs = ['sphinx', 'pypandoc']
+
 setup_args = {
     'name': 'pyuvdata',
     'author': 'Radio Astronomy Software Group',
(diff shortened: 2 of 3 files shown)


bhazelton:
A question for reviewers: Should anything be added to the readme about these options?

setup.py
```diff
@@ -101,7 +101,16 @@ def is_platform_mac():
...
-    'tests_require': ['pytest'],
+    'tests_require': ['pytest', 'pytest-cases'],
+    'extras_require': {
+        'casa': ['python-casacore'],
+        'healpix': ['astropy-healpix'],
+        'cst': ['pyyaml'],
+        'all': ['python-casacore', 'pyyaml', 'astropy-healpix'],
+        'dev': ['python-casacore', 'pyyaml', 'astropy-healpix', 'pytest',
+                'pytest-cases', 'pytest-cov', 'coverage', 'flake8',
+                'flake8-pytest', 'sphinx', 'pypandoc']
```
mkolopanis: The difference between `dev` and `all` is both test and docs related, do we want a `test` and a `doc` option?

setup.py
```diff
@@ -101,7 +101,16 @@ def is_platform_mac():
     'include_package_data': True,
     'setup_requires': ['pytest-runner', 'numpy>=1.15'],
     'install_requires': ['numpy>=1.15', 'six>=1.10', 'scipy', 'astropy>=2.0', 'h5py'],
-    'tests_require': ['pytest'],
+    'tests_require': ['pytest', 'pytest-cases'],
```
mkolopanis: Good addition, this probably should have been added long ago

setup.py
steven-murray: I think we could possibly refactor this a bit, so that we could easily just add lists here. Something like:

```
casa_rq = ['python-casacore']
healpix_rq = ['astropy-healpix']
cst_rq = ['pyyaml']
all = casa_rq + healpix_rq + cst_rq
test = ['pytest-cases', 'pytest-cov', 'coverage', 'flake8', 'flake8-pytest']
doc = ['sphinx', 'pypandoc']
dev = all + test + doc
```

bhazelton: sure, I can do that. I had thought about lists that we add but worried a little about making it less readable. But that may not be important.

bhazelton:
@steven-murray should `setuptools_scm` be in any of these lists? I ask because I see it in our conda environment for CI tests.

mkolopanis:
Oh I had to add that because python 2 couldn't install `pytest-cases` without it. It may, or may not, affect the result of this conversation but that's where it came from.

bhazelton:
ah ha! ok, thanks! I'm inclined to leave it out, if we're seriously worried about it we can delay this PR until the v2 PR is merged.

environment.yaml
```diff
@@ -0,0 +1,22 @@
+name: pyuvdata
+channels:
+  - conda-forge
```
steven-murray: Should conda-forge be the first channel here? Does that usurp priority? I would think it should be used as a second priority when needed. But maybe I'm wrong.

bhazelton: I copied it from the thing we run for CI. I think you're right, but we swapped it on purpose when we were having IERS issues and the default channel didn't give us the correct astropy version.

bhazelton:
if we merge this before the v2 branch, we need to take six out of the environment file in that PR.

environment.yaml
steven-murray: Ah OK. Cool.

steven-murray:
Shouldn't need `setuptools_scm` (yet).

bhazelton:
@steven-murray can you re-approve?

PR #763, opened 2020-02-11
Add sum of squares of weights to collapse output

mwilensky768:
## Description

This PR adds an output to `utils.collapse`, specifically `mean_collapse`, `absmean_collapse`, and `quadmean_collapse`, to return the sum of the squares of the weights on collapsing. 

## Motivation and Context

When the weights are not all equal to 1 or 0, this is necessary to standardize collapsed data and is not retrievable from what is returned already. SSINS, which subclasses UVFlag, will require this to make use of unevenly weighted data. I imagine other applications of UVFlag will as well.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

New feature checklist:

Breaking change checklist:  

Version change checklist:

### DIFF
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -1541,7 +1541,7 @@ def or_collapse(arr, weights=None, axis=None, return_weights=False,
         Whether to return dummy weights array.
         NOTE: the dummy weights will simply be an array of ones
     return_weights_square: bool
-        Whether to return the sum of the squares of the weights. Default is False.
+        NOT USED, but kept for symmetry with other collapsing functions.
     """
     if arr.dtype != np.bool:
         raise ValueError('Input to or_collapse function must be boolean array')
@@ -1571,7 +1571,7 @@ def and_collapse(arr, weights=None, axis=None, return_weights=False,
         Whether to return dummy weights array.
         NOTE: the dummy weights will simply be an array of ones
     return_weights_square: bool
-        Whether to return the sum of the squares of the weights. Default is False.
+        NOT USED, but kept for symmetry with other collapsing functions.
     """
     if arr.dtype != np.bool:
         raise ValueError('Input to and_collapse function must be boolean array')
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -1520,7 +1520,7 @@ def quadmean_collapse(arr, weights=None, axis=None, return_weights=False,
                         return_weights_square=return_weights_square)
     if return_weights and return_weights_square:
         return np.sqrt(out[0]), out[1], out[2]
-    elif return_weights:
+    elif return_weights or return_weights_square:
         return np.sqrt(out[0]), out[1]
     elif return_weights_square:
         return np.sqrt(out[0]), out[2]
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -1524,7 +1524,8 @@ def quadmean_collapse(arr, weights=None, axis=None, return_weights=False,
         return np.sqrt(out)
 
 
-def or_collapse(arr, weights=None, axis=None, return_weights=False):
+def or_collapse(arr, weights=None, axis=None, return_weights=False,
+                return_weights_square=False):
     """
     Collapse using OR operation.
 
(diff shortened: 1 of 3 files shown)


mwilensky768:
Currently I've just modified the functions in `utils` to return the item I want returned. I still need to change UVFlag's `waterfall` method to actually return this item. I'm wondering about whether it should be a default attribute of UVFlag objects (I think it has a range of uses), or if it should be optional. Thoughts appreciated.

mkolopanis:
I just want to make sure I understand. You're interested in adding a `weights_squared` parameter to `UVFlag`? If it is a useful quantity to carry around then I am totally for it. While I am the primary maintainer of `UVFlag` now I honestly am not too familiar with the uses for things like this. It would be great to have someone more hands on involved.

mwilensky768:
@mkolopanis yes something like that. I can try to post a couple lines of math justifying the need for such a quantity if desired. In English, a weighted sum of independent random variables (as implemented in `mean_collapse`, which I'm of the firm opinion is the correct way to do it) with identical variance will sum to a new random variable with variance proportional to the original variance, where the constant of proportionality is equal to the sum of the weights^2 divided by the square of the sum of the weights.

adampbeardsley:
So I think you want the plumbing to do that operation, but you don't necessarily need both weights and weights_squared around, right? I don't want to argue use cases -- if you really want both that's fine -- but I would think the new output weights being 1/var(output data) would be what you want.

adampbeardsley:
Also, for a little bit of context:

Returning the sum of the weights was chosen under the assumption that weights are typically 1/variance (for a minimum noise mean). Under that assumption, the resulting variance is 1/sum(weights), so if you weight by 1/variance again, the new weights are just the sum of the original weights.

Again, if you're doing something different, that's fine and we can figure out how to support it. Just want to be sure you're aware of what's there already.

mwilensky768:
I think keeping the sum of the squares extends code that is already in place to handle the case of general weights, which is a nice addition. I will spare you my thoughts on higher order cumulants (skewness etc.) since you do not want to argue about use cases.

adampbeardsley:
Ok, then I think my biggest concern is the naming. We previously used `return_weights` because, although it was the sum of the input weights, it was interpreted as the weights of the output.
But `return_weights_square` would seem to imply it's returning the square of the output weights.

Could we change them to be `return_weight_sum` and `return_weight_square_sum`? Is that too verbose for keywords? Would also require an API change. What does @mkolopanis think?

mwilensky768:
Does `return_square_weights` do any better? I feel like it's more confusing. If so, I think your suggestions are fine. API changes I guess can be icky for backwards compatibility. Not sure what is worse here.

adampbeardsley:
I think I was just overthinking this. Let's go with your original notation, and just make sure it's clear in docstrings (which I think it is).

Could you update the CHANGELOG? Once that's done, this should be good to go.

For UVFlag, I think it should be an optional parameter. I don't think most users will use it, and it would be confusing if you don't understand its uses.

mkolopanis:
This solution will work for me I think.

mwilensky768:
It seems like the process of adding an attribute to a UVFlag object, optional or not, is somewhat involved, but I think it is the right thing to do in this case. Is there a checklist of things to do when adding a parameter to a UVFlag object?

mwilensky768:
Ideally this would only be added as an attribute when the `type` was converted to `waterfall`

bhazelton:
This needs to be rebased on master to help tests pass.

bhazelton:
We're going to try to get this in and then add it as an optional parameter on UVFlag in a subsequent PR.

pyuvdata/utils.py
```diff
@@ -1496,16 +1512,24 @@ def quadmean_collapse(arr, weights=None, axis=None, return_weights=False):
...
-    if return_weights:
+    out = mean_collapse(np.abs(arr)**2, weights=weights, axis=axis,
+                        return_weights=return_weights,
+                        return_weights_square=return_weights_square)
+    if return_weights and return_weights_square:
+        return np.sqrt(out[0]), out[1], out[2]
+    elif return_weights or return_weights_square:
         return np.sqrt(out[0]), out[1]
+    elif return_weights_square:
+        return np.sqrt(out[0]), out[2]
```
bhazelton: It looks like this line isn't currently being covered by tests.

mwilensky768: Thanks! It looks like I managed to fix this.

PR #771, opened 2020-02-18
UVFlag strings

plaplant:
This PR makes sure that strings in UVFlag objects are handled properly when reading and writing files.

## Description

When writing strings to UVFlag objects on disk, strings were not being handled properly. This PR resolves that issue, as well as providing backwards compatibility for already-existing files (with deprecation warnings).

## Motivation and Context

As with UVH5 objects, strings should be cast to fixed-length ASCII strings using the numpy `np.string_` method (so they end up on-disk as `S<n>` datatypes, where `n` is the length of the string). Previously, strings were being written using the `.encode("utf8")` method, which are opaquely written as `np.object_`-type datasets. Unfortunately this approach does not provide the maximal amount of compatibility with the HDF5 implementations in other languages, which is a significant motivation for using the standard in the first place. (Cf. http://docs.h5py.org/en/stable/strings.html).

This PR also includes special handling of the `polarization_array` dataset, which is converted into a string when flags are collapsed along polarization. ~Additional testing must be done before merging this PR in to ensure this is still handled properly.~ Testing revealed the initial backwards compatibility handling was unnecessary because we're python3-only now, so only the writing has been changed significantly.

@mkolopanis @adampbeardsley I would love to have your feedback on how the polarization array is being handled, as well as thoughts on whether to re-write the test file. (Right now we get a lot of the aforementioned deprecation warnings and so I'd like to re-write the sample file, but I also didn't want to unilaterally change it without consulting you all first.)

## Types of changes

- [ ] Version change

## Checklist:

Breaking change checklist:  


pyuvdata/uvflag/uvflag.py
```diff
@@ -20,6 +20,40 @@
...
+    -----
+    This function is only designed to work on scalar datasets. Arrays of strings
+    should be handled differently.
+
+    """
+    if dataset.dtype.type is np.object_:
+        warnings.warn(
+            f"Strings in metadata of {filename} are not the correct type; "
+            "rewrite with `UVFlag.write` to ensure future compatibility. "
+            "Suppoort for reading these files will be removed in version 2.2.",
```
bhazelton: typo: Suppoort -> Support

mkolopanis: look at us using f strings. We python 3.6 now.

pyuvdata/uvflag/uvflag.py
```diff
@@ -1063,20 +1097,24 @@ def read(self, filename, history='', run_check=True, check_extra=True,
...
-                if isinstance(polarization_array[0], np.string_):
-                    polarization_array = np.asarray(polarization_array,
-                                                    dtype=np.str_)
+                polarization_array_type = header['polarization_array'][()].dtype.type
+                if polarization_array_type in (np.int, np.int32, np.int64):
+                    polarization_array = header['polarization_array'][()]
+                else:
+                    # we have a string-type polarization array from collapsing
+                    polarization_array = np.asarray(
+                        header['polarization_array'][()], dtype=str
```
adampbeardsley: sorry for my ignorance, but why does this not need the `_read_uvflag_string` call?

plaplant:
Thanks everyone for the input. I'll respond to each issue individually, but I wanted to make one general comment after doing some testing (more for the historical record). Now that we're python3-only, I don't think we need to be _as_ careful with string handling as in the past. We should still write strings to file by first casting using `np.string_`, but we no longer need special handling on reading (like we used to on UVH5 files) because even strings written the "old way" (i.e., using just `.encode("utf8")`) can be read back in without having to call the numpy `.tostring()` method before the `decode()`. (And as I discovered in my testing, there's actually even a weird corner case involving the empty string `''` that `.tostring()` messes up, but a straight `.decode()` handles just fine.) Writing a new UVFlag file with the branch as it is now can be read back in with current master, so evidently this isn't even a breaking change.

I think all this is to say that we _don't_ need to build in backwards compatibility/deprecation, and instead just make sure we write strings differently going forward. I'll go ahead and rewrite the test file to make sure it's conforming to the (still informal) spec, but that should be the biggest change that happens.

pyuvdata/uvflag/uvflag.py
plaplant: This one is different because it's saved as a length-1 array with `str` datatype, rather than a scalar string. I agree it's a little un-intuitive, but I think it's probably the best way to handle the fact that we still want it to be an array on the object.

Issue #773, opened 2020-02-20
Wrong number of blts during downsample


mkolopanis:
OS: ubuntu 18.04 
pyuvdata version: 2.0.1.dev26+gfbca23c4
python version: 3.8.1
Telescope: MWA phase 2 compact
obsids: `[1160576480,1160576848,1160576968,1160577096,1160577216,1160577336,1160577456,1160577584,1160577704,1160577824,1160578072,1160578192] `
when attempting to downsample in time a set of redundant baseline (baselines like (61, 58) see picture below) from 2.0s integration times to 4.0s integration times, the wrong number of blts is being returned by downsample in time. See attached traceback. 

```
uv.resample_in_time(4.0, only_downsample=True, keep_ragged=True)
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-42-ade2d3ae7c8f> in <module>
----> 1 uv.resample_in_time(4.0, only_downsample=True, keep_ragged=True)

~/src/anaconda/envs/py38/lib/python3.8/site-packages/pyuvdata/uvdata/uvdata.py in resample_in_time(self, target_time, only_downsample, only_upsample, blt_order, minor_order, keep_ragged, summing_correlator_mode, allow_drift)
   5655 
   5656         if downsample:
-> 5657             self.downsample_in_time(
   5658                 target_time,
   5659                 blt_order=blt_order,

~/src/anaconda/envs/py38/lib/python3.8/site-packages/pyuvdata/uvdata/uvdata.py in downsample_in_time(self, min_int_time, blt_order, minor_order, keep_ragged, summing_correlator_mode, allow_drift)
   5566 
   5567         # make sure we've populated the right number of baseline-times
-> 5568         assert temp_idx == temp_Nblts, ("Wrong number of baselines. Got {:d}, "
   5569                                         "expected {:d}. This is a bug, please "
   5570                                         "make an issue at https://github.com/"

AssertionError: Wrong number of baselines. Got 12218, expected 13776. This is a bug, please make an issue at https://github.com/RadioAstronomySoftwareGroup/pyuvdata/issues
```
![image](https://user-images.githubusercontent.com/8646829/74979293-975b8980-53eb-11ea-86c6-0b6475063aa2.png)

bhazelton:
That's exciting. I recall now writing that error message, but I don't recall the context. We probably need to look at this as a group.

For context, had you previously downsampled to the redundant set?

[...]

PR #785, opened 2020-03-11
Resample time factor

bhazelton:
## Description

Add support for averaging by an integer factor in time.

## Motivation and Context

Based on discussion in issue #773, this may behave more closely to human expectations than setting a minimum integration time.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/uvdata/tests/test_uvdata.py
+++ b/pyuvdata/uvdata/tests/test_uvdata.py
@@ -5299,6 +5299,7 @@ def test_downsample_in_time(resample_in_time_file):
     uv_object.phase_to_time(Time(uv_object.time_array[0], format="jd"))
     # reorder to make sure we get the right value later
     uv_object.reorder_blts(order="baseline", minor_order="time")
+    uv_object2 = uv_object.copy()
 
     # save some values for later
     init_data_size = uv_object.data_array.size
@@ -5311,7 +5312,7 @@ def test_downsample_in_time(resample_in_time_file):
     # change the target integration time
     min_integration_time = original_int_time * 2.0
     uv_object.downsample_in_time(
-        min_integration_time, blt_order="baseline", minor_order="time"
+        min_int_time=min_integration_time, blt_order="baseline", minor_order="time"
     )
 
     # Should have half the size of the data array and all the new integration time
@@ -5327,7 +5328,13 @@ def test_downsample_in_time(resample_in_time_file):
     out_ns = uv_object.get_nsamples(0, 1)
     assert np.isclose((init_ns[0, 0, 0] + init_ns[1, 0, 0]) / 2.0, out_ns[0, 0, 0])
 
-    return
+    # Compare doing it with n_times_to_avg
+    uv_object2.downsample_in_time(
+        n_times_to_avg=2, blt_order="baseline", minor_order="time"
+    )
+    assert uv_object.history != uv_object2.history
+    uv_object2.history = uv_object.history
+    assert uv_object == uv_object2
 
 
 @pytest.mark.filterwarnings("ignore:The xyz array in ENU_from_ECEF")
@@ -5338,7 +5345,6 @@ def test_downsample_in_time_partial_flags(resample_in_time_file):
     uv_object.phase_to_time(Time(uv_object.time_array[0], format="jd"))
     # reorder to make sure we get the right value later
     uv_object.reorder_blts(order="baseline", minor_order="time")
-
     # save some values for later
     init_wf = uv_object.get_data(0, 1)
     original_int_time = np.amax(uv_object.integration_time)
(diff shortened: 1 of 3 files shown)


pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -5327,7 +5327,13 @@ def test_downsample_in_time(resample_in_time_file):
     out_ns = uv_object.get_nsamples(0, 1)
     assert np.isclose((init_ns[0, 0, 0] + init_ns[1, 0, 0]) / 2.0, out_ns[0, 0, 0])
 
-    return
+    # Compare doing it with n_times_to_avg
+    uv_object2.downsample_in_time(
+        n_times_to_avg=2, blt_order="baseline", minor_order="time"
+    )
+    assert uv_object.history != uv_object2.history
```
plaplant: Why do the histories end up being different here? (Btw, I think this is a good rigorous check for making the histories equal.)

pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -5446,23 +5469,34 @@ def test_downsample_in_time_uneven_samples(resample_in_time_file):
...
+    uv_object2.history = uv_object.history
+    assert uv_object == uv_object2
 
 
 @pytest.mark.filterwarnings("ignore:The xyz array in ENU_from_ECEF")
 @pytest.mark.filterwarnings("ignore:The enu array in ECEF_from_ENU")
-def test_downsample_in_time_uneven_samples_discard_ragged(resample_in_time_file):
+def test_downsample_in_time_uneven_samples_keep_ragged(resample_in_time_file):
```
plaplant: Why did we change from keeping ragged to discarding them? Also, the test's docstring should be updated.

pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -5994,9 +6070,9 @@ def test_downsample_in_time_varying_integration_time(resample_in_time_file):
     uv_object.integration_time[inds01[-2:]] += initial_int_time
     uv_object.Ntimes = np.unique(uv_object.time_array).size
     min_integration_time = 2 * np.amin(uv_object.integration_time)
-    uvtest.checkWarnings(
-        uv_object.downsample_in_time, [min_integration_time], nwarnings=0
-    )
+    with pytest.warns(None) as record:
```
plaplant: Why do we explicitly check that no warnings are raised here? (I approve of the move to using `pytest.warns`, but just curious why we're doing this.)

pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -6400,10 +6476,17 @@ def test_resample_in_time_only_upsample(bda_test_file):
 
 @pytest.mark.filterwarnings("ignore:There is a gap in the times of baseline")
 def test_downsample_in_time_mwa():
-    """Test resample in time works with MWA data."""
+    """
+    Test resample in time works with numerical weirdnesses.
```
plaplant: Great docstring!

pyuvdata/uvdata/uvdata.py
```diff
@@ -4718,128 +4718,165 @@ def upsample_in_time(
...
-            & (
-                ~np.isclose(
-                    self.integration_time,
-                    min_int_time,
-                    rtol=self._integration_time.tols[0],
-                    atol=self._integration_time.tols[1],
+        if min_int_time is not None and n_times_to_avg is not None:
+            raise ValueError("Only one of min_int_time or n_times_to_avg can be set.")
+
+        if self.Ntimes == 1:
```
plaplant: Good catch 👍

pyuvdata/uvdata/uvdata.py
```diff
@@ -5026,10 +5069,16 @@ def downsample_in_time(
         self.check()
 
         # add to the history
-        history_update_string = (
-            " Downsampled data to {:f} second integration "
-            "time using pyuvdata.".format(min_int_time)
-        )
+        if min_int_time is not None:
+            history_update_string = (
+                " Downsampled data to {:f} second integration "
```
plaplant: Aha! That explains the different history messages above 😄

pyuvdata/uvdata/tests/test_uvdata.py
bhazelton: I see that you figured this out. Do you want me to add a comment?

pyuvdata/uvdata/tests/test_uvdata.py
bhazelton: For coverage. The test just above this one tests nearly the same thing with `keep_ragged=False`, so I switched this one to `True` to improve the coverage without increasing the number of tests. I'll fix the docstring.

pyuvdata/uvdata/tests/test_uvdata.py
bhazelton: I'm not perfectly sure, but I didn't want to remove an existing check. I think maybe it was because we wanted to make sure that we were doing things consistently when we changed the `integration time` and `time_array` (a warning is issued when there's inconsistencies between those 2 arrays).

bhazelton: I can take it out if you think we don't need it, or I can add a comment.

pyuvdata/uvdata/tests/test_uvdata.py
plaplant: Yeah, a quick comment might be helpful to those of us who form opinions before reading code :)

pyuvdata/uvdata/tests/test_uvdata.py
plaplant: Fair enough. Yes, I think a comment explaining we don't want to raise that warning would help.

Issue #783, opened 2020-03-10
One corrupted file breaks entire read


dstorer:
When reading in a whole list of files (specifically I've been using read_uvh5), if any one of the files is corrupted the entire read breaks. I propose adding an optional (defaulted off) try-catch block to catch potentially problematic files so that the rest of the read may continue. I think it would also make sense to print out a list of any files that had to be skipped. 

I will draft a PR that does this.

bhazelton:
was fixed in 788

PR #788, opened 2020-03-13
Test file on read

dstorer:
## Description

Add an option in the UVData.read() function to, when reading in a list of files, check each file before reading. This way, if one or more files are corrupted, the read call may proceed without crashing, and will instead provide a warning for each file that failed.

## Motivation and Context
This is in response to issue #783. For HERA we have run into problems with files occasionally missing metadata in their headers, and this causing our code to crash when trying to read in a whole night of data in one call. This fix will allow us to proceed without our code crashing, and instead get a warning that there was a corrupted file that was skipped.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/uvdata/tests/test_uvdata.py
+++ b/pyuvdata/uvdata/tests/test_uvdata.py
@@ -6810,27 +6810,37 @@ def test_multifile_read_check():
 
     # Test that the expected error arises
     with pytest.raises(KeyError) as cm:
-        uv.read(testfile, skip_bad_files=True)
+        uv.read(testfile, skip_bad_files=False)
     assert "Unable to open object (object 'ant_1_array' doesn't exist)" in str(cm.value)
 
-    # Test when the corrupted file is at the end
-    uvTrue = UVData()
-    uvTrue.read(uvh5_file)
-    fileList = [uvh5_file, testfile]
-    with pytest.warns(UserWarning) as cm:
-        uv.read(fileList, skip_bad_files=True)
-    # Check that a warning was issued
-    assert len(cm) == 1
-    # Check that the uncorrupted file was still read in
-    assert uv == uvTrue
-
-    # Test when the corrupted file is at the beggining
+    # Test when the corrupted file is at the beggining, skip_bad_files=False
+    uv = UVData()
     uvTrue = UVData()
     uvTrue.read(uvh5_file)
     fileList = [testfile, uvh5_file]
-    with pytest.warns(UserWarning) as cm:
+    with pytest.raises(KeyError) as cm:
+        with pytest.warns(UserWarning, match="Failed to read"):
+            uv.read(fileList, skip_bad_files=False)
+    assert "Unable to open object (object 'ant_1_array' doesn't exist)" in str(cm.value)
+    assert uv != uvTrue
+
+    # Test when the corrupted file is at the beggining, skip_bad_files=True
+    uv = UVData()
+    uvTrue = UVData()
+    uvTrue.read(uvh5_file)
+    fileList = [testfile, uvh5_file]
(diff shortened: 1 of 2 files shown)


bhazelton:
@dstorer Rather than having two copies of the `read` call, let's always use the try/catch block but if your keyword is False then re-raise the exception in the catch using `raise`.

I'm also not sure about the keyword name, maybe something more like `skip_bad_files`?

And what is the utility of the `file_num` variable?

pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -6792,3 +6793,113 @@ def test_multifile_read_errors(read_func, filelist):
...
+    # Test that the expected error arises
+    with pytest.raises(KeyError) as cm:
+        uv.read(testfile, check_file_status=True)
+    assert "Unable to open object (object 'ant_1_array' doesn't exist)" in str(cm.value)
+
+    # Test when the corrupted file is at the end
+    uvTrue = UVData()
+    uvTrue.read(uvh5_file)
+    fileList = [uvh5_file, testfile]
+    with pytest.warns(UserWarning) as cm:
```
bhazelton: If you add `match="your expected error message"` to the `pytest.warns` call it will check that the correct warning message was issued.

dstorer:
@bhazelton sure, I can make those first two changes easily. I included the `file_num` bit because it needs to read the first file to set the phase center before proceeding to the for loop where it reads the remaining files, so if the first (or first several) files wind up being skipped, it needs to know where to start the for loop. There may be a more clever way of dealing with that though. I considered just deleting files from the list if we skipped over them, but I'm not sure if that would create problems elsewhere.

pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -6792,3 +6793,123 @@ def test_multifile_read_errors(read_func, filelist):
...
+    testfile = os.path.join(DATA_PATH, "test", "zen.2458661.23480.HH.uvh5")
+    uv.read(uvh5_file)
+    uv.write_uvh5(testfile)
+    with h5py.File(testfile, "r+") as h5f:
+        del h5f["Header/ant_1_array"]
+
+    # Test that the expected error arises
+    with pytest.raises(KeyError) as cm:
+        uv.read(testfile, skip_bad_files=False)
+    assert "Unable to open object (object 'ant_1_array' doesn't exist)" in str(cm.value)
```
bhazelton: you don't need to change this, but FYI `pytest.raises` also can take the `match` parameter in the same way that `pytest.warns` does, so you don't have to write a separate assert.

pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -6792,3 +6793,123 @@ def test_multifile_read_errors(read_func, filelist):
...
+    assert uv == uvTrue
+
+    os.remove(testfile)
+
+    return
+
+
+def test_multifile_read_check_long_list():
+    """
+    Test setting check_file_status=True when reading in files for a list of length >2
```
bhazelton: check_file_status  -> skip_bad_files

pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -6792,3 +6793,123 @@ def test_multifile_read_errors(read_func, filelist):
         "Reading multiple files from class specific read functions is no "
         "longer supported."
     )
+
+
+def test_multifile_read_check():
+    """Test setting check_file_status=True when reading in files"""
```
bhazelton: check_file_status -> skip_bad_files

bhazelton:
I hadn't realized what `file_num` was doing. I think what you're doing is fine.

PR #790, opened 2020-03-18
Allow UVParameters to be equal if they contain NaNs in arrays in the same locations.

bhazelton:
## Description

Set equal_nan=True in all np.allclose tests in `UVParameter.__eq__` method.

## Motivation and Context

A problem came up in pyradiosky where two arrays were equal everywhere that they were not NaN and had NaNs in the same places, but our equality check failed because we didn't have the `equal_nan` parameter on `np.allclose` set to True. This fixes that and adds a test that fails on master.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 - `frequency_average` method on UVData to average data along the frequency axis.
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


bhazelton:
@plaplant just rebased this, can you re-approve?

PR #798, opened 2020-03-25
Add uvcalibrate checks

### LINKED ISSUE #684: uvcalibrate function doesn't test that times (or other parameters) match
We should be checking that the uvcal and uvdata objects are compatible.

bhazelton:
## Description

Add a number of checks to the `uvcalibrate` function in pyuvdata.utils to make sure the UVCal and UVData objects match along a number of axes. Include backwards compatibility via deprecation warnings.

## Motivation and Context

closes #684

## Types of changes

- [ ] Version change

## Checklist:

Breaking change checklist:


pyuvdata/utils.py
```diff
@@ -1518,27 +1529,258 @@ def uvcalibrate(
     undo : bool, optional
         If True, undo the provided calibration. i.e. apply the calibration with
         flipped gain_convention. Flag propagation rules apply the same.
+    override_ant_check : bool
```
nkern, line 1532: [minor] flipped signature -- docstring order here

pyuvdata/utils.py
```diff
@@ -1518,27 +1529,258 @@ def uvcalibrate(
...
 
     """
-    # deepcopy for not inplace
+    # copy for not inplace
     if not inplace:
-        uvdata = copy.deepcopy(uvdata)
+        uvdata = uvdata.copy()
+
+    uvdata_unique_nums = np.unique(np.append(uvdata.ant_1_array, uvdata.ant_2_array))
+    uvdata.antenna_names = np.array(uvdata.antenna_names)
```
nkern, line 1552: np.asarray could be better here, as it won't copy the data if it is already an array.

pyuvdata/utils.py
```diff
@@ -1518,27 +1529,258 @@ def uvcalibrate(
...
     UVData, optional
         Returns if not inplace
 
     """
-    # deepcopy for not inplace
+    # copy for not inplace
     if not inplace:
-        uvdata = copy.deepcopy(uvdata)
+        uvdata = uvdata.copy()
+
```
nkern, line 1550: could you add a comment here notifying that the next block of code is for antenna number checking

pyuvdata/utils.py
```diff
@@ -1490,6 +1499,8 @@ def uvcalibrate(
     Dterm_cal=False,
     delay_convention="minus",
     undo=False,
+    override_time_check=False,
```
nkern, line 1502: does it make more sense to just make this `time_check = True` and `ant_check = True`? I think this may be more consistent with extra checks in uvdata, but I may be wrong.

pyuvdata/utils.py
nkern: Ah, I see now, you are trying to get all antenna names **used in the data** not necessarily in the metadata. can you make a comment to make that a little clearer?

pyuvdata/utils.py
```diff
@@ -1518,27 +1529,258 @@ def uvcalibrate(
...
+                            "with calibration and flag the data for this antenna. "
+                            "Currently calibration will proceed, the data will not "
+                            "be flagged and not be calibrated. This behavior will "
+                            "deprected in version 2.2."
+                        )
+                    warnings.warn(warn_str, DeprecationWarning)
+
+    uvdata_times = np.unique(uvdata.time_array)
+    downselect_cal_times = False
+    if uvcal.Ntimes > 1:
```
nkern, line 1621: I'm wondering if it is also useful to make this `if uvcal.Ntimes > 1 and not override_time_check:`
the benefit is that it will entirely skip this next block if one wants to override the time check. This is nice b/c in my experience, some of these metadata checks can become take quite a while (the time check on uvdata for example was actually a non-negligible bottleneck in loading night-long uvdata waterfalls).

nkern: Same for the above ant_check, although not sure if thats as much of a memory / wall-time concern

pyuvdata/utils.py
```diff
@@ -1518,27 +1529,258 @@ def uvcalibrate(
...
+        freq_arr_match = np.allclose(
+            np.sort(uvcal.freq_array[0, :]),
+            np.sort(uvdata.freq_array[0, :]),
+            atol=uvdata._freq_array.tols[1],
+            rtol=uvdata._freq_array.tols[0],
+        )
+    except ValueError:
+        freq_arr_match = False
+
+    if freq_arr_match is False:
```
nkern, line 1706: for similar reasons, it might be nice to also have an `override_freq_check` parameter

pyuvdata/utils.py
bhazelton: No. We only allow a time override in the case that you have a single solution (Ntimes = 1) that you want to apply to a different time. Otherwise you have an array of solutions for each time integration and it doesn't really make sense to apply that to another unrelated data set -- first, it might have a different number of time integrations, second why would you want exactly the same sequence of solutions applied to an unrelated file? So the `override_time_check` is ignored if Ntimes > 1.

pyuvdata/utils.py
bhazelton: We were unaware of a case when you would want to do such a thing -- bandpass calibration is usually quite frequency specific, whereas transferring cals between different times is something we've all seen before.

pyuvdata/utils.py
nkern: I see, but my concern isn't so much about the physical reason and more about the computation required to do the check. Like I said with UVData, doing time checks across a nightly waterfall turned out to be a sizeable fraction of the load time, and it was helpful to be able to skip it if desired. 

Seems like there may also be good reason to be able to manually skip the checks (as is done in UVData to my understanding) in case one is doing something experimental. Of course they would be by default True so most anyone would be running them anyways.

bhazelton: Ok, we'll talk about it on the telecon tomorrow. You're welcome to join you like!

bhazelton: The feeling on the telecon is that you only end up in the slow part of the check (starting on line 1633 below) in very strange edge cases. If your UVCal times are the same as your UVData times then it's just an `np.allclose` call which is fast. If you have a single UVCal time or just a UVCal range you don't enter this slow part of the check.

pyuvdata/utils.py
bhazelton: The feeling on the telecon is that in very experimental modes like this the user can do the select on the objects before hand or even overwrite the frequency array on one of the objects if that's really what they want. The slow part of the check doesn't come up at all if the freq_array is matched between the UVCal and UVData objects.

pyuvdata/utils.py
nkern: I see, that makes sense to me

pyuvdata/utils.py
nkern: okay yup I'm fine w/ that too if the slow parts are only entered if the frequencies aren't matched

pyuvdata/utils.py
bhazelton: oops, thanks!

pyuvdata/utils.py
bhazelton: absolutely

pyuvdata/utils.py
bhazelton: hmm, that might make more sense, let's let @adampbeardsley weigh in.

adampbeardsley: Honestly I don't think it makes a big difference. I'd slightly lean toward Nick's suggestion just because it simplifies variable names and feels less like a double negative.

bhazelton: ok.

pyuvdata/utils.py
```diff
@@ -1518,27 +1529,260 @@ def uvcalibrate(
...
+                        f"Antenna {this_ant_name} has data on UVData but not on UVCal. "
+                        "override_ant_check is True, so the data associated with "
+                        "this antenna will be flagged."
+                    )
+                elif uvdata_ant_num in uvcal_unique_nums:
+                    warnings.warn(
+                        f"Antenna {this_ant_name} has data on UVData but not on UVCal. "
+                        f"Its antenna number is {uvdata_ant_num}, which is present "
+                        "on UVCal. Currently the data will be calibrated using the "
+                        "matching antenna number, but that will be deprecated in "
```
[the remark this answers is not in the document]
bhazelton: hmm, I take your point. The problem is that there are a lot of existing files that this will be broken for (because until very recently hera_cal was overwriting the antenna names). What if I change this to check all antennas and only use the number if none of the antenna names match?

pyuvdata/utils.py
```diff
@@ -1518,27 +1529,260 @@ def uvcalibrate(
     undo : bool, optional
         If True, undo the provided calibration. i.e. apply the calibration with
         flipped gain_convention. Flag propagation rules apply the same.
+    override_time_check : bool
+        Option to override the check that times match between the UVCal and UVData
+        objects. Only used in UVCal has only one time.
+    override_ant_check : bool
```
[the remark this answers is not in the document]
bhazelton: flag_missing is going away (it's deprecated). I updated the docstring to make that clear.

The behavior of `flag_missing=True` is the same as setting `ant_check=False`. The reason I want to get rid of flag_missing is that if `flag_missing=False` then the baseline is neither calibrated nor flagged, so you can end up with a mix of calibrated and uncalibrated unflagged baselines.

pyuvdata/utils.py
```diff
@@ -1518,27 +1529,260 @@ def uvcalibrate(
...
+                    )
+                    if flag_missing:
+                        warn_str += (
+                            " and will be flagged because flag_missing is True. "
+                            "The flag_missing keyword will be deprected in "
+                            "version 2.2, use override_ant_check instead."
+                        )
+                    else:
+                        warn_str += (
+                            ", set override_ant_check=True to proceed "
```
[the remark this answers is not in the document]
bhazelton: yes, it's a backward compatibility handling but I don't want to suggest using a deprecated keyword.

pyuvdata/utils.py
```diff
@@ -1518,27 +1529,260 @@ def uvcalibrate(
...
+                        atol=uvdata._time_array.tols[1],
+                        rtol=uvdata._time_array.tols[0],
+                    )
+                )
+                if wh_time_match[0].size > 0:
+                    uvcal_times_to_keep.append(uvcal.time_array[wh_time_match][0])
+                else:
+                    warnings.warn(
+                        f"Time {this_time} exists on UVData but not on UVCal. "
+                        "This will become an error in version 2.2",
```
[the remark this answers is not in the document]
bhazelton: nope.

pyuvdata/utils.py
```diff
@@ -1518,27 +1529,260 @@ def uvcalibrate(
...
+                    )
+            if len(uvcal_times_to_keep) < uvcal.Ntimes:
+                downselect_cal_times = True
+    elif uvcal.time_range is None:
+        if uvdata.Ntimes > 1 or not np.isclose(
+            uvdata_times,
+            uvcal.time_array,
+            atol=uvdata._time_array.tols[1],
+            rtol=uvdata._time_array.tols[0],
+        ):
```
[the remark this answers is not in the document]
bhazelton: So this block is only entered in uvcal.Ntimes = 1 and uvcal.time_range is None. In that case, if UVData has multiple times the times cannot match. If it only has one time, you only need an isclose (rather than an allclose) because it's only testing one time. And you don't need it in a try/except because it won't have an exception (the exception happens with the allclose if the uvcal time array is shorter than the update time array I think). I added some inline comments to try to clarify, see if they're enough.

pyuvdata/utils.py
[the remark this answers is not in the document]
adampbeardsley: To be sure I understand the proposed change:
- If no names match, use numbers
- If some names match, flag those that don't and use names for those that do.
- If all names match, use names

This sounds better to me. At least what is returned is self-consistent.

pyuvdata/utils.py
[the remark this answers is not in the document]
adampbeardsley: Ok, but I think we still need some more explanation here. After playing a little bit myself, it looks like if the number of times are equal, we line them up even though the times don't match and apply. But if UVCal has fewer times than UVData, you get a `ValueError: operands could not be broadcast together...`.  I think we need to explain that in the former case you're applying calibrations to unmatching times, and in the latter case we could probably throw a more informative error - maybe check the length of `uvcal_times_to_keep` and if it's smaller than `uvdata_times`, throw an error around line 1658?

pyuvdata/utils.py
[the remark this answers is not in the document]
bhazelton: ok, I made this change.

pyuvdata/utils.py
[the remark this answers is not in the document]
bhazelton: So I think this is handled down at line 1893 -- the downselect is not done if the select would result in a UVCal object that has fewer times than a UVData object. These lines are strictly for backwards compatibility, they should be removed in version 2.2. I guess you could still get an error if you passed in a UVCal object that even from the beginning had fewer times than the UVData object. I can add an error for that case.

bhazelton:
@adampbeardsley I think I've addressed your comments on this PR.

bhazelton:
I think this probably needs a changelog entry, which I will add. I don't think it needs a tutorial update.

PR #802, opened 2020-04-06
use utils.conj_pol when conjugating in select

### LINKED ISSUE #796: Should these lines up the CONJ_POL_DICT in utils.py
The referenced lines reverse (conjugate I think is the desire) polarizations during a `select` function execution with a length 3 bl tuple is given (ant1, ant2, pol) but the baseline is present as (ant2, ant2, pol_conj) in the uvdata object. Reversing the polarization will work for linear and circularly polarized data but this line would break for (pseudo-) Stokes I think?

https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/89d7796bb1acee786727198aa1eef1d66df103c2/pyuvdata/uvdata/uvdata.py#L4063-L4064

mkolopanis:
closes #796
## Description

removes lazy string reversing for calling `utils.conj_pol()` when performing select on uvdata, uvlfag, and miriad reads.
## Motivation and Context

The original code for this lazily assumed all polarizations were linear or circular but did not consider (pseudo) Stokes parameters.
## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/uvdata/miriad.py
+++ b/pyuvdata/uvdata/miriad.py
@@ -229,7 +229,7 @@ class Miriad(UVData):
                     else:
                         bl_str_list.append(str(bl[1]) + "_" + str(bl[0]))
                         if len(bl) == 3:
-                            bl_pols.add(bl[2][::-1])
+                            bl_pols.add(uvutils.conj_pol(bl[2].lower()))
 
                 if n_selects > 0:
                     # combine antpair_str_list and bl_str_list with an intersection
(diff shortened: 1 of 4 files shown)


pyuvdata/uvdata/miriad.py
```diff
@@ -229,7 +229,7 @@ def read_miriad(
                     else:
                         bl_str_list.append(str(bl[1]) + "_" + str(bl[0]))
                         if len(bl) == 3:
-                            bl_pols.add(bl[2][::-1])
+                            bl_pols.add(uvutils.conj_pol(bl[2].lower()))
```
adampbeardsley, line 232: This `lower()` is redundant with the `conj_pol` function. And should probably remove to be consistent with the other instances.

mkolopanis:
@bhazelton do you want a changelog update for this?

bhazelton:
I think so? it is a bug fix.

PR #805, opened 2020-04-15
added fast H5PY slice indexing for UVH5.read_uvh5()

nkern:
Added smarter h5py dataset slicing upon UVH5 read call.

## Description

 In general, if one is loading a subset of the blts axis it defaults to a fancy index, which is very slow for HDF5 (see [this stack overflow post](https://stackoverflow.com/questions/21766145/h5py-correct-way-to-slice-array-datasets)). This can be sped up if we are able to slice the datasets when possible. 

Furthermore, default behavior was to slice only one data axis at a time. This PR lets us slice as many axes simultaneously as possible, so we don't need to load in more data than what is requested.

This also exposes the `chunks` kwarg to `UVH5.write_uvh5` and `UVData.write_uvh5` for writing HDF5 files with custom chunking, which can significantly accelerate or degrade I/O performance depending on how the data are typically being read.

## Motivation and Context

Very slow read times when working with Nblts > 20k elements files but trying to select only a subset of the antennas.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

New feature checklist:

Breaking change checklist:

Version change checklist:

### DIFF
--- a/pyuvdata/uvdata/uvh5.py
+++ b/pyuvdata/uvdata/uvh5.py
@@ -227,13 +227,13 @@ def _get_slice_len(s, axlen):
     if s.stop is None:
         stop = axlen
     else:
-        stop = s.stop
+        stop = np.min([s.stop, axlen])
     if s.step is None:
         step = 1
     else:
         step = s.step
 
-    return (stop - start) // step
+    return (stop - 1 - start) // step + 1
 
 
 def _get_dset_shape(dset, indices):
@@ -293,11 +293,12 @@ def _index_dset(dset, indices):
     ndarray
         The indexed dset
     """
-    # get dset shape
-    dset_shape = _get_dset_shape(dset, indices)
+    # get dset and arr shape
+    dset_shape = dset.shape
+    arr_shape = _get_dset_shape(dset, indices)
 
     # create empty array of dset dtype
-    arr = np.empty(dset_shape, dtype=dset.dtype)
+    arr = np.empty(arr_shape, dtype=dset.dtype)
 
     # get arr and dest indices for each dimension in indices
     dset_indices = []
@@ -321,8 +322,8 @@ def _index_dset(dset, indices):
             elif isinstance(dset_inds[0], slice):
                 # this is a list of slices, need list of slice lens
                 slens = [_get_slice_len(s, dset_shape[i]) for s in dset_inds]
-                ssums = [sum(slens[:i]) for i in range(len(slens))]
-                arr_inds = [slice(ssums[i], ssums[i] + slens[i]) for i in range(len(slens))]
+                ssums = [sum(slens[:j]) for j in range(len(slens))]
+                arr_inds = [slice(ssums[j], ssums[j] + slens[j]) for j in range(len(slens))]
                 arr_indices.append(arr_inds)
                 dset_indices.append(dset_inds)
 
--- a/pyuvdata/uvdata/uvh5.py
+++ b/pyuvdata/uvdata/uvh5.py
@@ -333,8 +333,9 @@ def _index_dset(dset, indices):
             for freq_arr, freq_dset in zip(arr_indices[2], dset_indices[2]):
                 for pol_arr, pol_dset in zip(arr_indices[3], dset_indices[3]):
                     # index dset and assign to arr
-                    arr[blt_arr, spw_arr, freq_arr, pol_arr] = \
-                        dset[blt_dset, spw_dset, freq_dset, pol_dset]
+                    arr[blt_arr, spw_arr, freq_arr, pol_arr] = dset[
+                        blt_dset, spw_dset, freq_dset, pol_dset
+                    ]
 
     return arr
 
@@ -546,6 +547,7 @@ class UVH5(UVData):
         run_check_acceptability,
         data_array_dtype,
         keep_all_metadata,
+        multidim_slice,
     ):
         """
         Read the data-size arrays (data, flags, nsamples) from a file.
@@ -620,15 +622,15 @@ class UVH5(UVData):
 
         if min_frac == 1:
             # no select, read in all the data
+            inds = (np.s_[:], np.s_[:], np.s_[:], np.s_[:])
             if custom_dtype:
-                inds = (np.s_[:], np.s_[:], np.s_[:], np.s_[:])
                 self.data_array = _read_complex_astype(
                     dgrp["visdata"], inds, data_array_dtype
                 )
             else:
-                self.data_array = dgrp["visdata"][:, :, :, :]
-            self.flag_array = dgrp["flags"][:, :, :, :]
-            self.nsample_array = dgrp["nsamples"][:, :, :, :]
+                self.data_array = _index_dset(dgrp["visdata"], inds)
+            self.flag_array = _index_dset(dgrp["flags"], inds)
+            self.nsample_array = _index_dset(dgrp["nsamples"], inds)
         else:
             # do select operations on everything except data_array, flag_array
             # and nsample_array
@@ -639,8 +641,9 @@ class UVH5(UVData):
             # determine which axes can be sliced, rather than fancy indexed
             # max_nslice_frac of 0.1 yields slice speedup over fancy index for HERA data
             if blt_inds is not None:
-                blt_slices, blt_sliceable = _convert_to_slices(blt_inds,
-                                                               max_nslice_frac=0.1)
+                blt_slices, blt_sliceable = _convert_to_slices(
+                    blt_inds, max_nslice_frac=0.1
+                )
                 if blt_sliceable:
                     blt_inds = blt_slices
             else:
@@ -648,8 +651,9 @@ class UVH5(UVData):
                 blt_sliceable = True
 
             if freq_inds is not None:
-                freq_slices, freq_sliceable = _convert_to_slices(freq_inds,
-                                                                 max_nslice_frac=0.1)
+                freq_slices, freq_sliceable = _convert_to_slices(
+                    freq_inds, max_nslice_frac=0.1
+                )
                 if freq_sliceable:
                     freq_inds = freq_slices
             else:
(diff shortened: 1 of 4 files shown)


nkern:
Screen Shot 2020-04-15 at 6 17 25 AM

Example of a UVH5 read on 3 HERA IDR2.2 files calling 20 baselines a 1 pol with master (top) and with this branch (bottom). Roughly 1 OOM speed up

nkern:
This turns out to be a more subtle issue than I originally thought. I/O benchmarking is a little tricky because of disk cache, which it seems like can only be completely reset by re-booting the computer. (on a mac, `sudo purge` comes close but doesn't seem to completely clear my cache).

Anyways, here is some benchmarking trying to account for this
Screen Shot 2020-04-15 at 2 58 20 PM
What you are seeing is the new HDF5 slicing yielding a ~25% increase on a cold read (no cache). Interestingly, the new slicing is considerably faster upon re-read, which to me signals that it is using the disk cache more effectively than the master branch currently is.

Anyways, there are even more subtleties: the simultaneously slicing seems to yield speed ups only if the slices are well-matched to the dataset chunking shape (if it is chunked at all). If it is mismatched then it **actually slows it down considerably**. Given this, I think it makes sense to keep the simultaneous slicing off by default, but allow an experienced user to turn it on given an understanding of the data's chunking shape.

@plaplant let me know if that makes sense.

nkern:
Screen Shot 2020-04-15 at 11 20 40 PM

Here is an example on H1C IDR2 data how mismatched simultaneous slicing can slow the read in considerably (over 1 OOM!). In this final case, even though we are leading in only 1 pol vs. 4 pol, its near 30 times slower. In this case its better the load everything in first and then down select, which is currently the default.

plaplant:
@nkern thanks very much for the detailed profiling! I think this type of study is very valuable, and certainly something we'd like to do more of in the future for specific access patterns for analysis. It also makes sense that caching help you a lot, both at the OS and HDF5 levels. HDF5 in particular has some tricks for caching chunked datasets that help speed it up.

I think it makes sense for the profiling to be quite sensitive to the chunk size. Part of the tradeoff of chunking is that when you need to read data from a part of a chunk, HDF5 must read _all_ of the chunk. If you only need a relatively small fraction of that chunk, you'll waste a lot of I/O. On the other hand, we need to have chunking enabled for the compression filters that we use. So it makes sense that there isn't a one-size-fits-all chunk size, and potentially let the user specify what is best for what they are trying to do.

I'm happy to expose the chunk size to the user as an option when writing datasets. I had initially used the automatic values for simplicity's sake, but it seems like for our use cases HDF5 is choosing sub-optimal defaults. I also think that doing multi-dimensional slicing into datasets is interesting, and something that's worth including as an option. I'll do a more detailed review in the next couple of days, but I think the high-level ideas in this PR are excellent ones. Thanks!

nkern:
@plaplant great sounds good. Another thing I'm thinking about is that every time we re-write these files in our analysis pipe (say delay filtered or lst binned or xtalk subtracted) if we don't expose the chunking kwarg then we are writing the data w/ HDF5's sub-optimal auto-chunking, even if the raw data from the correlator are chunked well!

For posterity and others tuning in, here is the HDF5 group support page on chunking.
https://support.hdfgroup.org/HDF5/doc/Advanced/Chunking/
See "Pitfalls" for some ways in which chunking can *drastically* reduce performance (i.e. chunks that are too small or too large).

pyuvdata/uvdata/uvh5.py
```diff
@@ -767,6 +1074,7 @@ def write_uvh5(
         check_extra=True,
         run_check_acceptability=True,
         clobber=False,
+        chunks=True,
```
plaplant, line 1077: This is a nice addition. To make sure it's accessible to the user, it should be added to `UVData.write_uvh5` as well.

pyuvdata/uvdata/uvh5.py
```diff
@@ -136,6 +131,224 @@ def _write_complex_astype(data, dset, indices):
...
+    indexing, e.g. dset[indices].
+    This function specializes in repeated slices over the same axis,
+    e.g. if indices is [[slice(0, 5), slice(10, 15), ...], ..., ]
+    """
+    # get dset and arr shape
+    dset_shape = dset.shape
+    arr_shape = _get_dset_shape(dset, indices)
+
+    # create empty array of dset dtype
+    arr = np.empty(arr_shape, dtype=dset.dtype)
```
plaplant, line 310: I don't think you necessarily always want `dtype` of the array to be `dset.dtype`. For instance, with raw HERA correlator data, `dset.dtype` is a custom-defined type, whereas the output array should be `np.complex128` or `np.complex64`. We might just need to have special handling for this case, along the lines of what we do now.

pyuvdata/uvdata/uvh5.py
```diff
@@ -416,78 +631,163 @@ def _get_data(
...
+            self.nsample_array = _index_dset(dgrp["nsamples"], inds)
         else:
             # do select operations on everything except data_array, flag_array
             # and nsample_array
             self._select_metadata(
                 blt_inds, freq_inds, pol_inds, history_update_string, keep_all_metadata
             )
 
+            # determine which axes can be sliced, rather than fancy indexed
+            # max_nslice_frac of 0.1 yields slice speedup over fancy index for HERA data
```
plaplant, line 651: This is a good explanatory comment. Maybe even put a reference to the PR here, because you have the profiling plots?

plaplant (commented):
Thanks @nkern! I think this will be helpful for making reading uvh5 files more flexible, and the profiling certainly suggests that these change help a lot in certain situations.

On the whole, these changes look very good. We should make sure to get back to full test coverage, and I have individual comments mostly related to improving explanations rather than fundamental changes. Once they're addressed, I think this is good to go.

plaplant:
@nkern I forgot to mention in my review, but we should also update the changelog. We're adding the `chunks` keyword to `UVData.write_uvh5` and `multidim_index` to `UVData.read`. We also probably want to mention that reading uvh5 files is optimized for multidimensional slicing (provided it's well-matched to the chunk size).

pyuvdata/uvdata/uvh5.py
```diff
@@ -136,6 +131,224 @@ def _write_complex_astype(data, dset, indices):
     return
 
 
+def _convert_to_slices(indices, max_nslice_frac=0.1):
```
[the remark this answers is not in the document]
nkern: That's right. The idea here is that an indexing list, say, like `indices = [0, 1, 2, 3, 10, 11, 12, 13]` can be represented as `[slice(0, 4), slice(10, 14)]`, and performing `dset[slice(0, 4)]` and then `dset[slice(10, 14)]` in serial is faster than `dset[indices]`. This is still multi-slicing, but its still better than direct fancy indexing.

However, say we have an indices list that is not really representable in terms of slices (take for example a list of random integers) `[0, 10, 25, 55, 105, 235]`, then this could be represented as `[slice(0, 11, 10), slice(25, 56, 30), slice(105, 236, 130)]`, but this begins to be suboptimal as the number of elements increases. In the limit that we have Nslices = Nindices then the slice indexing is actually slower than the fancy indexing, so its best to just keep `indices` as is. That's what this parameter is trying to get at.

pyuvdata/uvdata/uvh5.py
```diff
@@ -136,6 +131,224 @@ def _write_complex_astype(data, dset, indices):
...
+    """
+    Convert list of indices to a list of slices.
+
+    Parameters
+    ----------
+    indices : list
+        A 1D list of (preferably monotonically increasing) integers for
```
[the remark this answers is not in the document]
nkern: Its not an assumption, and it might just be confusing, so I'll just take it out.

pyuvdata/uvdata/uvh5.py
nkern: Yeah as you can see I haven't changed the `_read_complex_astype` for this reason. I guess perhaps keep this as is until we have a need for making it more flexible? Lmk what you think.

plaplant: Yes, but it looks like `_read_complex_astype` calls this internally now. Thinking about it a little bit more, my guess is that when the assignment happens (e.g., when we do `output_array.real = _index_dset(dset["r"], indices)`) the array is initially an integer type for HERA-correlator raw data because that's the dataset dtype, which then gets coerced into the declared type (`np.real32` or `np.real64`, depending on whether the output type is `np.complex64` or  `np.complex128`). I don't think it's that much of a performance penalty, since we'd have to convert the type eventually. It's just less explicit now, so maybe we can add a comment that the type conversion will happen outside of this function.

nkern: Yes that's right--but I think that this is the same as it was before: `_read_complex_astype` would just assign `output_array.real` with `dset[...]`, which if it was an integer type would then be upcast into whatever dtype `output_array` was specified with. I can add a comment nonetheless. Where did you want this? In `_index_dset` or in `_read_complex_astype`?

plaplant: Yeah, maybe here where we make the placeholder output array, we just mention that it will natively be the same type as the data on disk, but may be upcast/type converted upon assignment.

plaplant:
Thanks very much @nkern! Looks like we've lost some test coverage, so I can add a few tests to get back to full coverage. I think that's the last remaining issue, and then we'll be good to go.

pyuvdata/uvdata/uvh5.py
nkern: Okay but just to be clear, the function `_index_dset` is simply meant to index the dset with the same dtype as its written. The specialty function `_read_complex_astype` is meant specifically to upcast whats index by `_index_dset` into complex type. So I were to put a comment it would be in `_read_complex_astype`. I think the function is fairly self explanatory as is, but I'll add a comment anyways for good measure :)

plaplant (approved):
@nkern thanks so much for the extra tests! I think this is good to go.

@bhazelton do you want to review too, or should we merge?

bhazelton:
I'm happy to merge. Thank you both!

PR #808, opened 2020-04-21
Speed up calfits reading by not deleting header items

jsdillon:
## Description
As a UVCal object is being created, fits header items were being deleted for bookkeeping reasons. For calfits files with very long histories (as was in the case in some H3C data) and small amounts of data, this turned out ot be increasing the runtime by roughly an OOM. Now they are deleted and the bookkeeping is done differently.

Bryna: I pulled Josh's logic out to a utility function that is now called by all three FITS readers (uvfits, calfits, beamfits).

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -169,6 +169,63 @@ def _fits_indexhdus(hdulist):
     return tablenames
 
 
+def _get_fits_extra_keywords(header, keywords_to_skip=None):
+    """
+    Get any extra keywords and return as dict.
+
+    Parameters
+    ----------
+    header : FITS header object
+        header object to get extra_keywords from.
+    keywords_to_skip : list of str
+        list of keywords to not include in extra keywords in addition to standard
+        FITS keywords.
+
+    Returns
+    -------
+    dict
+        dict of extra keywords.
+    """
+    # List standard FITS header items that are still should not be included in
+    # extra_keywords
+    std_fits_substrings = [
+        "HISTORY",
+        "SIMPLE",
+        "BITPIX",
+        "EXTEND",
+        "BLOCKED",
+        "GROUPS",
+        "PCOUNT",
+        "BSCALE",
+        "BZERO",
+        "NAXIS",
+        "PTYPE",
+        "PSCAL",
+        "PZERO",
+        "CTYPE",
+        "CRVAL",
+        "CRPIX",
+        "CDELT",
+        "CROTA",
+        "CUNIT",
+    ]
+
+    if keywords_to_skip is not None:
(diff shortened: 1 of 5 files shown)


pyuvdata/uvbeam/beamfits.py
```diff
@@ -298,41 +298,8 @@ def read_beamfits(
...
-                        primary_header.remove(key)
-
-            # find all the remaining header items and keep them as extra_keywords
-            for key in primary_header:
-                if key == "COMMENT":
-                    self.extra_keywords[key] = str(primary_header.get(key))
-                elif key != "":
-                    self.extra_keywords[key] = primary_header.get(key)
+
+            self.extra_keywords = uvutils._get_fits_extra_keywords(primary_header)
```
jsdillon, line 302: Is there any circumstance in which self.extra_keywords would already have entries? In which case, shouldn't we use update?

Also, does it matter that self.extra_keywords is actually a UVParameter and not a dictionary?

bhazelton: I don't think it should have entries already, this is in the code that reads a single file.

Actually `self.extra_keywords` is a property linked to `self._extra_keywords` which is the UVParameter. This is the right way to set it.

pyuvdata/utils.py
```diff
@@ -169,6 +169,63 @@ def _fits_indexhdus(hdulist):
...
+    ]
+
+    if keywords_to_skip is not None:
+        std_fits_substrings.extend(keywords_to_skip)
+
+    extra_keywords = {}
+    # find all the other header items and keep them as extra_keywords
+    for key in header:
+        if np.any([sub in key for sub in std_fits_substrings]):
+            continue
```
mkolopanis, line 220: Is this so if someone has `HISTORY_2_3_MINE"  in their header it won't get put into the fits keywords? Is there also a large precedent for this? I suppose at first glance I would have expected the logic to look like
```
if key in std_fits_substrings:
            continue
```

bhazelton: No. It's because NAXIS and the P & C keywords in the list are followed by various integers. I can do something fancier if you think it's needed.

mkolopanis: Ahh okay that actually makes a lot of sense. I think I had just forgotten about how fits does that.  Maybe add a comment so we (I) don't forget again?

bhazelton: sure, I'll add a comment.

bhazelton:
@mkolopanis I added the comments you asked for.

Issue #800, opened 2020-04-02
Testing speed up


mkolopanis:
More of a note for myself to look into, but can we use asynchronous testing to speed up the UVBeam testing. If we can push the beam healpix conversion to the background while other tests are executed this may help with testing times.
https://github.com/pytest-dev/pytest-asyncio

plaplant:
I think we might actually want to use parallel testing using [pytest-xdist](https://pypi.org/project/pytest-xdist/), rather than `pytest-asyncio`. I've toyed around with it locally doing, e.g., `pytest -n 8`, but gotten mixed results, largely because of file I/O issues. I think we'd have to move to using temporary directories for all I/O tests rather than writing files to the `data/test` directory, like we do now. But I'm definitely on board--if we can get it to work reliably, I think we can really speed up our test times.

mkolopanis:
yeah very interesting, this does seem like a better solution. There are temp directory fixtures build in to pytest too so the conversion should be "easy".

mkolopanis:
Looks like github action runners have a 2-cpu setup. Which would still provide some level of increase compared to single threading.

mkolopanis:
I'm looking into this a bit, starting with the temp_path fixture in pytest, but the miriad segfaults are popping up again.

plaplant:
Since the last time I fought with miriad segfaults, I realized we should probably define a private `_close()` method on the `Miriad` object that has a corresponding C++ function call that will flush and close the file. Right now it's only done when python calls the object's destructor, which the user has no control over. (Hence our current approach of "call `del`, run garbage collection, and cross our fingers it's actually closed.") We can then call that `_close()` method at the end of `write_miriad`, which (I hope) will prevent this segfault issue.

I'll take a quick stab at such a function definition, though I'm not sure of a robust way to test it besides removing the current song-and-dance from miriad tests and seeing if CI works...

mkolopanis:
If this temp stuff can reliably produce them I'd love to try that out paired with this update. That would inspire confidence for me at least, but yeah not really a way to definitively test it.

plaplant:
PRs #813 and #815 have significantly sped tests up and implemented `pytest-xdist`, so tests typically take less than 10 minutes per job now. There are probably some further improvements we could make, but I think we've made enough progress to close this issue for now.

PR #813, opened 2020-04-23
Speed up beam tests

bhazelton:
## Description

Use fixtures and cut sky healpix maps to speed up beam tests. Previously, the slowest beam test took ~154s on my laptop and the entire suite took 1401.83s = 23.36 minutes. Now the entire beam test suite runs in ~94s = 1.57 minutes on my laptop.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation change (documentation changes only)
- [ ] Version change
- [x] Build or continuous integration change

## Checklist:

- [x] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/master/.github/CONTRIBUTING.md).
- [x] My code follows the code style of this project.

Build or continuous integration change checklist:
- [ ] If required or optional dependencies have changed (including version numbers), I have updated the readme to reflect this.
- [ ] If this is a new CI setup, I have added the associated badge to the readme and to references/make_index.py (if appropriate).

PR #815, opened 2020-04-24
Update tests to use OS level temporary directories & secret miriad re-write

### LINKED ISSUE #812: Rewrite miriad interface using Cython/CFFI
Right now, we are interfacing with the underlying miriad routines using a series of hand-written wrapper functions using the [Python/C API](https://docs.python.org/3/c-api/index.html). While this has worked well on the whole, there are some corner cases (like properly closing files) where the handling of the objects can yield segmentation faults or other bits of user-unfriendliness. In particular the reference counting seems to be incorrect, and tracking down how it should be done is likely non-trivial.

One possible solution is to rewrite the miriad interface using [Cython](https://cython.readthedocs.io/en/latest/) or [CFFI](https://cffi.readthedocs.io/en/latest/), which (to my understanding) are ways to access the functionality of compiled C code without having to do reference counting or explicit interface function building by hand. Either of these options is likely a large amount of work, but would lead to more stability of the miriad interface. I haven't personally used either of these options in non-trivial cases, so any expertise or guidance in this area would be most appreciated.

mkolopanis:
Uses temporary directories for testing I/o
## Description

Added the use of the `tmp_path` fixture for testing I/O. This make all test write to an OS level tmp directory, the OS will handle cleanup for us now so we don't have to keep looking for old files and deleting them.

Rewrote the python miriad interface as a cython extension (`miriad_wrap.pyx`). The old pure cpp one was continuing to have segfaults and all around un-diagnosable issues when caught in certain cases. Most notably recently it is believed to not close file handlers after read/write execution though there is no definitive evidence to prove this. 

The new cython module runs ~1s slower over the course of the 60 test_miriad tests that my machine collected (skipped one requiring casa-core). At least two of the tests also read a uvfits file, write a miriad, then read said miriad and compare to the original uvfits. This leads me to believe the new module behaves comparably speed wise and also is collecting and writing the information correctly to dish.
## Motivation and Context

The motivation was to allow speed up on the way to #800 but also sparked further development.
closes #812 

I'm honestly not sure this falls under _any_ of the categories. Maybe bug fix because it does not seem to have the miriad segfualts anymore?

@bhazelton I think I could see arguments either way for a changelog update. So I'll leave it up to you to decide. It is a big change, but it is also all behind the scenes for users.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:


mkolopanis:
One big thing I have noticed is installation takes more time now, presumably because the cython has to read the extension and  write a ~10000 line cpp file.

mkolopanis:
I think this PR should be merged in after #813, and then we could implement `python-xdist` too in here if we want.

I think the bottleneck then would be in mwa corr fits tests.

mkolopanis:
Looks like we relied on some of those older files to ensure the overwriting worked correctly. I'll add in some lines to make sure these get addressed.

README.md
```diff
@@ -200,7 +200,7 @@ If you want to do development on pyuvdata, in addition to the other dependencies
 you will need the following packages:
 
 * pytest
-* pytest-cases
+* pytest-cases >= 1.12.1
```
plaplant, line 203: Should we also add cython to either this list or the main list of dependencies above? (Not sure if the typical user needs cython installed or not...)

pyuvdata/uvdata/src/miriad_wrap.cpp
```diff
@@ -1,738 +0,0 @@
-#include <Python.h>
```
plaplant, line 1: `miriad_wrap.cpp`... gone but not forgotten!

pyuvdata/uvdata/src/miriad_wrap.pyx
```diff
@@ -0,0 +1,578 @@
...
+
+cdef inline float MKBL(int i, int j):
+  return (i + 1) << 8 | (j + 1) if (i + 1 < 256 and j + 1 < 256) else (i + 1) * 2048 + (j + 1 + 65536)
+
+cdef inline void CHK_IO(int i) except *:
+  if (i != 0):
+    raise IOError("IO failed.")
+
+class MiriadError(RuntimeError):
+  pass
```
plaplant, line 108: So my read on this is that any error gets raised as a `RuntimeError`, but then nothing else happens. Is this the behavior that we want? Should we try to close the file if it's open? My guess is that this is essentially what was happening before, but we might want to think about whether this is what we'd like going forward.

pyuvdata/uvdata/src/miriad_wrap.pyx
```diff
@@ -0,0 +1,578 @@
...
+
+  elif type[0] == "r":
+    offset = INIT(item_hdl, real_item, H_REAL_SIZE)
+
+  elif type[0] == "d":
+    offset = INIT(item_hdl, dble_item, H_DBLE_SIZE)
+
+  elif type[0] == "c":
+    offset = INIT(item_hdl, cmplx_item, H_CMPLX_SIZE)
+
```
plaplant, line 151: Should we have an `else` block here for any other character? Or is that type of handling done somewhere else?

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -906,20 +840,22 @@ def test_miriad_telescope_locations():
...
     # test load_antpos w/ blank Miriad
     uv_in = Miriad()
     uv = aipy_extracts.UV(testfile)
     uv_in._load_antpos(uv)
     assert uv_in.antenna_positions is not None
 
     # cleanup
-    del uv, uv_in
-    gc.collect()
+    # uv.close()
```
plaplant, line 851: I think these lines are left over from me... we should get rid of them.

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -933,9 +869,9 @@ def test_miriad_integration_time_precision():
     assert uv_in == new_uv
 
     # cleanup
-    del new_uv, uv_in
-    gc.collect()
-    shutil.rmtree(write_file)
+    # del new_uv, uv_in
```
plaplant, line 872: These too.

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -971,14 +907,14 @@ def test_read_write_read_miriad_partial_bls(select_kwargs):
     assert uv_in == exp_uv
 
     # cleanup
-    del uv_in, full
-    gc.collect()
-    shutil.rmtree(write_file)
+    # del uv_in, full
```
plaplant, line 910: More (sorry)

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -994,10 +930,10 @@ def test_read_write_read_miriad_partial_antenna_nums():
     assert np.max(exp_uv.ant_2_array) == 0
     assert uv_in == exp_uv
 
-    # cleanup
-    del uv_in, exp_uv, full
-    gc.collect()
-    shutil.rmtree(write_file)
+    # # cleanup
```
plaplant, line 933: Even more

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -1033,16 +968,16 @@ def test_read_write_read_miriad_partial_times(select_kwargs):
     exp_uv = full.select(times=full_times, inplace=False, **select_kwargs)
     assert uv_in == exp_uv
 
-    # cleanup
-    del uv_in, full, exp_uv
-    gc.collect()
-    shutil.rmtree(write_file)
+    # # cleanup
```
plaplant, line 971: Again.

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -1056,15 +991,15 @@ def test_read_write_read_miriad_partial_pols(pols):
     exp_uv = full.select(polarizations=pols, inplace=False)
     assert uv_in == exp_uv
 
-    # cleanup
-    del uv_in, full
-    gc.collect()
-    shutil.rmtree(write_file)
+    # # cleanup
```
plaplant, line 994: And again.

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -1078,10 +1013,10 @@ def test_read_write_read_miriad_partial_ant_str():
     assert np.array([blp[0] == blp[1] for blp in uv_in.get_antpairs()]).all()
     exp_uv = full.select(ant_str="auto", inplace=False)
     assert uv_in == exp_uv
-
-    del uv_in
-    gc.collect()
-    shutil.rmtree(write_file)
+    #
+    # del uv_in
```
plaplant, line 1017: Running out of witty comments.

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -1093,9 +1028,9 @@ def test_read_write_read_miriad_partial_ant_str():
     exp_uv = full.select(ant_str="cross", inplace=False)
     assert uv_in == exp_uv
 
-    del uv_in
-    gc.collect()
-    shutil.rmtree(write_file)
+    # del uv_in
```
plaplant, line 1031: Here too.

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -1106,9 +1041,9 @@ def test_read_write_read_miriad_partial_ant_str():
     assert uv_in == full
 
     # cleanup
-    del uv_in, full, exp_uv
-    gc.collect()
-    shutil.rmtree(write_file)
+    # del uv_in, full, exp_uv
```
plaplant, line 1044: Same.

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -1202,15 +1139,15 @@ def test_read_write_read_miriad_partial_errors(err_type, select_kwargs, err_msg)
         uv_in.read(write_file, **select_kwargs)
     assert str(cm.value).startswith(err_msg)
 
-    del uv_in, full
-    gc.collect()
-    if os.path.exists(write_file):
-        shutil.rmtree(write_file)
+    # del uv_in, full
```
plaplant, line 1142: Once more.

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -1225,14 +1162,14 @@ def test_read_write_read_miriad_partial_error_special_cases():
     )
 
     # cleanup
-    del uv_in, full
-    gc.collect()
-    shutil.rmtree(write_file)
+    # del uv_in, full
```
plaplant, line 1165: As an aside, when we go through and get rid of these, we should make sure that we're closing any `_miriad.UV` objects with the new `.close()` method, rather than this hokey `del/gc.collect()` scheme.

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -1253,10 +1190,10 @@ def test_read_write_read_miriad_partial_with_warnings():
     )
     exp_uv = full.select(times=times_to_keep, inplace=False)
     assert uv_in == exp_uv
-
-    del uv_in
-    gc.collect()
-    shutil.rmtree(write_file)
+    #
```
plaplant, line 1193: More of my cruft.

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -1276,16 +1213,16 @@ def test_read_write_read_miriad_partial_with_warnings():
     exp_uv = full.select(blt_inds=blts_select, antenna_nums=ants_keep, inplace=False)
     assert uv_in != exp_uv
 
-    # cleanup
-    del uv_in, full
-    gc.collect()
-    shutil.rmtree(write_file)
+    # # cleanup
```
plaplant, line 1216: Maybe the last of it?

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -1418,21 +1344,21 @@ def test_rwr_miriad_antpos_issues():
 
     assert uv_in == uv_out
 
-    # cleanup
-    del uv_in, uv_out
-    gc.collect()
-    shutil.rmtree(write_file)
+    # # cleanup
```
plaplant, line 1347: Spoke too soon.

pyuvdata/uvdata/tests/test_uvh5.py
```diff
@@ -1457,21 +1462,25 @@ def test_initialize_uvh5_file_errors(uv_partial_write):
...
 
     # check that IOError is raised then when clobber == False
     with pytest.raises(IOError) as cm:
         partial_uvh5.initialize_uvh5_file(partial_testfile, clobber=False)
     assert str(cm.value).startswith("File exists; skipping")
 
+    # check we can write to it anyway
+    partial_uvh5.initialize_uvh5_file(partial_testfile, clobber=True)
+    captured = capsys.readouterr()
+    assert captured.out.startswith("File exists; clobbering")
```
plaplant, line 1476: Nice catch, thanks for adding this!

plaplant:
Also FWIW, I _definitely_ think we should update the changelog to say we migrated from the hand-written miriad wrapper to a cython one. It's true that it's hidden from view for most users, but if we have some weird issue down the road related to the wrapper it'd be good to be able to see when that change was introduced in the changelog. (And we should also highlight the hard work that went into making a new wrapper a reality 😄)

README.md
mkolopanis: It is listed as a build dependency in the `pyproject.toml` file. I'm not sure if it is necessary for a user though. I did _try_ to make a blank environment and pip install. Seemed to go okay but maybe someone else has a better test in mind?

pyuvdata/uvdata/src/miriad_wrap.pyx
mkolopanis: That's a pretty good point, we could wrap a lot of the function calls in try/except blocks and call a uv.close() if an error is raised.

pyuvdata/uvdata/src/miriad_wrap.pyx
mkolopanis: I'll add that in, a lot of other functions have it. this one should too.

pyuvdata/uvdata/tests/test_miriad.py
mkolopanis: So I wrote `_miriad.UV.__dealloc__` to call `self.close()` and I think when you call `del miriad_object` it will automatically call the dealloc block. But I'm not sure how that works if it is a UVData object. :thinking:

mkolopanis:
I am on board if we also want to add `python-xdist` here. @bhazelton any comments on the matter?

bhazelton:
I'm fine with it here or as a new PR, whatever you prefer. I agree about the changelog.

bhazelton:
I did wonder why you added `miriad_wrap.cpp` to the gitignore if you're deleting it from the repo. Does it get made again by cython or something?

mkolopanis:
yes the `cythonize` call in the setup.py has cython read the `pyx` file and dynamically create a `miriad_wrap.cpp` as you build the package. If it is not ignored git will probably complain a lot for those of us with developer installs.

README.md
bhazelton: This should also go in the `environment.yml` file that we point developers to in the readme.

setup.py
```diff
@@ -119,7 +130,7 @@ def is_platform_mac():
         "h5py",
         "setuptools_scm",
     ],
-    "tests_require": ["pytest", "pytest-cases"],
+    "tests_require": ["pytest", "pytest-xdist", "pytest-cases>=1.12.1"],
```
bhazelton, line 133: I think we should actually remove this line entirely. Calling tests from setup.py is deprecated: https://github.com/pypa/setuptools/issues/1684

pyuvdata/uvdata/src/miriad_wrap.pyx
mkolopanis: Also I hit plenty of errors during testing/creation that are not Runtime errors
the  `raise_miriad_error` and `except +raise_miriad_error` raise up c-level execution errors as a miriad error. So anything in the cython that will `raise` shows up as it's intended error.

turns out it is difficult to include the `e.what()` from c exceptions up into python. I tried to write a wrapper for it but couldn't figure it out and didn't want to distract myself too much.

plaplant: Thanks for looking into that. It's probably not worth investing more effort into this, especially because we weren't really handling errors properly before. Hopefully the Cython rewrite will make runtime errors even less common going forward.

mkolopanis:
Here's a fun timing runs against master for posterity:
```
(py3) matthew@bahamut [12:18:35] [~/src/pyuvdata] [master]
-> % hyperfine --warmup 2 "pytest -x pyuvdata/uvdata/tests/test_{miriad,aipy_extracts}.py"
Benchmark #1: pytest -x pyuvdata/uvdata/tests/test_{miriad,aipy_extracts}.py
  Time (mean ± σ):     21.122 s ±  0.497 s    [User: 20.780 s, System: 0.286 s]
  Range (min … max):   20.755 s … 22.484 s    10 runs


(py3) matthew@bahamut [12:26:37] [~/src/pyuvdata] [tests/use_tmpdir]
-> % hyperfine --warmup 2 "pytest -x pyuvdata/uvdata/tests/test_{miriad,aipy_extracts}.py"
Benchmark #1: pytest -x pyuvdata/uvdata/tests/test_{miriad,aipy_extracts}.py
 Time (mean ± σ):      22.240 s ±  0.870 s    [User: 21.885 s, System: 0.289 s]
  Range (min … max):   21.325 s … 23.961 s    10 runs
```

PR #825, opened 2020-05-04
Additional corr_fits speed-ups and enhancements

### LINKED ISSUE #830: mwa_corr_fits does not support metadata only reads
Using `/usr/bin/time -v` I've found that attempting a metadata-only read for 10G worth of MWA corr fits files takes about 50G of RAM. This seems unnecessary. It also appears that `UVData.read_mwa_corr_fits` does not have a `read_data` keyword, which may be the culprit i.e. the metadata-only functionality is not there to begin with.

mkolopanis:
A few more tweaks to corr fits and `uvdata.check`
adds metadata only reads to mwa_corr_fits
enables check related keyword features in read_mwa_corr_fits
closes #830 

## Description

unloaded some more corr_fits functions to the cython code.

## Motivation and Context

Does some explicit typing in cython (which is good) and a speed up of cable length correction which will be important if a file has many Nblts.

## Types of changes

- [ ] Version change

## Checklist:

- [x] I have updated the CHANELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,13 @@ All notable changes to this project will be documented in this file.
 - `propagate_flags` keyword for `UVData.frequency_average` which flags averaged samples if any contributing samples were flagged
 
 ### Changed
+- Updated formatting and added more explicit typing in corr_fits.pyx
+- Updated Nants calculations for speed up.
 - updated HERA telescope location to match the HERA defined center of array.
 - `utils.uvcalibrate` now incorporates many more consistency checks between the uvcal and uvdata object. The new keywords `time_check` and `ant_check` were added to control some of these checks.
 
-### Deprecated
 - `utils.uvcalibrate` will error rather than warn if some of the newly added checks do not pass, including if antenna names do not match between the objects, starting in version 2.2. In addition, the `flag_missing` keyword is deprecated and will be removed in version 2.2.
+### Deprecated
 
 ## [2.0.2] - 2020-4-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- `read_data` keyword to read_mwa_corr_fits, allows from metadata only reads of MWA correlator fits files.
 - `propagate_flags` keyword for `UVData.frequency_average` which flags averaged samples if any contributing samples were flagged
 
 ### Changed
@@ -12,9 +13,8 @@ All notable changes to this project will be documented in this file.
 - updated HERA telescope location to match the HERA defined center of array.
 - `utils.uvcalibrate` now incorporates many more consistency checks between the uvcal and uvdata object. The new keywords `time_check` and `ant_check` were added to control some of these checks.
 
-- `utils.uvcalibrate` will error rather than warn if some of the newly added checks do not pass, including if antenna names do not match between the objects, starting in version 2.2. In addition, the `flag_missing` keyword is deprecated and will be removed in version 2.2.
 ### Deprecated
-
+- `utils.uvcalibrate` will error rather than warn if some of the newly added checks do not pass, including if antenna names do not match between the objects, starting in version 2.2. In addition, the `flag_missing` keyword is deprecated and will be removed in version 2.2.
 ## [2.0.2] - 2020-4-29
 
 ### Added
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -954,13 +954,13 @@ miriad, and MWA correlator fits files.
 
 a) Reading the metadata of a uvfits, uvh5, miriad, or mwa corr fits file
 ********************************************************
-For uvh5, uvfits, miriad, and MWA correlator fits files, reading in the metadata results in a metadata only
+For uvh5, uvfits, and MWA correlator fits files, reading in the metadata results in a metadata only
 UVData object (which has every attribute except the data_array,
 flag_array and nsample_array filled out). For Miriad files, less of the
 metadata can be read without reading the data, but many of the attributes
 are available.
 
-FHD, and measurement set (ms) files do not support
+FHD and measurement set (ms) files do not support
 reading only the metadata
 (the read_data keyword is ignored for these file types).
 ::
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -949,18 +949,18 @@ and ``diff_vis`` methods.
 
 UVData: Working with large files
 --------------------------------
-To save on memory and time, pyuvdata supports reading only parts of uvfits, uvh5 and
-miriad files.
+To save on memory and time, pyuvdata supports reading only parts of uvfits, uvh5,
+miriad, and MWA correlator fits files.
 
-a) Reading the metadata of a uvfits, uvh5 or miriad file
+a) Reading the metadata of a uvfits, uvh5, miriad, or mwa corr fits file
 ********************************************************
-For uvh5 and uvfits files, reading in the metadata results in a metadata only
+For uvh5, uvfits, miriad, and MWA correlator fits files, reading in the metadata results in a metadata only
 UVData object (which has every attribute except the data_array,
 flag_array and nsample_array filled out). For Miriad files, less of the
 metadata can be read without reading the data, but many of the attributes
 are available.
 
-FHD, MWA correlator FITS files, and measurement set (ms) files do not support
+FHD, and measurement set (ms) files do not support
 reading only the metadata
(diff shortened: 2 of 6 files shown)


pyuvdata/uvdata/uvdata.py
```diff
@@ -897,7 +892,11 @@ def check(
             )
         if np.any(
             np.isclose(
-                np.linalg.norm(self.uvw_array[~autos], axis=1),
+                np.sqrt(
+                    self.uvw_array[~autos, 0] ** 2
+                    + self.uvw_array[~autos, 1] ** 2
+                    + self.uvw_array[~autos, 2] ** 2
+                ),
```
mkolopanis, line 899: Turns out doing this square and square root is faster than `numpy.linalg.norm` but I can change it back if that is desired, or more readable.

mkolopanis:
Have a changelog update ready, just saving it in case more changes are made.

PyxieLouStar (commented):
looks good!

pyuvdata/uvdata/mwa_corr_fits.py
```diff
@@ -616,9 +602,10 @@ def read_mwa_corr_fits(
             corr_ants_to_pfb_inputs, pfb_inputs_to_outputs, map_inds, conj,
         )
         # reorder data
-        self.data_array = self.data_array[:, :, map_inds]
-        self.nsample_array = self.nsample_array[:, :, map_inds]
-        self.flag_array = self.flag_array[:, :, map_inds]
+        self.data_array = np.take(self.data_array, map_inds, axis=2)
+        self.nsample_array = np.take(self.nsample_array, map_inds, axis=2)
+        self.flag_array = np.take(self.flag_array, map_inds, axis=2)
+        # self.flag_array = self.flag_array[:, :, map_inds]
```
bhazelton, line 608: should this line be removed?

bhazelton (commented):
Looks good, but I haven't perused the cython code at all.

pyuvdata/uvdata/mwa_corr_fits.py
mkolopanis: ah yeah probably should be.

pyuvdata/uvdata/mwa_corr_fits.py
```diff
@@ -457,12 +446,13 @@ def read_mwa_corr_fits(
         self.antenna_positions = antenna_positions_ecef - self.telescope_location
 
         # make initial antenna arrays, where ant_1 <= ant_2
-        ant_1_array = []
-        ant_2_array = []
-        for i in range(self.Nants_telescope):
-            for j in range(i, self.Nants_telescope):
-                ant_1_array.append(i)
-                ant_2_array.append(j)
+        ant_1_array, ant_2_array = np.transpose(
```
plaplant, line 449: Very clever! Can you just add a quick comment explaining that this is faster than the naive approach? (Similar to what you have in the `Nants_data` calculation is fine.)

pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
```diff
@@ -24,64 +26,94 @@ def input_output_mapping():
...
+              2 * out_ant1 * (out_ant1 + 1)
+              + 4 * out_ant2
+              + 2 * out_p1
+              + out_p2
+            )
+            map_inds[bls_ind * 4 + pol_ind] = data_index
+
   return map_inds, conj
+
+cpdef numpy.ndarray get_cable_len_diffs(
```
plaplant, line 95: Do you have to define the precision of the `ndarray` you're going to return? Or is it implicitly done as part of its definition inside the function?

plaplant (commented):
Looks great @mkolopanis! I just had a couple of small things.

pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
mkolopanis: So initially I could have sworn cython yelled at me when the precision was defined here, but now it is not. By default all returns are expected to be a "python object", so there is no harm in not defining the type or precision. But the less type coercion cython has to do the better it is usually. 

I'll add the dtype here because I personally like it. Even if it makes the code a little harder to read it must make the cython/C's time a little easier. But also going to throw this out there: it only save 4 microseconds.

plaplant (dismissed):
Looks good! I'll hold off on merging right now in case there are other changes we want to include or if other reviewers want to weigh in, but I think this is good to go.

pyuvdata/uvdata/tests/test_mwa_corr_fits.py
```diff
@@ -549,3 +549,31 @@ def test_flag_init_errors(flag_file_init, err_type, read_kwargs, err_msg):
...
         uv.read(flag_file_init, **read_kwargs)
     assert str(cm.value).startswith(err_msg)
+
+
+def test_read_metadata_only(tmp_path):
+    """
+    MWA correlator fits to uvfits loopback test.
+
+    Read in MWA correlator files, write out as uvfits, read back in and check
+    for object equality.
```
bhazelton, line 559: This docstring doesn't seem to match the logic below.

mkolopanis: ah yes thank you, copy/pasta errors.

mkolopanis:
Not sure what the deal with the QM tests are. Lots of pyqt5 runtime errors. Not sure how code changes here could not have caused them.

CHANGELOG.md
```diff
@@ -7,10 +7,12 @@ All notable changes to this project will be documented in this file.
 - `propagate_flags` keyword for `UVData.frequency_average` which flags averaged samples if any contributing samples were flagged
 
 ### Changed
+- Updated formatting and added more explicit typing in corr_fits.pyx
+- Updated Nants calculations for speed up.
 - `utils.uvcalibrate` now incorporates many more consistency checks between the uvcal and uvdata object. The new keywords `time_check` and `ant_check` were added to control some of these checks.
 
-### Deprecated
 - `utils.uvcalibrate` will error rather than warn if some of the newly added checks do not pass, including if antenna names do not match between the objects, starting in version 2.2. In addition, the `flag_missing` keyword is deprecated and will be removed in version 2.2.
+### Deprecated
```
bhazelton, line 15: did you mean to change this? I think the uvcalibrate line should be under Deprecated

bhazelton: Also, I think the new metadata only read for MWACorrFITS should be listed in the changelog and maybe added to the tutorial.

bhazelton (commented):
Looks good, just a couple of comments about the changelog.

CHANGELOG.md
mkolopanis: Hmm this is probably a re-base issue. Thanks for catching it. I'll add the section about metadata only reads and look at the tutorial to try to add that in too.

mkolopanis: Looks like in the tutorial, there is a list of files which are compatible with metadata only, and a single example of how. I'll update the wordage to reflect that mwa corr fits is under the supported set now.

docs/tutorial.rst
```diff
@@ -949,18 +949,18 @@ and ``diff_vis`` methods.
...
-To save on memory and time, pyuvdata supports reading only parts of uvfits, uvh5 and
-miriad files.
+To save on memory and time, pyuvdata supports reading only parts of uvfits, uvh5,
+miriad, and MWA correlator fits files.
 
-a) Reading the metadata of a uvfits, uvh5 or miriad file
+a) Reading the metadata of a uvfits, uvh5, miriad, or mwa corr fits file
 ********************************************************
-For uvh5 and uvfits files, reading in the metadata results in a metadata only
+For uvh5, uvfits, miriad, and MWA correlator fits files, reading in the metadata results in a metadata only
```
bhazelton, line 957: I'm don't think Miriad should be listed here because a miriad metadata only read doesn't result in a metadata only object that passes checks -- it's missing many required attributes even for a metadata only object. The details around Miriad are in the next sentence.

docs/tutorial.rst
```diff
@@ -949,18 +949,18 @@ and ``diff_vis`` methods.
...
 UVData object (which has every attribute except the data_array,
 flag_array and nsample_array filled out). For Miriad files, less of the
 metadata can be read without reading the data, but many of the attributes
 are available.
 
-FHD, MWA correlator FITS files, and measurement set (ms) files do not support
+FHD, and measurement set (ms) files do not support
```
bhazelton, line 963: take out the comma after FHD

bhazelton:
And I'm afraid this requires another rebase, sorry!!

bhazelton (approved):
Looks great. Thank you!!

PR #826, opened 2020-05-06
Use smaller but sufficient data types to reduce memory usage.

cjordan:
#135  Description
MWA data comes in 32-bit floats; when combining each pair of float to make a
complex number, the appropriate type is np.complex64, not np.complex128.

I don't know the details behind UVData's nsample_array, but at least in the MWA
context, this is only ever set to 1 or 0. It would be better to use bools or
8-bit ints, but pyuvdata insists that this type should be a float, and
np.float16 is the smallest.

I have verified with a run of SSINS over MWA data that the output is the same,
and the memory usage reduced from ~78 GB to ~44 GB.

## Motivation and Context
The current amount of memory required to use MWA data with pyuvdata is very high. This change (seemingly) improves the memory consumption for free.

## Types of changes
- [ ] Version change

## Checklist:

Bug fix checklist:

New feature checklist:

Breaking change checklist:

Version change checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 - `propagate_flags` keyword for `UVData.frequency_average` which flags averaged samples if any contributing samples were flagged
+- Reduced memory consumption when reading MWA data.
 
 ## [2.0.2] - 2020-4-29
 
--- a/pyuvdata/uvdata/mwa_corr_fits.py
+++ b/pyuvdata/uvdata/mwa_corr_fits.py
@@ -552,10 +552,10 @@ class MWACorrFITS(UVData):
 
         # read data into an array with dimensions (time, uv, baselines*pols)
         self.data_array = np.zeros(
-            (self.Ntimes, self.Nfreqs, self.Nbls * self.Npols), dtype=np.complex128
+            (self.Ntimes, self.Nfreqs, self.Nbls * self.Npols), dtype=np.complex64
         )
         self.nsample_array = np.zeros(
-            (self.Ntimes, self.Nfreqs, self.Nbls * self.Npols), dtype=np.float32
+            (self.Ntimes, self.Nfreqs, self.Nbls * self.Npols), dtype=np.float16
         )
         self.flag_array = np.full(
             (self.Ntimes, self.Nfreqs, self.Nbls * self.Npols), True


plaplant:
Hi @cjordan. Thanks for the PR! I think the idea is a good one, and we already support something similar for one of our other filetypes. The `UVData.read()` method has a `data_array_dtype` keyword, whose docstring is here:

https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/62f919a9c0af810e634f7e3bb42698486162df63/pyuvdata/uvdata/uvdata.py#L6486-L6490

I think it would make a lot of sense to pass this to the `read_mwa_corr_fits` function, which would then be used in the way that you've suggested to set the array's precision in memory. That way the user has some flexibility in deciding what precision they would like to use.

I'm a little hesitant to use half-precision floats for the nsample array. Although the raw data only use 0s or 1s, when data are averaged or combined it can in principle take on any value, which is the motivation behind using floats. Half-precision floats can lead to significant round-off errors, even if the arithmetic is being done in higher precision. My guess is that changing to single precision for the data will provide more significant memory savings anyway.

Do you mind doing some quick profiling using `np.complex64` and `np.float32` for the data and nsample arrays, versus `np.complex64` and `np.float16`, and report back? (It doesn't have to be anything fancy, just the level that you reported in your PR text is fine.) If it turns out to be significantly cheaper to use float16, we can consider supporting that as an option.

cjordan:
Thanks for getting back to me @plaplant. I've carried out a bunch of benchmarks with GNU time - this *is not* (at least on linux) the "time" command on the shell, it is accessible via /usr/bin/time for me.  e.g.

```
$ /usr/bin/time -v uptime
 10:26:46 up  2:33,  2 users,  load average: 0.58, 0.69, 0.61
        Command being timed: "uptime"
        User time (seconds): 0.00
        System time (seconds): 0.00
        Percent of CPU this job got: 6%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.01
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 3016
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 147
        Voluntary context switches: 2
        Involuntary context switches: 0
        Swaps: 0
        File system inputs: 32
        File system outputs: 0
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

So you can see that this takes up about 3 MB of memory and runs very quickly.

Benchmarks reading a 10GB MWA observation with pyuvdata:
1) `np.complex128`, `np.float32` (i.e. before this PR): 49661184 kbytes
2) `np.complex64`, `np.float32` (i.e. before this PR): 31286880 kbytes
3) `np.complex64`, `np.float16` (i.e. before this PR): 28661908 kbytes

So this shows that the peak usage is not actually that different when using the 16-bit floats. But this is only part of the story - bench 2 shows a usage typically about 17 GB, which flares up to its peak value, presumably as various copies are made elsewhere in the code. Bench 3 is typically around 14 GB.  My explanation is fairly simple - the nsample array has the same dimensions as the actual data.  As the raw data is stored as float32, when using `np.complex64` and `np.float32`, we should expect the memory usage to be about double the size of the observation (roughly true), and when using `np.complex64` and `np.float16`, we should expect the memory usage to be about 1.5 the size of the observation (also roughly true).  I guess if you'd want to confirm this, you'd have to do what I do and eyeball the memory usage in real time.

I can't comment on how "damaging" degrading the precision of the nsample array is, as I'm not that familiar with pyuvdata.  My interest is in using the associated SSINS code, and making it's memory usage manageable.  However, it seems that the performance of SSINS is not affected when using `np.float16`, so I'm encouraged that this is a good change.

plaplant:
@cjordan thanks very much for the detailed profiling! It sounds that the "typical" memory usage difference for `np.float16` versus `np.float32` is ~3 GB, or about 20%. It's not huge, but it's probably worth at least supporting a mode where the user can change the precision of the `nsample_array`, especially if they're not actually going to do any computation with it.

So I think my suggestions for the PR are these:
1. Use the existing `data_array_dtype` to control the precision of the `data_array` when calling `read_mwa_corr_fits`. This requires passing the argument from the `read` function "all the way down" to the actual reader in `mwa_corr_fits.py`.
2. Add a new keyword `nsample_array_dtype`, which must be one of `np.float16`, `np.float32`, or `np.float64` that controls the precision of the `nsample_array`. For this PR, we only have to plumb it through to the corr fits reader, but in principle we could imagine extending it to other file types that would support it. I'd prefer the default to be `np.float32`, to avoid round-off errors, but then the user would be able to opt for `np.float16` if appropriate (as it sounds like it is here).

Let me know if you'd like any help with implementing these changes. We should also probably write a quick test to verify the arrays have the precisions requested when using these keywords.

cjordan:
Hi @plaplant, thanks again for getting back to me.  If it were up to me, I would go for your option 2, but I'd rather not do this myself because I have not studied how everything fits together in pyuvdata.  For example, SSINS is currently using the generic `UVData.read` method, meaning if I wanted to use an optional argument to `read` to specify the nsample array dtype, I'd potentially be touching/breaking a lot of other code.

https://github.com/mwilensky768/SSINS/blob/7a8b2b3d56e9aff7bfc0e0e567ea8fef3e74d5cd/scripts/MWA_vis_to_SSINS.py#L11

I could do it, but I'd feel more comfortable if others could help here.  Please, let me know what you think!

mwilensky768:
This might be worth making a separate issue, but I'm also finding that when I do a metadata-only read of MWA corr fits files (10 GB total) that the total RAM consumption is spiking up to 50G somehow. This seems unnecessary.

mkolopanis:
There may be some unnecessary copying in places. I agree @mwilensky768 let's make a new issue and someone can take a memory profiler to it and work it out.

plaplant:
@cjordan sure, I'm happy to take over implementing these changes. As you said, it would touch a lot of code, and it might be tricky to get it right. If it's okay with you, it might be less messy to close this PR and have me make a fresh one. But I'll wait to hear back before I do anything.

cjordan:
Hi Paul, I don't mind - please proceed as you see fit. I'm not worried! And
thanks for your help!

plaplant:
Closing this in favor of #832.

PR #827, opened 2020-05-06
Add option to average rather than select in compress_by_redundancy

### LINKED ISSUE #670: Add option to `compress_by_redundancy` to average over rather than sample from the redundant group
Maybe similar to [#428](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/issues/428), but there is currently nothing in place to average over redundant baseline groups, only downselect/inflate.  This should be possible with the `get_baseline_redundancies` et al. functions?

bhazelton:
## Description

Add an option in compress_by_redundancy to average over the redundant baselines rather than selecting one of them. The final `data_array` is a weighted average using the `nsample_array` as the weights. The final `nsample_array` is a sum over the redundant baselines (so it can be larger than 1) and the final `flag_array` is an `or` over the redundant baselines.

## Motivation and Context

closes #670 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -1086,6 +1086,9 @@ def find_clusters(location_ids, location_vectors, tol):
     list of list of location_ids
 
     """
+    if location_vectors.ndim == 1:
+        location_vectors = location_vectors[:, np.newaxis]
+
     # For each baseline, list all others that are within the tolerance distance.
     adj_triu_mat = pdist(location_vectors) < tol
     adj = {}  # Adjacency dictionary
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -1086,8 +1086,8 @@ def find_clusters(location_ids, location_vectors, tol):
     list of list of location_ids
 
     """
-    if location_vectors.ndim == 1:
-        location_vectors = location_vectors[:, np.newaxis]
+    location_vectors = np.atleast_2d(location_vectors)
+    location_ids = np.asarray(location_ids)
 
     # For each baseline, list all others that are within the tolerance distance.
     adj_triu_mat = pdist(location_vectors) < tol
@@ -1132,7 +1132,7 @@ def get_baseline_redundancies(baselines, baseline_vecs, tol=1.0, with_conjugates
     baselines : array_like of int
         Baseline numbers, shape (Nbls,)
     baseline_vecs : array_like of float
-        Baseline vectors in meters, shape shape (Nbls, 3)
+        Baseline vectors in meters, shape (Nbls, 3)
     tol : float
         Absolute tolerance of redundancy, in meters.
     with_conjugates : bool
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -1086,8 +1086,10 @@ def find_clusters(location_ids, location_vectors, tol):
     list of list of location_ids
 
     """
-    location_vectors = np.atleast_2d(location_vectors)
+    location_vectors = np.asarray(location_vectors)
     location_ids = np.asarray(location_ids)
+    if location_vectors.ndim == 1:
+        location_vectors = location_vectors[:, np.newaxis]
 
     # For each baseline, list all others that are within the tolerance distance.
     adj_triu_mat = pdist(location_vectors) < tol
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -1068,6 +1068,58 @@ def get_lst_for_time(jd_array, latitude, longitude, altitude):
     return lst_array
 
 
+def find_clusters(location_ids, location_vectors, tol):
+    """
+    Find clusters of vectors (e.g. redundand baselines, times).
+
+    Parameters
+    ----------
+    location_ids : array_like of int
+        ID labels for locations.
+    location_vectors : array_like of float
+        location vectors, can be multidimensional
+    tol : float
+        tolerance for clusters
+
+    Returns
+    -------
+    list of list of location_ids
+
+    """
+    # For each baseline, list all others that are within the tolerance distance.
+    adj_triu_mat = pdist(location_vectors) < tol
+    adj = {}  # Adjacency dictionary
+
+    for bi, col in enumerate(squareform(adj_triu_mat)):
+        col[bi] = True
+        adj[location_ids[bi]] = location_ids[col]
+
+    # The adjacency list defines a set of graph edges.
+    # For each location b0, loop over its adjacency list ai \in adj[b0]
+    #   If adj[b0] is a subset of adj[ai], then ai is in a redundant group with b0
+    loc_gps = []
+    for k in adj.keys():
+        a0 = adj[k]
+        group = [k]
+        for a in a0:
+            if set(a0).issubset(adj[a]) and a not in group:
+                group.append(a)
+        group.sort()
+        loc_gps.append(group)
+
+    # Groups can be different lengths, but we need to take a unique over an axis
+    # to properly identify unique groups
+    # Pad out all the sub-lists to be the same length
+    pad = len(max(loc_gps, key=len))
+    loc_gps = np.array([i + [-1] * (pad - len(i)) for i in loc_gps])
+    # We end up with multiple copies of each redundant group, so remove duplicates
+    loc_gps = np.unique(loc_gps, axis=0).tolist()
+    # remove the dummy pad baselines from each list
+    loc_gps = [[bl for bl in gp if bl != -1] for gp in loc_gps]
+
+    return loc_gps
+
+
 def get_baseline_redundancies(baselines, baseline_vecs, tol=1.0, with_conjugates=False):
     """
     Find redundant baseline groups.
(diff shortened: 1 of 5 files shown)


bhazelton:
A question about this just occurred to me:
When we're averaging baselines by redundancy, should the `nsample_array` be summed across the averaged baselines? That seems appropriate since many values are going in, but it means the `nsample_array` will be different if the method is `average` vs `select`.

bhazelton:
On telecon, we agreed that the nsample_array should be summed. Also need to check that integration times are compatible.

pyuvdata/uvdata/uvdata.py
```diff
@@ -5563,14 +5570,178 @@ def compress_by_redundancy(self, tol=1.0, inplace=True, keep_all_metadata=True):
...
+                        if np.all(flags_to_avg):
+                            mask = np.zeros_like(flags_to_avg)
+                        else:
+                            mask = flags_to_avg
+
+                        vis_to_avg = np.ma.masked_array(vis_to_avg, mask=mask)
+
+                        nsample_to_avg = np.ma.masked_array(nsample_to_avg, mask=mask)
+
+                        avg_vis = np.average(vis_to_avg, weights=nsample_to_avg, axis=0)
```
mkolopanis, line 5714: Just want to verify  we are getting the behavior here that we want. When this call is invoked without explicitly calling the numpy.ma.average function it will count the weights of the masked items towards the total weights. If you only want to include the weights of the unmasked items we would want to us the `ma` module in numpy.

Ah there is also a `returned` keyword in the average function that can return the summed weights if desired.

See this example if this was confusing. I added a suggestion in case you wanted to move to the `ma` module's average. 
```
(py38) matthew@bahamut [9:00:27] [~] 
-> % ipython       
Python 3.8.1 (default, Jan  8 2020, 22:29:32) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.11.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np                                                                               

In [2]: this = np.arange(10)                                                                             

In [4]: this = np.ma.masked_array(this, mask=[1 if ind % 2 ==0 else 0 for ind in range(10)])             

In [6]: this                                                                                             
Out[6]: 
masked_array(data=[--, 1, --, 3, --, 5, --, 7, --, 9],
             mask=[ True, False,  True, False,  True, False,  True, False,
                    True, False],
       fill_value=999999)

In [7]: np.average(this)                                                                                 
Out[7]: 5.0

In [8]: this.mean()                                                                                      
Out[8]: 5.0

In [9]: np.average(this, weights=np.arange(10))                                                          
Out[9]: 3.6666666666666665

In [10]: np.sum(this*np.arange(10))/np.sum(np.arange(10))                                                
Out[10]: 3.6666666666666665

In [11]: np.ma.average(this, weights=np.arange(10))                                                      
Out[11]: 6.6

```
```suggestion
                        avg_vis = np.ma.average(vis_to_avg, weights=nsample_to_avg, axis=0)
```

bhazelton: oh, good catch. I think we do want the `ma` one here. I also want to go check the time & frequency averaging to make sure they're doing it correctly. I also want to add a check to catch this, let me think about how to do that.

bhazelton: Ha! so I made a test to catch this and it passed. Then I messed around on the command line and discovered that we're actually doing the right thing (accidentally) because we are passing a masked array as the weights, which gets us the correct behavior even when we use `np.average`. I think this should be changed or at least commented to make sure we don't confuse anyone in the future though.

My demonstration:
```
In [1]: import numpy as np                                                                                                    

In [2]: data = [1, 0, 0, 0, 0]                                                                                                

In [3]: flags = [False, True, True, True, True]                                                                               

In [4]: nsample = [1, 1, 1, 1, 1]                                                                                             

In [5]: masked_data = np.ma.masked_array(data = data, mask = flags)                                                           

In [6]: masked_data                                                                                                           
Out[6]: 
masked_array(data=[1, --, --, --, --],
             mask=[False,  True,  True,  True,  True],
       fill_value=999999)

In [7]: np.average(masked_data)                                                                                               
Out[7]: 1.0

In [8]: np.average(masked_data, weights=nsample)                                                                              
Out[8]: 0.2

In [9]: np.ma.average(masked_data, weights=nsample)                                                                           
Out[9]: 1.0

In [10]: masked_nsample = np.ma.masked_array(data=nsample, mask=flags)                                                        

In [11]: np.average(masked_data, weights=masked_nsample)                                                                      
Out[11]: 1.0

In [12]: np.ma.average(masked_data, weights=masked_nsample)                                                                   
Out[12]: 1.0
```

mkolopanis: ah interesting. nice catch there. Think of the terror if the weights and data arrays had different masks going in :scream:

bhazelton: The horror! I don't even know what should happen in that case.

bhazelton: I'll commit my test that checks this and change to using `np.ma.average` to make sure there's no confusion in the future.

bhazelton:
I still need to update the changelog, I was waiting until it seemed close before I did that to prevent too many conflicts.

pyuvdata/uvdata/uvdata.py
bhazelton: I just checked the time & frequency averaging, they look ok.

bhazelton:
@jburba This PR is a long delayed answer to an issue you made about averaging over redundant baselines. I'd welcome any comments you have if you have time to look at it.

pyuvdata/uvdata/uvdata.py
```diff
@@ -5563,14 +5570,179 @@ def compress_by_redundancy(self, tol=1.0, inplace=True, keep_all_metadata=True):
...
+                conj_group_inds = []
+                conj_group_times = []
+                for bl in reg_group:
+                    bl_inds = np.where(self.baseline_array == bl)[0]
+                    group_inds.extend(bl_inds)
+                    group_times.extend(self.time_array[bl_inds])
+                for bl in conj_group:
+                    bl_inds = np.where(self.baseline_array == bl)[0]
+                    conj_group_inds.extend(bl_inds)
+                    conj_group_times.extend(self.time_array[bl_inds])
```
mkolopanis, line 5621: We might be able to do this part without having to loop through every baseline with a `numpy.searchsorted` kind of like the uvw calculation. However, it would require the object to be in the `baseline,time` ordering to ensure that the `baseline_array` is sorted with baseline number increasing. But I'm not sure it would really save us too much. Probably worth some profiling to investigate later?

pyuvdata/uvdata/uvdata.py
```diff
@@ -5563,14 +5570,179 @@ def compress_by_redundancy(self, tol=1.0, inplace=True, keep_all_metadata=True):
...
+
+                    regular_orientation = np.array(
+                        [time_ind for time_ind in gp if time_ind < len(group_times)],
+                        dtype=np.int,
+                    )
+                    regular_inds = group_inds[np.array(regular_orientation)]
+                    conj_orientation = np.array(
+                        [time_ind for time_ind in gp if time_ind >= len(group_times)],
+                        dtype=np.int,
+                    )
```
mkolopanis, line 5665: The ` < len(group_times)` took me  second to understand, since that's the barrier between normally and conjugated baselines in the array, maybe we can add a inline comment to that effect?

pyuvdata/utils.py
```diff
@@ -1143,6 +1143,61 @@ def get_lst_for_time(jd_array, latitude, longitude, altitude):
...
+    tol : float
+        tolerance for clusters
+
+    Returns
+    -------
+    list of list of location_ids
+
+    """
+    if location_vectors.ndim == 1:
+        location_vectors = location_vectors[:, np.newaxis]
```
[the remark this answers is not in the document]
bhazelton: good catch, thanks!

pyuvdata/uvdata/uvdata.py
bhazelton: yeah, lets do that.

pyuvdata/uvdata/uvdata.py
bhazelton: Yes, I'll add a comment.

pyuvdata/uvdata/uvdata.py
bhazelton: should we make an issue?

pyuvdata/utils.py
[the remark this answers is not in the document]
bhazelton: hmm, `np.atleast_2d` adds the new axis as the front. And I can't just transpose it, because I don't want to transpose it if it's already 2d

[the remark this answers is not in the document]
bhazelton: So I'm using `np.asarray` to ensure it's an array but then keeping the `if` statement as is.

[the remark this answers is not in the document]
mkolopanis: Sounds like a good plan to me.

pyuvdata/uvdata/uvdata.py
mkolopanis: I've made one so we can come back to this (#863 )

PR #821, opened 2020-04-29
properly weight by nsample when downsampling in time

bhazelton:
## Description

This turned out to be more complicated than the I expected, because the integration time can also vary and that needs to be accounted for as well. I also found a couple of bugs in the process.

## Motivation and Context

closes #774 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation change (documentation changes only)
- [ ] Version change
- [ ] Build or continuous integration change

## Checklist:

- [x] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/master/.github/CONTRIBUTING.md).
- [x] My code follows the code style of this project.

Bug fix checklist:
- [x] My fix includes a new test that breaks as a result of the bug (if possible).
- [x] All new and existing tests pass.
- [x] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/master/CHANGELOG.md).

PR #832, opened 2020-05-11
Add array precision options for mwa_corr_fits files

plaplant:
## Description

This change is motivated by #826, and gives the user the opportunity to select data types for the `data_array` and `nsample_array` when reading in MWA corr_fits files. The previous hard-coded defaults of `np.complex128` and `np.float32`, respectively, consumed more memory than typically necessary, especially for instances where no manipulation of these arrays was being done (e.g., when RFI flagging with SSINS).

## Motivation and Context

Part of the motivation for this change is that the data for MWA corr_fits files is saved at single precision (`np.complex64`), so the previous default was artificially increasing the precision. Though subsequent calculations (e.g., applying calibration solutions) would benefit from this increased precision, analysis steps that only required reading the values without modifying them would not.

For the nsample array, the option of half-precision floats (`np.float16`) is added, though its use is discouraged (and mentioned in the docstrings). Floating-point round-off errors can be relatively large (~1e-3), but for raw corr_fits data that contains only 1s and 0s, there is no loss of precision. Care should be taken when averaging or down-selecting (e.g., LST-binning) to ensure that the calculation is being done at sufficient precision.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 - `background` keyword in `uvdata.set_lsts_from_time_array` to allow lst calculation in a background thread.
 - `read_data` keyword to read_mwa_corr_fits, allows from metadata only reads of MWA correlator fits files.
 - `propagate_flags` keyword for `UVData.frequency_average` which flags averaged samples if any contributing samples were flagged
+- `nsample_array_dtype` keyword for `UVData.read`, which sets the datatype for the nsample_array. Currently only used for mwa_corr_fits files.
 
 ### Changed
 - The `time_range` parameter on UVCal is no longer required and it is suggested to only set it if the Ntimes axis is length one.
@@ -18,6 +19,7 @@ All notable changes to this project will be documented in this file.
 - Updated Nants calculations for speed up.
 - updated HERA telescope location to match the HERA defined center of array.
 - `utils.uvcalibrate` now incorporates many more consistency checks between the uvcal and uvdata object. The new keywords `time_check` and `ant_check` were added to control some of these checks.
+- `data_array_dtype` keyword for `UVData.read` is now also respected by mwa_corr_fits files. Previously it was only used by uvh5 files.
 
 ### Deprecated
 - The UVCal methods `set_gain`, `set_delay`, `set_unknown_cal_type`, `set_sky`, and `set_redundant` have been made private. The public methods will be removed in version 2.2.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,22 +4,22 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- `nsample_array_dtype` keyword for `UVData.read`, which sets the datatype for the nsample_array. Currently only used for mwa_corr_fits files.
 - `utils` cython module to speed up some common utility functions.
 - Added support for installing the package on Windows.
 - 'background_lsts' keyword in `uvdata.read` to calculate lsts in the background while reading. Defaults to True.
 - `background` keyword in `uvdata.set_lsts_from_time_array` to allow lst calculation in a background thread.
 - `read_data` keyword to read_mwa_corr_fits, allows from metadata only reads of MWA correlator fits files.
 - `propagate_flags` keyword for `UVData.frequency_average` which flags averaged samples if any contributing samples were flagged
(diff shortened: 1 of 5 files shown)


plaplant:
@cjordan I've made this branch (`mwa_corr_fits_precision`) to implement the changes in your original PR with some added flexibility. You should now be able to do something like this:
```
uv.read_mwa_corr_fits(filelist, data_array_dtype=np.complex64, nsample_array_dtype=np.float16)
```
Let me know if that gives you the memory footprint you had before. Also note that these options will have to be propagated to modules that are reading these files, like SSINS.

pyuvdata/uvdata/uvdata.py
```diff
@@ -5904,8 +5904,8 @@ def read_mwa_corr_fits(
         end_flag=2.0,
         flag_dc_offset=True,
         phase_to_pointing_center=False,
-        phase_data=None,
-        phase_center=None,
+        data_array_dtype=np.complex64,
```
bhazelton, line 5907: Why is this defaulted differently than the `read` method or the `read_mwa_corr_fits` method on the MWACorrFITS object?

bhazelton:
@plaplant should we consider adding some options to the averaging methods to promote these arrays?

pyuvdata/uvdata/uvdata.py
plaplant: Good question. I think because I wasn't paying attention :)

I think given the raw data is single precision, I'd advocate for the default here and in the `MWACorrFITS.read_mwa_corr_fits` method to match that (`np.complex64`), but keep the default on the `UVData.read()` double-precision to avoid changing the current default on users. What are your thoughts?

plaplant:
@bhazelton that's an excellent suggestion re: datatype promotion. I'll go though and try to find all the points where we should do that.

pyuvdata/uvdata/uvdata.py
bhazelton: That sounds reasonable to me.

plaplant:
@bhazelton I added handling of the `nsample_array` precision in the `downsample_in_time` function, but I think that's the only place in `uvdata.py` that actually does arithmetic with it (rather than just down-selecting). I also added a test for it and some in-line comments.

bhazelton:
I think it's also averaged in `frequency_average`. I should also note that the logic in `downsample_in_time` is changed in PR #821, so whichever is merged second will need to do a careful rebase.

CHANGELOG.md
```diff
@@ -5,9 +5,11 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 - `propagate_flags` keyword for `UVData.frequency_average` which flags averaged samples if any contributing samples were flagged
+- `nsample_array_dtype` keyword for `UVData.read`, which sets the datatype for the nsample_array. Currently only used for mwa_corr_fits files.
```
bhazelton, line 8: This should be at the top -- reverse chronological order.

CHANGELOG.md
```diff
@@ -5,9 +5,11 @@ All notable changes to this project will be documented in this file.
...
 
 ### Changed
 - `utils.uvcalibrate` now incorporates many more consistency checks between the uvcal and uvdata object. The new keywords `time_check` and `ant_check` were added to control some of these checks.
+- `data_array_dtype` keyword for `UVData.read` is now also respected by mwa_corr_fits files. Previously it was only used by uvh5 files.
```
bhazelton, line 12: same comment here

cjordan:
Hey - I don't want to derail any useful conversation here, just want to say thanks for the work here. I haven't had time to test it through SSINS yet, but I'll do that when I can (hopefully over the next couple of days).

CHANGELOG.md
plaplant: Good catch, thanks! Should be fixed now.

plaplant:
@bhazelton I think this PR is ready to be looked at again when you get a chance.

pyuvdata/uvdata/uvdata.py
```diff
@@ -6004,9 +6024,9 @@ def read_mwa_corr_fits(
         end_flag=2.0,
         flag_dc_offset=True,
         phase_to_pointing_center=False,
-        phase_data=None,
-        phase_center=None,
```
bhazelton, line 6008: Good catch on these!

PR #834, opened 2020-05-12
Uvdata/timing tweaks, add `read_data` option to FHD

mkolopanis:
## Description

- Adds an option to do the LST calculation in the background during file reads.
- Adds a `read_data` option to FHD to allow metadata only reads of FHD files.
- some tweaks here and there to make a well oiled machine.

## Motivation and Context

For very large files, computing the lst_array in the background can save a little bit of time. However for trivially small files actually costs more time. Allow the user to decide. 

Currently defaults to `True` but we can change to `False`.
## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- 'background_lsts' keyword in `uvdata.read` to calculate lsts in the background while reading. Defaults to True.
 - `background` keyword in `uvdata.set_lsts_from_time_array` to allow lst calculation in a background thread.
 - `read_data` keyword to read_mwa_corr_fits, allows from metadata only reads of MWA correlator fits files.
 - `propagate_flags` keyword for `UVData.frequency_average` which flags averaged samples if any contributing samples were flagged
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,12 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- `background` keyword in `uvdata.set_lsts_from_time_array` to allow lst calculation in a background thread.
 - `read_data` keyword to read_mwa_corr_fits, allows from metadata only reads of MWA correlator fits files.
 - `propagate_flags` keyword for `UVData.frequency_average` which flags averaged samples if any contributing samples were flagged
 
 ### Changed
+- FHD now supports metadata only reads.
 - Updated formatting and added more explicit typing in corr_fits.pyx
 - Updated Nants calculations for speed up.
 - updated HERA telescope location to match the HERA defined center of array.
--- a/pyuvdata/uvdata/fhd.py
+++ b/pyuvdata/uvdata/fhd.py
@@ -580,7 +580,7 @@ class FHD(UVData):
                     pol_list.append(linear_pol_dict[pol])
             self.polarization_array = np.asarray(pol_list)
         else:
-            # Use Npols because for FHD, npol fully specifies whih pols to use
+            # Use Npols because for FHD, npol fully specifies which pols to use
             pol_strings = lin_pol_order[: self.Npols]
             self.polarization_array = np.asarray(
                 [linear_pol_dict[pol] for pol in pol_strings]
--- a/pyuvdata/uvdata/fhd.py
+++ b/pyuvdata/uvdata/fhd.py
@@ -130,7 +130,10 @@ class FHD(UVData):
             Read in the visibility, nsample and flag data. If set to False, only
             the metadata will be read in. Setting read_data to False results in
             a metadata only object. If read_data is False, an obs file must be
-            included in the filelist.
+            included in the filelist. Note that if read_data is False, Npols is
+            derived from the obs file and reflects the number of polarizations
+            used in the FHD run. If read_data is True, Npols is given by the
+            number of visibility data files provided in `filelist`.
         run_check : bool
             Option to check for the existence and proper shapes of parameters
             after after reading in the file (the default is True,
--- a/pyuvdata/uvdata/fhd.py
+++ b/pyuvdata/uvdata/fhd.py
@@ -105,6 +105,8 @@ class FHD(UVData):
         self,
         filelist,
         use_model=False,
+        background_lsts=True,
+        read_data=True,
         run_check=True,
         check_extra=True,
         run_check_acceptability=True,
@@ -116,11 +118,19 @@ class FHD(UVData):
         ----------
         filelist : array_like of str
             The list/array of FHD save files to read from. Must include at
-            least one polarization file, a params file and a flag file.
+            least one polarization file, a params file and a flag file. An obs
+            file is also required if `read_data` is False.
         use_model : bool
             Option to read in the model visibilities rather than the dirty
             visibilities (the default is False, meaning the dirty visibilities
             will be read).
+        background_lsts : bool
+            When set to True, the lst_array is calculated in a background thread.
+        read_data : bool
+            Read in the visibility, nsample and flag data. If set to False, only
+            the metadata will be read in. Setting read_data to False results in
+            a metadata only object. If read_data is False, an obs file must be
+            included in the filelist.
         run_check : bool
             Option to check for the existence and proper shapes of parameters
             after after reading in the file (the default is True,
@@ -145,6 +155,7 @@ class FHD(UVData):
         """
         datafiles = {}
         params_file = None
+        obs_file = None
         flags_file = None
         layout_file = None
         settings_file = None
@@ -173,6 +184,10 @@ class FHD(UVData):
                 if params_file is not None:
                     raise ValueError("multiple params files in filelist")
                 params_file = file
+            elif file.lower().endswith("_obs.sav"):
+                if obs_file is not None:
+                    raise ValueError("multiple obs files in filelist")
+                obs_file = file
             elif file.lower().endswith("_flags.sav"):
                 if flags_file is not None:
                     raise ValueError("multiple flags files in filelist")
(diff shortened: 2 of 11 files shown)


pyuvdata/uvdata/mwa_corr_fits.py
```diff
@@ -435,7 +438,8 @@ def read_mwa_corr_fits(
         self.Nblts = int(self.Nbls * self.Ntimes)
 
         # convert times to lst
-        self.set_lsts_from_time_array()
+        proc = self.set_lsts_from_time_array(background=background_lsts)
+        # self.set_lsts_from_time_array()
```
bhazelton, line 442: this line should be removed.

pyuvdata/uvdata/mwa_corr_fits.py
```diff
@@ -665,7 +669,14 @@ def read_mwa_corr_fits(
...
             self.data_array = np.conj(self.data_array)
 
+        # wait for LSTs if set in background
+        if proc is not None:
+            proc.join()
+
+        if not self.metadata_only:
             # reorder polarizations
+            # reorder pols calls check so much come after
+            # lst thread is re-joined.
```
bhazelton, line 678: typo: much -> must

pyuvdata/uvdata/uvdata.py
```diff
@@ -2230,17 +2255,20 @@ def unphase_to_drift(self, phase_frame=None, use_ant_pos=False):
...
-                    ant2_index = np.where(
-                        self.antenna_numbers == self.ant_2_array[bl_ind]
-                    )[0][0]
-                    self.uvw_array[bl_ind, :] = (
-                        ant_uvw[ant2_index, :] - ant_uvw[ant1_index, :]
-                    )
+                # instead of looping through every ind, find the spot in antenna number
+                # array where ant_num <= ant1 < ant_number and similary for ant2
+                # for all baselines in inds
+                # then find the uvw coordinate for all at the same time
```
bhazelton, line 2261: This comment makes it sound like the antenna numbers need to be in order, but I don't think that's the case because you do an argsort below. Maybe just update the comment.

pyuvdata/uvdata/uvdata.py
```diff
@@ -2437,17 +2465,21 @@ def phase(
...
-                    ant2_index = np.where(
-                        self.antenna_numbers == self.ant_2_array[bl_ind]
-                    )[0][0]
-                    self.uvw_array[bl_ind, :] = (
-                        frame_ant_uvw[ant2_index, :] - frame_ant_uvw[ant1_index, :]
-                    )
+                # instead of looping through every ind, find the spot in antenna number
+                # array where ant_num <= ant1 < ant_number and similary for ant2
+                # for all baselines in inds
+                # then find the uvw coordinate for all at the same time
```
bhazelton, line 2471: same comment here.

pyuvdata/uvdata/uvdata.py
```diff
@@ -6901,6 +6960,7 @@ def read(
                     read_data=read_data,
                     check_extra=check_extra,
                     run_check_acceptability=run_check_acceptability,
+                    background_lsts=background_lsts,
                 )
 
             elif file_type == "fhd":
```
bhazelton, line 6966: Looks like you didn't add the background_lsts for FHD. Was that on purpose?

mkolopanis: I need to look closer at the FHD read to make sure the lst calculation can be started further up. Currently it is done at the end and I just have not taken the time yet.

mkolopanis: That's right, this is why I did not add it. The check on the FHD read to ensure the telescope location is set properly:
```
        try:
            self.set_telescope_params()
        except ValueError as ve:
            warnings.warn(str(ve))

        # need to make sure telescope location is defined properly before this call
        self.set_lsts_from_time_array()
```

bhazelton: we can probably move that up too, I can take a look.

pyuvdata/uvdata/uvdata.py
mkolopanis: I updated the comments, hopefully they are clearer now.

mkolopanis:
some timings done on the phasings. 

unphasings with uvws saves ~30% time with the file i used.
unphasing with ant_pos saved 1 order of magnitude

phasing seems to be about 40-50% regardless of ant_pos/uvw.

[vectorize_phasing.pdf](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/4637280/vectorize_phasing.pdf)

pyuvdata/uvdata/fhd.py
```diff
@@ -205,17 +226,22 @@ def read_fhd(
...
+            # consistent with each other
+            vis_data = {}
+            for pol, file in datafiles.items():
+                this_dict = readsav(file, python_dict=True)
+                if use_model:
+                    vis_data[pol] = this_dict["vis_model_ptr"]
+                else:
+                    vis_data[pol] = this_dict["vis_ptr"]
+                this_obs = this_dict["obs"]
+            self.Npols = len(list(vis_data.keys()))
```
plaplant, line 244: Is it possible for this determination of `Npols` to be different than the one above? Should we do that consistency check?

pyuvdata/uvdata/fhd.py
```diff
@@ -576,7 +522,101 @@ def read_fhd(
...
+        lin_pol_order = ["xx", "yy", "xy", "yx"]
+        linear_pol_dict = dict(zip(lin_pol_order, np.arange(5, 9) * -1))
+        pol_list = []
+        if read_data:
+            for pol in lin_pol_order:
+                if pol in vis_data:
+                    pol_list.append(linear_pol_dict[pol])
+            self.polarization_array = np.asarray(pol_list)
+        else:
+            # Use Npols because for FHD, npol fully specifies whih pols to use
```
plaplant, line 583: `whih` -> `which`

pyuvdata/uvdata/tests/test_fhd.py
```diff
@@ -314,74 +334,46 @@ def test_read_fhd_warnings():
...
-    assert str(cm.value).startswith("multiple layout files in filelist")
-    os.remove(extra_layout_file)
-
-    extra_settings_file = testdir + testfile_prefix + "extra_settings.txt"
-    copyfile(testfiles[7], extra_settings_file)
-    with pytest.raises(ValueError) as cm:
-        fhd_uv.read(testfiles + [extra_settings_file])
-    assert str(cm.value).startswith("multiple settings files in filelist")
-    os.remove(extra_settings_file)
+@pytest.mark.parametrize(
```
plaplant, line 343: Nice parameterization to make tests more atomic 👍

pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -4902,6 +4902,23 @@ def test_lsts_from_time_with_only_unique():
     assert np.array_equal(full_lsts, uv.lst_array)
 
 
+@pytest.mark.filterwarnings("ignore:Altitude is not present in Miriad file")
+def test_lsts_from_time_with_only_unique_background():
+    """
+    Test `set_lsts_from_time_array` with only unique values is identical to full array.
+    """
+    miriad_file = os.path.join(DATA_PATH, "zen.2456865.60537.xy.uvcRREAA")
```
plaplant, line 4910: More a note to myself: if this is merged in after #839, we'll want to use the corresponding uvh5 test file instead so we can still test this on Windows. (If this PR is merged first, I'll fix the other PR in my rebase).

pyuvdata/uvdata/uvdata.py
```diff
@@ -2209,59 +2235,77 @@ def unphase_to_drift(self, phase_frame=None, use_ant_pos=False):
...
+        if phase_frame == "gcrs":
+            frame_telescope_locations = itrs_telescope_locations.transform_to(
+                getattr(coord, f"{phase_frame}".upper())(obstime=obs_times)
+            )
+        else:
+            frame_telescope_locations = itrs_telescope_locations.transform_to(
+                getattr(coord, f"{phase_frame}".upper())
+            )
+
+        # astropy 2 vs 3 use a different keyword name
```
plaplant, line 2255: Since we're astropy 3+ only now, can we just use the v3 keyword?

pyuvdata/uvdata/uvdata.py
```diff
@@ -2391,25 +2435,35 @@ def phase(
...
+        # need to re-add obstimes for a GCRS transformation
+        if phase_frame == "gcrs":
+            frame_telescope_locations = itrs_telescope_locations.transform_to(
+                getattr(coord, f"{phase_frame}".upper())(obstime=obs_times)
+            )
+        else:
+            frame_telescope_locations = itrs_telescope_locations.transform_to(
+                getattr(coord, f"{phase_frame}".upper())
+            )
+        # astropy 2 vs 3 use a different keyword name
```
plaplant, line 2455: Same comment here.

pyuvdata/uvdata/tests/test_fhd.py
mkolopanis: Going for that 1k test mark!

pyuvdata/uvdata/fhd.py
plaplant: @bhazelton do you know the answer to this? I'm not familiar with FHD.

bhazelton: It can be. For FHD, each polarization is written to a different file. If you provide data files, the number of polarizations will be the number of files you provided. If you do not provide the data files then you can't count it that way and you need to use the number in the `obs` structure, which tells you how many pols the FHD run that produced it used (which is an upper limit on the number you could pass it).

bhazelton: I can add a comment to this effect in the docstring or in the code if you think I should.

plaplant: Thanks for the info. It might help to clarify that when we do a metadata-only read, we get the number of available pols (if I'm understanding correctly), whereas for actual data, it's related to the number of files passed in.

CHANGELOG.md
```diff
@@ -4,10 +4,12 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- `background` keyword in `uvdata.set_lsts_from_time_array` to allow lst calculation in a background thread.
```
plaplant, line 7: Just a quick note: related to this I think we should also mention that we've added the `background_lsts` keyword for the data-reading methods and it's defaulted to `True`.

pyuvdata/uvdata/fhd.py
bhazelton: ok, I just pushed a commit with a comment about this in the docstring.

plaplant: Looks great, thanks!

PR #839, opened 2020-05-17
Add support for Windows

### LINKED ISSUE #778: rename `all` and `vars` in aipy_extracts.UV
because they shadow python builtin functions

### LINKED ISSUE #823: Release a Windows version (minus miriad)
We would like to make a release of the package for Windows, in an effort to support all major platforms. However, even with the cython-based rewrite of miriad in #815, it seems compiling the extension on Windows is not straightforward. (The underlying C code uses the `unistd.h` header file, which does not exist on Windows.) Nevertheless, most other filetypes and modules should be able to be used. In principle it seems possible to make a release for Windows that treats miriad as an optional filetype, similar to how we currently handle measurement sets.

Once we have a version that is Windows-friendly, we should also add it to our CI. If any developers have access to Windows machines for testing, help with this would be much appreciated. I myself have only been running inside a virtual machine, which makes robust testing difficult.

plaplant:
## Description

This PR adds support for building pyuvdata on Windows. Miriad is not supported on Windows, because it uses functions in the  UNIX-specific `<unistd.h>` header file, which has no direct analogue on Windows machines. Most other functionality should be accessible, including reading/writing of UVH5 and UVFITS files, as well as calibration and beam files.

## Motivation and Context

As part of this PR, the handling of the Miriad extension is changed slightly. `pip` does not try to build the extension if it detects that the platform is Windows, but continues to build it for Mac and Linux. Importing of the `_miriad` module is done at the top of `aipy_extracts.py`, which treats the module as optional (much like how `casacore` is treated for measurement sets). Although it may be possible for Mac and Linux users to have pyuvdata installed and not be able to read/write Miriad files, this is likely to occur only if the user is being purposely evil, e.g., removing the compiled `.so` file from their pip installation.

As a part of these changes, many long tests in `test_beamfits.py` and `test_calfits.py` were made more atomic, as well as adding more explicit error message checking in tests. In some cases, a UVH5 copy of an existing Miriad test file was made so that as many tests as possible could be run on Windows. I also encountered some strange behavior where the FITS writer would not overwrite/append an existing file on Windows (which may possibly be related to https://github.com/astropy/astropy/issues/7404), and so in some cases I used two test files to avoid these issues.

Because we now support Windows, we may want to consider incrementing the major version and making a new PyPI/conda release in the near future.

As a part of this PR, we also fixed up the `aipy_extracts.py` module to be flake8-compliant.

Closes #823. Closes #778.

## Types of changes

- [ ] Version change

## Checklist:

- [x] If this is a new CI setup, I have added the associated badge to the readme and to references/make_index.py (if appropriate).


pyuvdata/uvbeam/beamfits.py
```diff
@@ -863,3 +863,4 @@ def write_beamfits(
         hdulist.append(bandpass_hdu)
 
         hdulist.writeto(filename, overwrite=clobber)
+        hdulist.close()
```
mkolopanis, line 866: Nice catch here.

pyuvdata/uvcal/calfits.py
```diff
@@ -390,6 +390,7 @@ def write_calfits(
             hdulist.append(totqualhdu)
 
         hdulist.writeto(filename, overwrite=clobber)
+        hdulist.close()
```
mkolopanis, line 393: also here :tada:

pyuvdata/uvcal/tests/test_calfits.py
```diff
@@ -278,8 +312,55 @@ def test_extra_keywords(tmp_path):
...
+    )
+    uvtest.checkWarnings(
+        cal_in.write_calfits,
+        [testfile],
+        {"run_check": False, "clobber": True},
+        message=["key test_long_key in extra_keywords is longer than 8 characters"],
+    )
+    cal_in.extra_keywords.pop("test_long_key")
+
+    return
```
mkolopanis, line 360: With all the dict-key `pop`ing going on here and adding extra keywords, this might be another reasonable case for a parametrize call.

pyuvdata/uvcal/tests/test_fhd_cal.py
```diff
@@ -37,8 +38,12 @@ def test_read_fhdcal_write_read_calfits(tmp_path):
     fhd_cal.write_calfits(outfile, clobber=True)
     calfits_cal.read_calfits(outfile)
     assert fhd_cal == calfits_cal
+    del fhd_cal, calfits_cal
+    gc.collect()
```
mkolopanis, line 42: hmm a little nervous about additional explicit `gc` calls just to make things work again. But glad it is working. Maybe we should make an issue in case we want to come back to this and investigate.

pyuvdata/uvdata/aipy_extracts.py
```diff
@@ -15,20 +15,11 @@
...
-    import os
-
-    if os.environ.get("PYUVDATA_IGNORE_EXTMOD_IMPORT_FAIL", "") != "1":
-        raise
-
-    class UV(object):
-        pass
-
-    _miriad = UV()  # eww gross but it works
-    _miriad.UV = UV
```
mkolopanis, line 31: we don't actually need this silly catch-reassigning anymore since we don't have that weird "need pyuvdata installed to install pyuvdata" _feature_ anymore right? Just concerned about removing this work around if it is still needed.

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -294,6 +294,8 @@ def test_wronglatlon():
 
 
 def test_miriad_location_handling(tmp_path):
+    from pyuvdata.uvdata import aipy_extracts
```
mkolopanis, line 297: do we want to just make this an `importorskip` too?

pyuvdata/uvbeam/beamfits.py
plaplant: This is thanks to @bhazelton 😄

pyuvdata/uvcal/tests/test_calfits.py
plaplant: Good point, I can atomize/parametrize this test.

pyuvdata/uvcal/tests/test_fhd_cal.py
plaplant: I think this is a vestige of a first-attempt that may not actually be needed. I'll reorganize and see if tests still pass.

pyuvdata/uvdata/aipy_extracts.py
plaplant: Yeah, before we did `import pyuvdata.version` in the setup.py to generate versioning info on install, but now we do it in a saner way. I'm pretty sure we can get rid of this now. (Empirically installation seems to work without it...)

pyuvdata/uvdata/tests/test_miriad.py
plaplant: Good point, I think I have an `importorskip("pyuvdata._miriad")` at the top, but I can switch to `pyuvdata.aipy_extracts`.

mkolopanis: I was just thinking having both honestly

mkolopanis: but the aipy_extracts would also include the `_miriad` wouldn't it?

plaplant: Yeah I could do it either way, whichever you think is best. Can't hurt to be doubly sure, so I can change these to be `importorskip`.

mkolopanis: My first naive thought was just two `importorskips` at the top. I would say just test it out with the single one for aipy extracts and if that works we can go with that.

pyuvdata/uvcal/tests/test_calfits.py
plaplant: This is done now.

pyuvdata/uvcal/tests/test_fhd_cal.py
plaplant: I've split tests up more now, so I took out the `gc` statements.

pyuvdata/uvdata/tests/test_miriad.py
plaplant: Based on [the example in the pytest documentation](https://docs.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency), I've now assigned the result of the `importorskip` statements to the module we then import from.

mkolopanis: Nice looks much cleaner this way.

PR #840, opened 2020-05-18
Additional Cython optimizations

mkolopanis:
Some under the hood cython additions.
## Description

Made some optimizations to (less than important functions):
- baseline_to_antnums
- antnums_to_baseline
- xyz and latlonalt conversion
- ECEF and ENU conversions (not rotECEF)
- UVW phasing functions
## Motivation and Context

Many utility functions relied on numpy vectorization support but could still be improved with C level implementations. This will provide some speed up for computations on large input sets I believe based on some of the attached testing. 

## Types of changes

- [x] super secret behind the scenes C code

## Checklist:

- [x] I have updated the Changelog

### DIFF
--- a/pyuvdata/tests/test_utils.py
+++ b/pyuvdata/tests/test_utils.py
@@ -2,9 +2,7 @@
 # Copyright (c) 2018 Radio Astronomy Software Group
 # Licensed under the 2-clause BSD License
 
-"""Tests for common utility functions.
-
-"""
+"""Tests for common utility functions."""
 import os
 import copy
 
@@ -240,7 +238,6 @@ def test_enu_from_ecef_shape_errors(enu_ecef_info, shape_type):
 
 def test_enu_from_ecef_magnitude_error(enu_ecef_info):
     """Test ENU_from_ECEF input magnitude errors."""
-
     (
         center_lat,
         center_lon,
--- a/pyuvdata/tests/test_utils.py
+++ b/pyuvdata/tests/test_utils.py
@@ -107,7 +107,9 @@ def test_LatLonAlt_from_XYZ():
     pytest.raises(ValueError, uvutils.LatLonAlt_from_XYZ, ref_xyz[0:1])
 
 
-def test_ENU_tofrom_ECEF():
+@pytest.fixture(scope="module")
+def enu_ecef_info():
+    """Some setup info for ENU/ECEF calculations."""
     center_lat = -30.7215261207 * np.pi / 180.0
     center_lon = 21.4283038269 * np.pi / 180.0
     center_alt = 1051.7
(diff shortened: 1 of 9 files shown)


mkolopanis:
Some timings I did for the baseline_to_antnums and antnums_to_baseline. A good speed up for larges number ~50k inputs, and the overhead seems minimal (little to no impact) for single inputs. 
[bls_and_antnums.pdf](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/4645631/cython_tests.pdf)

mkolopanis:
Some additional timings from the enu/ecef changes 
[xyz_calcs.pdf](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/4647150/xyz_calcs.pdf)

mkolopanis:
Phasing calculations
[phase_uvw.pdf](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/4651672/phase_uvw.pdf)

mkolopanis:
here are some phasing runs with 60k blts
[phase_uvw_many_blts.pdf](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/4652176/phase_uvw_many_blts.pdf)

bhazelton:
Wow, the speedup on phasing is really impressive, about an order of magnitude!

mkolopanis:
but still drops in the ocean compared to the lst calculations

pyuvdata/tests/test_utils.py
```diff
@@ -145,66 +145,224 @@ def test_ENU_tofrom_ECEF():
...
+        x,
+        y,
+        z,
+        east,
+        north,
+        up,
+    )
+
 
+def test_xyz_from_latlonalt(enu_ecef_info):
```
plaplant, line 164: The atomic testing is excellent 🎉

pyuvdata/utils.py
```diff
@@ -817,16 +796,7 @@ def XYZ_from_LatLonAlt(latitude, longitude, altitude):
             "latitude, longitude and altitude must all have the same length"
         )
 
-    # see wikipedia geodetic_datum and Datum transformations of
-    # GPS positions PDF in docs/references folder
-    gps_n = gps_a / np.sqrt(1 - e_squared * np.sin(latitude) ** 2)
-    xyz = np.zeros((n_pts, 3))
-    xyz[:, 0] = (gps_n + altitude) * np.cos(latitude) * np.cos(longitude)
```
plaplant, line 824: I know it's a little outside the scope of this PR, but these rotation transformations are "almost magical" to me, in the sense that it's not obvious why we're doing this particular rotation. There is [this memo in the docs](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/master/docs/references/GPS.G1-X-00006_Datum_Transformations.pdf) that explains a little bit, but we may want to consider having some documentation that explains the difference between ECEF/rotECEF/ENU/EIEIO, etc. It's not worth doing here, and instead I'll make a separate issue about it.

pyuvdata/utils.pyx
```diff
@@ -0,0 +1,306 @@
...
+  cdef numpy.ndarray[ndim=1, dtype=numpy.int64_t] ant2 = np.empty(n, dtype=np.int64)
+  cdef long _min = baseline.min()
+  cdef int i
+  # make views as c-contiguous arrays of a known dtype
+  # effectivly turns the numpy array into a c-array
+  cdef numpy.int64_t[::1] _a1 = ant1
+  cdef numpy.int64_t[::1] _a2 = ant2
+  cdef numpy.int64_t[::1] _bl = baseline
+
+  with nogil:
```
plaplant, line 44: It's my understanding that sometimes it's not worth the overhead of releasing the GIL, but I think the profiling you've done has demonstrated that we don't pay a penalty, even for small-sized arrays. (Btw, thanks for doing the detailed profiling! It's very nice 👍 )

pyuvdata/utils.pyx
```diff
@@ -0,0 +1,306 @@
...
+    with nogil:
+      for i in range(n_pts):
+        gps_n = _gps_a / sqrt(1.0 - _e2 * sin(_lat[i]) ** 2)
+
+        _xyz[i, 0] = (gps_n + _alt[i]) * cos(_lat[i]) * cos(_lon[i])
+        _xyz[i, 1] = (gps_n + _alt[i]) * cos(_lat[i]) * sin(_lon[i])
+
+        _xyz[i, 2] = (_gps_b ** 2 / _gps_a ** 2 * gps_n + _alt[i]) * sin(_lat[i])
+
+    return xyz.squeeze()
```
plaplant, line 180: Is the `squeeze` needed here in case single numbers (as opposed to arrays) were passed in?

pyuvdata/utils.pyx
```diff
@@ -0,0 +1,306 @@
...
+    numpy.float64_t[::1] _alt,
+):
+  cdef int i
+  cdef int nblts = xyz.shape[0]
+  cdef numpy.ndarray[dtype=numpy.float64_t, ndim=2] enu = np.empty((nblts, 3), dtype=np.float64)
+
+  cdef numpy.ndarray[dtype=numpy.float64_t, ndim=1] xyz_center = _xyz_from_latlonalt(_lat, _lon, _alt)
+  cdef numpy.ndarray[dtype=numpy.float64_t, ndim=1] xyz_use = np.empty(3, dtype=np.float64)
+
+  cdef numpy.float64_t[:, ::1] _enu = enu
```
plaplant, line 197: If this is a memoryview as well, we should probably add a small comment as above.

plaplant (commented):
Very nice addition @mkolopanis! The timing information in particular is excellent (and thanks for all the speedups!). I just had a couple of small questions.

pyuvdata/utils.pyx
mkolopanis: We don't have to release it too. :thinking: Up to you. I think I agree that the penalty we pay isn't enough to make even single calls slower than the current implementation but I'm not married to nogil.

pyuvdata/utils.pyx
mkolopanis: I think so? This was definitely here in the old implementation and I believe it is for the case you're describing.

pyuvdata/utils.pyx
mkolopanis: i will add a comment. any `dtype_identifier[ slice indexing along dimensions]` syntax is a memory view. I'll make comments.

pyuvdata/utils.py
plaplant: I've now made #845 to track this issue.

pyuvdata/utils.pyx
plaplant: I think it's fine to keep in (b/c the profiling proves there isn't a significant penalty), just an idle musing I had.

pyuvdata/utils.pyx
plaplant: Sound good, thanks for clarifying.

mkolopanis:
Whenever you're ready for another look @plaplant no rush

plaplant (dismissed):
Thanks for adding the comments! This looks good to me.

mkolopanis:
I'm sorry :frowning: I found some spacing I didn't like.

plaplant (approved):

Issue #843, opened 2020-05-20
UVFlag add method


mwilensky768:
Would it be possible to make it so that when calling `UVFlag.__add__`, it iterates over the `_data_params` property, rather than only the `metric_array`, `weights_array`, or `flag_array` attributes? See this [line](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/609e6c9ffc719fc1f2882772058b61e47f38b5f8/pyuvdata/uvflag/uvflag.py#L1557).

This is important if people are using the optional `weights_square_array`, or if people subclass `UVFlag` and create additional `_data_params` (SSINS does this [here](https://github.com/mwilensky768/SSINS/blob/836716f424fbdcffef0c6c583ac41d66bed30a17/SSINS/incoherent_noise_spectrum.py#L419).)

mwilensky768:
I would do this with someone's blessing.

mkolopanis:
Yeah that seems like a reasonable change I think. I would bet this line is carry over from before UVFlag subclassed UVBase. As long as it does not affect the current behavior, it sounds like a good change.

mkolopanis:
did #844 close this?

mwilensky768:
Yes.

Issue #653, opened 2019-08-13
Make UVFlag add smarter


adampbeardsley:
Right now the UVFlag add method requires an axis to be specified and it blindly concatenates. We should make it smarter to check for compatibility and figure out which axis to add along, like UVData. Similar to UVData, we should also add a fast_concat method.

bhazelton:
Generally need to add more checking as is done with the other objects. In particular, make sure the right thing happens in the case of overlapping and differently sorted data.

see https://github.com/RadioAstronomySoftwareGroup/pyuvdata/issues/1102 and https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1105 for reference

PR #844, opened 2020-05-20
Make UVFlag.__add__ iterate over _data_params

mwilensky768:
This PR ensures that all data-like parameters are concatenated appropriately during addition of two UVFlag objects.

## Description

The data-like attributes to be concatenated were hardcoded in `__add__`. Now an iteration over `UVFlag._data_params` catches all of them, including the optional `weights_square_array`, if present.

## Motivation and Context

This enhancement ensures concatenation of the optional parameter `weights_square_array` if it exists, and is also a little friendlier to people subclassing UVFlag. See issue #843 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

New feature checklist:

Breaking change checklist:

Version change checklist:

### DIFF
--- a/pyuvdata/uvflag/uvflag.py
+++ b/pyuvdata/uvflag/uvflag.py
@@ -559,7 +559,8 @@ class UVFlag(UVBase):
         self._flag_array.required = True
         self._metric_array.required = False
         self._weights_array.required = False
-        self._weights_square_array.required = False
+        if self.weights_square_array is not None:
+            self.weights_square_array = None
 
         return
 
@@ -739,6 +740,7 @@ class UVFlag(UVBase):
                 not attr.required
                 and attr.value is not None
                 and attr.name != "x_orientation"
+                and attr.name != "weights_square_array"
             ):
                 attr.value = None
                 setattr(self, p, attr)
@@ -1005,9 +1007,6 @@ class UVFlag(UVBase):
                 self.weights_array = w
                 if return_weights_square:
                     self.weights_square_array = ws
-                    # Once this is calculated it needs to be required
-                    # until switch out of metric mode
-                    self._weights_square_array.required = True
             self.time_array, ri = np.unique(self.time_array, return_index=True)
             self.lst_array = self.lst_array[ri]
         if ((method == "or") or (method == "and")) and (self.mode == "flag"):
@@ -2541,8 +2540,6 @@ class UVFlag(UVBase):
                     self.weights_array = dgrp["weights_array"][()]
                     if "weights_square_array" in dgrp:
                         self.weights_square_array = dgrp["weights_square_array"][()]
-                        # Needs to be required now that it exists and is non-null
-                        self._weights_square_array.required = True
                 elif self.mode == "flag":
                     self.flag_array = dgrp["flag_array"][()]
 
--- a/pyuvdata/uvflag/uvflag.py
+++ b/pyuvdata/uvflag/uvflag.py
@@ -1550,16 +1550,13 @@ class UVFlag(UVBase):
             )
             this.Npols = len(this.polarization_array)
 
-        if this.mode == "flag":
-            this.flag_array = np.concatenate(
-                [this.flag_array, other.flag_array], axis=ax
-            )
-        elif this.mode == "metric":
-            this.metric_array = np.concatenate(
-                [this.metric_array, other.metric_array], axis=ax
-            )
-            this.weights_array = np.concatenate(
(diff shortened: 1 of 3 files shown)


mwilensky768:
Commit names look silly because I didn't understand how pre-commit works. Also wondering if we should tackle #653 while we're at it.

mkolopanis:
The issue you mentioned is hefty but if you feel up to it I would be glad to finally close it.

mwilensky768:
Well, the branch name is appropriate, but I realize I'd like this change merged in ~soon. Maybe we can do a quick merge now and work on the next part afterward?

mkolopanis:
Just want to pop in here. This will require a rebase now after the windows PR got merged in.

mwilensky768:
Rebase complete

pyuvdata/uvflag/uvflag.py
```diff
@@ -1550,16 +1550,13 @@ def __add__(
...
-                [this.flag_array, other.flag_array], axis=ax
-            )
-        elif this.mode == "metric":
-            this.metric_array = np.concatenate(
-                [this.metric_array, other.metric_array], axis=ax
-            )
-            this.weights_array = np.concatenate(
-                [this.weights_array, other.weights_array], axis=ax
+        # Will fail with AttributeError if the objects have different _data_params
+        # Might want a more straight-forward check ahead of time
```
mkolopanis, line 1554: This comment makes me a little nervous. I would prefer if the check was added as a part of this PR. We already know they must be the same `mode` and `type` which, as long as the user hasn't altered the object's state, forces the required parameters to be the same. So this may not actually be an issue for required parameters.

The other question is what to do if the optional parameters exist on one and not the other? 
Erroring seems extreme since it is an optional parameter. We could throw a warning, and then fill in the missing data with zeros? That seems a little weird though too but not sure what the best solution is.

mwilensky768: I think filling in a `weights_square_array` with zeros will cause problems down the line in the application I have in mind. I guess if this attribute was vestigial after a type conversion (metric to flag), and you were adding flag objects, maybe you would not want an error. Could be a warning in one case and an error in the other? Of course, as additional optional parameters are developed, we may have to rethink this line of reasoning.

mkolopanis: Yeah it definitely gets complicated :thinking: I'm okay with the warning/error differential depending on cases. But we'll still need to eloquently handle when parameters exist on one object and not the other. Maybe it just makes them incompatible in the case where they are not vestigial?

mwilensky768: I think that's a good start. I don't know what the best check is. Not sure how you feel about try/except. One way would be

```
for attr in this._data_params:
    try:
        setattr(this, attr, np.concatenate(...))
    except AttributeError:
        vestigial = some_check_for_vestigial (mode=='flag' for now?)
        if vestigial:
            warning
        else:
            raise ValueError("UVFlag objects are not compatible for addition")
```

mkolopanis: I think that could be a good start. I'm not a huge fan of the try/except (despite it being the _pythonic_ way). But could just have an `if not hasattr()` instead.

mwilensky768: Error message could of course be more verbose

mwilensky768: Oh, yeah `if not hasattr` is definitely simpler/fewer lines/easier to test

mwilensky768:
I've now added an attribute check that complains if `this` has a non-null-valued attribute where `other` has a null-valued one. Additionally, in order that the object passes `UVBase.check` at the end of `__add__`, I've made it so that when the `weights_square_array` is calculated, it becomes required, and then is forced to be unrequired if `mode` is switched to `flag`.

adampbeardsley:
Sorry it took so long for me to look at this. Overall it looks good, thanks for putting in the effort Mike!
My only concern is around switching whether `weights_square_array` is required. I'm not sure the `required` attribute is meant to change based on whether it exists (admittedly we change it based on the `mode`, but I think that's more akin to subclassing, which we probably should have done).
I suggest we add the line:
```
and attr.name != "weights_square_array"
```
back into `clear_unused_attributes`, and ensure any `weights_square_array` is removed when changing to flag mode.
Is that possible, and will it be sufficient for what you want?

mwilensky768:
@adampbeardsley I believe my newest commit accomplishes everyone's goals.

PR #859, opened 2020-06-04
Make sure all tests are run in CI on Windows

plaplant:
## Description
When we added Windows to CI, we noticed that not all tests were being run for python3.6 and 3.7. It seemed that only tests in the top-level `pyuvdata/tests` folder were being run, and it was not discovering tests inside of submodules. ~This PR changes how tests are invoked, making sure that all tests are run for all python versions.~ This PR changes how the package in installed, to make sure all recursive test directories are discovered.

## Motivation and Context

~It seemed odd that test discovery was behaving differently based on the python version (and only on Windows), especially because the version of `pytest` was the same. Also, previously we were running tests by using the `pytest --pyargs pyuvdata` construction in a separate folder, which may have affected how tests were discovered. Running `pytest` in the root directory seems to have fixed things.~
I think this issue came down to how the package was being handled in `setup.py`. Previously, we were explicitly listing the submodules to install, which included `pyuvdata.tests` but did _not_ include the submodule test directories (e.g., `pyuvdata.uvdata.tests`). Instead of explicitly listing packages, we switched to using the `find_namespace_packages()` function from setuptools, and that seems to have fixed the issue for all versions on all platforms.

## Types of changes

- [ ] Version change

## Checklist:

### DIFF
--- a/.github/workflows/macosx_windows_ci.yaml
+++ b/.github/workflows/macosx_windows_ci.yaml
@@ -60,7 +60,7 @@ jobs:
           mkdir -p empty
           cd empty
           source activate $ENV_NAME
-          pytest --pyargs pyuvdata --cov=pyuvdata --cov-config=../.coveragerc --cov-report xml:../coverage.xml
+          pytest --pyargs pyuvdata -c ../setup.cfg --cov=pyuvdata --cov-config=../.coveragerc --cov-report xml:../coverage.xml
           cd ..
       - name: Upload Coverage
         shell: bash
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ import os
 import io
 import sys
 import platform
-from setuptools import setup, Extension
+from setuptools import setup, Extension, find_namespace_packages
 
 from Cython.Distutils import build_ext
 from distutils.sysconfig import get_config_var
@@ -63,6 +63,12 @@ if is_platform_mac():
         if python_target < "10.9" and current_system >= "10.9":
             os.environ["MACOSX_DEPLOYMENT_TARGET"] = "10.9"
 
+# define the cython compile args, depending on platform
+if is_platform_windows():
+    extra_compile_args = ["/Ox"]
+else:
+    extra_compile_args = ["-O3"]
+
 global_c_macros = [
     ("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION"),
 ]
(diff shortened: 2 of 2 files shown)


.github/workflows/macosx_windows_ci.yaml
```diff
@@ -53,15 +53,12 @@ jobs:
...
+          python setup.py build_ext --inplace --force
       - name: Run Tests
         shell: bash
         run: |
-          mkdir -p empty
-          cd empty
           source activate $ENV_NAME
-          pytest --pyargs pyuvdata --cov=pyuvdata --cov-config=../.coveragerc --cov-report xml:../coverage.xml
-          cd ..
+          pytest --cov=pyuvdata --cov-config=.coveragerc --cov-report xml:coverage.xml
```
bhazelton, line 61: I want to make sure that we aren't undoing something by changing this setup. Presumably it was set up this way for a good reason, it's fine if that's no longer applicable but I want to make sure that's the case. I think Matt said it was @steven-murray who set it up this way, maybe he can comment?

steven-murray: It's hard to remember exactly, but I do remember having a lot of issues with running tests properly due to the structure of the repo. Since the code is not in a non-python-package `src/` directory, running `pytest` at top level would not run the _installed_ code, but rather the code directly from the repo. This can reasonably often cause problems.

I can't remember, but it may be that the `--pyargs` was a way to circumvent this, by forcing it to use the installed package rather than the repo. So I would be very careful in removing it.

bhazelton:
I guess the question is, in the context of CI builds where what we are trying to do is install the code from the repo and test it, is it important that it calls the installed package as opposed to the code in the directory?

.github/workflows/macosx_windows_ci.yaml
mkolopanis: It might be that the package is not properly installing tests then with their nested directory structure?

mkolopanis: Ah I may have an idea. It involved the `packages` keyword in the setuptools configuration. We could change to using `find_packages()` for `find_namespace_packages()` which are both imported from setuptools. These are similar but different tools to comb the pacakge subdirectories and install files it finds. It is also possible to list paths to exclude in their calls. But if the issue is that the tests are not getting installed properly it may fix that. Anecdotally it worked on a local install. I monitored my installed packages in the `site-packages` of a conda env and verified that no tests beyond the initial `tests` folder were installed before but making this edit caused all files and tests to be installed.

mkolopanis: But not sure if this preserves the pytest args set in setup.cfg then :thinking:

mkolopanis: ah the `-c` options allows us to point to a config file.

steven-murray:
@bhazelton  yes, it can be very important, especially when there are things like data files and relative paths to them. Sometimes the code in your repo will work, but if the files aren't installed properly it will break, but you'll miss it.

There are also other potential problems that I've encountered but can't recall right now.

@mkolopanis so does your solution actually install the package and test that then?

mkolopanis:
When looking on my personal machine, my proposed solution installed all the tests and files into the conda environment I experimented on and running tests in the non-source directory ran all tests.

steven-murray:
Awesome! As long as that is working, I approve.

plaplant:
@bhazelton sure, I can update the text. I was also going to squash these commits into a single one because I meandered about a bit before arriving at the proper solution.

PR #860, opened 2020-06-05
Improve UVParameter equality method for Quantity types

bhazelton:
## Description

Use `astropy.units.quantity.allclose` rather than `np.allclose` for checking the equality of Quantity parameters.

## Motivation and Context

This is mostly helpful for pyradiosky because SkyModel uses astropy Quantities for many of its parameters.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/parameter.py
+++ b/pyuvdata/parameter.py
@@ -121,11 +121,15 @@ class UVParameter(object):
                             "units are not equivalent"
                         )
                         return False
+                    if not isinstance(self.tols[1], units.Quantity):
+                        atol_use = self.tols[1] * self.value.unit
+                    else:
+                        atol_use = self.tols[1]
                     if not units.quantity.allclose(
                         self.value,
                         other.value,
                         rtol=self.tols[0],
-                        atol=self.tols[1],
+                        atol=atol_use,
                         equal_nan=True,
                     ):
                         print(
--- a/pyuvdata/parameter.py
+++ b/pyuvdata/parameter.py
@@ -105,7 +105,7 @@ class UVParameter(object):
             if not isinstance(self.value, other.value.__class__):
                 print(
                     f"{self.name} parameter value classes are different. Left is "
-                    f"{self.value.__class__}, right is {self.value.__class__}"
+                    f"{self.value.__class__}, right is {other.value.__class__}"
                 )
                 return False
             if isinstance(self.value, np.ndarray) and not isinstance(
--- a/pyuvdata/parameter.py
+++ b/pyuvdata/parameter.py
@@ -12,6 +12,7 @@ of metadata.
 
 """
 import numpy as np
+import astropy.units as units
 
 from . import utils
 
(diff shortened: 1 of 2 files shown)


pyuvdata/parameter.py
```diff
@@ -103,34 +104,47 @@ def __eq__(self, other):
...
             if not isinstance(self.value, other.value.__class__):
                 print(
-                    "{name} parameter value classes are different. Left is "
-                    "{lclass}, right is {rclass}".format(
-                        name=self.name,
-                        lclass=self.value.__class__,
-                        rclass=other.value.__class__,
-                    )
+                    f"{self.name} parameter value classes are different. Left is "
+                    f"{self.value.__class__}, right is {self.value.__class__}"
```
plaplant, line 108: I think this should be
```suggestion
                    f"{self.value.__class__}, right is {other.value.__class__}"
```

pyuvdata/parameter.py
```diff
@@ -103,34 +104,47 @@ def __eq__(self, other):
             # only check that value is identical
             if not isinstance(self.value, other.value.__class__):
                 print(
```
plaplant, line 106: Just in general, do we want these to be print statements? Should they instead be warnings? I'm thinking that these equality checks are typically done inside of a `.check()` call rather than explicitly invoked by the user, so that might help raise awareness that there's something wrong. We can leave them as prints, just a thought I had.

pyuvdata/tests/test_parameter.py
```diff
@@ -42,6 +43,62 @@ def test_array_equality_nans():
     assert param1 == param2
 
 
+@pytest.mark.parametrize("atol", [0.001, 1 * units.mm])
+def test_quantity_equality(atol):
```
plaplant, line 47: I think this is a great test, but I think we should also try using values that are not exactly equal, but within the tolerance. (E.g., send in 1001 mm or 2999 mm). Then for completeness/paranoia we can use values that are just outside of the tolerance, and make sure those get flagged as not equal.

pyuvdata/tests/test_parameter.py
```diff
@@ -42,6 +43,62 @@ def test_array_equality_nans():
...
+        name="p3", value=np.array([0, 1, 3]) * units.mm, tols=1 * units.mm
+    )
+    assert param1 != param3
+    param4 = uvp.UVParameter(
+        name="p4", value=np.array([0, 1, 3]) * units.Jy, tols=1 * units.mJy
+    )
+    assert param1 != param4
+
+
+def test_quantity_equality_nans():
```
plaplant, line 95: Nice test 👍

pyuvdata/parameter.py
bhazelton: yes, thanks for catching that!

pyuvdata/parameter.py
bhazelton: yeah, I've been wondering that for awhile. I think that's a big change and we should do it in another PR if we want to. Sometimes we expect non-equality, so a warning might be a little weird...

pyuvdata/tests/test_parameter.py
bhazelton: good idea

pyuvdata/parameter.py
bhazelton: These aren't actually called in the `check` method, they're called if you test objects for equality.

plaplant: Ah good point. And true, if we decide to change it we should do it in a different PR.

PR #867, opened 2020-06-24
Add a workflow to publish releases on pypi

### LINKED ISSUE #647: set up pypi deployments from ci based on github releases
e.g. https://circleci.com/blog/continuously-deploying-python-packages-to-pypi-with-circleci/

bhazelton:
## Description

Use a GitHub Action to publish releases on PyPI and to publish all pushes on TestPyPI to make sure things are working.

## Motivation and Context

closes #647 

## Types of changes

- [ ] Version change

## Checklist:

### DIFF
--- a/.github/workflows/publish-to-test-pypi.yaml
+++ b/.github/workflows/publish-to-test-pypi.yaml
@@ -43,7 +43,7 @@ jobs:
       shell: bash
       run: |
         source activate $ENV_NAME
-        python -m pep517.build --source --binary --out-dir dist/ .
+        python -m pep517.build --source --out-dir dist/ .
 
     - name: Publish to Test PyPI
       uses: pypa/gh-action-pypi-publish@master
--- a/.github/workflows/publish-to-test-pypi.yaml
+++ b/.github/workflows/publish-to-test-pypi.yaml
@@ -8,6 +8,8 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@master
+      with:
+        fetch-depth: 0
     - name: Set up Python 3.7
       uses: actions/setup-python@v1
       with:
--- a/.github/workflows/publish-to-test-pypi.yaml
+++ b/.github/workflows/publish-to-test-pypi.yaml
@@ -25,6 +25,7 @@ jobs:
         --source
         --binary
         --out-dir dist/
+        .
     - name: Publish to Test PyPI
       uses: pypa/gh-action-pypi-publish@master
       with:
--- a/.github/workflows/publish-to-test-pypi.yaml
+++ b/.github/workflows/publish-to-test-pypi.yaml
@@ -5,34 +5,52 @@ on: push
 jobs:
   build-n-publish:
     name: Build and publish to PyPI and TestPyPI
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
+    env:
+      ENV_NAME: pyuvdata_publish
+      PYTHON: 3.7
     steps:
     - uses: actions/checkout@master
       with:
         fetch-depth: 0
-    - name: Set up Python 3.7
-      uses: actions/setup-python@v1
-      with:
-        python-version: 3.7
-    - name: Install pep517
-      run: >-
-        python -m
-        pip install
-        pep517
-        --user
+
+    - name: Get Miniconda Linux
+      run: |
+        wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O $HOME/miniconda.sh;
+        bash $HOME/miniconda.sh -b -p $HOME/miniconda
+
+    - name: Setup paths
+      shell: bash
+      run: |
+        echo "::add-path::$HOME/miniconda/bin"
+        echo "::add-path::$HOME/miniconda/envs/test/bin"
+
+    - name: Setup Environment
+      run: ./ci/install_conda.sh
+
+    - name: Create environment variable
+      if: startsWith(github.ref, 'refs/tags/') != true
+      shell: bash
+      run: |
+        source activate $ENV_NAME
+        version=$(awk -F+ '{print $1}' <<< $(python setup.py --version | tail -1))
+        echo "::set-env name=SETUPTOOLS_SCM_PRETEND_VERSION::$version"
+
+    - name: Check environment variable
+      run: echo $SETUPTOOLS_SCM_PRETEND_VERSION
+
     - name: Build a binary wheel and a source tarball
-      run: >-
-        python -m
-        pep517.build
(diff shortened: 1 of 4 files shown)


bhazelton:
@steven-murray I have a question for you. The template I used for this workflow suggests uploading to the TestPyPI site with every push to make sure things are working generally in addition to uploading to the real PyPI site when tags are created. Unfortunately this doesn't work with the version numbers generated by `setuptools_scm` because the `+` in their names indicates that they are local versions and TestPyPI rejects them because it won't accept local versions. This has affected other `setuptools_scm` users (https://github.com/pypa/setuptools_scm/issues/342) and they solved it by removing the `+`. However, the `+` does seem to be "correct" in a certain sense. Do you think taking out the `+` is worth it to get the TestPyPI CI on every push? Or should we just run the part that builds the package but not attempt to push to either site unless there's a tag?

steven-murray:
The way I handled this in my own code (https://github.com/steven-murray/hmf)  was to wimp out and not use a local scheme, but this does lose some benefits. I think uploading to TestPyPI on every push is super useful, but it also has some drawbacks (eg. I've found it sometimes fails because two different open branches can have the same version number when you don't include the local-scheme). If there's a way to just prune the local-scheme to make it compatible with `PyPI` while keeping it unique I think that's the best solution.

bhazelton:
Ok, thanks. I think there's a way to do that (by replacing the `+` in our branch_scheme with a dot).

bhazelton:
@steven-murray hmm, my simple fix didn't work. Now I'm getting this error:
```
'2.0.3.dev174.gbc9d796e.publish-pypi' is an invalid value for Version. Error: Start and end with a letter or numeral containing only ASCII numeric and '.', '_' and '-'. See https://packaging.python.org/specifications/core-metadata for more information.
```
I guess I need to add a dot, dash or underscore at the end?

steven-murray:
Hmmm... not sure: https://www.python.org/dev/peps/pep-0440/ I wonder if any
part of the local scheme is allowed?

bhazelton:
yep, no dice. I'm going to undo all my changes to branch scheme and fall back on not uploading to testpypi but doing the build continually because I'm not willing to abandon the local scheme. Unless you have another suggestion that will keep the git hash and branch info in the version? Or if there's a way to modify this only for the testpypi build?

bhazelton:
@steven-murray maybe there's a way using the "SETUPTOOLS_SCM_PRETEND_VERSION" environment variable?
https://github.com/pypa/setuptools_scm/tree/v3.4.0#environment-variables

steven-murray:
Instead of using a python script to extract the version, why not use bash directly? Something like this:

```bash
version = $(python setup.py --version)
readarray -d + -t strarr <<< "$version"
SETUPTOOLS_SCM_PRETEND_VERSION = ${strarr[0]}
```

bhazelton:
Thank you @steven-murray. I had concluded that's what I needed to do but hadn't worked out the bash to do it yet.

bhazelton:
@steven-murray it's almost working, but the `python setup.py --version` call gives several output lines about cython stuff before it prints the version, so the following lines don't quite work. I can't quite test it on my mac because the `readarray` command doesn't work on macs. I'll keep working on it, but if you have insights I'd be happy to have them!

steven-murray:
Oh, shoot! Why is it printing out stuff about Cython?

steven-murray:
I think this will work:

```bash
SETUPTOOLS_SCM_PRETEND_VERSION=$(awk -F+ '{print $1}' <<< $(python setup.py --version | tail -1))
```

bhazelton:
@steven-murray it worked! Now I have a new problem: The wheel has a linux platform specific tag which is apparently not allowed by pypi. I think in the past I've just uploaded the sdist tar file but this workflow (that I copied from pypi documentation) is building a wheel as well and the testpypi site is rejecting it. The error is:
`Binary wheel 'pyuvdata-2.0.3.dev174-cp37-cp37m-linux_x86_64.whl' has an unsupported platform tag 'linux_x86_64'.`

Do you know how to make it not have that tag or should I just not make the wheel? My googling is not helping so far...

bhazelton:
I turned off building wheels and it worked!

bhazelton:
@pkgw It looks like we're having a problem with python-casacore on Mac OS. I think it's something we saw in the past on a previous version, it's causing errors like:
`TypeError: Expecting an object of type dict; got an object of type dict instead`
In that instance, it was an issue with which libraries were linked in the conda build I think?

I think this just started because a new version of casacore and python-casacore just started being used by our CI. I first saw it on this PR and I've since reproduced it on my own machine. It's only a problem on Mac, Linux seems to be fine.

Do you have any suggestions about how to fix this?

pkgw:
@bhazelton Hmm, it's not immediately obvious to me where the problem is coming from, but it looks like the conda-forge packages might need a bit of attention so I'll try to iterate them and we can see if that helps. Cf: https://github.com/conda-forge/casacore-feedstock/pull/45 to start.

If you can identify a test case that reproduces the error using just the conda-forge `casacore` or `python-casacore` packages, and you can file a GitHub issue on the respective `{python-}casacore-feedstock` repositories, that will be a better place to diagnose the problem and track it down.

steven-murray:
@bhazelton yeah I think for codes that aren't pure-python, you can't build wheels (or maybe it's just very hard to do it properly)

pkgw:
@steven-murray I believe that binary Linux wheels now need to be built with the "manylinux" framework, cf [pypa/manylinux](https://github.com/pypa/manylinux).

@bhazelton I think I've fixed the casacore macOS problem as of conda-forge/casacore-feedstock#46. If you can try a CI run that manages to install a `*_2` conda-forge build of casacore 3.3.0, I think it should work.

bhazelton:
@pkgw looks like it's fixed now, thank you!!!

steven-murray:
Just putting it out there that there's also the potential to automate the tagging via workflows. If you concede that everything that is pulled into master is ready to be used (as it's already tested etc), you can set up auto-tagging based on commit messages or something similar, and then essentially each PR auto creates a new PyPI/conda version. 

Downside is you don't get to control quite as well what things end up in what version (i.e. you can't lump different PRs into one release as easily). But if you think it's useful, I've kind of done it before.

bhazelton:
@plaplant yes, thanks for the reminder, I will update the notes.

@steven-murray I think that's a good thing to think about. I'm not quite ready to make that leap without more discussion, but maybe we can talk about it more in our next monthly telecon, which is July 15th. Can you make that one?

steven-murray:
Should be able to!

bhazelton:
@plaplant I updated the release instructions, so I think this is ready for another look when you have time.

PR #873, opened 2020-07-03
Add more robust error catching for skip_bad_files feature

dstorer:
## Description
Add functionality so that when reading a list of multiple files with skip_bad_files=True more types of exceptions can be caught and skipped. Previously it was only catching KeyErrors, now it also catches ValueError and OSError. 

## Motivation and Context
HERA files were being observed causing ValueErrors and OSErrors when being read in, and we wanted functionality to read long lists of files without the read being killed by a couple of corrupt files.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/uvdata/tests/test_uvdata.py
+++ b/pyuvdata/uvdata/tests/test_uvdata.py
@@ -7950,7 +7950,7 @@ def test_multifile_read_check(hera_uvh5, tmp_path):
 
 
 @pytest.mark.parametrize("err_type", ["KeyError", "ValueError"])
-def test_multifile_read_check_keyerror(hera_uvh5, tmp_path, err_type):
+def test_multifile_read_check_long_list(hera_uvh5, tmp_path, err_type):
     """
     Test KeyError catching by setting skip_bad_files=True when
     reading in files for a list of length >2
--- a/pyuvdata/uvdata/tests/test_uvdata.py
+++ b/pyuvdata/uvdata/tests/test_uvdata.py
@@ -8016,6 +8016,21 @@ def test_multifile_read_check_long_list(hera_uvh5, tmp_path, err_type):
 
     assert uvTest == uvTrue
 
+    # Test with corrupted file first in list, but with skip_bad_files=False
+    uvTest = UVData()
+    if err_type == "KeyError":
+        with pytest.raises(KeyError, match="Unable to open object"):
+            with pytest.warns(UserWarning, match="Failed to read"):
+                uvTest.read(fileList[0:4], skip_bad_files=False)
+    elif err_type == "ValueError":
+        with pytest.raises(ValueError, match="Nants_data must be equal to"):
+            with pytest.warns(UserWarning, match="Failed to read"):
+                uvTest.read(fileList[0:4], skip_bad_files=False)
+    uvTrue = UVData()
+    uvTrue.read([fileList[1], fileList[2], fileList[3]], skip_bad_files=False)
+
+    assert uvTest != uvTrue
+
     # Repeat above test, but with corrupted file in the middle of the list
     os.remove(fileList[0])
     uv2 = uv.select(times=np.unique(uv.time_array)[0:4], inplace=False)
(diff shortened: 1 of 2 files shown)


pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -7725,6 +7725,85 @@ def test_multifile_read_check_long_list(tmp_path):
 
     return
 
+def test_multifile_read_check_ValueError(tmp_path):
```
bhazelton, line 7728: This looks almost the same as the KeyError test, might be a good candidate for a `pytest.mark.parametrize`?

pyuvdata/uvdata/uvdata.py
```diff
@@ -7010,8 +7011,33 @@ def read(
                         background_lsts=background_lsts,
                     )
                     unread = False
-                except KeyError:
-                    warnings.warn("Failed to read {f}".format(f=filename[file_num]))
+                except KeyError as err:
+                    file_warnings = (
+                        file_warnings
+                        + "Failed to read {f} due to KeyError: {e}\n".format(
```
bhazelton, line 7017: Not a big deal, but this would be shorter if you use f-strings.

bhazelton:
@dstorer I think you need to rebase this to get an accurate coverage report.

pyuvdata/uvdata/uvdata.py
dstorer: Good suggestion - I did that!

pyuvdata/uvdata/tests/test_uvdata.py
dstorer: I'm not sure it would totally make sense, because what I'm doing to the test files to break them is different for each error type, but I can play around with trying it more if you'd like.

bhazelton: It's up to you, and it's just about reducing duplicated code. You can use an if statement to things slightly differently for each case if most of the code is the same. But if they're really different it might not be worth the effort.

dstorer: I went ahead and did it - you were right that it made things more efficient.

pyuvdata/uvdata/uvdata.py
```diff
@@ -7145,8 +7147,25 @@ def read(
...
+                    if skip_bad_files is False:
+                        raise
+                except ValueError as err:
+                    file_warnings = (
+                        file_warnings + f"Failed to read {f} due to ValueError: {err}\n"
+                    )
+                    file_num += 1
+                    if skip_bad_files is False:
+                        raise
+                # pragma: nocover
```
bhazelton, line 7164: This needs to be on the next line just after the colon to work properly (it tells the coverage calculator to ignore that line, and since it's an except block, to also ignore the lines inside the block).

pyuvdata/uvdata/uvdata.py
```diff
@@ -7194,8 +7213,30 @@ def read(
...
+                    except ValueError as err:
+                        file_warnings = (
+                            file_warnings
+                            + f"Failed to read {f} due to ValueError: {err}\n"
+                        )
+                        if skip_bad_files:
+                            continue
+                        else:
+                            raise
+                    # pragma: nocover
```
bhazelton, line 7234: Same thing here.

dstorer: Gotcha, my mistake. Just fixed it.

pyuvdata/uvdata/uvdata.py
```diff
@@ -7145,8 +7147,24 @@ def read(
...
+                    file_num += 1
+                    if skip_bad_files is False:
+                        raise
+                except ValueError as err:
+                    file_warnings = (
+                        file_warnings + f"Failed to read {f} due to ValueError: {err}\n"
+                    )
+                    file_num += 1
+                    if skip_bad_files is False:
+                        raise
```
bhazelton, line 7163: It looks like this line is not being covered by tests.

PR #875, opened 2020-07-07
Add ant string parsing to UVFlag

### LINKED ISSUE #841: ant_str keyword for UVFlag.select
The `UVFlag.select` method does not have an `ant_str` keyword, which is an extremely convenient one for sorting autos from crosses. It would be nice to have this keyword.

mkolopanis:
Adds `parse_ants` support to UVFlag select.
## Description

Moved `parse_ants` to a utility function to allow other UVBase subclasses besides UVData to call it if desired.
## Motivation and Context

closes #841 
## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:


mkolopanis:
I can add the changelog once we are ready. Not sure if there was more we wanted from this, or additional testing.

bhazelton:
@mwilensky768 I think this is solving an issue you made, care to comment?

pyuvdata/uvflag/tests/test_uvflag.py
```diff
@@ -2795,6 +2795,50 @@ def test_select(input_uvf, uvf_mode):
...
+            "There is no data matching ant_str=auto in this object.",
+        ),
+    ],
+)
+@pytest.mark.parametrize("uvf_mode", ["to_flag", "to_metric"])
+def test_select_parse_ants_errors(uvf_from_data, uvf_mode, select_kwargs, err_msg):
+    uvf = uvf_from_data
+    # used to set the mode depending on which input is given to uvf_mode
+    getattr(uvf, uvf_mode)()
+    if select_kwargs["ant_str"] == "auto":
```
mwilensky768, line 2836: This block looks counterintuitive. Is there something subtle here or is it a typo? We will be able to use ant_str to get autos, right?

mkolopanis: Sorry if it is not clear enough, this is a pretty contrived example in the test where I am fishing to raise and error in the `select` function. So I call select with `ant_str=auto` but before doing so call it with `ant_str=cross` to ensure the auto ant_str will not have any matching data on the object.

pyuvdata/utils.py
```diff
@@ -2071,3 +2072,256 @@ def apply_uvflag(
...
+    ant_str, ant_pairs_nums and polarizations are returned as None.
+
+    Parameters
+    ----------
+    uv : UVBase Object
+        A UVBased object that supports the following functions and parameters:
+         - get_ants
+         - get_antpairs
+         - get_pols
+         - x_orientation
```
mkolopanis, line 2094: I just noticed this one this morning, but not sure the best way to handle it. Currently the function is assuming the object has an attribute `x_orientation` to handle the polarization conventions, but I also added this keyword for x_orientation which is never used. Not sure which is the preferred way to handle x_orientation for this function.

bhazelton: `x_orientation` is not a required attribute on the objects, so I don't think you can assume it's set. In the case that it's not set, allowing someone to set it in this function is nice. I can see 2 options:

1.) only use the passed in parameter if the attribute is not set on the object (so the object's attribute overrides the parameter).
2.) Use the parameter regardless of what is set on the object (so that it overrides the object's attribute).

I think either is ok, but you need to be clear about it in the docstring.

mkolopanis: I think 2 is more inline with how some other function will operate. If it is set on an object, it should be passed as the parameter to the function.

mkolopanis: I implemented option 2. I believe.

pyuvdata/uvflag/tests/test_uvflag.py
mkolopanis: Just to explicitly clear it will be possible to use ant_str to get autos in general. Should behave exactly like the one in uvdata.

pyuvdata/utils.py
```diff
@@ -2071,3 +2072,254 @@ def apply_uvflag(
...
+        output in ant_pairs_nums. If ant_str has a minus sign as the first
+        character, 'all,' will be appended to the beginning of the string.
+        See the tutorial for examples of valid strings and their behavior.
+    print_toggle : bool
+        Boolean for printing parsed baselines for a visual user check.
+    x_orientation : str, optional
+        Orientation of the physical dipole corresponding to what is
+        labelled as the x polarization ("east" or "north") to allow for
+        converting from E/N strings. See corresonding parameter on UVData
+        for more details.
```
bhazelton, line 2110: my read of the code now is that if you don't pass this explicitly it won't be set, even if it is set on the object. Is that what we want?

mkolopanis: Yeah that's the way it is written currently. Could have it pick up the value from the object if it exists and input was None, but use the optional parameter to override if given?

bhazelton: I'd vote for that. Then you'll need to make it clear in the docstring that it overrides the value on the object.

bhazelton:
@mwilensky768 are you happy with this now?

mwilensky768:
Haven't totally been keeping up but it looks like it does what I want it to do, which was select using ant_str. Thank you very much for pushing on this.

pyuvdata/utils.py
bhazelton: @mkolopanis do you want to make this change or just leave it as is?

mkolopanis: Oh yes I do want to make this change. Thanks for the ping.

bhazelton: I think you mean `override` not `overwrite`. It won't actually change the value on the object will it?

mkolopanis: good catch definitely a typo.

bhazelton:
This looks good to me now. Can you update the changelog? I guess this will be added to the tutorial when the UVFlag section is written.

mkolopanis:
Let's make sure to add it to the big UVFlag wishlist in the sky!

PR #878, opened 2020-07-08
Improve versioning for test PyPI deployment

plaplant:
## Description
This PR creates a unique identifier for a branch when uploading to the test PyPI server as part of CI.

## Motivation and Context
The current way that version numbers are constructed for uploads to the test PyPI server only keeps the "main" version information, and discards the "local" component (which includes the git hash and branch name in our scheme). This led to collisions between branches that were both the same number of commits from the last release, and the test PyPI server does not allow uploading files that have the same name. So, we would like to include the git hash as part of the "dev" tag to make the identifier unique. However, PyPI enforces the versioning scheme described by [PEP 440](https://www.python.org/dev/peps/pep-0440/), which means that only numbers are permitted. So this PR converts the hex characters in the git hash to numeric values and adds it to the end of the commit number, taking care to add leading zeros to the existing numeric values of the hash to avoid hash collisions.

## Types of changes

- [ ] Version change

## Checklist:

### DIFF
--- a/ci/make_dev_version.sh
+++ b/ci/make_dev_version.sh
@@ -25,6 +25,11 @@
 # As an added benefit, this approach also handles tagged release versions (such
 # as "2.0.3" with no trailing "dev" information), and does not mangle the
 # version information.
+#
+# Examples showing the conversion from pyuvdata version to pypi version:
+#   1.0.2 -> 1.0.2
+#   1.0.2.dev1+g0123456.branch.name -> 1.0.2.dev100010203040506
+#   1.0.2.dev197+g9aac2d38.branch.name -> 1.0.2.dev1970910101202130308
 
 # get version from python
 ver=$(python setup.py --version | tail -1)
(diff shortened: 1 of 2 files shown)


ci/make_dev_version.sh
```diff
@@ -0,0 +1,61 @@
...
+# version number that is acceptable to the testpypi server. According to PEP
+# 440, only numeric values are allowed in the ".dev" field, and testpypi does
+# not allow for "local" information (such as the git version). To get around
+# these constraints, we substitute the hex characters in the git hash for the
+# corresponding numeric values (e.g., a -> 10, b -> 11, etc.). We also prepend
+# numbers with a leading "0" (e.g., 1 -> 01) to avoid hash collisions.
+#
+# As an added benefit, this approach also handles tagged release versions (such
+# as "2.0.3" with no trailing "dev" information), and does not mangle the
+# version information.
```
steven-murray, line 27: Would be really nice to have a few examples here, eg

```
# v1.0.2.dev23+g1af3e56 --> v1.0.2dev2301101503140506
# v1.0.2 --> v1.0.2
```

plaplant: Good suggestion! I'll add a few.

PR #880, opened 2020-07-08
Replace `checkWarnings` with a new function similar to `pytest.warns`

### LINKED ISSUE #766: Use pytest.warns to check warning messages rather than our own checkWarnings function
This would naturally also take care of the example called out in #745.

See: https://docs.pytest.org/en/latest/warnings.html#warns

### LINKED ISSUE #836: Use of mutable defaults in checkWarnings.
#835 got me hunting and I did find a spot with poor choice of defaults. The list and dict defaults used in the checkWarnings call could cause issues with repeated calls.

https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/4c1c67efe488e8ea77bc2755ae68357aecad7920/pyuvdata/tests/__init__.py#L33-L41

bhazelton:
## Description

Wrote a new function `check_warnings` that functions similarly to `pytest.warns` but has support for multiple warnings. Can be used as a drop-in replacement for `pytest.warns`, but it will error if there are extra warnings not included in the call to `check_warnings`.

Also updates `checkWarnings` to not use mutable defaults and deprecates it.

~This builds off #520, it can't be merged before that PR and will need to be rebased once that PR is merged.~

## Motivation and Context

closes #766 
closes #836 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

Breaking change checklist:

### DIFF
--- a/pyuvdata/tests/__init__.py
+++ b/pyuvdata/tests/__init__.py
@@ -25,9 +25,18 @@ class WarningsChecker(warnings.catch_warnings):
     A context manager to check raised warnings.
 
     Adapted from pytest WarningsRecorder and WarningsChecker.
+
+    Parameters
+    ----------
+    expected_warning : list of Warnings
+        List of expected warnings.
+    match : list of str or regex
+        List of strings to match warnings to.
+
     """
 
     def __init__(self, expected_warning, match):
+        """Check inputs and initalize object."""
         super().__init__(record=True)
         self._entered = False
         self._list = []
@@ -79,7 +88,14 @@ class WarningsChecker(warnings.catch_warnings):
         return len(self._list)
 
     def pop(self, cls):
-        """Pop the first recorded warning, raise exception if not exists."""
+        """
+        Pop the first recorded warning, raise exception if not exists.
+
+        Parameters
+        ----------
+        cls : Warning
+            Warning class to check for.
+        """
         for i, w in enumerate(self._list):
             if issubclass(w.category, cls):
                 return self._list.pop(i)
@@ -179,7 +195,9 @@ def check_warnings(expected_warning, match=None, nwarnings=None, *args, **kwargs
 
     Similar to ``pytest.warns``, but allows for specifying multiple warnings.
     It also better matches warning strings when the warning uses f-strings or
-    formating.
+    formating. Can be used as a drop-in replacement for ``pytest.warns`` if
+    only one warning is issued (if more are issued they will need to be added
+    to the input lists for this to pass).
 
     Note that unlike the older checkWarnings function, the warnings can be passed
     in any order, they do not have to match the order the warnings are raised
@@ -193,9 +211,9 @@ def check_warnings(expected_warning, match=None, nwarnings=None, *args, **kwargs
     expected_warning : list of Warnings or Warning
         List of expected warnings. If a single warning type or a length 1 list,
         will be used for the type of all warnings.
-    match : list of str or str
-        List of strings to match warnings to. If a str or a length 1 list, will
-        be used for all warnings.
+    match : str or regex or list of str or regex
+        List of strings or regexes to match warnings to. If a str or a length 1
+        list, will be used for all warnings.
     nwarnings : int, optional
         Option to specify that multiple of a single type of warning is expected.
         Only used if category and match both only have one element.
--- a/pyuvdata/tests/__init__.py
+++ b/pyuvdata/tests/__init__.py
@@ -154,9 +154,11 @@ class WarningsChecker(warnings.catch_warnings):
                         )
                     elif self.match is not None:
                         for record in self:
-                            if re.compile(self.match[warn_i]).search(
+                            if str(record.message).startswith(
+                                self.match[warn_i]
+                            ) or re.compile(self.match[warn_i]).search(
                                 str(record.message)
-                            ) or str(record.message).startswith(self.match[warn_i]):
+                            ):
                                 if issubclass(
                                     record.category, self.expected_warning[warn_i]
                                 ):
--- a/pyuvdata/tests/__init__.py
+++ b/pyuvdata/tests/__init__.py
@@ -5,17 +5,241 @@
 """Setup testing environment, define useful testing functions.
 
 """
-import warnings
+import re
 import sys
+import warnings
 
 from astropy.utils import iers
 
 import pyuvdata.utils as uvutils
 
 __all__ = [
+    "check_warnings",
     "clearWarnings",
     "checkWarnings",
 ]
+
+
+class WarningsChecker(warnings.catch_warnings):
+    """
+    A context manager to check raised warnings.
+
+    Adapted from pytest WarningsRecorder and WarningsChecker.
+    """
+
+    def __init__(self, expected_warning, match):
+        super().__init__(record=True)
+        self._entered = False
+        self._list = []
+
+        assert isinstance(expected_warning, list)
+
+        msg = "exceptions must be derived from Warning, not %s"
+        if expected_warning is None:
+            expected_warning_list = expected_warning
+        elif isinstance(expected_warning, list):
+            for exc in expected_warning:
+                if not issubclass(exc, Warning):
+                    raise TypeError(msg % type(exc))
+            expected_warning_list = expected_warning
+        elif issubclass(expected_warning, Warning):
+            expected_warning_list = [expected_warning]
+        else:
+            raise TypeError(msg % type(expected_warning))
+
+        msg = "match must be a str, not %s"
+        if match is None:
+            match_list = None
+        elif isinstance(match, list):
+            for exc in match:
+                if not isinstance(exc, str):
+                    raise TypeError(msg % type(exc))
+            match_list = match
+        elif isinstance(match, str):
+            match_list = [match]
+        else:
+            raise TypeError(msg % type(match))
+
+        self.expected_warning = expected_warning_list
+        self.match = match_list
+
+    @property
+    def warnlist(self):
+        """The list of recorded warnings."""
+        return self._list
+
+    def __getitem__(self, i: int):
+        """Get a recorded warning by index."""
+        return self._list[i]
+
+    def __iter__(self):
+        """Iterate through the recorded warnings."""
+        return iter(self._list)
+
+    def __len__(self):
+        """The number of recorded warnings."""
+        return len(self._list)
+
+    def pop(self, cls):
+        """Pop the first recorded warning, raise exception if not exists."""
+        for i, w in enumerate(self._list):
+            if issubclass(w.category, cls):
+                return self._list.pop(i)
+        __tracebackhide__ = True
+        raise AssertionError("%r not found in warning list" % cls)
+
+    def clear(self) -> None:
+        """Clear the list of recorded warnings."""
+        self._list[:] = []
+
+    def __enter__(self):
+        if self._entered:
+            __tracebackhide__ = True
+            raise RuntimeError("Cannot enter %r twice" % self)
+        _list = super().__enter__()
+        # record=True means it's None.
+        assert _list is not None
+        self._list = _list
+        warnings.simplefilter("always")
+        warnings.filterwarnings("ignore", message="numpy.dtype size changed")
+        warnings.filterwarnings("ignore", message="numpy.ufunc size changed")
+
+        # Filter iers warnings if iers.conf.auto_max_age is set to None, as we
+        # do in testing if the iers url is down. See conftest.py for more info.
+        if iers.conf.auto_max_age is None:
+            warnings.filterwarnings("ignore", message="failed to download")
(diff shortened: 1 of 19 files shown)


pyuvdata/tests/__init__.py
```diff
@@ -5,19 +5,245 @@
...
+
+    Adapted from pytest WarningsRecorder and WarningsChecker.
+    """
+
+    def __init__(self, expected_warning, match):
+        super().__init__(record=True)
+        self._entered = False
+        self._list = []
+
+        assert isinstance(expected_warning, list)
```
mkolopanis, line 35: This assert statement appears to be in direct conflict with the `if/else` statements that follows. One of the cases is a list instance but the others could never trigger. Or am I reading this incorrectly?

pyuvdata/tests/__init__.py
```diff
@@ -5,19 +5,245 @@
...
+    def __enter__(self):
+        if self._entered:
+            __tracebackhide__ = True
+            raise RuntimeError("Cannot enter %r twice" % self)
+        _list = super().__enter__()
+        # record=True means it's None.
+        assert _list is not None
+        self._list = _list
+        warnings.simplefilter("always")
+        warnings.filterwarnings("ignore", message="numpy.dtype size changed")
```
mkolopanis, line 104: should we add a small comment here about why we filter these for our tests?

pyuvdata/tests/__init__.py
```diff
@@ -5,19 +5,245 @@
...
+                            if str(record.message).startswith(
+                                self.match[warn_i]
+                            ) or re.compile(self.match[warn_i]).search(
+                                str(record.message)
+                            ):
+                                if issubclass(
+                                    record.category, self.expected_warning[warn_i]
+                                ):
+                                    break
+                        else:
```
mkolopanis, line 166: I've been looking for a good excuse to use `for/else` statements :tada:

pyuvdata/tests/__init__.py
```diff
@@ -5,19 +5,245 @@
...
+    """
+    Assert that code raises a particular set of warnings, used as a context manager.
+
+    Similar to ``pytest.warns``, but allows for specifying multiple warnings.
+    It also better matches warning strings when the warning uses f-strings or
+    formating.
+
+    Note that unlike the older checkWarnings function, the warnings can be passed
+    in any order, they do not have to match the order the warnings are raised
+    in the code.
```
mkolopanis, line 185: This is my favorite part :heart:

pyuvdata/tests/__init__.py
```diff
@@ -5,19 +5,245 @@
...
+
+        super().__exit__(exc_type, exc_val, exc_tb)
+
+        # Built-in catch_warnings does not reset entered state so we do it
+        # manually here for this context manager to become reusable.
+        self._entered = False
+
+        # only check if we're not currently handling an exception
+        if exc_type is None and exc_val is None and exc_tb is None:
+            if self.expected_warning is not None:
```
mkolopanis, line 139: I'm a little confused as to the use of the object in the case where `self.expected_warning is None` though. Just feels like it does nothing. Is that the behavior we want?

pyuvdata/tests/__init__.py
```diff
@@ -5,19 +5,245 @@
...
+
+    Note that unlike the older checkWarnings function, the warnings can be passed
+    in any order, they do not have to match the order the warnings are raised
+    in the code.
+
+    Parameters
+    ----------
+    expected_warning : list of Warnings or Warning
+        List of expected warnings. If a single warning type or a length 1 list,
+        will be used for the type of all warnings.
```
mkolopanis, line 191: I may have to go back and check, but it seems like you only need a list of all the unique types of warnings created by a test since for each record, the list of expected types is looped through until a matching one is found.

pyuvdata/tests/__init__.py
```diff
@@ -5,19 +5,245 @@
...
+    """
+    __tracebackhide__ = True
+    if not isinstance(expected_warning, list):
+        expected_warning_list = [expected_warning]
+    else:
+        expected_warning_list = expected_warning
+    if not isinstance(match, list):
+        match_list = [match]
+    else:
+        match_list = match
```
mkolopanis, line 208: Should we use the `list()` function here to cast as a list? What if the inputs are tuples, sets, or numpy arrrays?

pyuvdata/tests/__init__.py
```diff
@@ -5,19 +5,245 @@
...
+    ):
+        raise ValueError(
+            "If expected_warning and match both have more than one element, "
+            "they must be the same length."
+        )
+
+    if len(expected_warning_list) > 1 or len(match_list) > 1:
+        nwarnings = max(len(expected_warning_list), len(match_list))
+    elif nwarnings is None:
+        nwarnings = 1
```
mkolopanis, line 223: This block is the first time `nwarnings` appears, and it overwrites the definition of the variable if there is more than one warning. Should we even have it as an optional kwarg if we discard user input?

pyuvdata/tests/__init__.py
```diff
@@ -5,19 +5,245 @@
...
+
+    if len(expected_warning_list) < nwarnings:
+        expected_warning_list = expected_warning_list * nwarnings
```
mkolopanis, line 226: So this block would work assuming that expected warning list is length 1 and all warnings are the same type. But I don't think we necessarily need to require the list to have an entry for each warning, just one entry for each type since the underlying object will iterate until a matching type is found.

pyuvdata/tests/__init__.py
```diff
@@ -5,19 +5,245 @@
...
+    if len(match_list) < nwarnings:
+        match_list = match_list * nwarnings
```
mkolopanis, line 228: Same comment here about the matching of lengths thing. This could be a power feature, or a terrible bug :thinking:

pyuvdata/tests/__init__.py
mkolopanis: or perhaps `list(uvutils._get_iterable())` to ensure a single string is not broken up?

mkolopanis: Oh that wouldn't necessarily work either, might need a special case of `isinstance( ???, str)` if you do what I suggested.

pyuvdata/tests/__init__.py
bhazelton: sure, will do.

pyuvdata/tests/__init__.py
bhazelton: Ha! I kind of hate them because I always have to think hard about what they mean, but they are useful sometimes.

pyuvdata/tests/__init__.py
bhazelton: It can be used to check that the right kind of warning is emitted without checking the warning message. I included it because `pytest.warns` supports that and I wanted to be able to make this be a drop in replacement. But I'm willing to take it out and error if the warning message isn't provided.

pyuvdata/tests/__init__.py
bhazelton: I thought I was writing it so that the message and error type had to match (so that I use the message index to get the error index I'm checking against), but I may have made a mistake.

pyuvdata/tests/__init__.py
bhazelton: hmm, I should probably be more careful. I was expecting to get either a string or a list. I can add some checking for that.

pyuvdata/tests/__init__.py
bhazelton: So the point (in my head) of `nwarnings` is the case where you have a lot of only one warning over and over. In the docstring I tried to make it clear that `nwarnings` was only used if both the error and message only had one element. I don't like that with the current function you have to specify nwarnings when you've already specified how many by listing all the errors and messages.

pyuvdata/tests/__init__.py
bhazelton: I guess my goal was to only give people a pass on specifying all the warnings if they were all the same (via the nwarnings parameter). I suppose we could allow people to supply a list to nwarnings that specifies how many of each type of warning.

pyuvdata/tests/__init__.py
bhazelton: hmm, you're right. I think this is debugging cruft.

pyuvdata/tests/__init__.py
mkolopanis: Ah okay I think I get it now. I was not necessarily thinking about this case when I read the block.

pyuvdata/tests/__init__.py
mkolopanis: Right, that loop happens down in line 145. But even the length checking happens inside of this if statement and since there is no `else` to this if I am just confused why we have it.

pyuvdata/tests/__init__.py
mkolopanis: Further inspection looks like I missed a `warn_i` index into the `expected_warnings` list. should work the way you intend I think.

pyuvdata/tests/__init__.py
bhazelton: I said the wrong thing before. This approach is used in `pytest.warns` if you don't want to assert anything about the warning but want to get back the list of warning objects so that you can do your own tests on them. I think this is a pretty infrequent use case, but I suppose it could be used to check that there are no warnings -- I need to test that usage though and make sure it actually works for that.

mkolopanis: Oh okay I see. thanks for the clarification.

mkolopanis:
@bhazelton I think all my concerns have been addressed now. Did you want another look or still planning some work work?

bhazelton:
I'm going to check if there's a good way to test for no warning and if so, document it in the docstring.

bhazelton:
Should I update the changelog? It's weird because this is not part of our API, but we are deprecating something...

mkolopanis:
Hm this is a tough one. I agree it's not part of the normal API, but I would bet a good number of HERA repos use this functionality. Changelog update might be good just for the sake of clarity?

pyuvdata/tests/__init__.py
bhazelton: So it turns out I was wrong, but I think that functionality is useful, so I made some modifications so that if you pass `None` as the first argument it will assert that no warnings are raised by the code inside the context manager.

bhazelton:
I'm happy either way. If @plaplant wants to take a look great, but if he's busy I think it's in good shape. Thanks for all the comments, it's definitely better than where it started.

bhazelton:
@jsdillon we're seeing a bunch of what `hera_cal` testing errors on this PR and other ones. It looks like IERS issues. Our practice has evolved on how best to manage those, happy to talk about our current approach if it would be helpful.

jsdillon:
Sure thing. What do I need to do?

bhazelton:
I'll have to look at `hera_cal` to see how you're doing it. I just suspect it's different because we aren't getting errors on pyuvdata tests right now.

bhazelton:
@jsdillon Try removing line 22 in your `conftest.py` (the one that reads `iers_a = iers.IERS_A.open(iers.IERS_A_URL)`). That seems like the most obvious difference between what we have and hera_cal.

plaplant:
@bhazelton @mkolopanis I'm happy to take a look, but probably won't get to it till next week at the earliest. I haven't looked extensively but the bits I've seen seem good to me 👍

pyuvdata/tests/__init__.py
```diff
@@ -5,19 +5,260 @@
...
+
+    To assert that there are no warnings raised by some code, set `expected_warning`
+    to None (i.e. `with check_warnings(None):`)
+
+    Parameters
+    ----------
+    expected_warning : list of Warnings or Warning
+        List of expected warnings. If a single warning type or a length 1 list,
+        will be used for the type of all warnings.
+    match : list of str or str
```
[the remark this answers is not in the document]
bhazelton: Yes, thanks. The regex part is copied exactly from `pytest.warns` and I think it will match things in the middle of the warning string. But I have found that it doesn't work properly for warnings that have fstrings or values inserted with a format call. I don't understand why it has that problem, but adding the `startswith` check has fixed it.

pyuvdata/uvbeam/tests/test_cst_beam.py
```diff
@@ -109,7 +109,7 @@ def test_read_yaml_override(cst_efield_2freq):
     beam1 = cst_efield_2freq
     beam1.telescope_name = "test"
 
-    with pytest.warns(
+    with uvtest.check_warnings(
```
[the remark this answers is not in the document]
bhazelton: Yeah, I liked that. I added some text to that effect in the docstring to make it clearer.

bhazelton:
@mkolopanis signed off, so I'm going to merge.

PR #882, opened 2020-07-09
Remove UVFlag Nblts requirement under the hood

### LINKED ISSUE #661: UVFlag sometimes has no sense of Nblts
Currently Nblts are used synonymously with the time axis for "antenna" and "waterfall" type objects. This was done for ease when using functions like `select`. It may be reasonable to fix this up and add if/else blocks around the nblts to account for objects with no baselines.

This _should_ be pretty obscured from the user but probably still philosophically incorrect the way it is implemented.

mkolopanis:
When UVFlag was converted to a UVBased object, the "Nblts" axis was kept around under the hood as a comparison parameter for many tests even when the object did not have a physical sense of baselines. This removes that to make the object more pedantically self consistent. 
## Description

Added logic statements to check the type of the UVFlag object when a comparison vs Nblts is necessary to compare against Ntimes in waterfall and antenna types.

## Motivation and Context

fixed #661 

## Types of changes

- [ ] Version change
- [x] Pedantic verbiage and self-consistency

## Checklist:

### DIFF
--- a/pyuvdata/uvflag/uvflag.py
+++ b/pyuvdata/uvflag/uvflag.py
@@ -1897,7 +1897,7 @@ class UVFlag(UVBase):
         blt_inds : array_like of int, optional
             The baseline-time indices to keep in the object. This is
             not commonly used.
-        blt_inds : array_like of int, optional
+        ant_inds : array_like of int, optional
             The antenna indices to keep in the object. This is
             not commonly used.
 
--- a/pyuvdata/uvflag/uvflag.py
+++ b/pyuvdata/uvflag/uvflag.py
@@ -588,8 +588,8 @@ class UVFlag(UVBase):
         self._Nants_data.required = True
         self._Nbls.required = False
         self._Nspws.required = True
+        self._Nblts.required = False
 
-        self.Nblts = self.Ntimes
         desc = (
             "Floating point metric information, "
             "has shape (Nants_data, Nspws, Nfreqs, Ntimes, Npols)."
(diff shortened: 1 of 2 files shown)


mkolopanis:
Not sure if we're 100% ready here, but can you give it a look over @adampbeardsley

pyuvdata/uvflag/uvflag.py
```diff
@@ -1856,7 +1856,10 @@ def _select_preprocess(
             blt_inds = uvutils._get_iterable(blt_inds)
             if np.array(blt_inds).ndim > 1:
                 blt_inds = np.array(blt_inds).flatten()
-            history_update_string += "baseline-times"
+            if self.type == "baseline":
+                history_update_string += "baseline-times"
+            else:
+                history_update_string += "times"
```
adampbeardsley, line 1862: I wonder if we should have a separate keyword `t_inds`, and error if you set `blt_inds` when not type `baseline`? Or maybe at least warn?

adampbeardsley:
This isn’t related to the PR, but I noticed while looking at it… `_select_preprocess` has all arguments, but the docstring says everything is optional.
Also noticed blt_inds is documented twice… second one should be ant_inds?

bhazelton:
If it's like the one on UVData, the `_select_preprocess` arguments are all allowed to be `None` -- maybe that's what is meant by optional?

mkolopanis:
To be honest it is pretty much copy-pasta from UVData implementation, but yeah I think the optionalness implies it can be none. Maybe we should default the values to `None` though to actually make them optional.

pyuvdata/uvflag/uvflag.py
mkolopanis: Yeah, maybe. But at the same time time it only appears in functions that presumably a regular user would never call and the `blt_ind`/`t_ind` would just introduce another input/output to the function which end up slicing into the same arrays anyway. I can add it if you'd like though.

bhazelton:
I don't think you should default them, this is a private helper method, the defaulting is done in the method that calls this one. Since this is a private method, the docstring isn't even in the docs. But you can update the docstring if it's confusing to developers.

pyuvdata/uvflag/uvflag.py
bhazelton: This is a private helper method. If there's a concern about user input it should be handled in the public method that calls this one.

mkolopanis:
I went ahead and fixed up the docstrings. Regarding the parameters, no one has even brought up confusion before on other select functions which are almost identical to this one. I'm prone to leaving it until broader concerns are raised?

PR #895, opened 2020-07-29
Github action yaml change. Mostly aesthetics 

mkolopanis:
A cleaner way to set up Github Actions conda environments.
## Description

uses a github action to set up the miniconda environment rather than logic blocks to download for each operating system.
Sets a default shell to `bash -l {0}` which logs the runner user into the bash shell.

## Motivation and Context

Wanted to see if others think it looks nicer compared to the old github action yaml.

## Types of changes

- [ ] Version change

## Checklist:

### DIFF
--- a/.github/workflows/macosx_windows_ci.yaml
+++ b/.github/workflows/macosx_windows_ci.yaml
@@ -17,56 +17,47 @@ jobs:
         os: [macos-latest, windows-latest]
         python-version: [3.6, 3.7, 3.8]
     steps:
+      - name: Set ENV NAME
+        run: |
+          if [[ "${{ runner.os }}" = "Windows" ]]; then
+            echo "::set-env name=ENV_NAME::pyuvdata_tests_windows"
+          else
+            echo "::set-env name=ENV_NAME::pyuvdata_tests"
+          fi
+
       - uses: actions/checkout@master
         with:
           fetch-depth: 1
 
-      - name: Install Miniconda
-        shell: bash
+      - name: Setup Miniconda
+        uses: goanpeca/setup-miniconda@v1.6.0
+        with:
+          auto-update-conda: "true"
+          miniconda-version: "latests"
+          python-version: ${{ matrix.python-version }}
+          environment-file: ci/${ENV_NAME}.yml
+          activate-environment: ${ENV_NAME}
+
+      - name: Conda Info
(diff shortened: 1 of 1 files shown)


.github/workflows/macosx_windows_ci.yaml
```diff
@@ -18,52 +20,44 @@ jobs:
...
+        with:
+          auto-update-conda: "true"
+          miniconda-version: "latest"
+          python-version: ${{ matrix.python-version }}
+          environment-file: ci/${{ steps.env_name.outputs.ENV_NAME }}.yml
+          activate-environment: ${{ steps.env_name.outputs.ENV_NAME }}
+
+      - name: Conda Info
+        run: |
+          conda info -a
```
bhazelton, line 43: I think we also want a `conda list` call, I very often want to see which exact versions of packages were installed.

.github/workflows/macosx_windows_ci.yaml
```diff
@@ -6,8 +6,10 @@ jobs:
     name: Run Tests
     env:
       PYTHON: ${{ matrix.python-version }}
-      OS: ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
+    defaults:
+     run:
+       shell: bash -l {0}
```
bhazelton, line 12: what does this do?

mkolopanis: so the `-l` adds the `login` flag to the bash shell. I've attached the clip from the man page about what that option does. The real reason I added it is because conda doesn't always seem to work correctly on windows without it. Nothing I have noticed on this particular repository but I have on 21cmsense. Inspecting the examples from the miniconda setup action's github page they also used the `-l` option in their examples.

>When bash is invoked as an interactive login shell, or as  a  non-interactive  shell  with  the
       --login  option,  it first reads and executes commands from the file /etc/profile, if that file
       exists.  After reading that file, it looks for ~/.bash_profile, ~/.bash_login, and  ~/.profile,
       in  that order, and reads and executes commands from the first one that exists and is readable.
       The --noprofile option may be used when the shell is started to inhibit this behavior.

bhazelton: maybe add a comment that it's required to make windows work properly?

PR #908, opened 2020-08-22
Properly scale MWA digital gains

PyxieLouStar:
MWA digital gains have been scaled by a factor of 64 before being recorded in the metafits file. This fix properly accounts for that scaling in `mwa_corr_fits`.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Version change

## Checklist:

- [X] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/master/.github/CONTRIBUTING.md).
- [X] My code follows the code style of this project.

Bug fix checklist:
- [X] All new and existing tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ## [2.1.1] - 2020-8-07
 
 ### Added
+- Adjustment to digital gain removal from mwa_corr_fits files to account for a division by 64 due to a bit selection shift.
 - Options to remove coarse band shape and digital gains from mwa_corr_fits files.
 - Support for cotter flags in mwa_corr_fits files.
 - Added read-only support for MIR files, adding the `Mir` class to `UVData`.
--- a/pyuvdata/uvdata/mwa_corr_fits.py
+++ b/pyuvdata/uvdata/mwa_corr_fits.py
@@ -437,8 +437,7 @@ class MWACorrFITS(UVData):
             antenna_names = meta_tbl["TileName"][1::2]
             antenna_flags = meta_tbl["Flag"][1::2]
             cable_lens = np.asarray(meta_tbl["Length"][1::2]).astype(np.str_)
-            # the digital gains have been multiplied by a factor of 64
-            dig_gains = meta_tbl["Gains"][1::2, :].astype(np.float64) / 64
+            dig_gains = meta_tbl["Gains"][1::2, :].astype(np.float64)
 
             # get antenna postions in enu coordinates
             antenna_positions = np.zeros((len(antenna_numbers), 3))
@@ -705,7 +704,11 @@ class MWACorrFITS(UVData):
             if remove_dig_gains:
                 # get gains for included coarse channels
                 coarse_inds = np.where(np.isin(coarse_chans, included_coarse_chans))[0]
-                dig_gains = dig_gains[:, coarse_inds]
+                # during commissioning a shift in the bit selection in the digital
+                # receiver was implemented which effectively multiplies the data by
+                # a factor of 64. To account for this, the digital gains are divided
+                # by a factor of 64 here. For a more detailed explanation, see PR #908.
+                dig_gains = dig_gains[:, coarse_inds] / 64
                 dig_gains = np.repeat(dig_gains, num_fine_chans, axis=1)
                 dig_gains1 = dig_gains[self.ant_1_array, :]
                 dig_gains2 = dig_gains[self.ant_2_array, :]
--- a/pyuvdata/uvdata/mwa_corr_fits.py
+++ b/pyuvdata/uvdata/mwa_corr_fits.py
@@ -437,7 +437,8 @@ class MWACorrFITS(UVData):
             antenna_names = meta_tbl["TileName"][1::2]
             antenna_flags = meta_tbl["Flag"][1::2]
             cable_lens = np.asarray(meta_tbl["Length"][1::2]).astype(np.str_)
-            dig_gains = meta_tbl["Gains"][1::2, :]
+            # the digital gains have been multiplied by a factor of 64
+            dig_gains = meta_tbl["Gains"][1::2, :].astype(np.float64) / 64
 
             # get antenna postions in enu coordinates
             antenna_positions = np.zeros((len(antenna_numbers), 3))
@@ -710,8 +711,11 @@ class MWACorrFITS(UVData):
                 dig_gains2 = dig_gains[self.ant_2_array, :]
                 dig_gains1 = dig_gains1[:, :, np.newaxis]
                 dig_gains2 = dig_gains2[:, :, np.newaxis]
+                if self.data_array.dtype != np.complex128:
+                    self.data_array = self.data_array.astype(np.complex128)
                 self.data_array = self.data_array / (dig_gains1 * dig_gains2)
-
+                if self.data_array.dtype != data_array_dtype:
+                    self.data_array = self.data_array.astype(data_array_dtype)
             # divide out coarse band shape
             if remove_coarse_band:
                 # get coarse band shape
--- a/pyuvdata/uvdata/tests/test_mwa_corr_fits.py
+++ b/pyuvdata/uvdata/tests/test_mwa_corr_fits.py
@@ -607,7 +607,7 @@ def test_remove_dig_gains():
     with fits.open(filelist[0]) as meta:
         meta_tbl = meta[1].data
         antenna_numbers = meta_tbl["Antenna"][1::2]
-        dig_gains = meta_tbl["Gains"][1::2, :]
+        dig_gains = meta_tbl["Gains"][1::2, :].astype(np.float64) / 64
     reordered_inds = antenna_numbers.argsort()
     dig_gains = dig_gains[reordered_inds, :]
     dig_gains = dig_gains[:, np.array([23])]


gsleap:
Hi @PyxieLouStar, @andrew5au and I saw this PR and were curious as to the origins of the 64 factor in the digital gains- we would like to understand where this number comes from if you have any info on that? Thanks!

PyxieLouStar:
@gsleap and @andrew5au, thanks for your interest! The factor of 64 is coming out of COTTER. Some exploration by @mwilensky768 revealed that the digital pfb gains COTTER was dividing out did not match the gains recorded in the metafits file. I look through the COTTER code and found that metafits gains were being divided by a facter of 64. The code was accompanied by this comment: `The digital pfb gains are multiplied with 64 to allow more careful fine tuning of the gains.` I'm now searching for more documentation in order to understand where this factor of 64 might be coming from.

mwilensky768:
Hi all. I have some plots to deliver.

cotter_dig_gain_compare

This plot is an autocorrelation for a COTTER-made uvfits file, where the blue is uncorrected, and the orange has had the digital gains divided out, as COTTER sees them.

pyuvdata_dig_gain_compare

This is the same plot, but with the pyuvdata reader, and put on a log scale so that we can see the corrected version a little better. These have been corrected by the digital gains as listed in the metafits file, which ends up giving factors of ~64^2 for an autocorrelation, hence the log scale.

cotter_dig_gain_ratio

This plot is the ratio of the curves on the first plot (uncorrected COTTER / corrected COTTER). We can see some clear stair-steps demarcating the different coarse bands which had slightly different digital gains.

pyuvdata_ratio

This plot is the ratio of the curves on the _second_ plot (uncorrected pyuvdata / corrected pyuvdata). We can see it is identical to the previous plot, just with the vertical axis rescaled. This is because, as @PyxieLouStar mentioned COTTER rescales the gains listed in the metafits file before dividing by them.

PyxieLouStar:
The digital gains as recorded in the metafits files are accurate. The factor of 64 comes from a shift in the bit selection in the digital receiver. After the polyphase filter bank, the data is downselected from 16 + 16 bit real/imaginary pairs to 5 + 5 bit pairs. Initially (MWA 32T) bits 09-05 inclusive were selected. In this range, the digital gains were limited to values 1-7. To make possible greater resolution in the digital gains, during commissioning (2013) the bit selection was changed from bits 09-05 to bits 15-11. This change in bit selection is equivalent to dividing the data by a factor of 64. Dividing the digital gains by the factor of 64 before dividing the gains out of the data thus accounts for the bit selection. Selection of bits 15-11 and division by 64 of the digital gains has been the practice for all MWA phase 1 & 2 operations.

The following correspondences from Brian Crosse (brian.crosse@curtin.edu.au) and @andreww5au describe this in greater detail.

---
The output of the coarse PFB inside the receiver gives us a 'twos compliment', 16 bit real + 16 bit imaginary number for each coarse (1.28MHz) channel, however there is/was only enough communications link speed to send 5 bits real and 5 bits imaginary to the control building.  Of each 16 output bits, the bits sent were numbers 09 through 05 inclusive.  Since the signal levels were typically only toggling the bottom few bits, we could only set digital gains in the (around) 1 to 7 range before saturating.  The PFB contents were multiplied by these prior to the 5 bit, bit selection.

MSB                                                                         LSB
15 14 13 12 11 10 09 08 07 06 05 04 03 02 01 00

So to repeat, while still in 16 bits, an integer multiple is applied to each of the real and imaginary parts that is allowed to be different for each of the 256 possible channels and then the 5 bits for transport are cut off and sent.

These gain values by coarse channel number were originally:

1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 3 3 3 3 3 3 3 3 3 3 3 3 3 4 4 4 4 4 4 4 5 5 5 5 5 5 5 5 5 5 5 5 5 5 5 5 5 5 5 5 5 5 5 5 5 5 5 6 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7 7

The values were selected by Chris Williams to roughly flatten out the frequency dependant drop-off in power seen on the original 32t array.

It's important to note that the drop-off was a function (mostly) of the cable length in the 32T array.  In the 128T array there are two other important factors; the cable type (RG6 vs LMR-400) and the presence or absence of the whitening filter.  This prompted us to try for greater resolution in the gain selections.

What was done, was to change which of the 16 bits was selected to be sent from bits 09-05 to bits 15-11.  That is, we moved the bit selection up by 6 bits (a factor of 64 in values).  At that point we would have identical behaviour iff we increased the digital gains by exactly a factor of 64.  That way the same initial (pre-gain) values would end up in the 5 bits that made it out the fibre.  I believe there was a short time when that was done, but very quickly we moved to a more complicated system where the digital gain varied by cable flavor and, of course, needed to contain values 64x larger to have the same effect as the original gains.  These new values did not need to be as granular any more since we had more 'fine tuning' available in the new range.

So the specific answer to your question is that a factor of 64 was introduced in late 2012 (or early 2013 by the time it was activated?) when the selection of which bits out of the coarse PFB were sent to the correlator changed.

---
The change that Brian's described, in the way that the ADFB module inside the receiver applies the digital gains, happened well before the MWA finished commissioning, so you won't see any metafits files with those old gain values (256 numbers ranging from 1-7).

What you will find is a few changes in how we calculated gain profiles for each type of cable. At first, up to mid 2014, we used Chris's original numbers (in Brian's email below), multiplied by 64, for every type of cable. That gave exactly the same digital output from the receiver as before the ADFB bit selection was changed:

These numbers were:

{64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,6
4,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,64,128,128,128,128,128,128,128,128,128,128,128,128,128,128,128,192,192,192,192,192,192,192,192,192,192,192,192,192,256,256,256,256,256,256,256,320,320,320,320,320,320,320,320,320,320,320,320,320,320,320,320,320,320,320,320,320,320,320,320,320,320,320,384,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448,448}

Note that the first ~90 channels have essentially zero signal, so when deciding on digital gains for those channels, the gain is irrelevant. Chris picked the lowest possible value (1), so after the ADFB firmware changes, that became 64 using the new firmware - but resulted in the exactly the same 5 bit output values.

In July 2014, we calculated new digital gains to smooth the band shape a little better, using the fact that we now had less granularity (not just a number from 1-7), and customising the fit for each type of cable. This time, as before, we chose the minimum legal gain (1) for the first ~90 coarse channels that had zero signal:

For example, for RG6_90 cables, the new values were:

{1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,367,362,354,339,312,270,219,173,140,123,115,108,100,91,83,78,76,78,80,82,82,81,80,79,80,82,85,88,89,89,88,87,86,87,89,91,92,92,92,91,89,89,88,89,90,91,91,90,89,87,86,85,85,85,85,85,85,84,83,81,80,78,78,77,77,77,77,76,74,73,71,70,69,69,69,68,68,67,66,65,64,64,64,64,64,65,65,65,64,64,64,65,65,66,67,67,68,68,69,70,71,72,73,74,75,76,77,77,78,79,80,82,83,84,85,86,87,88,89,90,92,93,95,97,99,100,102,104,106,108,110,113,116,119,121,124,127,130,133,136,140,145,149,154,158,161,165,168,172,175,179,182,186,189,193,196,200,204,207,211,214,218,221,225,228,232,235,238,240,241,243,245,247,250,253,255,258,260,261,262,262,263,264,267,270,274,278,282,285,289,293,298,304,311,318,325,332,338,343,348,352,355,358,361,363,365,366,367,368,368}

Note that while it looks at first glance like this set has numbers that are 1/64th of the gain values before that date, it's only because of the first ~90 channels, that have no signal. The values for the usable channels are still in the same rough range (~100 to ~300) as before, but with a smoother profile, so the RF power stays the same.

In 2016, we went through another process of modelling the gains to correct for cable losses, and the new RG6_90 gains became:

1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,687,565,446,329,225,147,98,76,69,67,64,58,51,45,41,41,43,45,47,46,45,44,43,44,47,49,51,51,51,50,50,50,51,53,55,56,56,56,55,54,54,54,56,57,58,58,58,57,56,56,56,56,57,59,60,60,60,59,58,58,57,58,60,61,62,62,62,61,61,60,61,61,63,64,65,65,65,64,64,64,64,65,66,68,68,69,69,68,68,68,69,70,72,73,74,75,75,75,75,76,77,79,80,82,83,84,84,84,85,86,88,90,92,94,95,97,98,99,101,104,106,109,113,116,118,121,124,127,130,134,138,143,148,152,157,162,166,171,176,182,189,195,202,140,136,170,213,224,110,96,179,211,233,178,148,250,272,202,195,251,257,293,285,239,239,369,375,380,385,389,371,394,399,406,403,416,427,431,434,438,441,446,453,462,471,483,496,510,526,543,561,583,611,644,682,723,767,809,849,886,920,954,985,1013,1039,1060,1076,1087,1092}

Again, the overall power level, in the usable coarse channels, is roughly the same, although there's larger numbers at the very high end(~230 and up), where there is also little usable signal.

Apart from gain values for unusable channels with zero signal, nothing to do with the digital gains has changed by a factor of 64 since well before commissioning finished in mid 2013.

PR #926, opened 2020-09-16
Fix references to the master branch, remove uses of master

bhazelton:
## Description

Remove references to the master branch (since we have converted to main). Also remove uses of the term master in as many places as possible throughout.

## Types of changes

- [ ] Version change

## Checklist:


plaplant (commented):
Thanks @bhazelton! This is a great change. I think all of these replacements are good, but there were a couple of other locations that I noticed should probably also change.
1. FWIW, it looks like the [GitHub actions/checkout repo has renamed their default branch to `main`](https://github.com/actions/checkout), so it might be worth updating the .github/workflow YAMLs to reflect that.
2. Several of the pages in the `docs` folder still have references to `master` instead of `main` in urls to memos etc. I used `git grep master` to find these instances.

bhazelton:
@plaplant Thanks, I didn't know the checkout was now main, I fix those. The ones I found (using `git grep master`) were pointing at other repos that haven't had their default branch changed, but I might have missed some in the noise. Do you have specific files for me to look at again?

docs/index.rst
```diff
@@ -1,309 +1,2 @@
-.. pyuvdata documentation master file, created by
```
plaplant, line 1: Does this mean that we don't actually need this file to make the docs?

plaplant (approved):
Thanks so much @bhazelton! I just have a small question, but I don't think it actually requires any code changes.

docs/index.rst
bhazelton: We do not. I had initially excluded it from the repo entirely in the .gitignore, which made it much less likely that it could get screwed up. But @dannyjacobs didn't like that when you clicked on the "edit on github" button at the top of RTD it gave you a 404 error, so he made a stub file. I'd still prefer not to have it, every time I build the docs locally on my machine I have to be careful to make sure I don't commit the changes to that file, which is what the .gitignore is supposed to protect me against. And I clearly made a mistake a few months ago and committed changes that I shouldn't have.

plaplant: Can we add it to the `.gitignore` now that it's back to being a stub? I don't mind either way, we can also merge as is.

bhazelton: That won't do anything because git is tracking it already. `.gitignore` only affects untracked files. The only way to really fix the problem is to remove it from the repo, but then we'll go back to having a 404 error if you click the "edit on github" link on RTD. @dannyjacobs do you still really care that that link doesn't 404? Because this stub only fixes that link on the main docs page, not on any of the subpages, which currently give a 404 error.

bhazelton: apparently there is a way to remove that link: https://docs.readthedocs.io/en/stable/guides/remove-edit-buttons.html

plaplant: In that case, I'd vote for getting rid of this file (and adding it to the `.gitignore`), and the "edit on github" link. @dannyjacobs is that okay with you?

bhazelton: Ok, done. It was already in the .gitignore. I have verified that the link is gone on RTD.

bhazelton:
hmm, the test_pypi publish is breaking. I just started seeing the same thing on pyradiosky, I don't understand what's going on. It says the API token doesn't work, but I haven't changed anything.

plaplant (approved):
The pypi thing is weird, but I'm okay with merging if that's the only thing that's failing.

bhazelton:
I just re-ran it and it worked, must have been some kind of outage issue.

PR #931, opened 2020-09-30
Only publish tagged releases to testpypi to prevent package bloat.

bhazelton:
## Motivation and Context

Publishing every push to test.pypi resulted in the package eventually exceeding the 10GB size limit. There does not appear to be any way to set a retention time on releases, nor is there any interface for deleting releases other than by hand on their website. That is painful because it required me to copy the name of the release into an “are you sure?” box for 179 devel type releases. So this PR makes it so we only publish actual releases on test.pypi.

I made an issue on the python package github repo, asking them to clarify this problem in the docs (https://github.com/pypa/packaging.python.org/issues/777). It is possible that there is a better way to do what we are doing, if they come back with a better suggestion we can revisit this.

## Types of changes

- [ ] Version change

## Checklist:

### DIFF
--- a/.github/workflows/publish-to-test-pypi.yaml
+++ b/.github/workflows/publish-to-test-pypi.yaml
@@ -47,6 +47,7 @@ jobs:
         python -m pep517.build --source --out-dir dist/ .
 
     - name: Publish to Test PyPI
+      if: startsWith(github.event.ref, 'refs/tags')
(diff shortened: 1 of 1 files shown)


plaplant (approved):
Thanks @bhazelton! Shame there isn't better infrastructure for more regular testing, but sounds like it's much more sustainable to only do this on releases.

PR #937, opened 2020-10-14
Speedup to reading multiple files

AaronParsons:
Modifies UVData.read to concatenates a list of uv files in one shot, instead of incrementally one by one. This should make concatenation run in O(N) time for fast_concat and O(N logN) time for standard concat, instead of O(N^2) as it does now.

## Description
Changed the loop in UVData.read to instantiate and check all the UVData objects, then concatenate them once at the end. UVdata.fast_concat was modified to operate on lists of UVData objects to allow it to concatenate them all in one shot. Because UVData.__add__ was too much work to modify this same way, I instead used a binary merge-down to make this run in O(N log N) time.

## Motivation and Context
Loading and concatenating files takes too long as implemented because each pair of files is concatenated incrementally, one by one.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Version change

## Checklist:

- [X] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/.github/CONTRIBUTING.md).
- [X] My code follows the code style of this project.

Bug fix checklist:
- [X] All new and existing tests pass.
- [X] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/CHANGELOG.md).

New feature checklist:
- [X] I have added or updated the docstrings associated with my feature using the [numpy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [X] All new and existing tests pass.
- [X] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/CHANGELOG.md).

### DIFF
--- a/pyuvdata/uvdata/uvdata.py
+++ b/pyuvdata/uvdata/uvdata.py
@@ -5,9 +5,11 @@
 """Primary container for radio interferometer datasets."""
 import os
 import copy
-import numpy as np
+from collections.abc import Iterable
 import warnings
 import threading
+
+import numpy as np
 from astropy import constants as const
 import astropy.units as units
 from astropy.time import Time
@@ -18,8 +20,6 @@ from ..uvbase import UVBase
 from .. import parameter as uvp
 from .. import telescopes as uvtel
 from .. import utils as uvutils
-
-from collections.abc import Iterable
 
 __all__ = ["UVData"]
 
@@ -3478,7 +3478,7 @@ class UVData(UVBase):
 
     def fast_concat(
         self,
-        others,
+        other,
         axis,
         inplace=False,
         phase_center_radec=None,
@@ -3501,7 +3501,7 @@ class UVData(UVBase):
 
         Parameters
         ----------
-        others : UVData object or list of UVData objects
+        other : UVData object or list of UVData objects
             UVData object or list of UVData objects which will be added to self.
         axis : str
             Axis to concatenate files along. This enables fast concatenation
@@ -3552,28 +3552,29 @@ class UVData(UVBase):
         ValueError
             If other is not a UVData object, axis is not an allowed value or if
             self and other are not compatible.
+
         """
         if inplace:
             this = self
         else:
             this = self.copy()
-        if issubclass(others.__class__, this.__class__):
+        if not isinstance(other, (list, tuple, np.ndarray)):
             # if this is a UVData object already, stick it in a list
-            others = [others]
+            other = [other]
         # Check that both objects are UVData and valid
         this.check(
             check_extra=check_extra,
             run_check_acceptability=run_check_acceptability,
             strict_uvw_antpos_check=strict_uvw_antpos_check,
         )
-        for other in others:
-            if not issubclass(other.__class__, this.__class__):
-                if not issubclass(this.__class__, other.__class__):
+        for obj in other:
+            if not issubclass(obj.__class__, this.__class__):
+                if not issubclass(this.__class__, obj.__class__):
                     raise ValueError(
                         "Only UVData (or subclass) objects can be "
                         "added to a UVData (or subclass) object"
                     )
-            other.check(
+            obj.check(
                 check_extra=check_extra,
                 run_check_acceptability=run_check_acceptability,
                 strict_uvw_antpos_check=strict_uvw_antpos_check,
@@ -3591,10 +3592,10 @@ class UVData(UVBase):
                     phase_frame=orig_phase_frame, use_ant_pos=use_ant_pos
                 )
 
-            for other in others:
-                if other.phase_type != "drift":
+            for obj in other:
+                if obj.phase_type != "drift":
                     warnings.warn("Unphasing other UVData object to drift")
-                    other.unphase_to_drift(
+                    obj.unphase_to_drift(
                         phase_frame=orig_phase_frame, use_ant_pos=use_ant_pos
                     )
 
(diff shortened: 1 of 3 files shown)


AaronParsons:
@bhazelton I pushed the version of the code that exposes the problem I'm having with `UVData.__add__`.

nkern:
Thanks for getting this in @AaronParsons, beat me to it :)

I think this looks good. One concern, which is not so much relevant to the changes made here but to how `fast_concat` works, is that the ordering of the final data is set by the ordering of the files (I presume for efficiency). However, the normal __add__ function does sort the data if the files are mis-ordered (e.g. by increasing frequency, time), which could lead to some confusion for ppl used to the normal `read` capabilities.  We could do one of two things: 1. make this more clear in the `axis` kwarg in the `read_*` docstrings (something like `Note: the ordering of the final data matches the ordering of the input files`) or 2. order the data along the desired `axis` before returning the object. Not sure how much overhead that would incur, though. My main concern is that an unordered list of sub-band files and/or time-chunk files using this `fast_concat` feature could cause problems with some steps that assume monotonicity of those axes.

bhazelton:
I am comfortable with either a documentation update or a reordering. It will be some overhead but shouldn't be crazy. I think some testing on large datasets would really help establish what that looks like. We already have all the relevant reordering methods, so we could just test calling those after the read or not. I hope that it would be pretty minimal overhead if they're already in the right order (still some just because of checking it out).

@nkern @anguta @jsdillon I would also really like to see some time testing on this branch vs the main branch with a large number of HERA files. And for both `fast_concat` and for the regular add. I'd like to document those in this PR. I will note that I am currently trying to see if I can speed up the regular add even more, but I'm not sure yet if that is going to work out.

AaronParsons:
Are we close to a positive review? I've confirmed this dramatically speeds up reading auto files (1862 of them) by a factor of at least 10.

pyuvdata/uvdata/uvdata.py
```diff
@@ -3473,7 +3478,7 @@ def __iadd__(
 
     def fast_concat(
         self,
-        other,
+        others,
```
mkolopanis, line 3481: I am slightly concerned about changing the name of a positional argument. While not necessary it _can_ be called as a keyword argument in python. I think it would be wise to leave the `other` keyword in the function call, but then throw a deprecation warning stating `others` is the prefered call signature and `other` will stop working (in some release). The `other` parameter can then just be assigned to `others` inside.

pyuvdata/uvdata/uvdata.py
```diff
@@ -3552,27 +3557,31 @@ def fast_concat(
             this = self
         else:
             this = self.copy()
+        if issubclass(others.__class__, this.__class__):
+            # if this is a UVData object already, stick it in a list
+            others = [others]
```
mkolopanis, line 3562: I could imagine a (terrible) situation where this check could fail if say `this` is a HeraData object and `other` is a UVData object. But you can't add those two things together really anyway. But then the iteration where classes are checked again fail but not for the reasons that are desired. 

I wonder if this check should just straight up compare if others is a list/tuple/set (or other well behaved iterable).

mkolopanis (commented):
I think this looks really good. thanks for the work on this @AaronParsons . I have a few  small concerns. The one regarding the function signature I am not too worried about but I'd prefer some input on the matter from @plaplant or @bhazelton .

pyuvdata/uvdata/uvdata.py
bhazelton: Good point. Should I add `other` as a named parameter then in order to catch someone using it with it's name?

mkolopanis: I think that sounds pretty easy. Just cast the default to None and check if it gets populated during a call?

pyuvdata/uvdata/uvdata.py
bhazelton: sure, that makes sense.

pyuvdata/uvdata/uvdata.py
bhazelton: So it turns out this doesn't quite work. I added `other` as a named parameter, but if you pass it but don't pass anything for `others` it errors with `TypeError: fast_concat() missing 1 required positional argument: 'others'`.

bhazelton: We can just leave it named `other` even though it takes a list. That has the advantage of not changing the API but it makes the parameter naming less clear. Is there a way in python to tell if something was passed by position or name?

mkolopanis: That's a good point. I had not considered TypeError from missing a positional. We definitely could leave it, or change it. Probably won't make much difference in the end. We could always make a lot of noise when the release comes out to warn about the api change. I'm not sure if there is a way to tell honestly.

bhazelton: Hmm, I did some googling and can't tell that there's a way to detect calling a positional argument by name. Interestingly, starting in python 3.8 you can define parameters as positional only (you can already define parameters as keyword only in python 3.*).

plaplant: I think it's better to keep the API as it is, and just point out in the docstring that we support either a single object or an iterable for `other` (as we do now). But I don't feel super strongly, we could also change the API and make lots of noise during the release process.

pyuvdata/uvdata/uvdata.py
plaplant: I agree that we should check to see if `other` is already iterable, and build a single-element list if it's not. The consensus online seems to be to call the `iter()` method inside of a `try` and fail gracefully, so maybe something like this:
```
try:
    others = iter(other)
except TypeError:
    others = iter([other])

for other in others:
    # check that subclasses match, then combine objects
```
I'm wondering if the above might not work if we take multiple passes through the iterator (like first to do a compatibility/object test, then to actually combine objects), but I guess we could always test it and find out. 🙂 

It's also possible to check for `isinstance(other, collections.abs.Iterable)`, but that apparently misses objects that have a `__getitem()__` method but not an `__iter()__` method. Not sure if that's an edge case worth catching or not, but that check might be more robust than building a potentially-single-use iterator.

plaplant (commented):
Thanks @AaronParsons! This is a really great speed up. I've just added my $0.02 on the API change and iterator check. Otherwise I think this is good to go.

pyuvdata/uvdata/uvdata.py
bhazelton: So testing against `collections.abs.Iterable` lets in strings, which I don't think we want. My initial solution to this was to use `isinstance(other, (list, tuple, np.ndarray))`. I guess that could miss some legit things, but I feel like it should be sufficient.

pyuvdata/uvdata/uvdata.py
bhazelton: I just put it back to `other` to avoid the API change.

pyuvdata/uvdata/uvdata.py
bhazelton: I implemented it with the `isinstance(other, (list, tuple, np.ndarray))` for now. Happy to test something else if you want @plaplant, but since UVData object implement an `__iter__` method and somehow match `isinstance(other, collections.abs.Iterable)` neither of those worked out.

mkolopanis: I think that solution should work fine. yeah having `__iter__` on the object makes it tough.

plaplant (approved):
Thanks @AaronParsons and @bhazelton! I think this is good to go 👍

PR #948, opened 2020-11-16
Add `extra_keywords` to UVFlag

no-lex:
## Description

This PR adds support for the `extra_keywords` dict (already implemented in UVData) in UVFlag.

## Motivation and Context

Having `extra_keywords` dict in UVFlag appears to be the most straightforward way for SSINS (which inherits the UVFlag class for its INS object) to hold object-specific metadata.

Implementing `extra_keywords` in UVFlag allows the `INS` object to use the same semantics as the `SS` object, which uses the already implemented `extra_keywords` functionality for UVData.

The SSINS issue specifically referencing this is https://github.com/mwilensky768/SSINS/issues/78, and the context for this issue is at https://github.com/mwilensky768/SSINS/issues/73
## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/uvflag/tests/test_uvflag.py
+++ b/pyuvdata/uvflag/tests/test_uvflag.py
@@ -610,11 +610,11 @@ def test_read_write_nocompress_flag(uvdata_obj, test_outfile):
 def test_read_write_extra_keywords(uvdata_obj, test_outfile):
     uv = uvdata_obj
     uvf = UVFlag(uv, label="test")
-    uvf.extra_keywords = {"keyword1": 1}
+    uvf.extra_keywords = {"keyword1": 1, "keyword2": "string"}
     uvf.write(test_outfile, clobber=True, data_compression=None)
     uvf2 = UVFlag(test_outfile)
-    print("UVF2:" + str(uvf2.extra_keywords))
-    assert uvf2.extra_keywords["keyword1"] is not None
+    assert uvf2.extra_keywords["keyword1"] == 1
+    assert uvf2.extra_keywords["keyword2"] == "string"
 
 
 @pytest.mark.filterwarnings("ignore:The uvw_array does not match the expected values")
--- a/pyuvdata/uvflag/tests/test_uvflag.py
+++ b/pyuvdata/uvflag/tests/test_uvflag.py
@@ -158,11 +158,9 @@ def test_init_bad_mode(uvdata_obj):
 
 @pytest.mark.filterwarnings("ignore:The uvw_array does not match the expected values")
 def test_init_uvdata(uvdata_obj):
-    # for uvf initialization below
-    testdict = {"keyword1": 1, "keyword2": 2}
     uv = uvdata_obj
     uvf = UVFlag(
-        uv, history="I made a UVFlag object", label="test", extra_keywords=testdict
+        uv, history="I made a UVFlag object", label="test"
     )
     assert uvf.metric_array.shape == uv.flag_array.shape
     assert np.all(uvf.metric_array == 0)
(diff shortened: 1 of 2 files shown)


mkolopanis (commented):
This is looking like a good start. I would think we should also plumb the extra keywords through the read and write functions, which can easily be accomplished by mirroring what is done in the `uvh5.py` file. I suppose it is not a hard requirement or mine, but feels like it would be a good time to add this feature. @bhazelton @adampbeardsley strong opinions?

bhazelton:
There are new astropy deprecation warnings that are breaking some tests unrelated to the changes in this PR. We'll need to fix those in another PR and get it merged before this one can go.

bhazelton:
@no-lex We just merged in a fix for the astropy 4.2 changes that were breaking some tests on this PR. If you rebase this, I think those will be fixed.

no-lex:
@bhazelton It appears that the PR now passes the tests since being rebased, so besides the `CHANGELOG` I think this PR is ready to go.

One thing that we talked about earlier was the acceptability of adding `extra_keywords` to the `clear_unused_attributes()` function; this was done to allow `extra_keywords` to be optional while also allowing it to be handled by files. Is that OK or does it need to be worked in a different way?

bhazelton:
@mkolopanis @mwilensky768 can you comment on the question about the `clear_unused_attributes` method? I know almost nothing about it.

mwilensky768:
In my experience, this is the right thing to do with `clear_unused_attributes` if it is deleting your optional parameters, however @no-lex and I think this is a bandaid over a deeper problem.

It seems to me like there is a slowly bloating case statement in `clear_unused_attributes` that is just collecting optional items that we don't want accidentally cleared later on. It is unclear to me why `clear_unused_attributes` is called in certain places, such as in `read`. Presumably if some optional data is contained in a file, it should be read in without being cleared later since it was intended to go with the object. I think that we should have a think about the purpose of `clear_unused_attributes` and maybe come up with a cleaner way to handle optional parameters.

@mkolopanis what do you think?

mkolopanis:
I think that is probably the best way forward. `clear_unused_attributes` was a holdover from when `UVFlag` was not a subclass of `UVBase` back in the wild wild west of flagging. It was a quick and dirty way to just get rid of things that didn't make sense anymore when you change to a different representation.. We could just be more explicit about the parameters that are no longer applicable during a change also get deleted instead of this function. This probably sounds like more work but the book keeping would be a lot better.

Another  approach is to rework `UVFlag` object and then subclass it with each of the different types waterfall, antenna, baseline. The subclass would handle the transition logic, then re-populate a UVFlag object with all parameters on the subclass. As I write this it sounds a little crazy, but I'm imagining something similar to how the different file formats are read/written in `UVData`. This would be...a lot more work and I'm not even sure if it is worth doing, we might get ourselves into more trouble this way. I also worry about memory bloat with an approach like this, but that would require some detailed profiling.

Option A definitely seems the easiest way to implement this and it's even possible to transition to B later if we wanted to do that.
Also I'm not married to either of these and welcome other ideas.

mkolopanis:
I'm going to start an issue for this though. The `clear_unused_attributes` problem is way beyond the scope of this PR and sliding `extra_keywords` into that function is fine given the state of the code.

mkolopanis:
@bhazelton any idea why the pypi action is not even showing even if it is passed?

mkolopanis (approved):
This all looks pretty good to me, thanks for the addition @no-lex

bhazelton:
no, I don't know, but I can override it.

mkolopanis:
Ah in our haste we forgot a changelog entry. @no-lex would you mind making a new PR to update the changelog please?

no-lex:
@mkolopanis Yep, I put it up at https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/961

PR #956, opened 2020-12-04
Fix handling of astropy Time parameters for astropy 4.2

### LINKED ISSUE #953: Fix support for astropy Time UVParameters
pyradiosky uses astropy Time objects in UVParameter instances, which worked until astropy 4.2 dropped. Now the check method in UVBase is failing on those parameters with this traceback:

```
    def check(
        self, check_extra=True, run_check_acceptability=True, ignore_requirements=False
    ):
        """
        Check that required parameters exist and have the correct shapes.
    
        Optionally, check that the values are acceptable.
    
        Parameters
        ----------
        check_extra : bool
            If true, check shapes and values on all parameters,
        run_check_acceptability : bool
            Option to check if values in parameters are acceptable.
        ignore_requirements : bool
            Do not error if a required parameter isn't set.
            This allows the user to run the shape/acceptability checks
            on parameters in a partially-defined UVData object.
    
        """
        if check_extra:
            p_check = list(self.required()) + list(self.extra())
        else:
            p_check = list(self.required())
    
        for p in p_check:
            param = getattr(self, p)
            # Check required parameter exists
            if param.value is None:
                if ignore_requirements:
                    continue
                if param.required is True:
                    raise ValueError("Required UVParameter " + p + " has not been set.")
            else:
                # Check parameter shape
                eshape = param.expected_shape(self)
                # default value of eshape is ()
                if eshape == "str" or (eshape == () and param.expected_type == "str"):
                    # Check that it's a string
                    if not isinstance(param.value, str):
                        raise ValueError(
                            "UVParameter " + p + " expected to be " "string, but is not"
                        )
                else:
                    # Check the shape of the parameter value. Note that np.shape
                    # returns an empty tuple for single numbers.
                    # eshape should do the same.
                    try:
>                       np.shape(param.value)

../pyuvdata/pyuvdata/uvbase.py:322: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

args = (<Time object: scale='utc' format='iso' value=2015-03-01 00:00:00.000>,), kwargs = {}
relevant_args = (<Time object: scale='utc' format='iso' value=2015-03-01 00:00:00.000>,)

>   ???

<__array_function__ internals>:6: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <Time object: scale='utc' format='iso' value=2015-03-01 00:00:00.000>, function = <function shape at 0x7fd8e0c3c4d0>
types = (<class 'astropy.time.core.Time'>,), args = (<Time object: scale='utc' format='iso' value=2015-03-01 00:00:00.000>,)
kwargs = {}

    def __array_function__(self, function, types, args, kwargs):
        """Wrap numpy functions that make sense."""
        if function in self._APPLICABLE_FUNCTIONS:
            if function is np.broadcast_to:
                # Ensure that any ndarray subclasses used are
                # properly propagated.
                kwargs.setdefault('subok', True)
            elif (function in {np.atleast_1d,
                               np.atleast_2d,
                               np.atleast_3d}
                  and len(args) > 1):
                return tuple(function(arg, **kwargs) for arg in args)
    
            if self is not args[0]:
                return NotImplemented
    
            return self._apply(function, *args[1:], **kwargs)
    
        if self is args[0]:
            # Call the method rather than _apply(function.__name__),
            # since classes could override the method.
            name = {np.amax: 'max',
                    np.amin: 'min'}.get(function, function.__name__)
            method = getattr(self, name, None)
            if method is not None:
>               return method(*args[1:], **kwargs)
E               TypeError: 'tuple' object is not callable

../../../opt/anaconda3/envs/rasg/lib/python3.7/site-packages/astropy/utils/shapes.py:247: TypeError
```

It appears that we now need to call `np.shape` on the Time object's value (so change to `np.shape(param.value.value)` for the case of Time objects).

bhazelton:
## Description

Fix handling of astropy Time UVParameters in the UVBase `check` method, which broke with astropy 4.2.

## Motivation and Context

fixes #953 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/tests/test_uvbase.py
+++ b/pyuvdata/tests/test_uvbase.py
@@ -5,6 +5,7 @@
 """Tests for uvbase object.
 
 """
+from astropy.time import Time
 import pytest
 import numpy as np
 
@@ -79,6 +80,13 @@ class UVTest(UVBase):
 
         self._location = uvp.LocationParameter(
             "location", description="location", value=np.array(ref_xyz)
+        )
+
+        self._time = uvp.UVParameter(
+            "time",
+            description="astropy Time object",
+            value=Time("2015-03-01 00:00:00", scale="utc"),
+            expected_type=Time,
         )
 
         self._optional_int1 = uvp.UVParameter(
--- a/pyuvdata/uvbase.py
+++ b/pyuvdata/uvbase.py
@@ -10,6 +10,7 @@ Subclassed by UVData and Telescope.
 import copy
 import warnings
 
+from astropy.time import Time
 import numpy as np
 
 from . import parameter as uvp
@@ -318,13 +319,16 @@ class UVBase(object):
                     # Check the shape of the parameter value. Note that np.shape
                     # returns an empty tuple for single numbers.
                     # eshape should do the same.
-                    if not np.shape(param.value) == eshape:
+                    if isinstance(param.value, Time):
+                        this_shape = np.shape(param.value.value)
+                    else:
+                        this_shape = np.shape(param.value)
+
+                    if not this_shape == eshape:
                         raise ValueError(
-                            "UVParameter {param} is not expected shape. "
-                            "Parameter shape is {pshape}, expected shape is "
-                            "{eshape}.".format(
-                                param=p, pshape=np.shape(param.value), eshape=eshape
-                            )
+                            f"UVParameter {param} is not expected shape. "
+                            f"Parameter shape is {this_shape}, expected shape is "
+                            f"{eshape}."
                         )
                     if eshape == ():
                         # Single element


mkolopanis (approved):
Looks simple enough! thanks @bhazelton. Can't help but wonder if this is a bug in astropy somehow but at least this workaround is sufficient.

PR #963, opened 2020-12-07
cython util update

mkolopanis:
Updated the cython utilities to perform better under wider ranges of inputs.
## Description

Moved the fast axis dimension to be nblts for all sizes of inputs (before the 3 dimenions was the fast one). This provides significant improvements over the last version especially for larger arrays sizes.

## Motivation and Context

See my attached PDF for timing profiling. "Python" is the original implementation, "Cython" the current, and "Cython 2" the implementation in this pr. 
[xyz_calcs.pdf](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/5655578/xyz_calcs.pdf)

I guess it can be considered a bug fix, I fixed the bug that was my poor design last time I wrote this file.

Do we want a changelog update for this? I think there is considerable evidence that the phasing and ECEF/ENU code was actually _slowed down_ last time for large number of `Nblts`. That being said, I didn't want to focus too much on the phasing given the major overhaul that is coming in with the multi-source update.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/utils.pyx
+++ b/pyuvdata/utils.pyx
@@ -52,44 +52,39 @@ cpdef tuple baseline_to_antnums(numpy.ndarray[ndim=1, dtype=numpy.int64_t] basel
       _a1[i] = (_bl[i] - (_a2[i] + 1)) // 256 - 1
   return ant1, ant2
 
-
 @cython.boundscheck(False)
 @cython.wraparound(False)
 cdef numpy.ndarray[dtype=numpy.int64_t] _antnum_to_bl_2048(
-numpy.ndarray[ndim=1, dtype=numpy.int64_t] ant1,
-numpy.ndarray[ndim=1, dtype=numpy.int64_t] ant2,
+numpy.int64_t[::1] ant1,
+numpy.int64_t[::1] ant2,
 ):
-  cdef unsigned long n = ant1.size
-  cdef unsigned int i
+  cdef unsigned long n = ant1.shape[0]
+  cdef Py_ssize_t i
   cdef numpy.ndarray[ndim=1, dtype=numpy.int64_t] baselines = np.empty(n, dtype=np.int64)
   # make views as c-contiguous arrays of a known dtype
   # effectivly turns the numpy array into a c-array
   cdef numpy.int64_t[::1] _bl = baselines
-  cdef numpy.int64_t[::1] _a1 = ant1
-  cdef numpy.int64_t[::1] _a2 = ant2
 
   for i in range(n):
-    _bl[i] = 2048 * (_a1[i] + 1) + (_a2[i] + 1) + 2 ** 16
+    _bl[i] = 2048 * (ant1[i] + 1) + (ant2[i] + 1) + 2 ** 16
 
   return baselines
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
 cdef numpy.ndarray[dtype=numpy.int64_t] _antnum_to_bl_256(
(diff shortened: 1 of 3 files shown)


mkolopanis:
I can re-run the timings with larger number of iterations, 7 is small

mkolopanis:
Here's the same notebook but with more iterations. Once the number of inputs exceed ~1000 the numpy implementation of LLA from XYZ actually ends up being better. We could make a parallel implementation once the number of points is large enough, but I suspect the XYZ/LLA calculations are more often than not used for a single point like the array center. Still easy to implement that if we'd like.
[xyz_calcs.pdf](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/5666948/xyz_calcs.pdf)

pyuvdata/utils.pyx
```diff
@@ -118,105 +112,110 @@ cpdef numpy.ndarray[dtype=numpy.int64_t] antnums_to_baseline(
...
+cpdef numpy.ndarray[numpy.float64_t, ndim=2] _ENU_from_ECEF(
     numpy.float64_t[:, ::1] xyz,
     numpy.float64_t[::1] _lat,
     numpy.float64_t[::1] _lon,
     numpy.float64_t[::1] _alt,
 ):
-  cdef int i
-  cdef int nblts = xyz.shape[0]
-  cdef numpy.ndarray[dtype=numpy.float64_t, ndim=2] enu = np.empty((nblts, 3), dtype=np.float64)
+    cdef Py_ssize_t i
```
plaplant, line 182: I don't really care about indentation per se, but should this be 2 spaces instead of 4 to be consistent with the rest of the file?

pyuvdata/utils.pyx
```diff
@@ -228,88 +227,114 @@ cpdef numpy.ndarray[dtype=numpy.float64_t] _ECEF_FROM_ENU(
...
+        xyz[2, i] = cos_lat * enu[1, i] + sin_lat * enu[2, i] + xyz_center[2]
+
+    return _xyz
 
 # inital_uvw is a memoryviewed array as an input
 @cython.boundscheck(False)
 @cython.wraparound(False)
-cpdef numpy.ndarray[dtype=numpy.float64_t] _phase_uvw(
+cpdef numpy.ndarray[dtype=numpy.float64_t, ndim=2] _phase_uvw(
     numpy.float64_t ra,
```
plaplant, line 269: Should we be catching cases (either here or in `utils.py`) where `ra` and `dec` are passed as arrays? I know that doesn't make sense for how it's usually called, but we may want to sanitize the input (or at least make sure the user isn't doing anything unexpected).

plaplant (commented):
Thanks for the great work @mkolopanis! I just had a couple of small comments.

pyuvdata/utils.pyx
mkolopanis: yeah that's a good point, I think we would need to do this in `utils.py`. I'll make sure it is sanitized against length 1 arrays, or scalars as arrays.

mkolopanis: So it looks like in `utils.py`, we expect inputs to be `float` type but don't do any explicit checking. I can add these checks in, but it will be a little more work for the multi-source branch to pull out later. Up to your discretion

plaplant: Let's not worry about it then. Probably not a huge deal, especially because in the docstring we explicitly say it's a `float` and not an `ndarray`.

plaplant (approved):
Thanks @mkolopanis!

PR #965, opened 2020-12-09
Relaxed type comparisons

### LINKED ISSUE #947: Comparison with numpy dtypes
Should we update this comparison to use `np.integer`? `np.int` is basically just a wrapper around the built-in `int` class whereas `numpy.integer` is a baseclass for all numpy-based integers of different precisions.
https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/eedb1e10ef2af8e4a019e7df2abed012d23c6471/pyuvdata/uvdata/uvdata.py#L5094

by extension should we alter this line to compare against numpy.floating?https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/eedb1e10ef2af8e4a019e7df2abed012d23c6471/pyuvdata/uvdata/uvdata.py#L2768

possible there are other places too. It hasn't caused any issues yet, but I can imagine an issue showing up where a non-standard integer is passed to these functions.

mkolopanis:
Relaxes the requirement that `expected_type` for a UVParameter object must match precision. Also relaxes some check against explicit precision in `if/else` clauses.
closes #947
## Description

Makes `expected_type` a tuple of more generic type classes based on inputs (e.g. `expected_type=float` gets cast to `expected_type=(float, numpy.floating)` this allows the parameter value to change precision without breaking a `UVBase.check` call. 
These generic expected types **do not** change the actual precision of the parameter value, just let it change. There are tests is mwa_corr_fits which I've linked that explicitly test against input precisions.
https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/90537f78230d3d34f5db4d39a9f2a18373435437/pyuvdata/uvdata/tests/test_mwa_corr_fits.py#L565-L582

I've added a `strict_type_check` kwarg when instantiating a `UVParameter` object. This negates the generic conversion and requires the value on the parameter to match exactly the input type when a `UVBase.check` call is made.

## Motivation and Context

Some `if/else` statements were checking if inputs were a very specific precision during `isinstance` this should alleviate that in the case where a  user inputs an unusual precision. 

TODO: 
- changelog once we are ready to merge.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

Breaking change checklist:

### DIFF
--- a/pyuvdata/uvbase.py
+++ b/pyuvdata/uvbase.py
@@ -12,6 +12,7 @@ import warnings
 
 from astropy.time import Time
 import numpy as np
+from astropy.units import Quantity
 
 from . import parameter as uvp
 from . import __version__
@@ -356,8 +357,12 @@ class UVBase(object):
                                         " be: " + str(param.expected_type)
                                     )
                         else:
-                            # Array
-                            if not isinstance(param.value.item(0), param.expected_type):
+                            # Array or quantity
+                            if isinstance(param.value, Quantity):
+                                check_val = param.value.item(0).value
+                            else:
+                                check_val = param.value.item(0)
+                            if not isinstance(check_val, param.expected_type):
                                 raise ValueError(
                                     "UVParameter " + p + " is not the appropriate"
                                     " type. Is: "
--- a/pyuvdata/uvbase.py
+++ b/pyuvdata/uvbase.py
@@ -377,7 +377,7 @@ class UVBase(object):
                                         )
                                     else:
                                         # matches expected type
-                                        continue
+                                        continue  # pragma: no cover
                                 else:
                                     # Expected type is not a Quantity subclass
                                     # Assuming it is a data type like float, int, etc
--- a/pyuvdata/uvbase.py
+++ b/pyuvdata/uvbase.py
@@ -85,7 +85,7 @@ class UVBase(object):
 
         # String to add to history of any files written with this version of pyuvdata
         self.pyuvdata_version_str = (
-            "  Read/written with pyuvdata version: " + __version__ + "."
+            f"  Read/written with pyuvdata version: {__version__ }."
         )
 
     def prop_fget(self, param_name):
@@ -231,8 +231,9 @@ class UVBase(object):
                 other_required.append(p)
             if set(self_required) != set(other_required):
                 print(
-                    "Sets of required parameters do not match. Left is {lset},"
-                    " right is {rset}".format(lset=self_required, rset=other_required)
+                    "Sets of required parameters do not match. "
+                    f"Left is {self_required},"
+                    f" right is {other_required}."
                 )
                 return False
 
@@ -245,8 +246,9 @@ class UVBase(object):
                     other_extra.append(p)
                 if set(self_extra) != set(other_extra):
                     print(
-                        "Sets of extra parameters do not match. Left is {lset},"
-                        " right is {rset}".format(lset=self_extra, rset=other_extra)
+                        "Sets of extra parameters do not match. "
+                        f"Left is {self_extra},"
+                        f" right is {other_extra}."
                     )
                     return False
 
@@ -260,8 +262,8 @@ class UVBase(object):
                 other_param = getattr(other, p)
                 if self_param != other_param:
                     print(
-                        "parameter {} does not match. Left is {},"
-                        " right is {}".format(p, self_param.value, other_param.value)
+                        f"parameter {p} does not match. Left is {self_param.value},"
+                        f" right is {other_param.value}."
                     )
                     p_equal = False
             return p_equal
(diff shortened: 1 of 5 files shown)


pyuvdata/uvbase.py
```diff
@@ -356,8 +357,12 @@ def check(
                                         " be: " + str(param.expected_type)
                                     )
                         else:
-                            # Array
-                            if not isinstance(param.value.item(0), param.expected_type):
+                            # Array or quantity
+                            if isinstance(param.value, Quantity):
+                                check_val = param.value.item(0).value
+                            else:
+                                check_val = param.value.item(0)
```
bhazelton, line 364: I think this will pass if the parameter is set to be an Astropy Quantity but the expected type is "float". Do we want that? Or should we make it so that the expected type should be a Quantity to pass? I'm concerned that in pyradiosky we actually set the expected type to be a Quantity in some cases and that might fail here?

mkolopanis: oh yeah that is an interesting point. This is more about evaluating the type of quantity, and just letting quantities slide through unnoticed but still supported. :thinking: I'm more than willing to remove it.

mkolopanis: how do you feel about this. I'm not sure the logic is 100% there at the moment, but basically it adds an additional check if the `expected_type` is a subclass of Quantity, if it is not and the `expected_type` is a data type like `float`, `int`, etc, then it will fall back to the other check.


<details>

```
                            if isinstance(param.value, Quantity):
                                # user put expected type as a type of quantity object
                                # not a more generic type of number.
                                if issubclass(param.expected_type, Quantity):
                                    # Verify the param is an instance
                                    # of the specific Quantity type
                                    if not isinstance(param.value, param.expected_type):
                                        raise ValueError(
                                            f"UVParameter {param} is a Quantity object "
                                            "but not the appropriate type. "
                                            f"Is {type(param.value)} but "
                                            f"expected {param.expected_type}."
                                        )
                                   else:
                                       continue
                                else:
                                    # expected type is not a type of Quantity
                                    # assuming it is a date a type, continue with check below
                                    check_val = param.value.item(0).value
                            else:
                                check_val = param.value.item(0)
                            if not isinstance(check_val, param.expected_type):
                                raise ValueError(
                                    "UVParameter " + p + " is not the appropriate"
                                    " type. Is: "
                                    + str(param.value.dtype)
                                    + ". Should be: "
                                    + str(param.expected_type)
                                )
```
</details>

bhazelton: That looks reasonably good to me. I'll note that we should use `param.name` or `p` to in the print statement rather than just `param` because that gives a useless nonsense string (I actually changed that in #967 because I was trying to debug and couldn't figure out which parameter it was erroring on).

I'm a little confused about the part that reads:
```
# expected type is not a type of Quantity
# assuming it is a date a type, continue with check below
check_val = param.value.item(0).value
```

I think there's a typo in the first comment line, but I think you're assuming it's a Time object? I don't really understand why...

mkolopanis: I was thinking more like you have an array of quantities, but you set the expected type to `float` not `Quantity`. That would trigger the case with the comment you just mentioned. Maybe the overloading of the word "type" isn't helping here.

mkolopanis:
regarding the coverage of that one `continue` line see this issue from coverage.py. Specifically the linked comment explains why we see this. https://github.com/nedbat/coveragepy/issues/198#issuecomment-399705984

pyuvdata/uvbase.py
bhazelton: ok, so it has nothing to do with a Time object? I guess I was confused by the comment line that read "assuming it is a date a type"

mkolopanis: correct, I was trying to find words to describe the different kids of numbers (float,int, etc), data types is what came to mind.

pyuvdata/uvbase.py
```diff
@@ -356,8 +358,34 @@ def check(
...
+                                            f"expected {param.expected_type}."
+                                        )
+                                    else:
+                                        # matches expected type
+                                        continue
+                                else:
+                                    # Expected type is not a Quantity subclass
+                                    # Assuming it is a data type like float, int, etc
+                                    # continuing with check below
+                                    check_val = param.value.item(0).value
```
bhazelton, line 385: Ok, now I think I understand what you're doing but I'm not sure if it's what we want. I think this will allow a situation to pass where you specified that the expected type was a float and the actual parameter is a Quantity. Do we want that to pass?

mkolopanis: Right that's with what would happen. I had the though that choice was up to the user :thinking:

Quantities definitely add a layer of complexity. This would let you like "upgrade" a parameter to a quantity (like the situation you detailed), but never the other way if the type is expected to be a quantity specifically. I personally would prefer to have maybe a new type of parameter specifically for quantites that would track the expected type like it is now, but maybe also the equivalency class of the unit. But not sure that was really in the scope here, and probably needs more flushing out anyway.

bhazelton:
On the continue issue, that's a known problem, just mark it as not needing to be covered with `# pragma: no cover`

bhazelton (approved):
Looks good. Hurray for fstrings!

PR #943, opened 2020-11-04
Fix handling of FHD antenna positions and Cotter telescope frame

bhazelton:
## Description

Properly treat FHD antenna positions as being in the relative rotated ECEF frame, just like uvfits antenna positions. Add a test against raw MWA correlator files and a Cotter derived file to verify the conversion, which fails on `main`.

Also fix the handling of the telescope frame for Cotter. Cotter leaves out this information, which led to using the known telescope location. But that meant that the reference position for the antenna positions changed without the antenna positions being properly updated. Now we detect that it is a Cotter file and set the telescope frame to 'ITRF' as it should be.

## Motivation and Context

Fixes #942 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation change (documentation changes only)
- [ ] Version change
- [ ] Build or continuous integration change

## Checklist:

- [x] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/.github/CONTRIBUTING.md).
- [x] My code follows the code style of this project.

Bug fix checklist:
- [x] My fix includes a new test that breaks as a result of the bug (if possible).
- [x] All new and existing tests pass.
- [x] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/CHANGELOG.md).

PR #967, opened 2020-12-13
Add support for future array shapes

### LINKED ISSUE #912: Update UVH5 memo to include flexible spectral window discussion
Once #909 is merged, we will technically have changed the UVH5 file specification (in a good way!). We should update the UVH5 memo to reflect that the file format can support multiple spectral windows, and that there are several new items in the header that have be added (or have changed shape).

### LINKED ISSUE #922: Some Header datasets in UVH5 memo lack units etc
The UVH5 memo specifies a number of `Header` datasets for a UVH5 file.  The datatypes and dataspaces (dimensions) of these datasets is well defined in the memo, but some of them lack information regarding what units should be used and occasionally other details.  Here is a list (that ended up being longer than I thought it would be when I started):

1. `phase_type`: Needs more clarity.  The two specified values are `phased` and `drift`, but it's not clear what `phased` means.  Is it that the telescope has physically tracked the phase center for the duration of the observation or that the fringe phase has been "stopped"?  This feels like a field that has been specified for fixed antennas but needs more fleshing out for steerable antennas.
1. `uvw_array`: Lack units. It has no `Nfreqs` dimension it can't be wavelengths, but is it meters or nanoseconds or ???
1. `antenna_positions`: Lacks units.  The frame is specified as "relative to the _telescope location_ (in the ITRF frame)", but "_telescope location_", despite being in italics, is not defined elsewhere other than perhaps the `latitude`, `longitude`, `altitude` (LLA) fields.  The sample file `zen.2458432.34569` has antenna positions such as `(5488676, -105, -3240024)`, which seems like ECEF coordinates in meters that have been rotated in longitude to the longitude of the LLA. The "LLA to ECEF" reference should also be specified (WGS84?).
1. `dut1`: Lacks units.  Most likely it's seconds, but should be explicitly stated as such for completeness.
1. `earth_omega`: Why degrees (per day) when all other angles are radians (other than `latitude`/`longitude` which are specified in degrees so commonly that it's tolerable to store them as degrees in the data file)?  I'm not advocating a change to this, but just pointing out that it would be nice to give some historical context for this design decision.
1. `gst0`: Lacks units.  Many possibilities come to mind, no "obvious" default unit for this (IMHO). Also specified as "Greenwich sidereal time at midnight on reference date", but is that Greenwich or "local" midnight?  Not clear whether this is apparent or mean sidereal time.  [Why is this even here?  Isn't this derivable from e.g. `time_array` values and `dut1`?]
1. `rdate`: This string field lacks format/resolution details.  Should be more clearly specified (even if its specified as a free-form "you figure it out" format).
1. `x_orientation`: This is already being discussed in #854, but free form text seems suboptimal IMHO.
1. `uvplane_reference_time`: Lacks units and origin (and frame if you want to get relativistic)
1. `phase_center_frame`: Not clear what the two specified values mean (`gcrs` and `icrs`).  Also not clear how `uvw_array` can be "phased to" a frame.  Isn't `(u,v,w)` just `B dot S` irrespective of any phasing that's been applied to the data?
1. `lst_array`: A brief explanation of the benefits of including this metadata-that-can-be-derived-from-other-metadata would be useful IMHO.
1. `extra_keywords`: This Header "subgroup" specifies that it is for the storage of "extra" (presumably non-required).  It says that this group contains datasets whose names are the "extra keywords" and that the extra values "are expected to be scalars, but UVH5 makes no formal restriction on this".  I think it would be good to specify further that they be comprised of HDF5 datatypes when feasible (e.g. as opposed to Python or Numpy "pickles", as is the case for sample file `zen.2458432.34569`).

### LINKED ISSUE #924: Conjugation and UVW conventions are not specified in UVH5 memo
The UVH5 memo describes the `/Header/ant_1_array` and `/Header/ant_2_array` datasets that contain an antenna index for each baseline-time sample in `/Data/visdata` (et al.), but the memo does not specify the conjugation convention or UVW convention.  Is the visibility supposed to be `ant1 * conj(ant2)` or `conj(ant1) * ant2` or is it intentionally left unspecified?  Likewise, for the UVW values, are they `pos(ant1) - pos(ant2)` or `pos(ant2) - pos(ant1)`?  Are there any restrictions such as `ant1 <= ant2` for any given baseline-time sample?

### LINKED ISSUE #936: Antenna positions in `pyuvdata/data/zen.2458432.34569.uvh5` use incorrect coordinate frame
As mentioned in #922, the `pyuvdata/data/zen.2458432.34569.uvh5` file has antenna positions that appear to be in an Earth Centered frame that is rotated from the ECEF ITRF frame.  This differs from all that other UVH5 files in `pyuvdata/data` in that 1) it is the only file that has Earth Centered antenna positions and 2) the coordinate frame is rotated such that the "X" axis points to the local meridian rather than 0 longitude.  If this file exists as an intentional "error case" it should be given a more descriptive name so that it will not be mistaken as an example of correct antenna positions.  If it's not an intentional error, this file should be corrected or removed to avoid confusion.

### LINKED ISSUE #939: Add a method on UVData objects to allow users to convert to the future array sizes (no spw axis, channel_width length Nfreqs)
This would add a flag on the object called something like `future_array_sizes` and a method to convert the object to have the future array sizes. It would also require logic in nearly every method to handle both the current and future sizes (of course we have to write this all at some point anyway).

This would allow for a graceful transition when the user wanted to make that transition, rather than when we release our next generation version. It would also allow us to find all the places that need to change and start testing it now.

When we release the version that drops support for the old array shapes, the method will continue to exist but be a no-op with deprecation warnings for a few major versions so users can remove the call to that method on their own time scale.

### LINKED ISSUE #983: Add versioning info for UVH5 file format
As we add/change functionality (e.g. array shapes, support for flexible spws & multiple objects) having a versioning system would be helpful. Especially since other arrays are considering using the format.

bhazelton:
## Description

- Add `use_future_array_shapes` method to `UVData` to allow users to switch over to using the future array shapes now, as well as `use_current_array_shapes` method to revert to the standard shapes (mostly for testing).
- Add a new parameter on `UVData` objects to track whether or not the object is using the future shapes.
- Add support for objects with the future shapes to all UVData methods and to `utils.uvcalibrate` and `utils.apply_flag`.
- Add support for initializing UVFlag objects from UVData objects with the future shapes.
- Add support for writing uvfits, miriad and uvh5. For uvh5, the arrays are written in the future shapes and those shapes are supported on the read.
- Added a `reorder_freqs ` method.

Other cleanup:
- Reordered the methods in miriad.py to be in a sensible order (so methods are defined before they are called).
- Converted several assertions that tested user inputs in miriad.py to checks with ValueErrors.
- Moved one method in uvh5.py to define it before it was called.
- Added a bunch of testing & handling for flexible spectral windows in various methods (`add`, `fast_concat`).

Still todo:
- [x] implement uvh5 version tracking
- [x] update uvh5 memo to describe versioning.
- [x] update tutorial. Need to add section on `reorder_freqs` as well as a section on `use_future_array_shapes`.
- [x] update changelog to mention `reorder_freqs` and uvh5 versioning.
- [x] potentially move `test_flex_spw_read` from `test_mir.py` to `test_uvdata.py`

## Motivation and Context

closes #939
closes #983 
closes #912 
closes #922 
closes #924 
closes #936

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:


bhazelton:
@mkolopanis something is going crazy with pytest on this PR. I don't think it has anything to do with my changes.

plaplant:
@bhazelton I am also seeing lots of test failures on the current `main` build that was kicked off from merging #943. Locally I've fixed it by downgrading pytest (from 6.2.0 to 6.1.2). I'll make a quick PR to that effect to see if that fixes things on our CI.

pyuvdata/uvdata/uvh5.py
```diff
@@ -532,12 +561,16 @@ def _read_header(
...
         self.spw_array = header["spw_array"][:]
 
+        if self.freq_array.ndim == 1:
+            assert np.asarray(header["channel_width"]).size == self.freq_array.size
+            self._set_future_array_shapes()
+
         # Pull in the channel_width parameter as either an array or as a single float,
         # depending on whether or not the data is stored with a flexible spw.
-        if self.flex_spw:
+        if self.flex_spw or np.asarray(header["channel_width"]).ndim == 1:
```
kartographer, line 573: This is admittedly bit of a corner case, but is there ever a situation where `flex_spw = True` and `np.asarray(header["channel_width"]).ndim != 1:`? I suspect the answer is no (seems like that case should throw an error), in which case I might ask if there is any advantage in using this new logic instead of the original one that just checks `flex_spw`.

pyuvdata/uvdata/uvdata.py
```diff
@@ -689,6 +694,73 @@ def metadata_only(self):
...
+        self.channel_width = np.zeros(self.Nfreqs, dtype=np.float) + self.channel_width
+
+    def use_current_array_shapes(self):
+        """
+        Change the array shapes of this object to match the current future shapes.
+
+        This method sets allows users to convert back to the current array shapes.
+        This method sets the `future_array_shapes` parameter on this object to False.
+        """
+        if self.Nspws == 1:
```
kartographer, line 738: I think this needs handling for `flex_spw` explicitly -- one can have a single SPW that is irregularly sampled/varying channel widths (i.e., due to flagging of RFI). In that case, I think one would handle this array in the identical way that one would deal with `freq_array`.

pyuvdata/uvdata/uvdata.py
```diff
@@ -689,6 +694,73 @@ def metadata_only(self):
...
+
+        """
+        self._set_future_array_shapes()
+        if not self.metadata_only:
+            for param_name in self._data_params:
+                setattr(self, param_name, (getattr(self, param_name))[:, 0, :, :])
+
+        self.freq_array = self.freq_array[0, :]
+
+        self.channel_width = np.zeros(self.Nfreqs, dtype=np.float) + self.channel_width
```
kartographer, line 729: Same comment as (directly) below -- needs a slightly different handling for when `flex_spw = True`

pyuvdata/uvdata/uvh5.py
bhazelton: Yes, there is. We agreed that we want the uvh5 files to have the same shapes as the object in memory. So if you call `set_future_array_shapes` and then write a uvh5 file, it may have only one SPW but the channel width will be an array. In that case we read it in properly and set the `future_array_shapes` boolean on the object. In the future, of course, this will always be the case but we will have to support old files indefinitely.

pyuvdata/uvdata/uvdata.py
bhazelton: Ok, I think this is a case where I was conflating `flex_spw` with `Nspws>1`, I didn't think about having one spw with variable channel widths within it.

pyuvdata/uvdata/uvdata.py
bhazelton: It probably should and I'll change it, but I think the tests are passing because in the case that `self.channel_width` is already an array, this doesn't affect it at all (just adds zeros to every channel width).

pyuvdata/uvdata/uvdata.py
```diff
@@ -4575,7 +4954,11 @@ def _select_metadata(
 
         if freq_inds is not None:
             self.Nfreqs = len(freq_inds)
-            self.freq_array = self.freq_array[:, freq_inds]
+            if self.future_array_shapes:
+                self.freq_array = self.freq_array[freq_inds]
+                self.channel_width = self.channel_width[freq_inds]
```
kartographer, line 4959: I think this is another place that needs handling for `flex_spw` (with respect to `channel_width`) -- I must have missed adding handling for this in the original PR.

pyuvdata/uvdata/uvdata.py
```diff
@@ -3732,40 +4062,59 @@ def fast_concat(
...
-            this.Nfreqs = sum([this.Nfreqs] + [obj.Nfreqs for obj in other])
+            if this.future_array_shapes:
+                this.freq_array = np.concatenate(
+                    [this.freq_array] + [obj.freq_array for obj in other]
+                )
+                this.channel_width = np.concatenate(
+                    [this.channel_width] + [obj.channel_width for obj in other]
+                )
+                this.Nfreqs = sum([this.Nfreqs] + [obj.Nfreqs for obj in other])
+            else:
```
kartographer, line 4073: Another place where I think I failed to add handling for the `channel_width` array with `flex_spw=True` -- would be good to add this in there as well (feel free to hand these off to me, since they are my oversights).

pyuvdata/uvdata/uvdata.py
bhazelton: Yes, thanks for finding that! I'm bothered that tests passed without that. Means I'm probably missing a test that I need to add.

pyuvdata/uvdata/uvdata.py
```diff
@@ -3192,6 +3409,15 @@ def __add__(
         else:
             fnew_inds = []
 
+        # if channel width is an array and there's any overlap in freqs,
+        # check extra params
+        if this.future_array_shapes or this.Nspws > 1:
```
kartographer, line 3414: I think the latter part of this should be `or this.flex_spw == True` rather than `or this.Nspws > 1`.

pyuvdata/uvdata/uvdata.py
```diff
@@ -3232,6 +3458,14 @@ def __add__(
                     rtol=this._lst_array.tols[0],
                     atol=this._lst_array.tols[1],
                 )
+            elif a == "_channel_width" and this.future_array_shapes:
```
kartographer, line 3461: Another case where some added handling for `flex_spw=True` and `this.future_array_shapes=False` would be good (gah, seems like I missed a number of these...).

pyuvdata/uvdata/uvdata.py
```diff
@@ -2415,7 +2587,10 @@ def unphase_to_drift(self, phase_frame=None, use_ant_pos=False):
                 / const.c.to("m/s").value
                 * self.freq_array.reshape(1, self.Nfreqs)
             )
-            phs = np.exp(-1j * 2 * np.pi * (-1) * w_lambda[:, None, :, None])
+            if self.future_array_shapes:
+                phs = np.exp(-1j * 2 * np.pi * (-1) * w_lambda[:, :, None])
+            else:
+                phs = np.exp(-1j * 2 * np.pi * (-1) * w_lambda[:, None, :, None])
```
kartographer, line 2593: Not an actual problem, but there's a good chunk of the `phase` method (and a few others related to w-projection tasks) that's been updated in PR #979 -- would be good to make sure that our two PRs don't inadvertently collide! 😅

pyuvdata/uvdata/uvdata.py
bhazelton: yes, I just found this one and some others. I'm realizing there's no testing for the add or fast concat (and maybe the select) methods with flexible spws. I can add some, can you point me at some good files to use?

pyuvdata/uvdata/uvdata.py
bhazelton: yep, found that one just now too.

pyuvdata/uvdata/uvdata.py
bhazelton: yes, I found this one too.

pyuvdata/uvdata/uvdata.py
bhazelton: Yeah, I suspect we'll have to merge one and then adjust the other after a rebase. If we have good testing coverage I think we'll catch things but we'll need to be careful.

pyuvdata/uvdata/miriad.py
```diff
@@ -41,1486 +41,1529 @@ def _pol_to_ind(self, pol):
...
-                        warn_string = (
-                            warn_string + "for {telescope_name} in known "
-                            "telescopes. Using altitude value from known "
-                            "telescopes and lat/lon from "
-                            "file.".format(telescope_name=telescope_obj.telescope_name)
-                        )
-                        warnings.warn(warn_string)
+            # Need the array direction here since channel_width is always supposed
+            # to be > 0, but channels can be in decending freq order
+            if self.future_array_shapes:
```
kartographer, line 1533: The if/else logic seems to lead to the same outcome -- the math operation appears to be the same. Is something missing here?

bhazelton: Oops, yes. The second one should have be `freq_dir = np.sign(np.diff(freq_array_use[0, ([0, -1])]))` since the freq_array has 2 dimensions. Not sure why that didn't cause a problem in tests.

pyuvdata/uvdata/uvdata.py
kartographer: I suspect the either the MIR test file (`sma_test.mir`) or the CARMA-SZA test file (`carma_miriad`) would work, since they both have multiple spws in the dataset that you could in principle select on and then add back together. Again, if you want, I'm happy to try to add that in there, since they're effectively leftovers from my PR.

pyuvdata/uvdata/miriad.py
```diff
@@ -41,1486 +41,1529 @@ def _pol_to_ind(self, pol):
...
+        uv.add_var("npol", "i")
+        uv["npol"] = self.Npols
+
+        #####################################################
+        # Frequency information here
+        uv.add_var("nspect", "i")
+        uv["nspect"] = self.Nspws
+
+        if self.future_array_shapes:
+            freq_array_use = self.freq_array
```
kartographer, line 1495: You might need to similarly add something like this for `channel_width` -- it may only get used when `flex_spw=True`, but the array shape will of course be different when `future_array_shapes=True`.

pyuvdata/uvdata/uvdata.py
bhazelton: Ok, I'd be happy for you to. I think I have all the logic updated properly, I'll push that up. I do want to make sure we're exercising everything so I'd like to see that your new tests fail before my updated logic is applied (so you'd need to check out the commit before my update or just not pull my latest commit at first). In particular, we need to test cases with one flexible spw.

pyuvdata/uvdata/miriad.py
kartographer: Whoops -- I take that back. `channel_width` is already of shape `("Nfreqs",)`... my bad!

pyuvdata/uvdata/uvdata.py
kartographer: Yeah, I think I understand now why it's passing -- `channel_width` was always set to be of shape `(Nfreqs,)` for flexible spw datasets, presumably because it didn't make sense to _add_ a vestigial axis where none existed before.

pyuvdata/uvdata/miriad.py
bhazelton: oh, I see. This isn't needed because I make it 1D in either case earlier in the code.

kartographer:
Hi @bhazelton -- as you've already noticed, I've gone through the PR, trying to keep an eye on anything related to `flex_spw`. Overall I think it looks great, most of the comments that I've added are relatively minor (and I'll try to get those tests together ASAP!).

pyuvdata/uvdata/uvh5.py
```diff
@@ -1018,6 +1081,11 @@ def read_uvh5(
                 strict_uvw_antpos_check,
             )
 
+        # For now, always use current shapes when data is read in, even if the file
```
plaplant, line 1084: This comment jogged my memory about the UVH5 versioning discussed in #983. Let's decide what scheme we'd like to implement and add it as part of this PR (since this is a significant change!).

pyuvdata/uvdata/uvdata.py
```diff
@@ -122,9 +123,7 @@ def __init__(self):
         self._Nspws = uvp.UVParameter(
             "Nspws",
             description="Number of spectral windows "
-            "(ie non-contiguous spectral chunks). "
-            "More than one spectral window is not "
-            "currently supported.",
+            "(ie non-contiguous spectral chunks). ",
```
plaplant, line 126: Thanks again @kartographer for making this description obsolete 🎉

pyuvdata/uvdata/uvdata.py
```diff
@@ -689,6 +694,76 @@ def metadata_only(self):
...
+
+        This method sets allows users to convert to the planned array shapes changes
+        before the changes go into effect. This method sets the `future_array_shapes`
+        parameter on this object to True.
+
+        """
+        self._set_future_array_shapes()
+        if not self.metadata_only:
+            for param_name in self._data_params:
+                setattr(self, param_name, (getattr(self, param_name))[:, 0, :, :])
```
plaplant, line 725: Very clever! This took me a few seconds to parse in my head, can we add a quick comment explicitly saying that we're removing the length-1 `spw` axis for data-like parameters?

plaplant (commented):
Thanks so much @bhazelton! This is a tremendous amount of work! I mostly have a few small questions/comments.

pyuvdata/uvflag/uvflag.py
```diff
@@ -1198,11 +1198,20 @@ def to_baseline(
         if self.type == "waterfall":
             # Populate arrays
             if self.mode == "flag":
-                arr = np.zeros_like(uv.flag_array)
+                if issubclass(uv.__class__, UVData) and uv.future_array_shapes:
```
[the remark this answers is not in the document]
bhazelton: I think we should leave it to future work, there's more that needs to be done on UVFlag, UVCal and UVBeam to support flexible spectral windows as well as the future array shapes.

pyuvdata/uvdata/miriad.py
```diff
@@ -41,1486 +41,1526 @@ def _pol_to_ind(self, pol):
...
-        uv["longitu"] = self.telescope_location_lat_lon_alt[1].astype(np.double)
-        uv.add_var("nants", "i")
+        # select on time range
+        if time_range is not None:
+            # type check
+            err_msg = (
+                "time_range must be a len-2 list of Julian Date floats, "
+                "Ex: [2458115.2, 2458115.6]"
+            )
+            assert isinstance(time_range, (list, np.ndarray)), err_msg
```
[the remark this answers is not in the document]
bhazelton: Ok, I'll push a fix shortly to clean these all up.

pyuvdata/uvdata/uvh5.py
```diff
@@ -89,7 +89,7 @@ def _read_complex_astype(dset, indices, dtype_out=np.complex64):
         raise ValueError(
             "output datatype must be one of (complex, np.complex64, np.complex128)"
         )
-    dset_shape = _get_dset_shape(dset, indices)
+    dset_shape, indices = _get_dset_shape(dset, indices)
```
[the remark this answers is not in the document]
bhazelton: I'll fix that, thanks for pointing it out.

pyuvdata/uvdata/uvh5.py
```diff
@@ -532,12 +561,16 @@ def _read_header(
             proc = self.set_lsts_from_time_array(background=background_lsts)
 
         # get frequency information
-        self.freq_array = header["freq_array"][:, :]
+        self.freq_array = header["freq_array"][:]
         self.spw_array = header["spw_array"][:]
 
+        if self.freq_array.ndim == 1:
+            assert np.asarray(header["channel_width"]).size == self.freq_array.size
```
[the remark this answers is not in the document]
bhazelton: Yes, good idea.

pyuvdata/uvdata/uvh5.py
bhazelton: sounds good!

kartographer:
@bhazelton -- I think I've finished putting in my changes, including the test coverage for the `add` and `fast_concat` methods when `flex_spw=True`. Take a look and let me know if there's anything that needs adjusting, otherwise I'll leave this PR in your hands again 🙂

pyuvdata/uvdata/uvdata.py
```diff
@@ -2460,6 +2460,206 @@ def reorder_blts(
...
+        ----------
+        spw_order : str or array_like of int
+            A string describing the desired order of spectral windows along the
+            frequecy axis. Allowed strings include `number` (sort on spectral window
+            number) and `freq` (sort on median frequency). A '-' can be appended
+            to signify descending order instead of the default ascending order,
+            e.g., if you have SPW #1 and 2, and wanted them ordered as [2, 1],
+            you would specify `-number`. Alternatively, one can supply an index array
+            of length Nspws that specifies the new order. Default is to apply no
+            sorting of spectral windows.
```
bhazelton: This last comment confuses me, it looks like the default is `spw_order="number"`, which I believe means it will by default sort by spws.

pyuvdata/uvdata/uvdata.py
```diff
@@ -2460,6 +2460,206 @@ def reorder_blts(
...
+        channel_order : str or array_like of int
+            A string describing the desired order of spectral windows along the
```
bhazelton, line 2493: should this line read "A string describing the desired order of frequency channels..." instead?

kartographer: Fixed in 6e0eeb9 (ah, the dangers of blindly copying and pasting...).

pyuvdata/uvdata/tests/test_mir.py
```diff
@@ -124,7 +148,9 @@ def test_write_mir(uv_in_uvfits, err_type=NotImplementedError):
         mir_uv.write_mir("dummy.mir")
 
 
-def test_flex_spw_read(tmp_path):
+@pytest.mark.filterwarnings("ignore:LST values stored in this file are not ")
+@pytest.mark.parametrize("future_shapes", [True, False])
+def test_flex_spw_read(tmp_path, future_shapes):
```
bhazelton, line 153: I wonder whether this test should be moved to `test_uvdata` because it is effectively designed to test UVData functionality, it just happens to use a MIR file because it's a file with flexible spectral windows. It probably doesn't matter too much, but if I was looking for the test that tested this functionality I wouldn't look in this file right away.

pyuvdata/uvdata/uvh5.py
bhazelton: Ok @plaplant, I think this PR is ready for you to take over to implement the uvh5 versioning.

pyuvdata/uvdata/tests/test_mir.py
kartographer: I'd certainly be amenable to moving it over to `test_uvdata` -- it started in `test_mir` because that just so happened to be the first file format that supported it in early testing, but it doesn't need to stay that way.

bhazelton:
@plaplant do you want to update the uvh5 memo with info on versioning in this PR or a subsequent one?

plaplant:
@bhazelton yes, I'm working on local changes, they should be done in the next day or so.

plaplant:
@david-macmahon I've updated the UVH5 memo as part of this PR. I've tried to address the topics raised in #922, #924, and #912 as best I can. If you get a chance to take a look, I'd be very interested in hearing your thoughts!

@bhazelton if you don't mind taking a look at the memo for accuracy (especially as it pertains to flexible spws and future array shapes, Secs. 5.2.2 and 5.3.1), I'd appreciate it!

bhazelton:
@plaplant the sections you pointed me to look good, but I don't think the writer in this branch is following the recommended best practice of writing V1.0 files. It's following the array shapes of the object. Should we add code to convert to future shapes for writing?

docs/references/uvh5_memo.tex
```diff
@@ -140,51 +152,71 @@ \subsection{Required Parameters}
...
   corresponding to the center of an integration. This is a one-dimensional array
   of size Nblts. (\textit{time\_array})
 \item \textbf{integration\_time}: \textit{float} An array of the length of time
   in seconds of an integration. This is a one-dimensional array of size
   Nblts. (\textit{time\_array})
 \item \textbf{freq\_array}: \textit{float} An array of the frequencies stored in
-  the file in Hertz. This is a two-dimensional array of size (Nspws,
-  Nfreqs). (\textit{freq\_array})
+  the file in Hertz. This is a one-dimensional array of size
+  (Nfreqs). (\textit{freq\_array})
```
bhazelton, line 174: Maybe we should list the changes (including this one) in the version history section so that if someone needs to read an older version they know what to expect?

docs/references/uvh5_memo.tex
```diff
@@ -269,19 +306,35 @@ \subsection{Visdata Dataset}
...
+$j$ is as follows:
+\begin{equation}
+\mathcal{V}(u_j - u_i, v_j - v_i) = \int \dd{l} \dd{m} I(l, m) g_i(l, m) e^{-2\pi i (u_i l + v_i m)} g_j^*(l, m) e^{2\pi i (u_j l + v_j m)}.
+\end{equation}
+That is, the baseline vector defined by the $uvw$ coordinates is directed from
+antenna $i$ to antenna $j$ (so the baseline vector can be computed as
+$\vb{r_j} - \vb{r_i}$, where $\vb{r}$ is the position vector of a given
+antennas), and the data corresponding to antenna $j$ is conjugated. Note that if
+a file is generated with the opposite convention, it is usually sufficient to
+multiply $uvw$ coordinates by $-1$ to generate a self-consistent dataset.
```
bhazelton, line 325: I think you'd want to conjugate the data as well when you do this. At least that's true if the data are matched to the other convention.

docs/references/uvh5_memo.tex
```diff
@@ -306,6 +359,93 @@ \subsection{Nsamples Dataset}
...
+Version 1.0 of UVH5 represents a significant change in the way that the data
+arrays (\verb+visdata+, \verb+flags+, and \verb+nsamples+) and metadata arrays
+are stored. The previously vestigial spectral-window axis is removed, meaning
+that data arrays are rank-3 instead of rank-4. Explicitly, these arrays have
+shape (Nblts, Nfreqs, Npols), where Nfreqs includes the number of channels
+across all spectral windows. The description of data and metadata in the body of
+this memo assumes the Version 1.0 specification. Although \verb+pyuvdata+ plans
+to indefinitely support files written with the previous convention (i.e., having
+an explicit spectral window-axis), UVH5 files should be written such that they
+conform to Version 1.0.
```
bhazelton, line 447: I think we should add here that the freq_array went from rank 2 to rank 1

david-macmahon:
@plaplant These changes to the UVH5 memo look great.  It will be nice when these changes and the `ant_1_array`/`ant_2_array` changes from PR #994 eventually get merged.  Here are some more comments (ordered as the relevant topics appear in the memo):

1. Kind of nit pick, but the description of `time_array` refers to the "center of an integration".  IMHO, the word "center" is already very overloaded (pointing center, phase center, delay center, etc) so maybe it would be clearer to replace "center of an integration" with "temporal midpoint of the corresponding baseline's integration"?
2. Somewhat similar nit pick for `integration_time`: replace "An array of the length of time..." with "An array of the duration...".  This avoids A) having "length" near "array" when it doesn't refer the the array's length, and B) using the word "time" in the definition of `integration_time`.
3. `freq_array`: Be more explicit about **all** frequencies, "An array of all the frequencies (for all spectral windows) stored in the file in Hertz."
4. `antenna_positions`: I find "reference position of the radio telescope array" to be less ambiguous than "telescope location" (which sounds like it could be any one of the array's antennas).  I guess interferometrists like us think of the array as being "the telescope" and antennas as mere components of said telescope, but single dish people certainly call their lone antenna a telescope (e.g. it's the GBT, not the GBA :)).  I think this is essentially the antennas' ECEF coordinates minus the ECEF coordinates of the array reference position.  I'm not 100% clear on the best terminology, but I think "IRTF frame" is slightly redundant.  Maybe it would be clearer to refer to it as "the ITRS ECEF frame translated to the array reference position", but maybe that has its own confusion inducing properties.
5. `flex_spw`: Can this be an HDF5 Bool type or does it need to be stored using the H5Enum type described in Appendix C?  Is it basically an error to have this present and set to False (vs not being present at all)?
6. The `Conjugation Convention" section is super helpful!  Adding a reference for RIME would be nice (for completeness).
7. The "Version History" is helpful.  Adding a table to summarize the various version and their changes would be good.  Does the lack of a `version` dataset imply that a UVH5 file could be any "pre-1.0" version?  Maybe better emphasize that UVH5 writers are encouraged to conform to the latest UVH5 version (currently version 1.0) and UVH5 readers are encouraged to support all versions?
8. Regarding strings, as discussed in appendix A, it could be useful to show the output of `h5dump -d Header/<dataset_name> -A <filename>` for valid string scalar and array datasets.  Restricting to ASCII is probably best for interoperability (e.g. conversion to MIRIAD and/or UVFITS), but ASCII is a bit "primitive" in comparison to UTF-8.  Perhaps it could be relaxed slightly to say UTF-8 encoded strings are permissible, but for maximal flexibility (e.g. for use with python2 and/or conversion to MIRIAD/UVFITS) restricting strings to ASCII is recommended?
9. Compression: The section on flags talks about LZF compression, but other compression schemes are fairly ubiquitous.  Is it worth mentioning some of these others explicitly?  And/or maybe state that LZF is one of several available compressor options and mention of it here (i.e. in the flags section) is not intended as an exhaustive list of all possible compression schemes?

That's all I have so far.  Now I have to go update my UVH5 writer.  That might lead to some more questions.

plaplant:
@david-macmahon I've updated the UVH5 memo based on feedback. I think I've presented everything in a way that's self consistent, but let me know if there's anything that can be improved. I also updated the sample UVH5 file mentioned in #936, so the antenna positions/numbers/names metadata should all be correct.

@bhazelton I played with the idea of forcing v1.0 (i.e., future array shape) files to be written, even if the user hadn't opted to `use_future_array_shapes`, but backed off because it led to a big dip in coverage (because we weren't testing writing files the old way!) and the hera_cal tests were failing. While I do think we want to write v1.0 files as much as possible, I also think in the interest of merging this in and making a new release soon, we should make it more voluntary at first. But I'm open to hearing other ideas about it.

Assuming we're okay with how things are, I think this PR is ready to go from my point of view (provided tests and coverage are good).

mkolopanis (commented):
Thanks @plaplant @bhazelton @kartographer for all the hard work on this PR. I think things are looking really great. However I do have some concerns about the lengths of some of the tests, especially new tests added in this PR. I've always been of the mindset that the more atomic we can test the better for identifying failures. There are some older tests that got touched on too that are quite long as well. Some of these tests seems like they could easily be parametrized or broken up into individual checks, others not so much.

I probably won't push very hard on this if there is good reason not to break them up though.

docs/tutorial.rst
```diff
@@ -1243,6 +1279,54 @@ just specify a desired order by passing an array of baseline-time indices.
...
+
+  # Use the ``select_spw`` keyword to sort only one spectral window.
+  # Now the frequencies in spectral window 1 are in descending order but the frequencies
+  # in spectral window 2 are in ascending order
+  >>> uv.reorder_freqs(select_spw=1, channel_order="-freq")
+  >>> print(np.min(np.diff(uv.freq_array[0, np.nonzero(uv.flex_spw_id_array == 1)])) <= 0)
+  True
+
+  >>> print(np.min(np.diff(uv.freq_array[0, np.nonzero(uv.flex_spw_id_array == 2)])) >= 0)
+  True
```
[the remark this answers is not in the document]
bhazelton: oh, they should. I missed this one in the rebase.

bhazelton:
@mkolopanis I agree with you on the test lengths. I don't have time to fix them right now, though, and I'd really like to get a new version out soon. Would it be ok to move this to a new issue?

mkolopanis:
Sure I don't mind punting that for a new issue.

mkolopanis (approved):
In light of the new issue I think this is good to go!

PR #972, opened 2020-12-16
Fix ReadTheDocs

bhazelton:
## Description

Minor bug fixes in docs build for newer astropy versions.
Add CI to build docs for every PR to ensure we know when doc building fails.
See the RTD build for this branch here: https://pyuvdata.readthedocs.io/en/fix_rtd/index.html

## Motivation and Context

The docs build broke on RTD because of a newer version of astropy. Unfortunately, I didn't catch this earlier because RTD still had `master` as our default branch rather than `main` so it wasn't actually building the docs on the `main` branch. I've fixed that on the RTD settings but RTD failures always come after a PR is merged, it doesn't actually tell us about problems in real time. To address that issue, I added a github actions workflow to build our docs for every push & PR so we can catch problems more quickly.

## Types of changes

- [ ] Version change

## Checklist:

- [x] If this is a significant change to the readme or other docs, I have checked that they are rendered properly on ReadTheDocs. (you may need help to get this branch to build on RTD, just ask!)

### DIFF
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -46,7 +46,6 @@ jobs:
 
       - name: Build docs
         run: |
-          source activate $ENV_NAME
           cd docs
           make html
 
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8, 3.9]
 
     steps:
       - uses: actions/checkout@master
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -27,7 +27,7 @@ jobs:
         with:
           auto-update-conda: true
           miniconda-version: "latest"
-          python-version: ${{ strategy.matrix.python-version }}
+          python-version: ${{ matrix.python-version }}
           environment-file: ci/${{ env.ENV_NAME }}.yml
           activate-environment: ${{ env.ENV_NAME }}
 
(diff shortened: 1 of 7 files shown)


.github/workflows/build_docs.yaml
```diff
@@ -0,0 +1,50 @@
...
+     run:
+       # Adding -l {0} helps ensure conda can be found properly.
+       shell: bash -l {0}
+    env:
+      ENV_NAME: docs
+      PYTHON: ${{ matrix.python-version }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.7, 3.8, 3.9]
```
mkolopanis, line 18: do we find value in building against all three python versions? Genuine question, I would have naively only done one.

bhazelton: We can probably get away with one, just as in the linter. Which one should I use?

bhazelton: We currently use 3.7 on RTD. I think we should use the same on this CI as RTD, but not sure which one to pick for both.

mkolopanis: I don't personally see any advantage in a specific version. If 3.7 works just fine why change it up?

bhazelton:
@mkolopanis I think this PR is ready when you have time to take a look.

.github/workflows/build_docs.yaml
```diff
@@ -0,0 +1,48 @@
...
+    strategy:
+      fail-fast: false
+
+    steps:
+      - uses: actions/checkout@master
+        with:
+          fetch-depth: 1
+
+      - name: Setup Miniconda
+        uses: conda-incubator/setup-miniconda@v2.0.0
```
mkolopanis, line 24: we could change this from v2.0.0 to master. :thinking: but then I worry about what if a feature is added/changed and messes with workflows? Really that isn't such a big deal because we can make a quick fix PR. Thoughts?

mkolopanis (commented):
Looks great. One very small question about the version of make conda action we have pinned (that yes I pinned that version and now have doubts).

.github/workflows/build_docs.yaml
bhazelton: I could go either way, but I'd vote to make it a new PR to change to master if that's ok.

PR #976, opened 2021-01-02
Speedup find_clusters

aelanman:
## Description

1. Break long baseline lists into chunks to avoid overrunning memory.
2. Structure adjacency as a list of sets to avoid doing type conversions inside loops.
3. Convert adjacency sets to frozensets to make them hashable.
4. Identify redundant baselines as those who all share the same adjacency set.
4. When baselines are added to redundant groups, remove them from the remaining groups to be checked. This saves on loop time and also removes the need to do `np.unique` on the final list.

## Motivation and Context

Happy new year!

The current redundancy finder relies on a function `find_clusters`, which groups points in 3D space that are all within a specified distance of each other. It does this by constructing a graph with the points as vertices and edges defined whenever two vertices are within the specified distance. The graph structure is expressed as a set of "adjacency lists" -- A list of lists where the k-th list gives the indices of vertices connected to vertex k. The vertices in each adjacency list are called "neighbors" of k.

Redundant groups are identified as _maximal cliques_ -- subsets of vertices such that each vertex in the subset is connected to each other, and no other vertices can be added while maintaining this "fully-connected" property. Apparently the task of identifying all maximal cliques in a graph if a well-known NP-complete problem. The existing approach is to loop over the neighbors (y) of each vertex (x) and identify when a x's adjacency list is a subset of y's. In that case, it means y is connected to all of the same points as x, and so it gets included in a group.

There is one other rule for identifying redundant groups that should be in place. In addition to being a maximal clique, a group of redundant baselines should also be disconnected from the rest of the graph such that no vertex is connected to a vertex outside the clique. I'll refer to these as "isolated cliques."

In practice, this means we can identify redundant groups as sets of vertices that all have the same adjacency list (if each vertex is considered adjacent to itself). Rather than looking at subsets then, we can look at equivalence, which can be a lot faster if we use hashing:
```
In [3]: s2 = frozenset(range(1000))

In [4]: s3 = frozenset(range(1000))

In [5]: %timeit s2 == s3
17.5 µs ± 308 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [6]: %timeit s2.__hash__() == s3.__hash__()
221 ns ± 1.71 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

Under the described changes, only isolated cliques will be counted as groups of redundant baselines, so we never end up with a baseline in multiple redundant groups. Such "ambiguous baselines" are detected when there are adjacency lists that are not in the final list of isolated cliques. Tolerance should always be chosen to produce a graph made up only of isolated cliques.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:


pyuvdata/utils.py
```diff
@@ -1081,9 +1081,64 @@ def get_lst_for_time(jd_array, latitude, longitude, altitude):
...
+
+    # Require all adjacency lists to be isolated maximal cliques:
+    if strict:
+        if not all(sorted(st) in loc_gps for st in adj):
+            raise ValueError("Non-isolated cliques found in graph.")
+
+    return loc_gps
+
+
+def find_clusters(location_ids, location_vectors, tol, strict=False):
```
bhazelton, line 1139: Please add the new `strict` parameter to the Parameters section of the docstring.

pyuvdata/utils.py
```diff
@@ -1081,9 +1081,64 @@ def get_lst_for_time(jd_array, latitude, longitude, altitude):
     return lst_array
 
 
-def find_clusters(location_ids, location_vectors, tol):
+def _adj_list(vecs, tol, n_blocks=None):
```
bhazelton, line 1084: Do you want to plumb the `n_blocks` parameter up to the `find_clusters` function? If not, I'm not sure we should have it in the function signature here.

bhazelton (commented):
This looks great @aelanman, thank you! Just a couple of doctring comments.

pyuvdata/utils.py
aelanman: The n_blocks keyword here is used by unit tests. Otherwise, the case of `n_blocks > 1` won't get tested.

pyuvdata/utils.py
aelanman: Whoops, thanks!

aelanman:
@bhazelton Thanks for the review! I've responded to each comment.

pyuvdata/utils.py
bhazelton: Ok. But if it's not accessible from the code users will call, what's the point in having it as an option in the helper function?

aelanman: The test calls this function directly, and I don't know of another way to set the parameter except by making it a keyword. It shouldn't be used by end users.

bhazelton: No, I'm asking why is this an option in the helper function, and why there's an if statement for `n_blocks` being specified if no one can actually specify that. In other words, why handle a user specifying `n_blocks` if they actually cannot specify it.

aelanman: The users can't specify it, but it does get specified by a test. So the keyword is just there for the test.

Normally, `n_blocks` is only larger than 1 if the len(vecs) > 2**10. I wanted to test that the function works correctly when the vector is is broken up into blocks, and for that to happen I either need to artificially set the number of blocks or make a very large `vecs`. I included this keyword just for that test.

I guess it wouldn't hurt to make it available to users.

bhazelton: Ah, ok. I was trying to understand what the value was, but a way to test that more than one block works without making the test take forever is a good reason. I'm ok with it as is then, I just didn't understand the purpose.

bhazelton:
@aelanman I think this is ready to go, there's just a small linting issue. I can fix it if you want me to.

bhazelton (approved):
Looks good. Thanks!

PR #986, opened 2021-01-22
Mwa memory bloat

### LINKED ISSUE #954: Large memory required for gpubox file reading
I am unable to read all of the gpubox files for one observation on our cluster due to the large memory requirements needed to read in the gpubox files.

- Time/freq resolution is 0.5s/40kHz
- Each gpubox file is less than 2Gb
- There are 24 channels with 2 gpubox files each

Below is a sample script. I am only reading in approx 1/4 of the gpubox files at once in this example.
```
from pyuvdata import UVData
import numpy as np
UV = UVData()

# Construct the list of files
data_path = '/fred/oz048/MWA/data/1061316296/rts_thermal_noise/'
filelist = [data_path + i for i in ['1061316296.metafits',
     '1061316296_thermal01_00.fits','1061316296_thermal01_01.fits',
     '1061316296_thermal02_00.fits','1061316296_thermal02_01.fits',
     '1061316296_thermal03_00.fits','1061316296_thermal03_01.fits',
     '1061316296_thermal04_00.fits','1061316296_thermal04_01.fits',
     '1061316296_thermal05_00.fits','1061316296_thermal05_01.fits',
     '1061316296_thermal06_00.fits','1061316296_thermal06_01.fits']]
UV.read(filelist, correct_cable_len=True, phase_to_pointing_center=True, flag_init=False,data_array_dtype=np.complex64)
write_file = '/fred/oz048/MWA/data/1061316296/rts_thermal_noise/1061316296_1.uvfits'
UV.write_uvfits(write_file, spoof_nonessential=True)
```

Below is the CPU usage and memory allocation for a job run on the above script. Blue indicates user processes (calculations), green indicates system processes (I/O), yellow indicates memory, and dashed yellow is the allocated memory. The job was killed before all files were read. 
Screen Shot 2020-12-04 at 9 49 41 pm

mkolopanis:
Moved fits reading code to own function. Some cython clean up as well. 

I actually think there are a few type casting lines that I left in that now _cannot_ be hit by coverage (where previously we would cast up to double precision then back but now the casting is done in cache during an inplace division operation).

## Description

Moved the inner reading loop to its own function, this provides better clues to the garbage collector to close handles and release memory.
Change the `generate_map` function in the corr_fits extension to operate inplace on the `map_inds` and `conj` arrays. These were already being provided as inputs to the function so it seemed silly for them to also be outputs.
Touched up the cable length differencing code. Made inputs memoryviews and tweaked the actual difference to run in C not python. notebook provided for reference on development. 
[mwa_corr_fits.pdf](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/5857191/mwa_corr_fits.pdf)

Notebook from memory profiling. This current implementation looks at least as good as reading each file individually then `fast_concat`ing them together. The latter method started breaking down on test files as the `lst_array` would being to differ when trying to read > 10 files.
[mwa_memory_bloat.pdf](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/5857195/mwa_memory_bloat.pdf)

## Motivation and Context

Closes #954  which identified potential memory leak when reading in MWA corr fits files. Always looking to make the cython better when I touch the corr first code.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
+++ b/pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
@@ -44,6 +44,18 @@ cpdef void generate_map(
   numpy.int32_t[:] map_inds,
   numpy.npy_bool[:] conj,
 ):
+  """Compute the map between pfb inputs and antenna numbersself.
+
+  This function operates on input `map_inds` and `conj` arrays inplace.
+
+  Parameters
+  ----------
+  map_inds : 1D numpy array of type int32
+    The array into which mapping indices will be populated.
+  conj : 1D numpy array of type np.bool_
+    The array into which indices of baselines to conjugate will be populated.
+
+  """
   cdef int ant1, ant2, p1, p2, pol_ind, bls_ind, out_ant1, out_ant2
   cdef int out_p1, out_p2, ind1_1, ind1_2, ind2_1, ind2_2, data_index
 
@@ -106,6 +118,25 @@ cpdef numpy.ndarray[ndim=1, dtype=numpy.float64_t] get_cable_len_diffs(
   numpy.int_t[:] ant2_array,
   numpy.ndarray cable_lens,
 ):
+  """Computer the difference in cable lengths for each baseline.
+
+  Parameters
+  ----------
+  Nblts : int
+    The number of baseline times
+  ant1_array : numpy array of type int_t
+    Array of antenna 1 numbers for each baseline.
+  ant2_array : numpy array of type int_t
+    Array of antenna 2 numbers for each baseline.
+  cable_lens : numpy array
+    Array of strings of the length of the cable for each antenna.
+
+  Returns
+  -------
+  cable_diffs : numpy array of type float64
+    Array of lenght Nblts with the difference of cable lengths for each baseline.
+
+  """
   cdef Py_ssize_t i
   cdef int n_cables = cable_lens.shape[0]
   cdef numpy.float64_t[::1] cable_array = np.zeros(n_cables, dtype=np.float64)
@@ -117,7 +148,6 @@ cpdef numpy.ndarray[ndim=1, dtype=numpy.float64_t] get_cable_len_diffs(
   # check if the cable length already has the velocity factor applied
   # this loop is very python-y but there is no real gain to trying to make it C-esque
   for i in range(n_cables):
-
     if cable_lens[i][:3] == "EL_":
       cable_array[i] = float(cable_lens[i][3:])
     else:
--- a/pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
+++ b/pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
@@ -37,14 +37,17 @@ cpdef dict input_output_mapping():
 
   return pfb_inputs_to_outputs
 
-cpdef tuple generate_map(
+@cython.boundscheck(False)
+@cython.wraparound(False)
+cpdef void generate_map(
   dict ants_to_pf,
-  dict in_to_out,
-  numpy.ndarray[ndim=1, dtype=numpy.int32_t] map_inds,
-  numpy.ndarray[ndim=1, dtype=numpy.npy_bool] conj
+  numpy.int32_t[:] map_inds,
+  numpy.npy_bool[:] conj,
 ):
   cdef int ant1, ant2, p1, p2, pol_ind, bls_ind, out_ant1, out_ant2
   cdef int out_p1, out_p2, ind1_1, ind1_2, ind2_1, ind2_2, data_index
+
+  cdef dict in_to_out = input_output_mapping()
 
   for ant1 in range(128):
     for ant2 in range(ant1, 128):
@@ -56,15 +59,13 @@ cpdef tuple generate_map(
           pol_ind = int(2 * p1 + p2)
           bls_ind = int(128 * ant1 - ant1 * (ant1 + 1) / 2 + ant2)
           # find the pfb input indices for this combination
-          (ind1_1, ind1_2) = (
-            ants_to_pf[(ant1, p1)],
-            ants_to_pf[(ant2, p2)],
-          )
+          ind1_1 = ants_to_pf[(ant1, p1)]
+          ind1_2 = ants_to_pf[(ant2, p2)]
+
           # find the pfb output indices
-          (ind2_1, ind2_2) = (
-            in_to_out[(ind1_1)],
-            in_to_out[(ind1_2)],
-          )
+          ind2_1 = in_to_out[ind1_1]
+          ind2_2 = in_to_out[ind1_2]
+
           out_ant1 = int(ind2_1 / 2)
           out_ant2 = int(ind2_2 / 2)
           out_p1 = ind2_1 % 2
@@ -95,30 +96,35 @@ cpdef tuple generate_map(
             )
             map_inds[bls_ind * 4 + pol_ind] = data_index
 
-  return map_inds, conj
+  return
 
+@cython.boundscheck(False)
+@cython.wraparound(False)
 cpdef numpy.ndarray[ndim=1, dtype=numpy.float64_t] get_cable_len_diffs(
   int Nblts,
-  numpy.ndarray[dtype=numpy.int_t, ndim=1] ant1_array,
-  numpy.ndarray[dtype=numpy.int_t, ndim=1] ant2_array,
+  numpy.int_t[:] ant1_array,
+  numpy.int_t[:] ant2_array,
   numpy.ndarray cable_lens,
 ):
-  cdef int i
-  cdef numpy.ndarray[ndim=1, dtype=numpy.float64_t] cable_array = np.zeros(len(cable_lens), dtype=np.float_)
-  cdef numpy.ndarray[ndim=1, dtype=numpy.float64_t] cable_diffs = np.zeros(Nblts, dtype=np.float_)
-
+  cdef Py_ssize_t i
+  cdef int n_cables = cable_lens.shape[0]
+  cdef numpy.float64_t[::1] cable_array = np.zeros(n_cables, dtype=np.float_)
+  cdef numpy.float64_t[::1] cable_diffs = np.zeros(Nblts, dtype=np.float_)
   # "the velocity factor of electic fields in RG-6 like coax"
   # from MWA_Tools/CONV2UVFITS/convutils.h
   cdef float v_factor = 1.204
 
   # check if the cable length already has the velocity factor applied
-  for i in range(len(cable_lens)):
-    if cable_lens[i][0:3] == "EL_":
+  # this loop is very python-y but there is no real gain to trying to make it C-esque
+  for i in range(n_cables):
+
+    if cable_lens[i][:3] == "EL_":
       cable_array[i] = float(cable_lens[i][3:])
     else:
       cable_array[i] = float(cable_lens[i]) * v_factor
 
   # build array of differences
-  cable_diffs = cable_array[ant2_array] - cable_array[ant1_array]
+  for i in range(Nblts):
+    cable_diffs[i] = cable_array[ant2_array[i]] - cable_array[ant1_array[i]]
 
-  return cable_diffs
+  return np.asarray(cable_diffs)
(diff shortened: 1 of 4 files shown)


mkolopanis:
@nicholebarry would you mind testing this out to see if it help alleviate the issue you saw on AWS too?

bhazelton:
If those lines cannot be reached, should we just remove them? If not, we should mark them not to be included in coverage reports with a comment explaining why.

bhazelton:
@plaplant are you willing to take a look at this, especially for the cython parts?

mkolopanis:
> If those lines cannot be reached, should we just remove them? If not, we should mark them not to be included in coverage reports with a comment explaining why.

Yeah this is a great question. I left them in mostly for insurance right now but I don't mind removing them.

mkolopanis:
There is one additional line that is now not covered (the functio nattached). Turns out we only use this inside the cython code, so I just moved the call to be inside of the cython (since it's calling a cython function). I left the python hook though because I didn't know if it is useful as a public function and didn't want to just remove functionality.
https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/b91af8b8d682b01c3fde865980bdbd176a355c5f/pyuvdata/uvdata/mwa_corr_fits.py#L22-L31

pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
```diff
@@ -95,30 +96,35 @@ cpdef tuple generate_map(
...
+  numpy.int_t[:] ant2_array,
   numpy.ndarray cable_lens,
 ):
-  cdef int i
-  cdef numpy.ndarray[ndim=1, dtype=numpy.float64_t] cable_array = np.zeros(len(cable_lens), dtype=np.float_)
-  cdef numpy.ndarray[ndim=1, dtype=numpy.float64_t] cable_diffs = np.zeros(Nblts, dtype=np.float_)
-
+  cdef Py_ssize_t i
+  cdef int n_cables = cable_lens.shape[0]
+  cdef numpy.float64_t[::1] cable_array = np.zeros(n_cables, dtype=np.float_)
```
plaplant, line 111: I know that `np.float_` is the same thing as `np.float64`, but I have a slight preference for being explicit in the number of bits (though I understand I'm a bit of an old-timer in that regard). Also a big fan of using memory views 👍

pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
```diff
@@ -95,30 +96,35 @@ cpdef tuple generate_map(
...
   # "the velocity factor of electic fields in RG-6 like coax"
   # from MWA_Tools/CONV2UVFITS/convutils.h
   cdef float v_factor = 1.204
 
   # check if the cable length already has the velocity factor applied
-  for i in range(len(cable_lens)):
-    if cable_lens[i][0:3] == "EL_":
+  # this loop is very python-y but there is no real gain to trying to make it C-esque
+  for i in range(n_cables):
+
```
plaplant, line 120: I think we can get rid of this blank line.

pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
```diff
@@ -37,15 +37,18 @@ cpdef dict input_output_mapping():
 
   return pfb_inputs_to_outputs
 
-cpdef tuple generate_map(
+@cython.boundscheck(False)
+@cython.wraparound(False)
+cpdef void generate_map(
```
plaplant, line 42: It might be worth mentioning (either in a comment or a docstring) that `map_inds` and `conj` are update "in-place", so to speak, without explicitly being returned to the user. I know the `void` return type and name hints at that, but it might be good to be explicit (for people like me who got confused at first...).

pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
```diff
@@ -37,15 +37,18 @@ cpdef dict input_output_mapping():
...
-  dict in_to_out,
-  numpy.ndarray[ndim=1, dtype=numpy.int32_t] map_inds,
-  numpy.ndarray[ndim=1, dtype=numpy.npy_bool] conj
+  numpy.int32_t[:] map_inds,
+  numpy.npy_bool[:] conj,
 ):
   cdef int ant1, ant2, p1, p2, pol_ind, bls_ind, out_ant1, out_ant2
   cdef int out_p1, out_p2, ind1_1, ind1_2, ind2_1, ind2_2, data_index
 
+  cdef dict in_to_out = input_output_mapping()
```
plaplant, line 50: Related to your comment on the PR, I think it makes sense to move the mapping call here (in case the compiler can be smart about inlining/calling the function more efficiently). But I would vote for keeping the python function around as well, since it doesn't seem to be harming anything and it would be an API-level change to remove it (so we'd have to go through the deprecation rigmarole).

pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
```diff
@@ -95,30 +96,35 @@ cpdef tuple generate_map(
...
     else:
       cable_array[i] = float(cable_lens[i]) * v_factor
 
   # build array of differences
-  cable_diffs = cable_array[ant2_array] - cable_array[ant1_array]
+  for i in range(Nblts):
+    cable_diffs[i] = cable_array[ant2_array[i]] - cable_array[ant1_array[i]]
 
-  return cable_diffs
+  return np.asarray(cable_diffs)
```
plaplant, line 130: Nice to cast the memory view to an array to make it more user-friendly 👍

pyuvdata/uvdata/mwa_corr_fits.py
```diff
@@ -681,14 +709,11 @@ def read_mwa_corr_fits(
             # these are the indices for the data corresponding to the initial
             # antenna/pol pair
             # generate a mapping index array
-            map_inds = np.zeros((self.Nbls * self.Npols), dtype=np.int32)
+            map_inds = np.zeros((self.Nbls * self.Npols), dtype=np.int32, order="C",)
```
plaplant, line 712: Is the `order` keyword needed  for a 1-d array?

pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
```diff
@@ -37,15 +37,18 @@ cpdef dict input_output_mapping():
...
 
-cpdef tuple generate_map(
+@cython.boundscheck(False)
+@cython.wraparound(False)
+cpdef void generate_map(
   dict ants_to_pf,
-  dict in_to_out,
-  numpy.ndarray[ndim=1, dtype=numpy.int32_t] map_inds,
-  numpy.ndarray[ndim=1, dtype=numpy.npy_bool] conj
+  numpy.int32_t[:] map_inds,
```
plaplant, line 44: Since you're defining these in python using `np.zeros` (and are thus contiguous), is there a benefit to declaring these memory views as, e.g., `numpy.int32_t[::1]`? (Genuinely curious, I haven't done the benchmarking myself.)

plaplant (commented):
Thanks @mkolopanis! This is a nice addition to have, and looks to make reading corr fits files much more efficient. I just have a few small comments.

pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
mkolopanis: I think I remember the use of `np.float_` having a difference precision on windows too. Thanks for pointing this out, I'll try to change any instances of it.

pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
mkolopanis: Sounds like a plan to me. I we _could_ turn it into a c++ map too which would make the call even a little more efficient, but I have not really used it before and don't know if there's a good way to go back from a map to a dict. Either way I don't think it is a particularly troublesome part of the code so I have been in favor of leaving it.

pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
mkolopanis: This isn't 100% related, but it turns out that if you compile to C and return a memoryview it often casts back to an array and everything is fine. if you compile to C++ it returns a memoryview object, so I've been trying to get in the habit of casting it explicitly when the code says it should be an array.

pyuvdata/uvdata/mwa_corr_fits.py
mkolopanis: Probably not? But I just want to ensure the memory will be the way it is expected when using memoryviews as inputs.

pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
mkolopanis: Do you mean the `map_inds` being a memoryview now? The overhead is (allegedly) _slightly_ lower for memoryviews, and allows for direct C access to the arrays. This way we don't have to bloat up the function with additional declarations of memoryviews. I have found in some tests in other places that memoryview inputs are a little faster. But I think my preference is still in the anecdotal category. The cython docs do emphasize that the memoryviews are "newer and better" and more flexible over all.

mkolopanis:
The error on this conda build literally says just try again. I'm going to restart if there are no objections?
>CondaHTTPError: HTTP 000 CONNECTION FAILED for url <https://conda.anaconda.org/conda-forge/osx-64/repodata.json>
Elapsed: -
>An HTTP error occurred when trying to retrieve this URL.
>HTTP errors are often intermittent, and a simple retry will get you on your way.
>'https://conda.anaconda.org/conda-forge/osx-64'

pyuvdata/uvdata/corr_fits_src/corr_fits.pyx
plaplant: I meant more that we use the `::1` indexing syntax, rather than just `:`. I believe the former further requires the memory to be C-contiguous (if I'm interpreting the docs correctly), but like the initial declaration discussed in the python code below, I don't think it really matters for a 1-d array. Let's leave it as is.

mkolopanis: Oh yeah that's a good point. especially since I'm just using `[:]`. it doesn't matter. We should just pick one, would you prefer `[::1]` and leave the python ordering or remove the python ordering?

plaplant: Let's index it using `::1` here (maybe with a comment explaining that it requires a C-contiguous array), and keep the `order` argument in python-land. Probably better to be explicit and model best practices, even if it is a moot point for 1d array.

pyuvdata/uvdata/mwa_corr_fits.py
```diff
@@ -741,15 +770,15 @@ def read_mwa_corr_fits(
...
-                if self.data_array.dtype != np.complex128:
-                    self.data_array = self.data_array.astype(np.complex128)
-                self.data_array = self.data_array / (dig_gains1 * dig_gains2)
+
+                self.data_array /= (
+                    dig_gains[self.ant_1_array, :, np.newaxis]
+                    * dig_gains[self.ant_2_array, :, np.newaxis]
+                )
+
                 if self.data_array.dtype != data_array_dtype:
```
plaplant, line 779: @mkolopanis I think the data type checking here (and the other one below) are not needed anymore. The initial declaration happens using the `data_array_dtype` argument, and the rest of the manipulations involve array reshaping and/or inplace updates. If we wanted to be paranoid we could change these to `assert` statements, but I'm also okay with removing them.

plaplant:
We should also probably add an explicit test for the `input_output_mapping` function. I can't think of a clever way to do it besides reimplementing the cython code (at which point we're essentially asking "Did cython compile this correctly?"), but I think it's good coding practice to have explicit tests for all public functions. I can add this quickly.

pyuvdata/uvdata/mwa_corr_fits.py
plaplant: I went ahead and removed these statements while I was adding the new test.

plaplant:
@bhazelton do you mind taking a look at this PR? I don't feel comfortable approving it now that I've contributed to it. But I think it's in a good state and ready to be merged.

mkolopanis:
Also I never added a change log, this probably worthy unlike most of my under the hood contributions. I was just waiting to add one until it is ready.

bhazelton (commented):
I think it looks good. Let's get the changelog update on here and we can merge.

bhazelton (approved):
Looks great. Thank you @mkolopanis!!

PR #990, opened 2021-02-02
Fixes for numpy v1.20

plaplant:
## Description

As noted in [the release notes for numpy v1.20.0](https://numpy.org/doc/stable/release/1.20.0-notes.html#deprecations), several datatypes such as `np.float` have started issuing `DeprecationWarnings`. This PR makes a lot of changes, in both the code and tests, to handle data types more precisely and make the warnings go away.

## Motivation and Context

This PR includes several different changes, which I've tried to summarize here:
* Explicitly using a type such as `np.float64` for array data types when creating new ones or using `asarray` instead of the generic `np.float`. (e.g., `np.zeros(Nvals, dtype=np.float)` -> `np.zeros(Nvals, dtype=np.float64)`). This includes integer types as well.
* Casting scalars with the python builtins `float` and `int` instead of `np.float` and `np.int`.
* Using `np.str_` instead of `np.str` for doing type comparisons with `isinstance`.
* Changing the `expected_type` value for some parameters to make type checks pass.

~This last point is something I'd like to discuss more generally, because it feels like things are a little looser than I'd like. In the past, it seemed that `np.float` would match the class of both `float` and `np.float64` (or maybe it was just how we were doing the checks?), and now we need to be a little more explicit. I just want to make sure everyone is happy how we do them going forward. (I am not wedded to how they are being done right now, as my primary motivation was making the warnings go away rather than doing things The Right Way.)~ There were several changes made to how type checking and handling were done, and now all parameter types use python builtins (e.g., `float`, `int`, etc.). This relies on the generic typing introduced in #965 to handle specific types when specified by the user, and more general types outside of that.

~Potentially unrelated to this, our tests seem to be failing when built using numpy v1.19.x. When we work out a fix for that, we might want to include it as part of this PR. Running locally, this branch passes all tests using numpy v1.20.0 and v1.19.2, so I don't believe we need to change the numpy requirement on the repo.~ This issue was resolved by using the `oldest-supported-numpy` metapackage in the `pyproject.toml` file.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/parameter.py
+++ b/pyuvdata/parameter.py
@@ -94,21 +94,20 @@ class UVParameter(object):
     description : str
         Description of the data or metadata in the object.
     expected_type
-        The type that the data or metadata should be.
-        Default is np.int or str if form is 'str'
+        The type that the data or metadata should be. Default is int or str if
+        form is 'str'.
     acceptable_vals : list, optional
         List giving allowed values for elements of value.
     acceptable_range: 2-tuple, optional
         Tuple giving a range of allowed magnitudes for elements of value.
     tols : float or 2-tuple of float
-        Tolerances for testing the equality of UVParameters. Either a
-        single absolute value or a tuple of relative and absolute values to
-        be used by np.isclose()
+        Tolerances for testing the equality of UVParameters. Either a single
+        absolute value or a tuple of relative and absolute values to be used by
+        np.isclose()
     strict_type_check : bool
-        When True, the input expected_type is used exactly, otherwise a more generic
-        type is found to allow changes in precions
-        or to/from numpy dtypes to not break checks.
-
+        When True, the input expected_type is used exactly, otherwise a more
+        generic type is found to allow changes in precicions or to/from numpy
+        dtypes to not break checks.
     """
 
     def __init__(
@@ -136,10 +135,12 @@ class UVParameter(object):
         self.form = form
         if self.form == "str":
             self.expected_type = str
+            self.strict_type = True
         else:
             self.expected_type = _get_generic_type(
                 expected_type, strict_type_check=strict_type_check,
             )
+            self.strict_type = strict_type_check
         self.acceptable_vals = acceptable_vals
         self.acceptable_range = acceptable_range
         if np.size(tols) == 1:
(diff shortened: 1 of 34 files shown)


pyuvdata/parameter.py
```diff
@@ -326,14 +326,14 @@ def expected_shape(self, uvbase):
...
-        elif isinstance(self.form, (int, np.integer)):
+        elif isinstance(self.form, (int, np.int_)):
             # Fixed shape, just return the form
             return (self.form,)
         else:
             # Given by other attributes, look up values
             eshape = ()
             for p in self.form:
-                if isinstance(p, (int, np.integer)):
+                if isinstance(p, (int, np.int_)):
```
mkolopanis, line 336: These lines caught my eye. Any particular reason you removed the more general `np.integer` here? I think the isinstance is fine using the generic base class. Basically it let's users use not 64bit integers as the elements to "form".

plaplant: Just saw this too. I don't think I should have made this change, it should go back to being `np.integer`.

pyuvdata/uvdata/src/miriad_wrap.pyx
```diff
@@ -200,7 +200,7 @@ cpdef hread_init(int item_hdl) except +raise_miriad_error:
   else:
     raise RuntimeError("unknown item type.")
 
-cpdef hwrite(int item_hdl, int offset, val, str type) except +raise_miriad_error:
+cpdef int hwrite(int item_hdl, int offset, val, str type) except *:
```
bhazelton, line 203: I don't know enough about cython to know what this means and why it should change.

bhazelton (commented):
This looks great, thank you @plaplant! I just have one minor docstring comment and a question.

pyuvdata/uvdata/src/miriad_wrap.pyx
mkolopanis: Typing returns helps optimize the hand off and reduce overhead between c calls in cython. This was also on my radar. Beat me to it 😂

bhazelton:
I don't think this needs a changelog update, but @plaplant can you update the original (top) description of the PR based on how it evolved (so that if we come back to it in the future we can see what was done more clearly).

pyuvdata/uvdata/src/miriad_wrap.pyx
bhazelton: Great. What about the change from `+raise_miriad_error` -> `*`

mkolopanis: Ah yeah that's about how exceptions get propagated up to python. I'm making a bunch of changes like this in my branch too. The short of it is I don't think the raise_miriad_error did anything anyway

bhazelton (approved):

pyuvdata/uvdata/src/miriad_wrap.pyx
plaplant: Yeah, I ~stole~ borrowed some of this from #989. I had to get rid of the reference to `np.int` in the type checking, and included other changes for this function. And yes, the `raise_miriad_error` was just a `pass` statement, so I don't think it actually did anything.

PR #466, opened 2018-10-01
Writems

tammojan:
Casacore 2.2.0 introduced a function `default_ms()` which is very useful for writing out measurement sets. Things to add later:
 * Tests
 * Writing FEED table
 * Checking all assumptions (currently only phased is implemented)
 * ...

I won't have much time to work on this, which is why I'm submitting this PR with only half functionality.

PR #995, opened 2021-03-10
Write measurement sets

bhazelton:
## Description

Add a measurement set write method to UVData, based on the tools available in the `casacore` package.  The updated module  has the ability to write multi-spw and multi-field (or in the parlance of pyuvdata, `multi_phase_center`) data sets, tested against some limited manual checking in CASA. The method for reading measurement sets has also undergone a long-overdue update which allows for multi-spw, multi-field data sets to be read in.

A small, related update -- a new feature has been added to the  `check` method to see if UVW coordinates are flipped relative to that expected (that is, multiplying the values by -1 brings the recorded and expected values into agreement), and if found, to fix the UVW coordinates and appropriately complex conjugate `data_array`. This addition was motivated in part because of conflicting standards with the MS format with different writers, which inevitably mean that both conjugation standards (ant2-ant1 and ant1-ant2) exist _without_ any keyword, etc. that documents which conjugation scheme is used.

## Motivation and Context

This is an overhaul of #466 to essentially bring that code up to date with the repo, and add the requisite testing, etc. to the PR.

## Types of changes

## Checklist:

New feature checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,14 +37,17 @@ tested with ATCA files
 * uvfits conforms to AIPS memo 117 (as of March 2020).  It is tested against
 FHD, CASA, and AIPS. However AIPS is limited to <80 antennas and CASA uvfits
 import does not seem to support >255 antennas.
-* CASA measurement sets (TODO: Add more here about how we test these).
+* CASA measurement sets, primarily conforming to [CASA Memo 229](https://casa.nrao.edu/Memos/229.html), with some elements taken from the proposed v3.0 format documented in [CASA Memo 264](https://casacore.github.io/casacore-notes/264.html). Measurement sets are tested against
+VLA and MWA data sets, (the latter filled via cotter), with some manual verification
+haven been performed against ALMA and SMA data sets, the latter filled using the `importuvfits` task of CASA.
+tested against ALMA-filled datasets and with SMA datasets
 * uvh5 is an HDF5-based file format defined by the HERA collaboration,
 details in the [uvh5 memo](docs/references/uvh5_memo.pdf). Note that this is a
 somewhat new format, so it may evolve a bit
 but we will strive to make future versions backwards compatible with the current format.
 It is probably not compatible with other interferometric HDF5 files defined by other groups.
 * FHD (read-only support, tested against MWA and PAPER data)
-* MIR (read-only support, tested agains SMA data)
+* MIR (read-only support, tested against SMA data)
 * MWA correlator FITS files (read-only support, tested against Cotter outputs and FHD)
 
 ## UVCal file formats
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ Required:
 
 Optional:
 
-* python-casacore >= 2.2.0 (for working with CASA measurement sets)
+* python-casacore >= 3.1.0 (for working with CASA measurement sets)
 * astropy-healpix (for working with beams in HEALPix formats)
 * pyyaml (for working with settings files for CST beam files)
 * novas and novas_de405 (for using the NOVAS library for astrometry)
--- a/README.md
+++ b/README.md
@@ -6,9 +6,9 @@
 [![DOI](http://joss.theoj.org/papers/10.21105/joss.00140/status.svg)](https://doi.org/10.21105/joss.00140)
 
 pyuvdata defines a pythonic interface to interferometric data sets.
-Currently pyuvdata supports reading and writing of miriad, uvfits, and uvh5 files
-and reading of CASA measurement sets and FHD
-([Fast Holographic Deconvolution](https://github.com/EoRImaging/FHD)) visibility save files.
+Currently pyuvdata supports reading and writing of miriad, uvfits, CASA measurement sets
+and uvh5 files and reading of FHD
+([Fast Holographic Deconvolution](https://github.com/EoRImaging/FHD)) visibility save files, SMA Mir files and MWA correlator FITS files.
 
 API documentation and a tutorial is hosted on [ReadTheDocs](https://pyuvdata.readthedocs.io).
 
@@ -34,17 +34,18 @@ Also can convert raw data quality metrics into flags using thresholding.
 ## UVData File standard notes
 * miriad has been thoroughly tested with aipy-style miriad files and minimally
 tested with ATCA files
-* uvfits conforms to AIPS memo 117 (as of May 2015).  It is tested against
+* uvfits conforms to AIPS memo 117 (as of March 2020).  It is tested against
 FHD, CASA, and AIPS. However AIPS is limited to <80 antennas and CASA uvfits
 import does not seem to support >255 antennas.
+* CASA measurement sets (TODO: Add more here about how we test these).
 * uvh5 is an HDF5-based file format defined by the HERA collaboration,
 details in the [uvh5 memo](docs/references/uvh5_memo.pdf). Note that this is a
 somewhat new format, so it may evolve a bit
 but we will strive to make future versions backwards compatible with the current format.
 It is probably not compatible with other interferometric HDF5 files defined by other groups.
 * FHD (read-only support, tested against MWA and PAPER data)
-* CASA measurement sets (read-only support)
 * MIR (read-only support, tested agains SMA data)
+* MWA correlator FITS files (read-only support, tested against Cotter outputs and FHD)
 
 ## UVCal file formats
 * calfits: a new format defined in pyuvdata, details in the [calfits_memo](docs/references/calfits_memo.pdf).
@@ -68,7 +69,6 @@ metadata, details here: [cst settings file](docs/cst_settings_yaml.rst)
 * MirParser: object and python interface for MIR file format (beta version)
 
 ## Known Issues and Planned Improvements
-* UVData: add support for writing CASA measurement sets
 * UVBeam: support phased-array antenna beams (e.g. MWA beams).
 * UVFlag: Adding requires a high level knowledge of individual objects. (see [issue #653](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/issues/653))
 
@@ -224,7 +224,8 @@ Testing of `UVFlag` module requires the `pytest-cases` plug-in.
 # API
 The primary interface to data from python is via the UVData object. It provides
 import functionality from all supported file formats (UVFITS, Miriad, UVH5, FHD,
-CASA measurement sets) and export to UVFITS, Miriad, and UVH5 formats and can
+CASA measurement sets, SMA Mir, MWA correlator FITS) and export to UVFITS, Miriad,
+CASA measurement sets and UVH5 formats and can
 be interacted with directly. Similarly, the primary calibration, beam, and flag
 interfaces are via the UVCal, UVBeam, and UVflag objects. The attributes of the UVData,
 UVCal, UVBeam, and UVFlag objects are described in the UVData Parameters, UVCal Parameters,
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ Required:
 
 Optional:
 
-* python-casacore (for working with CASA measurement sets)
+* python-casacore >= 2.2.0 (for working with CASA measurement sets)
 * astropy-healpix (for working with beams in HEALPix formats)
 * pyyaml (for working with settings files for CST beam files)
 * novas and novas_de405 (for using the NOVAS library for astrometry)
--- a/pyuvdata/uvdata/ms.py
+++ b/pyuvdata/uvdata/ms.py
@@ -459,6 +459,7 @@ class MS(UVData):
     def write_ms(
         self,
         filepath,
+        force_phase=False,
         clobber=False,
         run_check=True,
         check_extra=True,
@@ -472,6 +473,9 @@ class MS(UVData):
         ----------
         filepath : str
             The MS file path to write to.
+        force_phase : bool
+            Option to automatically phase drift scan data to zenith of the first
+            timestamp.
         clobber : bool
             Option to overwrite the file if it already exists.
         run_check : bool
@@ -502,6 +506,33 @@ class MS(UVData):
                 print("File exists; clobbering")
             else:
                 raise IOError("File exists; skipping")
+
+        # TODO: the phasing requirement can probably go away once we have proper
+        # multi-object support.
+        if self.phase_type == "phased":
+            pass
+        elif self.phase_type == "drift":
+            if force_phase:
+                print(
+                    "The data are in drift mode and do not have a "
+                    "defined phase center. Phasing to zenith of the first "
+                    "timestamp."
+                )
+                phase_time = Time(self.time_array[0], format="jd")
+                self.phase_to_time(phase_time)
+            else:
+                raise ValueError(
+                    "The data are in drift mode. "
+                    "Set force_phase to true to phase the data "
+                    "to zenith of the first timestamp before "
+                    "writing a uvfits file."
+                )
+        else:
+            raise ValueError(
+                "The phasing type of the data is unknown. "
+                "Set the phase_type to drift or phased to "
+                "reflect the phasing status of the data"
+            )
 
         nchan = self.freq_array.shape[1]
         npol = len(self.polarization_array)
--- a/pyuvdata/uvdata/ms.py
+++ b/pyuvdata/uvdata/ms.py
@@ -443,6 +443,7 @@ class MS(UVData):
         nrows = len(message)
         history_table.addrows(nrows)
 
+        # the first two lines below break on python-casacore < 3.1.0
         history_table.putcol("APP_PARAMS", np.asarray(app_params)[:, np.newaxis])
         history_table.putcol("CLI_COMMAND", np.asarray(cli_command)[:, np.newaxis])
         history_table.putcol("APPLICATION", application)
--- a/pyuvdata/uvdata/ms.py
+++ b/pyuvdata/uvdata/ms.py
@@ -140,7 +140,7 @@ class MS(UVData):
         antenna_id_table = np.arange(nfeeds_table, dtype=np.int32)
         feed_table.putcol("ANTENNA_ID", antenna_id_table)
 
-        beam_id_table = np.ones(nfeeds_table, dtype=np.int32)
+        beam_id_table = -1 * np.ones(nfeeds_table, dtype=np.int32)
         feed_table.putcol("BEAM_ID", beam_id_table)
 
         beam_offset_table = np.zeros((nfeeds_table, 2, 2), dtype=np.float64)
@@ -174,6 +174,9 @@ class MS(UVData):
         receptor_angle_table = np.zeros((nfeeds_table, nfeed_pols), dtype=np.float64)
         feed_table.putcol("RECEPTOR_ANGLE", receptor_angle_table)
 
+        spectral_window_id_table = -1 * np.ones(nfeeds_table, dtype=np.int32)
+        feed_table.putcol("SPECTRAL_WINDOW_ID", spectral_window_id_table)
+
         feed_table.done()
 
     def _write_ms_field(self, filepath):
@@ -199,6 +202,71 @@ class MS(UVData):
         field_table.putcell("REFERENCE_DIR", 0, phasedir)
         field_table.putcell("NAME", 0, self.object_name)
 
+    def _write_ms_pointing(self, filepath):
+        """
+        Write out the pointing information into a CASA table.
+
+        Parameters
+        ----------
+        filepath : str
+            path to MS (without POINTING suffix)
+
+        """
+        if not casa_present:  # pragma: no cover
+            raise ImportError(no_casa_message) from casa_error
+
+        pointing_table = tables.table(
+            filepath + "::POINTING", ack=False, readonly=False
+        )
+        # TODO: we probably have to fix this for steerable dishes
+        nants = np.max(self.antenna_numbers) + 1
+        antennas = np.arange(nants, dtype=np.int32)
+        times = np.asarray(
+            [
+                Time(t, format="jd", scale="utc").mjd * 3600.0 * 24.0
+                for t in self.time_array
+            ]
+        )
+        intervals = self.integration_time
+        ntimes = len(times)
+
+        antennas_full = np.tile(antennas, ntimes)
+        times_full = np.repeat(times, nants)
+        intervals_full = np.repeat(intervals, nants)
+
+        nrows = len(antennas_full)
+        assert len(times_full) == nrows
+        assert len(intervals_full) == nrows
+
+        pointing_table.addrows(nrows)
+
+        pointing_table.putcol("ANTENNA_ID", antennas_full)
+        pointing_table.putcol("TIME", times_full)
+        pointing_table.putcol("INTERVAL", intervals_full)
+
+        name_col = np.asarray(["ZENITH"] * nrows, dtype=np.bytes_)
+        pointing_table.putcol("NAME", name_col)
+
+        num_poly_col = np.zeros(nrows, dtype=np.int32)
+        pointing_table.putcol("NUM_POLY", num_poly_col)
+
+        time_origin_col = times_full
+        pointing_table.putcol("TIME_ORIGIN", time_origin_col)
+
+        # we always "point" at zenith
+        direction_col = np.zeros((nrows, 2, 1), dtype=np.float64)
+        direction_col[0] = np.pi / 2
+        pointing_table.putcol("DIRECTION", direction_col)
+
+        # just reuse "direction" for "target"
+        pointing_table.putcol("TARGET", direction_col)
+
+        # set tracking info to `False`
+        tracking_col = np.zeros(nrows, dtype=np.bool_)
+        pointing_table.putcol("TRACKING", tracking_col)
+
+        pointing_table.done()
+
     def _write_ms_spectralwindow(self, filepath):
         """
         Write out the spectral information into a CASA table.
@@ -223,7 +291,10 @@ class MS(UVData):
         )
 
         sw_table.addrows()
+        sw_table.putcell("NAME", 0, "WINDOW0")
+        sw_table.putcell("REF_FREQUENCY", 0, self.freq_array[0, 0])
         sw_table.putcell("CHAN_FREQ", 0, self.freq_array[0])
+        sw_table.putcell("MEAS_FREQ_REF", 0, 1)
         # TODO fix for future array shapes
         chanwidths = np.ones_like(self.freq_array[0]) * self.channel_width
         sw_table.putcell("CHAN_WIDTH", 0, chanwidths)
@@ -277,6 +348,14 @@ class MS(UVData):
         if not casa_present:  # pragma: no cover
             raise ImportError(no_casa_message) from casa_error
 
+        pol_str = uvutils.polnum2str(self.polarization_array)
+        feed_pols = {feed for pol in pol_str for feed in uvutils.POL_TO_FEED_DICT[pol]}
+        pol_types = [pol.lower() for pol in sorted(feed_pols)]
+        pol_tuples = np.asarray(
+            [(pol_types.index(i), pol_types.index(j)) for i, j in pol_str],
+            dtype=np.int32,
+        )
+
         pol_table = tables.table(filepath + "::POLARIZATION", ack=False, readonly=False)
         pol_table.addrows()
         pol_table.putcell(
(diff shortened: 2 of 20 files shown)


pyuvdata/uvdata/ms.py
```diff
@@ -119,6 +119,63 @@ def _write_ms_antenna(self, filepath):
...
+
+        feed_table = tables.table(filepath + "::FEED", ack=False, readonly=False)
+
+        nfeeds_table = np.max(self.antenna_numbers) + 1
+        feed_table.addrows(nfeeds_table)
+
+        antenna_id_table = np.arange(nfeeds_table, dtype=np.int32)
+        feed_table.putcol("ANTENNA_ID", antenna_id_table)
+
+        beam_id_table = np.ones(nfeeds_table, dtype=np.int32)
```
bhazelton, line 143: since we're not filling out a beam table, I think this should be filled with -1s. From the doc:
> BEAM ID Beam identifier. Points to an optional BEAM sub-table defining the primary beam and polarization response for this FEED. A value of -1 indicates that no associated beam response is defined.

pyuvdata/uvdata/ms.py
```diff
@@ -119,6 +119,63 @@ def _write_ms_antenna(self, filepath):
...
+        feed_table.putcol("BEAM_ID", beam_id_table)
+
+        beam_offset_table = np.zeros((nfeeds_table, 2, 2), dtype=np.float64)
```
bhazelton, line 146: I need to understand what this means. I'm worried about just setting it to zero for pointing arrays like MWA.

pyuvdata/uvdata/ms.py
```diff
@@ -119,6 +119,63 @@ def _write_ms_antenna(self, filepath):
...
+        feed_table.putcol("BEAM_OFFSET", beam_offset_table)
+
+        feed_id_table = np.zeros(nfeeds_table, dtype=np.int32)
```
bhazelton, line 149: I think this needs to be zeros for one of the arms ('x'?) and 1 for the other. It's hard to interpret from the docs (they're self-referential) but the docs do say:
> Feeds are numbered from 0 on each separate antenna

pyuvdata/uvdata/ms.py
```diff
@@ -119,6 +119,63 @@ def _write_ms_antenna(self, filepath):
...
+
+        pol_response_table = np.dstack(
+            [np.eye(2, dtype=np.complex64) for n in range(nfeeds_table)]
+        ).transpose()
+        feed_table.putcol("POL_RESPONSE", pol_response_table)
+
+        position_table = np.zeros((nfeeds_table, 3), dtype=np.float64)
+        feed_table.putcol("POSITION", position_table)
+
+        receptor_angle_table = np.zeros((nfeeds_table, nfeed_pols), dtype=np.float64)
```
bhazelton, line 174: There's something we need to understand better here, that should be different depending on whether our x_orientation is 'north' or 'east'. I'm not sure if it only comes into this `receptor_angle` or whether it also comes into the `pol_response`.

pyuvdata/uvdata/ms.py
```diff
@@ -126,11 +754,303 @@ def find_not_ascii(s):
...
-    def write_ms(self):
-        """Write ms: Not yet supported."""
+            for idx, spw_id in enumerate(self.spw_array):
+                if self.flex_spw:
+                    sel_mask = self.flex_spw_id_array == spw_id
+                else:
+                    sel_mask = np.ones(self.Nfreqs, dtype=bool)
+
+                temp_dict = {
+                    "r%i" % ((self.Nspws * jdx) + 1): temp_array[None, jdx, sel_mask, :]
```
e-koch, line 988: @kartographer -- Can you double check if this is right? That should produce 8 SPWs in the output MS. Otherwise, the dictionary key is the same for all SPWs.
```suggestion
                    "r%i" % ((self.Nspws * idx) + 1): temp_array[None, jdx, sel_mask, :]
```

kartographer: Whoops -- yeah, the above was missing a piece of the indexing scheme (should have been `(self.Nspws * jdx) + 1 + idx`), can you see if it works now?

pyuvdata/uvdata/ms.py
```diff
@@ -126,11 +754,304 @@ def find_not_ascii(s):
...
+                    "timestamp."
+                )
+                phase_time = Time(self.time_array[0], format="jd")
+                self.phase_to_time(phase_time)
+            else:
+                raise ValueError(
+                    "The data are in drift mode. "
+                    "Set force_phase to true to phase the data "
+                    "to zenith of the first timestamp before "
+                    "writing a uvfits file."
```
plaplant, line 950: This looks like a copy-paste typo (should be `measurement set` instead of `uvfits`)

kartographer:
@bhazelton -- barring a small amount of clean-up, I think the MS writer is ready to move to the next step. Round-trip w/ MIRIAD, MIR, UVFITS all seem to work as expected, with both multi-spw and multi-field data sets. Any objections moving this out of draft?

bhazelton:
Awesome! I have not objections to moving it out of draft stage and I'll try to read it soon. Can you update the PR description?

pyuvdata/uvdata/ms.py
bhazelton: @kartographer do you understand what this means?

pyuvdata/uvdata/ms.py
bhazelton: @kartographer have you learned anything about this?

pyuvdata/uvdata/ms.py
```diff
@@ -55,6 +115,672 @@ class MS(UVData):
...
+        Parameters
+        ----------
+        filepath : str
+            path to MS (without FIELD suffix)
+
+        """
+        if not casa_present:  # pragma: no cover
+            raise ImportError(no_casa_message) from casa_error
+
+        # TODO revisit for multi object!
```
bhazelton, line 383: Is this still an issue? I thought I saw a test using multiple fields.

pyuvdata/uvdata/ms.py
```diff
@@ -55,6 +115,672 @@ class MS(UVData):
...
+            path to MS (without POINTING suffix)
+
+        """
+        if not casa_present:  # pragma: no cover
+            raise ImportError(no_casa_message) from casa_error
+
+        pointing_table = tables.table(
+            filepath + "::POINTING", ack=False, readonly=False
+        )
+        # TODO: we probably have to fix this for steerable dishes
```
bhazelton, line 577: Is this still a TODO? If so, what line does it refer to?

pyuvdata/uvdata/ms.py
```diff
@@ -55,6 +115,672 @@ class MS(UVData):
...
+        name_col_desc = tableutil.makearrcoldesc(
+            "TELESCOPE_LOCATION",
+            self.telescope_location[0],
+            shape=[3],
+            valuetype="double",
+        )
+        observation_table.addcols(name_col_desc)
+        observation_table.putcell("TELESCOPE_LOCATION", 0, self.telescope_location)
+
+        # should we test for observer in extra keywords?
```
bhazelton, line 748: should we?

pyuvdata/uvdata/ms.py
```diff
@@ -126,11 +886,715 @@ def find_not_ascii(s):
...
+            raise ValueError(
+                "The phasing type of the data is unknown. "
+                "Set the phase_type to drift or phased to "
+                "reflect the phasing status of the data"
+            )
+
+        # Initialize a skelton measurement set
+        ms = self._init_ms_file(filepath)
+
+        # All all the rows we need up front, which will allow us to fill the columns
```
bhazelton, line 1094: All all -> Add all (I think?)

pyuvdata/uvdata/ms.py
```diff
@@ -126,11 +886,715 @@ def find_not_ascii(s):
...
+            )
+
+        if os.path.exists(filepath):
+            if clobber:
+                print("File exists; clobbering")
+            else:
+                raise IOError("File exists; skipping")
+
+        # TODO: the phasing requirement can probably go away once we have proper
+        # multi-object support.
```
bhazelton, line 1065: What are we missing on this front?

pyuvdata/uvdata/ms.py
```diff
@@ -126,11 +886,715 @@ def find_not_ascii(s):
...
+                if col == "WEIGHT_SPECTRUM":
+                    temp_weights[idx :: self.Nspws] = np.median(
+                        temp_array[:, sel_mask], axis=1
+                    )
+
+                val_dict.update(temp_dict)
+            ms.putvarcol(col, {key: val_dict[key] for key in sorted(val_dict.keys())})
+
+        # TODO: If/when UVData objects can store visibility noise estimates, update
+        # the code below to capture those.
```
bhazelton, line 1133: Fascinating. I didn't realize there was a place for these in measurement sets. We actually have something like this for HERA.

pyuvdata/uvdata/ms.py
```diff
@@ -126,11 +886,715 @@ def find_not_ascii(s):
...
+        )
+
+        # FITS uvw direction convention is opposite ours and Miriad's.
+        # CASA's convention is unclear: the docs contradict themselves,
+        # but after a question to the helpdesk we got a clear response that
+        # the convention is antenna2_pos - antenna1_pos, so the convention is the
+        # same as ours & Miriad's
+        ms.putcol("UVW", np.repeat(self.uvw_array, self.Nspws, axis=0))
+        if self.multi_phase_center:
+            # We have to do an extra bit of work here, because CASA won't accept arb
```
bhazelton, line 1159: arb -> arbitrary

pyuvdata/uvdata/ms.py
```diff
@@ -126,11 +886,715 @@ def find_not_ascii(s):
...
+            # values for field ID (rather, the ID number matches to the row number in
+            # the FIELD subtable). When we write out the fields, we use sort so that
+            # we get reproduce the same ordering here
```
bhazelton, line 1162: we get reproduce -> we reproduce

pyuvdata/uvdata/ms.py
```diff
@@ -126,11 +886,715 @@ def find_not_ascii(s):
...
+
+        Raises
+        ------
+        ValueError
+            If the MS file contains data from multiple subarrays.
+        """
+        tb_main = tables.table(filepath, ack=False)
+        time_arr = tb_main.getcol("TIME")
+        # N.b., EXPOSURE is what's needed for noise calculation, but INTERVAL defines
+        # the time period over which the data are collected
```
bhazelton, line 1330: Interesting! Does that mean it supports visibilities that have been integrated together with a gap between them? How does this interact with the observation times?

pyuvdata/uvdata/ms.py
```diff
@@ -126,11 +886,715 @@ def find_not_ascii(s):
...
+            check_mask = has_data[blt_idx]
+            check_idx = blt_idx[check_mask]
+
+            # Loop through the metadata fields we intend to populate
+            for arr, name in zip(arr_tuple, name_tuple):
+                if np.any(check_mask):
+                    if not np.all(data_dict[key][name][check_mask] == arr[check_idx]):
+                        # This is presently a warning, although it could conceivably
+                        # be made into an error. If doing so, we probably want to change
+                        # the above to 'is_close' rather than strict equality.
```
bhazelton, line 1520: Hmm. One part of me prefers warnings on read because they don't totally block users, but I worry that warnings are easily ignored. What value gets filled in if they vary? The first value? The most common value?

Another option would be to have a keyword that allows the user to turn this into a warning rather than an error but it defaults to erroring so that the user has to make a choice to ignore it.

pyuvdata/uvdata/ms.py
```diff
@@ -144,6 +1608,7 @@ def read_ms(
         strict_uvw_antpos_check=False,
         use_old_phase=False,
         fix_phase=False,
+        ignore_single_chan=True,
```
bhazelton, line 1611: These two parameters are not documented in the docstring.

pyuvdata/uvdata/ms.py
```diff
@@ -290,105 +1786,81 @@ def read_ms(
...
+        station_names = np.asarray(tb_ant.getcol("STATION"))[ant_good_position].tolist()
+        self.antenna_diameters = tb_ant.getcol("DISH_DIAMETER")[ant_good_position]
+
+        # importuvfits measurement sets store antenna names in the STATION column.
+        # cotter measurement sets store antenna names in the NAME column, which is
+        # inline with the MS definition doc. In that case all the station names are
+        # the same.
+        # TODO: Potentially flip which is the default -- use ant_names if filled and
+        # unique. This does seem to fail though on "day2_TDEM0003_10s_norx_1src_1spw".
+        # See the test "test_read_ms_read_uvfits".
```
bhazelton, line 1841: Depending on what clever words are written earlier I'd support this swap as long as we can come up with some way to catch cases like that file (which is a CASA tutorial test file I believe).

pyuvdata/uvdata/ms.py
```diff
@@ -290,105 +1786,81 @@ def read_ms(
...
+        self.antenna_positions = relative_positions
 
         tb_ant.close()
+
+        self._set_phased()
+
+        # set LST array from times and itrf
+        proc = self.set_lsts_from_time_array(background=background_lsts)
+
+        # CASA weights column keeps track of number of data points averaged.
```
bhazelton, line 1862: what code does this comment go with?

pyuvdata/uvdata/ms.py
```diff
@@ -399,62 +1871,138 @@ def read_ms(
...
+        self.data_array = np.expand_dims(self.data_array, 1)
+        self.nsample_array = np.expand_dims(self.nsample_array, 1)
+        self.flag_array = np.expand_dims(self.flag_array, 1)
+        self.freq_array = np.expand_dims(self.freq_array, 0)
+
         # order polarizations
-        self.reorder_pols(order=pol_order)
+        self.reorder_pols(order=pol_order, run_check=False)
+
+        # This is a thing
```
bhazelton, line 1984: What kind of thing?

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -192,6 +192,68 @@ def test_read_write_read_carma(tmp_path):
...
+            "bfmask in extra_keywords is a list, array or dict",
+            "Cannot fix the phases of multi phase center datasets, as they were not "
+            "supported when the old phasing method was used, and thus, there "
+            "is no need to correct the data.",
+        ],
+    ):
+        uv_in.read(carma_file, fix_old_proj=True)
+
+    # Dump the extra keywords here, since they aren't written to MS format anyways
+    uv_in.extra_keywords = {}
```
bhazelton, line 224: Is there not a way to write generic header keywords to ms files?

pyuvdata/uvdata/tests/test_ms.py
```diff
@@ -19,17 +19,37 @@
 
 
 @pytest.fixture(scope="session")
-def nrao_uv():
+def nrao_uv_main():
+    # This file is known to be made with CASA's importuvfits task, but apparently the
+    # history table is missing, so we can't infer that from the history. This means that
+    # the uvws are not flipped/data is not conjugated as they should be. Fix that.
```
bhazelton, line 25: Maybe we should add a check to read_uvfits that is similar to the check at the end of read_ms to see if a flip needs to happen.

pyuvdata/uvdata/tests/test_ms.py
```diff
@@ -19,17 +19,37 @@
...
+
     uvobj = UVData()
     testfile = os.path.join(DATA_PATH, "day2_TDEM0003_10s_norx_1src_1spw.ms")
     uvobj.read(testfile)
+    # uvobj.uvw_array *= -1
+    # uvobj.data_array = np.conj(uvobj.data_array)
```
bhazelton, line 31: Oh, wait. If these lines are out then should we take out the earlier comment?

pyuvdata/uvdata/tests/test_ms.py
```diff
@@ -234,13 +269,20 @@ def test_multi_files(casa_uvfits, axis):
...
+
     filesread = [testfile1, testfile2]
     # test once as list and once as an array
     if axis is None:
         filesread = np.array(filesread)
 
     uv_multi.read(filesread, axis=axis)
-    # Casa scrambles the history parameter. Replace for now.
+    # uv_multi.uvw_array *= -1
+    # uv_multi.data_array = np.conj(uv_multi.data_array)
```
bhazelton, line 283: remove these commented lines. If they go with the prior comment maybe remove it too?

bhazelton (commented):
This is awesome! Thank you for all the work @kartographer!

pyuvdata/uvdata/ms.py
kartographer: I believe this column is used for handling pointing offsets for an individual primary beam, i.e. when the primary beam isn't pointed at the phase center of the array. But because we set BEAM_ID=-1 above (and we don't have a BEAM table _yet_), none of this gets used at the moment. At some point further down the road, one ought to be able to fill some of that in with a `UVBeam` object.

pyuvdata/uvdata/ms.py
kartographer: From what I can tell through experimentation with ALMA and other data sets, this seems to be the polarization orientation of the feed relative to horizonal/vertical of the (ideal) primary. I don't know that what the corresponding attribute in `UVData` would be -- it's not actually clear that it really exists (`x_orientation` is related, but seems more to document convention rather than specifics of individual pol response of antennas).

pyuvdata/uvdata/ms.py
kartographer: Nope, that's a stale comment, I'll get rid of it.

pyuvdata/uvdata/ms.py
```diff
@@ -55,6 +115,672 @@ class MS(UVData):
...
+        # The required_ms_desc returns the defaults for a CASA MS table
+        ms_desc = tables.required_ms_desc("MAIN")
+
+        # The tables have a different choice of dataManagerType and dataManagerGroup
+        # based on a test ALMA dataset and comparison with what gets generated with
+        # a dataset that comes through importuvfits.
+        ms_desc["FLAG"].update(
+            dataManagerType="TiledShapeStMan", dataManagerGroup="TiledFlag",
+        )
+        # TODO: Check if MEASINFO:Ref should be ITRF or J2000
```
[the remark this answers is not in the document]
bhazelton: Weird. Ok!

pyuvdata/uvdata/ms.py
bhazelton: Yeah, I don't think that's `x_orientation`. Thanks for running that down!

pyuvdata/uvdata/ms.py
kartographer: It is, although not quite in the way its written. I think this actually requires some thought on how to implement in UVData -- it basically requires being able to separately handle where the phase center is vs where the primary beam center is. Right now, its set so that the two are the same, so interestingly enough it's fine for steerable dishes, but NOT for something more "PAPER"-like. I'll migrate this down below to where the DIRECTION and TARGET columns get written, since that's the place where things will need to get updated.

pyuvdata/uvdata/ms.py
kartographer: CASA does not appear to support "unphased" data, so we'll still need to keep this check (I actually need to add a check for when there is unphased data in a multi-obj data set).

pyuvdata/uvdata/ms.py
bhazelton: Ok. I think that currently UVData doesn't have information on where the primary beam center is, just on where the phase center is. You can of course change the phase center of the data set in later processing (although only at the coarseness of the integration). This is an issue in the MWA as well, because it has (electronically) steerable tiles. It would be good to be able to capture that information, we have had some ideas about that in the past (using a different object) that we should talk through.

pyuvdata/uvdata/ms.py
bhazelton: Ok, that sounds reasonable. Can we update the comment?

pyuvdata/uvdata/ms.py
kartographer: If you can think of it, CASA has an N-dimensional table ready to store said information for you, whether you like it or not ;)

pyuvdata/uvdata/ms.py
kartographer: Hypothetically, yes. If that's the case, then the TIME column would actually need to get updated, since it always refers to the temporal midpoint of the visibility, with INTERVAL setting the total length of the window. I believe this gets checked against things that are indexed by time (the one immediate example that pops into mind are the entries in POINTING, but I'm sure there are others).

kartographer: I've got no idea how it'd interact with the other fields in the overall MS structure, but from what's documented, I think it should support that (and in its current state, I don't think it would cause any issues with the data sets that the writer produces).

pyuvdata/uvdata/ms.py
kartographer: Yup, will do.

pyuvdata/uvdata/ms.py
kartographer: I think I might prefer the latter option (making this an error rather than a warning that can be switched off). There are probably some clever things that we can do about chosing the "best" value, but I might just prefer to default to the first one for now (since it's fastest), and worry about something smarter further down the road.

pyuvdata/uvdata/ms.py
kartographer: Heh, something that got deleted during code refactoring...

pyuvdata/uvdata/ms.py
kartographer: The most thingiest thing that you've ever seen!

pyuvdata/uvdata/tests/test_miriad.py
kartographer: Not that I can easily see -- one might be able to add arbitrary keywords to the full table itself (and I can try to experiment with that a bit), but I have no idea how CASA will behave with those in there. Let me give it a try and see what happens...

pyuvdata/uvdata/tests/test_ms.py
kartographer: So the check is in `ms.py` (the "this is a thing" comment marks where it is currently) -- I can certainly add a similar check to UVFITS, although if we're talking about expanding it to multiple filetypes, perhaps this is something that should be added to `UVData`? Either as a standalone method or as a piece of code in `check` that can be activated with a keyword?

bhazelton: Sure, that's probably the right way to implement it. I guess that this can be flipped easily because the images just work out.

pyuvdata/uvdata/tests/test_miriad.py
kartographer: Okay, I take back my previous comments. Arbitrary generic header keywords for everyone!

README.md
```diff
@@ -34,17 +34,18 @@ Also can convert raw data quality metrics into flags using thresholding.
 ## UVData File standard notes
 * miriad has been thoroughly tested with aipy-style miriad files and minimally
 tested with ATCA files
-* uvfits conforms to AIPS memo 117 (as of May 2015).  It is tested against
+* uvfits conforms to AIPS memo 117 (as of March 2020).  It is tested against
 FHD, CASA, and AIPS. However AIPS is limited to <80 antennas and CASA uvfits
 import does not seem to support >255 antennas.
+* CASA measurement sets (TODO: Add more here about how we test these).
```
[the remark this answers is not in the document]
kartographer: Fixed in de7365a7.

pyuvdata/uvdata/ms.py
```diff
@@ -55,6 +115,672 @@ class MS(UVData):
...
+            keywords={"CATEGORY": []},
+        )
+        ms_desc["WEIGHT"].update(
+            dataManagerType="TiledShapeStMan", dataManagerGroup="TiledWgt",
+        )
+        ms_desc["SIGMA"].update(
+            dataManagerType="TiledShapeStMan", dataManagerGroup="TiledSigma",
+        )
+
+        # The ALMA default for he next set of columns from the MAIN table use the
```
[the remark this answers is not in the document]
kartographer: Fixed in 3068e067.

pyuvdata/uvdata/ms.py
```diff
@@ -55,6 +115,672 @@ class MS(UVData):
...
+            ant_names_table[num] = self.antenna_names[ai]
+
+        # There seem to be some variation on whether the antenna names are stored
+        # in the NAME or STATION column (importuvfits puts them in the STATION column
+        # while Cotter and the MS definition doc puts them in the NAME column).
+        # The MS definition doc suggests that antenna names belong in the NAME column
+        # and the telescope name belongs in the STATION column (it gives the example of
+        # GREENBANK for this column.) so we follow that here.
+        # TODO: Karto, STATION == pad, NAME == Antenna name. Write something clever here
+        # about it.
```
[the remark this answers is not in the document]
kartographer: Added in 3068e067. Can't promise that much cleverness, but it's at least intelligible.

pyuvdata/uvdata/ms.py
```diff
@@ -55,6 +115,672 @@ class MS(UVData):
...
+            1, 3
+        )
+        ant_pos_table = np.zeros((nants_table, 3), dtype=np.float64)
+        for ai, num in enumerate(self.antenna_numbers):
+            ant_pos_table[num, :] = ant_pos_absolute[ai, :]
+
+        antenna_table.putcol("POSITION", ant_pos_table)
+        if self.antenna_diameters is not None:
+            ant_diam_table = np.zeros((nants_table), dtype=self.antenna_diameters.dtype)
+            # This is here is suppress an error that arises when one has antenans of
```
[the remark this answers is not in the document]
kartographer: Fixed in 3068e067.

pyuvdata/uvdata/ms.py
kartographer: This got deleted as part of moving the flip check out of `read_ms`.

pyuvdata/uvdata/tests/test_miriad.py
kartographer: Alright, added in 3068e067, the full dict in `extra_keywords` is now stored under the main MS table under the name `pyuvdata_extra`.

pyuvdata/uvdata/tests/test_ms.py
kartographer: Migraded, and the commented-out code has been deleted.

pyuvdata/uvdata/tests/test_ms.py
kartographer: Deleted.

pyuvdata/uvdata/tests/test_ms.py
kartographer: Deleted.

pyuvdata/uvdata/ms.py
kartographer: Eliminated in 3068e067.

pyuvdata/uvdata/ms.py
kartographer: We should! Added in 3068e067.

pyuvdata/uvdata/ms.py
kartographer: Whoops -- corrected in 3068e067.

pyuvdata/uvdata/ms.py
kartographer: Comment updated in 3068e067.

pyuvdata/uvdata/ms.py
kartographer: English is hard. Commenting code -- harder. Fixed in 3068e067.

pyuvdata/uvdata/ms.py
kartographer: Updated in 3068e067.

pyuvdata/uvdata/ms.py
kartographer: Documentation added in 3068e067.

pyuvdata/uvdata/ms.py
kartographer: So I did attempt this swap earlier, but it ended up throwing an error across several tests outside of `test_ms` that I couldn't immediately chase down. I haven't made a second attempt _yet_, although I could probably be talked into it.

pyuvdata/uvdata/ms.py
kartographer: Removed in 3068e067.

kartographer:
Thanks @bhazelton! I've gone through and updated the code in response to the comments (save for one small item), and I've added the last bit of test coverage that needed to get added.

plaplant (approved):
Thanks so much @kartographer!! This is amazing!

PR #1014, opened 2021-04-29
Add functionality for selecting on LSTs and an LST range

### LINKED ISSUE #593: add option to select on LST in addition to time (JD)
The underlying fear is that the mapping between JD in the BLT array and LST in that array has to be done correctly by the user.  

Not critical, but nice.

### LINKED ISSUE #618: add get_lst method (paralleling get_times)
for Danny (and maybe others)

plaplant:
## Description

This PR adds functionality for selecting data in a UVData object based on a series of specific LSTs or a range of LST, analogously to how times are handled. There is also support for select-on-read of LSTs for UVFITS and UVH5 files.

## Motivation and Context

There is a long-standing issue requesting this functionality. We have finally gotten around to adding it. As a part of this PR, we  also added a `get_lsts` method, which is analogous to the `get_times` method. It seemed close enough in spirit to just address.

Still to do:
- [x] Add a test for selecting data that wraps around LST=0.
- [x] Update the tutorial to reflect new functionality.
- [x] Add tests for doing a select-on-read for UVFITS and UVH5.
- [x] Update changelog.

Closes #593.
Closes #618.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/uvdata/uvdata.py
+++ b/pyuvdata/uvdata/uvdata.py
@@ -5233,10 +5233,24 @@ class UVData(UVBase):
 
             time_blt_inds = np.zeros(0, dtype=np.int64)
             for jd in times:
-                if jd in self.time_array:
-                    # TODO: maybe use a more tolerant check
+                if np.any(
+                    np.isclose(
+                        self.time_array,
+                        jd,
+                        rtol=self._time_array.tols[0],
+                        atol=self._time_array.tols[1],
+                    )
+                ):
                     time_blt_inds = np.append(
-                        time_blt_inds, np.where(self.time_array == jd)[0]
+                        time_blt_inds,
+                        np.where(
+                            np.isclose(
+                                self.time_array,
+                                jd,
+                                rtol=self._time_array.tols[0],
+                                atol=self._time_array.tols[1],
+                            )
+                        )[0],
                     )
                 else:
                     raise ValueError(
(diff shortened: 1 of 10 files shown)


jpober:
Have we tested the case where you read multiple nights of data into a single UV object (so the same lsts show up multiple times) and then select an lst or lst range?  Would we expect the logic here to work in that case?  It's not something that would come up with a JD select so it might not be fully supported...

mkolopanis:
just based on how the index matching is done, i would naively expect it to work. However it would be good to actually test that.

plaplant:
@jpober yes, I added a test that indeed selects the same LSTs from different days if you have them in the same object.

I think this PR is ready for a look whenever someone gets a chance.

pyuvdata/uvdata/uvdata.py
```diff
@@ -5158,17 +5213,28 @@ def _select_preprocess(
...
 
+        if times is not None:
             times = uvutils._get_iterable(times)
             if np.array(times).ndim > 1:
                 times = np.array(times).flatten()
 
             time_blt_inds = np.zeros(0, dtype=np.int64)
             for jd in times:
                 if jd in self.time_array:
+                    # TODO: maybe use a more tolerant check
```
bhazelton, line 5237: Should we turn this into an issue?

bhazelton (commented):
Thanks for tackling this! It looks really good to me, but we need to update the changelog.

pyuvdata/uvdata/uvdata.py
plaplant: Thanks for catching this! I've changed it to use `np.isclose` to test for membership, rather than `in`.

bhazelton (approved):
Looks good, thanks!

PR #1018, opened 2021-06-02
Remove polarization warning on multi-pol read

### LINKED ISSUE #888: polarization concatenation order and possible errors in writing output files
When concatenating files with different polarizations, the order in which they are read in (using the `read` method), or possibly added, can generate the error 

`Combined polarizations are not evenly spaced. This will make it impossible to write this data out to some file types.`

If the read order is `['xx','yy','xy','yx']` (AIPS ordering), this error is not generated.  The underlying problem is apparently that uvfits requires a specific order if all 4 are present.

Suggest allowing a read in any order (since pyuvdata doesn't appear to fundamentally care), but enforcing AIPS ordering on writing uvfits files (or other formats which require a particular order).

plaplant:
## Description

There is an issue where in the middle of reading files and combining along polarization, a spurious warning could be issued if the files were not read in using AIPS order. This is somewhat unfriendly to the user, and so we have removed this warning. We have also made it so files where the polarization order matters (currently just UVFITS) will reorder the polarization array on-the-fly, so that the user does not need to worry about polarizations being potentially out of order.

## Motivation and Context

This warning was originally issued in the case that the user wanted to write out to UVFITS, as it is the only file format we write that is particular about polarization order and spacing. However, this warning was overzealous, and would print out mid-read without checking the final object. We have made it so there is no warning, and instead an error is raised if the polarization axis is incompatible with writing a UVFITS file.

Closes #888.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:


bhazelton (commented):
This looks great! It looks like we don't sort the pols in the add method like we do the frequencies. Should we maybe add that?

plaplant:
@bhazelton I think we do sort it maybe? It looks like the relevant part of `__add__` is here:

https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/1ba3ccd13be24fcadc1e1dc470e7d5fcef14d3c4/pyuvdata/uvdata/uvdata.py#L3939-L3942

I think the `np.argsort` will reorder the polarizations by their absolute value, because we then index into the polarization axis of the combined arrays using `p_order`. We also have part of a test that explicitly adds the polarizations in the wrong order, and checks that they match the ordered original:

https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/1ba3ccd13be24fcadc1e1dc470e7d5fcef14d3c4/pyuvdata/uvdata/tests/test_uvdata.py#L3179-L3186

bhazelton:
oops, I guess I missed the pol reordering, glad it's there.

bhazelton (approved):
Looks good, thanks! I note that I don't think this needs a changelog entry because it is only removing a warning that is often spurious, it isn't changing functionality.

PR #1031, opened 2021-07-07
Add support for parameters that can be ignored in __eq__ methods

plaplant:
## Description
This PR adds an `allowed_failures` keyword argument to the `__eq__` method of UVBase objects (and subclasses) which allows the user to specify optional parameters that are allowed to be different in equality checks. It also adds the `filename` parameter automatically for UVData objects, which avoids recent test failures inadvertently introduced with the new parameter.

## Motivation and Context
We wanted to make the `filename` attribute optional-ish for equality checks. Now if the attribute is different between two objects, the user will be notified with a printed message, but the equality check will not fail.

## Types of changes
- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+
+## [Unreleased]
+
+### Changed
+- The `__eq__` method on UVBase objects (and subclasses) now supports an
+`allowed_failures` keyword for parameters which are allowed to be unequal
+without failing the check.
+
+
 ## [2.2.0] - 2021-6-26
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ All notable changes to this project will be documented in this file.
 ### Changed
 - The `__eq__` method on UVBase objects (and subclasses) now supports an
 `allowed_failures` keyword for parameters which are allowed to be unequal
-without failing the check.
+without failing the check. This includes the `filename` parameter of
+`UVData` objects by default.
 
 
 ## [2.2.0] - 2021-6-26
--- a/pyuvdata/uvbase.py
+++ b/pyuvdata/uvbase.py
@@ -279,12 +279,13 @@ class UVBase(object):
                 p_check = self_required
 
             if allowed_failures is not None:
-                for p in allowed_failures:
-                    if not p.startswith("_"):
-                        p = "_" + p
-                    if p not in self_required:
-                        if p in p_check:
-                            p_check.remove(p)
+                for i, param in enumerate(allowed_failures):
+                    if not param.startswith("_"):
+                        param = "_" + param
+                        allowed_failures[i] = param
+                    if param not in self_required:
+                        if param in p_check:
+                            p_check.remove(param)
 
             p_equal = True
             for p in p_check:
@@ -300,7 +301,7 @@ class UVBase(object):
             if allowed_failures is not None:
                 for p in allowed_failures:
                     self_param = getattr(self, p)
-                    other_param = getattr(self, p)
+                    other_param = getattr(other, p)
                     if self_param != other_param:
                         print(
                             f"parameter {p} does not match. Left is {self_param.value},"
--- a/pyuvdata/uvbase.py
+++ b/pyuvdata/uvbase.py
@@ -234,7 +234,7 @@ class UVBase(object):
         check_extra : bool
             Option to specify whether to include all parameters, or just the
             required ones. Default is True.
-        allowed_failues : list of str, optional
+        allowed_failures : list of str, optional
             List of parameter names that are allowed to fail while still passing
             an overall equality check. These should only include optional
             parameters.
--- a/pyuvdata/uvbase.py
+++ b/pyuvdata/uvbase.py
@@ -223,12 +223,26 @@ class UVBase(object):
         for a in extra_list:
             yield a
 
-    def __eq__(self, other, check_extra=True):
+    def __eq__(self, other, check_extra=True, allowed_failures=None):
         """
-        Equal if classes match and parameters are equal.
+        Test if classes match and parameters are equal.
 
-        If check_extra is True, include all parameters, otherwise only include
-        required parameters.
+        Parameters
+        ----------
+        other : class
+            Other class instance to check
+        check_extra : bool
+            Option to specify whether to include all parameters, or just the
+            required ones. Default is True.
+        allowed_failues : list of str, optional
+            List of parameter names that are allowed to fail while still passing
+            an overall equality check. These should only include optional
+            parameters.
+
+        Returns
+        -------
+        bool
+            Whether the two instances are equivalent.
         """
         if isinstance(other, self.__class__):
             # only check that required parameters are identical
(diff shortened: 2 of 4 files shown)


pyuvdata/uvbase.py
```diff
@@ -275,6 +297,17 @@ def __eq__(self, other, check_extra=True):
...
+
+            if allowed_failures is not None:
+                for p in allowed_failures:
+                    self_param = getattr(self, p)
+                    other_param = getattr(other, p)
+                    if self_param != other_param:
+                        print(
+                            f"parameter {p} does not match. Left is {self_param.value},"
+                            f" right is {other_param.value}."
+                        )
```
bhazelton, line 309: should we add something to this to clarify that this is an allowed failure so people don't worry about it?

pyuvdata/uvbase.py
```diff
@@ -260,11 +274,19 @@ def __eq__(self, other, check_extra=True):
                         f" right is {other_extra}."
                     )
                     return False
-
                 p_check = self_required + self_extra
             else:
                 p_check = self_required
 
+            if allowed_failures is not None:
+                for i, param in enumerate(allowed_failures):
```
mkolopanis, line 282: We never actually do any input verification on `allowed_failures`, should we? It's a lot of work for something so far down in the weeds but if only a string is given (maybe because I only want one to be allowed to fail) then this will iterate through the characters

bhazelton: yeah, we probably should make sure it's a list first.

pyuvdata/uvbase.py
mkolopanis: I think it could be helpful, otherwise it just looks like the failure of required parameters

pyuvdata/uvdata/uvdata.py
```diff
@@ -2574,6 +2574,40 @@ def check(
...
         return True
 
+    def __eq__(self, other, check_extra=True, allowed_failures=None):
+        """
+        Test for equality between two objects.
+
+        By default, this will add the `filename` attribute to the list of
+        parameters allowed to be not equal while passing an overall equality
+        check. The user is permitted to add other attributes to this list with
+        the `allowed_failures` keyword argument.
```
bhazelton, line 2584: I just realized that this approach does not let a user require that filenames are the same because we always add that parameter. Another option would be to set the default like this: `allowed_failures=["filename"]`. The downside there being that the user would have to include `"filename"` in the list themselves, but if a user is modifying the default to this magic method they probably know what they are doing.

mkolopanis: Also it seems a little silly to me that someone would need the filename to be equal, but I'm sure there's a reasonable case for it. Isn't it more important all the (meta)data is the same in the end?

kartographer: I can think of a _potential_ (but not yet supported) use-case where you want the `UVData` object to have only the metadata, and load the actual visibilities on demand. Something like this is implemented in the `MirParser` class -- may be useful for dealing w/ larger file sizes in the future. But again, it's not actually a supported thing (yet...).

mkolopanis: I think I'm a little confused about the filename here though? Wouldn't all the metadata being equal tell you it's the same dataset?

kartographer: I suppose that's probably true -- or rather, you'd have to have something that _only_ acted on the data but not the metadata. Only thing that comes to mind would be applying flagging or gains, although I don't know if performing those actions touches the metadata in some way that I'm not aware of.

mkolopanis: I think I just don't understand the point you're trying to make.

plaplant: If I understand @kartographer's point correctly, I think it's similar to a change I made for partial writing in UVH5: 

https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/4c66208245db29be8b9920434d6d48d23b29e367/pyuvdata/uvdata/uvh5.py#L1600-L1677

In this case, we're checking that the partially-written file on disk matches the object in memory, but unless the user is overwriting the file (bad idea!), the `filename` won't match. I explicitly ignore that parameter when testing equality. But I think the case mentioned is sort of the inverse, where you want to make sure the filename in the partially-loaded object matches the one you ask to fetch more data from (right)? In that case, the function could maybe just check that `filename` matches explicitly.

pyuvdata/uvbase.py
plaplant: Okay, I can tweak the message that's printed to make it explicit that it's okay they're different.

pyuvdata/uvbase.py
plaplant: Sure, I can turn it into a list if it's a string. Good suggestion.

pyuvdata/uvdata/uvdata.py
kartographer: Sorry, I think I'm getting a little overly in the weeds on this -- let me take it back a step. I can imagine a case where one might want to know _if_ to UVData objects came from the same file versus two different files (even if the data they contain is identical). These cases that I can imagine are potentially corner-cases, but I think you want to have the user be able to expose that capability if it was so desired.

I think the core issue here isn't so much that this particular parameter (`UVData.filename`) is excluded by default, it's that `allowed_failures=None` is being treated as synonymous with some default settings, rather than having `allowed_failures=['filename']` actually being set explicitly in the function definition. I can see users wanting to have the control afforded by the latter, wherever possible.

mkolopanis: Okay I think I see that. We could make the default a tuple `("filename")` that would be immutable at least and get around the really awful things that happen with mutable kwargs in functions :thinking: Though that definitely be different than most of the rest of the codebase.

pyuvdata/uvbase.py
plaplant: Updated in 00cb301.

pyuvdata/uvbase.py
plaplant: Updated in 00cb301.

pyuvdata/uvdata/uvdata.py
plaplant: Addressed in 00cb301.

bhazelton: This looks good now, thank you! I'm not sure that you still need the longer explanation in the main part of the docstring since the default is made clear in the Parameters section.

plaplant: Good point, removed in 2b832011.

CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+
+## [Unreleased]
+
+### Changed
+- The `__eq__` method on UVBase objects (and subclasses) now supports an
+`allowed_failures` keyword for parameters which are allowed to be unequal
+without failing the check.
```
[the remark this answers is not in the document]
plaplant: Good suggestion, done in 141b431.

bhazelton (approved):
Looks good to me!

PR #1035, opened 2021-07-20
add a method to Initialize a UVCal object from the metadata in a UVData object

### LINKED ISSUE #753: UVCal method to inherit metadata from a UVData object
Be able to copy all relevant metadata from a UVData object into a UVCal object.

bhazelton:
## Description

Added a new method to UVCal to initialize a UVCal object from an UVData object.

## Motivation and Context

closes #753

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/uvcal/tests/conftest.py
+++ b/pyuvdata/uvcal/tests/conftest.py
@@ -11,6 +11,7 @@ import numpy as np
 from pyuvdata.data import DATA_PATH
 from pyuvdata import UVCal
 import pyuvdata.tests as uvtest
+import pyuvdata.uvcal.tests.test_fhd_cal as test_fhd_cal
 
 
 @pytest.fixture(scope="session")
@@ -113,3 +114,62 @@ def delay_data_inputflag_future(delay_data_inputflag_future_main):
     yield delay_object
 
     del delay_object
+
+
+@pytest.fixture(scope="session")
+def fhd_cal_raw_main():
+    """Read in raw FHD cal."""
+    fhd_cal = UVCal()
+    with uvtest.check_warnings(
+        UserWarning, match="Telescope location derived from obs lat/lon/alt values"
+    ):
+        fhd_cal.read_fhd_cal(
+            test_fhd_cal.cal_testfile,
+            test_fhd_cal.obs_testfile,
+            layout_file=test_fhd_cal.layout_testfile,
+            settings_file=test_fhd_cal.settings_testfile,
+            raw=True,
+        )
+
+    yield fhd_cal
+
+    del fhd_cal
+
+
+@pytest.fixture(scope="function")
+def fhd_cal_raw(fhd_cal_raw_main):
+    """Make function raw FHD cal object."""
+    fhd_cal = fhd_cal_raw_main.copy()
+
+    yield fhd_cal
+
+    del fhd_cal
+
+
+@pytest.fixture(scope="session")
+def fhd_cal_fit_main():
+    """Read in fit FHD cal."""
+    fhd_cal = UVCal()
+    with uvtest.check_warnings(
+        UserWarning, match="Telescope location derived from obs lat/lon/alt values"
+    ):
+        fhd_cal.read_fhd_cal(
+            test_fhd_cal.cal_testfile,
+            test_fhd_cal.obs_testfile,
+            layout_file=test_fhd_cal.layout_testfile,
+            settings_file=test_fhd_cal.settings_testfile,
+            raw=False,
+        )
+    yield fhd_cal
+
+    del fhd_cal
+
+
+@pytest.fixture(scope="function")
+def fhd_cal_fit(fhd_cal_fit_main):
+    """Make function fit FHD cal object."""
+    fhd_cal = fhd_cal_fit_main.copy()
+
+    yield fhd_cal
+
+    del fhd_cal
--- a/pyuvdata/uvcal/tests/test_fhd_cal.py
+++ b/pyuvdata/uvcal/tests/test_fhd_cal.py
@@ -39,24 +39,20 @@ settings_file_multi = [
 ]
 
 
-@pytest.mark.filterwarnings("ignore:Telescope location derived from obs lat/lon/alt")
 @pytest.mark.parametrize("raw", [True, False])
-def test_read_fhdcal_raw_write_read_calfits(raw, tmp_path):
+def test_read_fhdcal_write_read_calfits(raw, fhd_cal_raw, fhd_cal_fit, tmp_path):
     """
     FHD cal to calfits loopback test.
 
     Read in FHD cal files, write out as calfits, read back in and check for
     object equality.
     """
-    fhd_cal = UVCal()
+    if raw:
+        fhd_cal = fhd_cal_raw
+    else:
+        fhd_cal = fhd_cal_fit
+
     calfits_cal = UVCal()
-    fhd_cal.read_fhd_cal(
-        cal_testfile,
-        obs_testfile,
-        layout_file=layout_testfile,
-        settings_file=settings_testfile,
-        raw=raw,
-    )
 
     filelist = [cal_testfile, obs_testfile, layout_testfile, settings_testfile]
 
@@ -68,7 +64,19 @@ def test_read_fhdcal_raw_write_read_calfits(raw, tmp_path):
     calfits_cal.read_calfits(outfile)
     assert fhd_cal == calfits_cal
 
-    # check metadata only read
+
+@pytest.mark.filterwarnings("ignore:Telescope location derived from obs lat/lon/alt")
+@pytest.mark.parametrize("raw", [True, False])
+def test_read_fhdcal_metadata(raw, fhd_cal_raw, fhd_cal_fit):
+    """
+    Test FHD cal metadata only read.
+    """
+    if raw:
+        fhd_cal_full = fhd_cal_raw
+    else:
+        fhd_cal_full = fhd_cal_fit
+
+    fhd_cal = UVCal()
     fhd_cal.read_fhd_cal(
         cal_testfile,
         obs_testfile,
@@ -77,24 +85,25 @@ def test_read_fhdcal_raw_write_read_calfits(raw, tmp_path):
         raw=raw,
         read_data=False,
     )
-    calfits_cal2 = calfits_cal.copy(metadata_only=True)
+
+    fhd_cal2 = fhd_cal_full.copy(metadata_only=True)
 
     # this file set has a mismatch in Nsources between the cal file & settings
     # file for some reason. I think it's just an issue with the files chosen
-    assert fhd_cal.Nsources != calfits_cal2.Nsources
-    fhd_cal.Nsources = calfits_cal2.Nsources
+    assert fhd_cal.Nsources != fhd_cal2.Nsources
+    fhd_cal.Nsources = fhd_cal2.Nsources
 
     # there is a loss in precision for float auto scale values in the
     # settings file vs the cal file
     assert (
         fhd_cal.extra_keywords["autoscal".upper()]
-        != calfits_cal2.extra_keywords["autoscal".upper()]
+        != fhd_cal2.extra_keywords["autoscal".upper()]
     )
-    fhd_cal.extra_keywords["autoscal".upper()] = calfits_cal2.extra_keywords[
+    fhd_cal.extra_keywords["autoscal".upper()] = fhd_cal2.extra_keywords[
         "autoscal".upper()
     ]
 
-    assert fhd_cal == calfits_cal2
+    assert fhd_cal == fhd_cal2
 
     fhd_cal.read_fhd_cal(
         cal_testfile,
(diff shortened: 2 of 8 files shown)


bhazelton:
@jsdillon I'd love your feedback on this.

jsdillon (changes requested):
In general, looks great! I think this will help us simplify some `hera_cal` functions and will serve our purposes over there.

I've got a handful of minor comments and suggestions, including possible simplifications (the code as written is super general, which is maybe what you want for pyuvdata, but it does incur some long-term support costs that might not be totally worthwhile... it's a judgement call).

pyuvdata/uvcal/uvcal.py
```diff
@@ -3351,6 +3356,436 @@ def _convert_to_filetype(self, filetype):
...
+        """
+        Initialize this object based on a UVData object.
+
+        Parameters
+        ----------
+        uvdata : UVData object
+            The UVData object to initialize from.
+        gain_convention : str
+            What gain convention the UVCal object should be initialized to
+            ("minus" or "plus").
```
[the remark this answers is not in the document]
bhazelton: oops, you're right. There is a `delay_convention`  that is `minus` or `plus` but that's different.

pyuvdata/uvcal/uvcal.py
```diff
@@ -3351,6 +3356,436 @@ def _convert_to_filetype(self, filetype):
...
+        metadata_only : bool
+            Option to only initialize the metadata. If False, this method also
+            initializes the data-like arrays with the appropriate sizes.
+        include_uvdata_history : bool
+            Option to include the history from the uvdata object in the uvcal history.
+        cal_type : str
+            What cal_type the UVCal object should be initialized to
+            ("gain", or "delay").
+        times : array_like of float, optional
+            Calibration times in decimal Julian date. If None, use all unique times from
```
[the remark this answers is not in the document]
bhazelton: Hmm. I'm a little hesitant to make this so specifically tuned to BDA data. You can always pass in a time array that does exactly what you want. If we want this to have special defaults for BDA data I think I need to add another keyword to let users indicate that.

pyuvdata/uvcal/uvcal.py
```diff
@@ -3351,6 +3356,436 @@ def _convert_to_filetype(self, filetype):
...
+            "telescope_name",
+            "telescope_location",
+            "antenna_numbers",
+            "antenna_names",
+            "antenna_positions",
+            "Nants_telescope",
+            "Nants_data",
+            "x_orientation",
+        ]
+        if self.cal_type != "delay" and not self.wide_band:
```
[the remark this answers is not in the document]
bhazelton: It is complicated, but it'll get better once the future array shapes change is made. We actually have increasing interest in this object from non-HERA users, so I don't want to only make this work for HERA.

pyuvdata/uvcal/uvcal.py
```diff
@@ -3351,6 +3356,436 @@ def _convert_to_filetype(self, filetype):
...
+        ----------
+        uvdata : UVData object
+            The UVData object to initialize from.
+        gain_convention : str
+            What gain convention the UVCal object should be initialized to
+            ("minus" or "plus").
+        cal_style : str
+            What calibration style the UVCal object should be initialized to
+            ("sky" or "redundant").
+        future_array_shapes : bool
```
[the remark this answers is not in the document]
bhazelton: That's an interesting idea, it probably would make things simpler. I'll check with the core devel team about what they think.

pyuvdata/uvcal/uvcal.py
```diff
@@ -3351,6 +3356,436 @@ def _convert_to_filetype(self, filetype):
...
+                self.integration_time = integration_time
+        self.Ntimes = self.time_array.size
+        self.set_lsts_from_time_array()
+
+        if time_range is not None:
+            self.time_range = time_range
+
+        if jones is None:
+            if np.all(uvdata.polarization_array < -4):
+                self.jones_array = np.array([-5, -6])
```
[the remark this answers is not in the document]
bhazelton: Ok, I'll add handling for that case.

pyuvdata/uvcal/uvcal.py
```diff
@@ -436,6 +437,7 @@ def __init__(self):
             "Nsources", required=False, expected_type=int, description=desc
         )
 
+        # TODO add units. wavelengths??
```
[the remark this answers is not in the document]
bhazelton: Yes, thanks for reminding me. I think we probably want this to be in meters to be in line with other pyuvdata units. I think this is currently only really used by FHD, I need to figure out what units they come in from FHD with.

pyuvdata/uvcal/uvcal.py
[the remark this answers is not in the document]
bhazelton: I guess we could test to see if the integration time was varying -- that would indicate that there might be BDA (or some other screwiness).

pyuvdata/uvcal/uvcal.py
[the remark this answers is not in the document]
bhazelton: I turned this into an issue, I'll remove this TODO.

pyuvdata/uvcal/uvcal.py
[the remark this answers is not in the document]
bhazelton: We discussed this on our telecon today. We think that allowing the user to pass in the times they want make this flexible enough and don't want to add more options to allow for any possible desired thing with BDA. In our discussion we agreed that using the `integration_time` array to detect baselines with different levels of averaging would be the better approach for the HERA case, so you might want to do that instead of what you showed in that link.

pyuvdata/uvcal/uvcal.py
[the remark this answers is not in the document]
bhazelton: We want to keep it this way because we want to nudge users to use the future array shapes on UVCal, even if their UVData object doesn't use the future shapes.

jsdillon (dismissed):
Thanks for the changes. I approve!

bhazelton:
I still need to update the changelog and maybe the tutorial. Will try to get those in soon.

bhazelton:
@jsdillon FYI I turned this into a classmethod instead to make it a one-liner and I added a section to the tutorial on it.

pyuvdata/uvcal/tests/conftest.py
```diff
@@ -113,3 +114,62 @@ def delay_data_inputflag_future(delay_data_inputflag_future_main):
     yield delay_object
 
     del delay_object
+
+
+@pytest.fixture(scope="session")
```
plaplant, line 119: Nice use of fixtures for these files 👍

pyuvdata/uvcal/tests/test_fhd_cal.py
```diff
@@ -77,24 +85,25 @@ def test_read_fhdcal_raw_write_read_calfits(raw, tmp_path):
...
-    fhd_cal.Nsources = calfits_cal2.Nsources
+    assert fhd_cal.Nsources != fhd_cal2.Nsources
+    fhd_cal.Nsources = fhd_cal2.Nsources
 
     # there is a loss in precision for float auto scale values in the
     # settings file vs the cal file
     assert (
         fhd_cal.extra_keywords["autoscal".upper()]
-        != calfits_cal2.extra_keywords["autoscal".upper()]
+        != fhd_cal2.extra_keywords["autoscal".upper()]
```
plaplant, line 100: This is a nit-pick, but are the approximately true? Like would they pass a `numpy.isclose` test? If so I might advocate for doing that before overwriting the values so the object equality will pass.

pyuvdata/uvcal/uvcal.py
```diff
@@ -1390,7 +1394,7 @@ def get_gains(self, ant, jpol=None, squeeze_pol=True):
         Returns
         -------
         complex ndarray
-            Gain solution of shape (Nfreqs, Ntimes, Npol) or (Nfreqs, Ntimes)
+            Gain solution of shape (Nfreqs, Ntimes, NJones) or (Nfreqs, Ntimes)
```
plaplant, line 1397: Another nit-pick, but I think the parameter is `Njones` (i.e., lower case `j`). I think we should be consistent in all the docstrings.

pyuvdata/uvcal/uvcal.py
```diff
@@ -3351,6 +3355,446 @@ def _convert_to_filetype(self, filetype):
...
+        if not issubclass(type(uvdata), UVData):
+            raise ValueError("uvdata must be a UVData (or subclassed) object.")
+
+        # re-initialize to make sure we have an empty object
+        uvc = cls()
+
+        if cal_type == "gain":
+            uvc._set_gain()
+        elif cal_type == "delay":
+            uvc._set_delay()
```
plaplant, line 3513: Should we be raising an error if `cal_type` doesn't belong to one of these? (I forget if we support cal objects that are essentially "unknown"...)

pyuvdata/uvcal/uvcal.py
```diff
@@ -3351,6 +3355,446 @@ def _convert_to_filetype(self, filetype):
...
+        if time_range is not None:
+            uvc.time_range = time_range
+
+        if jones is None:
+            if np.all(uvdata.polarization_array < -4):
+                if uvdata.Npols == 1 and uvdata.polarization_array[0] > -7:
+                    # single pol data, make a single pol cal object
+                    uvc.jones_array = uvdata.polarization_array
+                else:
+                    uvc.jones_array = np.array([-5, -6])
```
plaplant, line 3760: Very clever! One small question: do we have to worry about the order of the `jones_array`? I forget if we use it to index anything...

pyuvdata/uvcal/tests/test_uvcal.py
```diff
@@ -2585,3 +2585,846 @@ def test_set_antpos_from_telescope_errors(gain_data, modtype, tmp_path):
...
+    # but that requirement is not enforced. Set it to None for this test
+    uvc.time_range = None
+
+    uvc2 = uvc.copy(metadata_only=True)
+
+    uvc_new = UVCal.initialize_from_uvdata(
+        uvd, uvc.gain_convention, uvc.cal_style, future_array_shapes=uvcal_future_shapes
+    )
+
+    # antenna positions are different by ~6cm or less. The ones in the uvcal file
```
plaplant, line 2629: Very informative comments 🙂

pyuvdata/uvcal/uvcal.py
```diff
@@ -3351,6 +3355,446 @@ def _convert_to_filetype(self, filetype):
...
+            uvc.jones_array = np.asarray(jones)
+        uvc.Njones = uvc.jones_array.size
+
+        uvc.ant_array = np.union1d(uvdata.ant_1_array, uvdata.ant_2_array)
+
+        uvc.history = "Initialized from a UVData object with pyuvdata."
+        if include_uvdata_history:
+            uvc.history += " UVData history is: " + uvdata.history
+
+        if not metadata_only:
```
plaplant, line 3782: I think "initialize with zeros" is the correct default. But maybe we want to make this explicit to the user in the docstring? (I could imagine the user potentially thinking the flag array might be transferred, for instance...)

plaplant (commented):
Very excellent @bhazelton! Thanks so much for this nice functionality and tons of new tests. I think the core functionality is great, I just have a few small suggestions around documentation.

pyuvdata/uvcal/uvcal.py
bhazelton: good catch! I found a few more and fixed them all.

pyuvdata/uvcal/uvcal.py
bhazelton: yes. We're planning to remove the "unknown" one, so I'm going to have it error if it's not "gain" or "delay"

pyuvdata/uvcal/uvcal.py
bhazelton: We only use it to index the data-like arrays. So when the user fills out those arrays they need to line them up with the `jones_array`, but I don't think we need to worry about anything in this method.

pyuvdata/uvcal/uvcal.py
bhazelton: I have this in the docstring under the `metadata_only` parameter:
```
        metadata_only : bool
            Option to only initialize the metadata. If False, this method also
            initializes the data-like arrays to zeros (or False for the flag_array)
            with the appropriate sizes.
```
Should I add something at the top level? The default is `metadata_only=True`.

pyuvdata/uvcal/uvcal.py
plaplant: Sounds good, thanks for clarifying.

pyuvdata/uvcal/uvcal.py
plaplant: Ah thanks, didn't realize it was in the parameter docstring. Because the default is `metadata_only=True`, then I don't think we need extra documentation beyond this.

plaplant (approved):
Thanks @bhazelton! Looks great. I'll leave it up to you to merge in given this affects the tutorial (and #1139).

PR #1042, opened 2021-07-28
Ensure that flag_arrays are boolean

### LINKED ISSUE #1039: The expected_type of the flag_array on UVData and UVCal objects is `(int, numpy.integer)` instead of `bool`
The expected_type of the flag_array on UVData and UVCal objects is `(int, numpy.integer)` even though they are being set to `bool` because in the `_get_generic_type` method `issubclass(bool, (int, np.integer))` is True.

Is this the expected behavior? I would have thought that we'd want it to be a `bool` so that flag_arrays filled with integers would not pass the check.

I think we might want to add `(bool, np.bool_)` at the top of the for loop over types in `_get_generic_type`. If I do that, the flag_array expected_type becomes `(bool, np.bool_)` and all the existing tests pass.

bhazelton:
## Description

- Added booleans to parameter types and fixed equality checking for arrays with strict types.
- Added tests that integer flag arrays fail checks.
- Some test fixture cleanup in test_uvdata.py

## Motivation and Context

fixes #1039 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/parameter.py
+++ b/pyuvdata/parameter.py
@@ -53,6 +53,7 @@ def _get_generic_type(expected_type, strict_type_check=False):
         return expected_type
 
     for types in [
+        (bool, np.bool_),
         (float, np.floating),
         (np.unsignedinteger),  # unexpected but just in case
         (int, np.integer),
--- a/pyuvdata/parameter.py
+++ b/pyuvdata/parameter.py
@@ -245,7 +245,7 @@ class UVParameter(object):
                         print(
                             f"{self.name} parameter has incompatible types. Left "
                             f"requires {type(self.value)}, right is "
-                            "{other.expected_type}"
+                            f"{other.expected_type}"
                         )
                         return False
                 if other.strict_type:
--- a/pyuvdata/parameter.py
+++ b/pyuvdata/parameter.py
@@ -191,8 +191,9 @@ class UVParameter(object):
                             return False
                     elif not isinstance(self.value.item(0), other.expected_type):
                         print(
-                            f"{self.name} parameter has incompatible dtypes. Left is "
-                            f"{self.value.dtype}, right is {other.expected_type}"
+                            f"{self.name} parameter has incompatible dtypes. Left "
+                            f"requires {self.expected_type}, right is "
+                            f"{other.value.dtype}"
                         )
                         return False
                 elif other.strict_type:
@@ -200,7 +201,7 @@ class UVParameter(object):
                     if not isinstance(other.value.item(0), self.expected_type):
                         print(
                             f"{self.name} parameter has incompatible dtypes. Left is "
-                            f"{other.value.dtype}, right is {other.expected_type}"
+                            f"{self.value.dtype}, right requires {other.expected_type}"
                         )
                         return False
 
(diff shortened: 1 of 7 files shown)


pyuvdata/parameter.py
```diff
@@ -192,6 +176,34 @@ def __eq__(self, other):
...
+                            f"{self.name} parameter has incompatible dtypes. Left is "
+                            f"{self.value.dtype}, right is {other.expected_type}"
+                        )
+                        return False
+                elif other.strict_type:
+                    # types must match in the other direction
+                    if not isinstance(other.value.item(0), self.expected_type):
+                        print(
+                            f"{self.name} parameter has incompatible dtypes. Left is "
+                            f"{other.value.dtype}, right is {other.expected_type}"
```
mkolopanis, line 203: We're using `other` for both left and right here. Should right be `self.expected_type`?

pyuvdata/parameter.py
```diff
@@ -225,6 +237,24 @@ def __eq__(self, other):
...
+                            f"{self.name} parameter has incompatible types. Left is "
+                            f"{type(self.value)}, right is {other.expected_type}"
+                        )
+                        return False
+                if other.strict_type:
+                    # types must match in the other direction
+                    if not isinstance(other.value, self.expected_type):
+                        print(
+                            f"{self.name} parameter has incompatible types. Left is "
+                            f"{type(other.value)}, right is {other.expected_type}"
```
mkolopanis, line 254: same comment here about `other` being both left and right.

mkolopanis (commented):
This looks really good. Thanks for cleaning up the tests as well. I have a question about some docstrings but should be good to go after that.

pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -255,6 +255,28 @@ def hera_uvh5_split(hera_uvh5_split_main):
     yield uv1_copy, uv2_copy, uvfull_copy
 
 
+@pytest.fixture(scope="session")
+def hera_uvh5_xx_main():
+    """Read in a HERA uvh5 file."""
+    hera_uvh5_xx = UVData()
+    hera_uvh5_xx.read_uvh5(os.path.join(DATA_PATH, "zen.2457698.40355.xx.HH.uvcA.uvh5"))
+
+    return hera_uvh5_xx
```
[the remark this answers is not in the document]
bhazelton: me neither, happy to add some cleanup.

pyuvdata/parameter.py
bhazelton: Yes, I fixed this one and the one before to be more clear.

bhazelton:
@mkolopanis I think I addressed your comments, but now the linter and doctest are failing on the initial linux spinup. I noticed that those two jobs have the os specified to be "linux" whereas the regular tests use a docker image.

mkolopanis:
Interesting that we specifically set the OS there, I would think we would be agnostic to it. Maybe they are older copy/pasta?

pyuvdata/parameter.py
```diff
@@ -225,6 +238,25 @@ def __eq__(self, other):
...
                     return False
             else:
+                # check to see if strict types are used
+                if self.strict_type:
+                    # types must match
+                    if not isinstance(self.value, other.expected_type):
+                        print(
+                            f"{self.name} parameter has incompatible types. Left "
+                            f"requires {type(self.value)}, right is "
+                            "{other.expected_type}"
```
mkolopanis, line 248: looks like we're missing an `f` here?
```suggestion
                            f"{other.expected_type}"
```

bhazelton:
@mkolopanis I fixed the CI issues on this PR

PR #1046, opened 2021-08-25
Add methods to set data, flags, and nsamples

### LINKED ISSUE #886: Consider making `put_*` methods matching `get_*` methods
A request form James, needs more thinking.

plaplant:
## Description

This PR adds the ability for the user to set data, flags, and nsamples arrays for UVDat objects. This is essentially the inverse of the `UVData.get_data()` et al. methods. By providing an appropriately sized array and the corresponding keys that would fetch data, the user is able to overwrite the data in the arrays without having to index into the datasets by hand.

## Motivation and Context

This PR leverages the fancy dataset indexing that was originally developed for performing optimized reading of UVH5 data. It has been made more general to support setting data in arrays, but the reading portion of the methods should work the same as before.

Still to do:
* add more test coverage
* add examples to the tutorial
* update the CHANGELOG

Closes #886.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -310,8 +310,15 @@ slower integral implementation.
 
 UVData: Quick data access
 -------------------------
-A small suite of functions are available to quickly access numpy arrays of data,
-flags, and nsamples.
+A small suite of functions are available to quickly access the underlying numpy
+arrays of data, flags, and nsamples. Although the user can perform this indexing
+by hand, several convenience functions exist to easily extract specific subsets
+corresponding to antenna-pair and/or polarization combinations. There are three
+specific methods that will return numpy arrays: ``get_data``, ``get_flags``, and
+``get_nsamples``. When possible, these methods will return numpy MemoryView
+objects, which is relatively fast and adds minimal memory overhead. There are
+also corresponding methods ``set_data``, ``set_flags``, and ``set_nsamples``
+which will overwrite sections of these datasets with user-provided data.
 
 a) Data for single antenna pair / polarization combination.
 ************************************************************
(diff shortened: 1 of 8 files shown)


pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -10948,4 +10948,324 @@ def test_eq_allowed_failures_filename_string(bda_test_file, capsys):
...
+    Test an error is raised when a key has too many values.
+    """
+    uv = hera_uvh5
+
+    ant1 = np.unique(uv.antenna_numbers)[0]
+    ant2 = np.unique(uv.antenna_numbers)[1]
+    data = uv.get_data(ant1, ant2, squeeze="none", force_copy=True)
+    with pytest.raises(ValueError) as cm:
+        uv.set_data(data, (ant1, ant2, "xx", "foo"))
+    assert str(cm.value).startswith("no more than 3 key values can be passed")
```
bhazelton, line 11115: you could just use the `match` keyword in the pytest.raises call (here and below).

bhazelton (commented):
It seems like there's a lot of copy and pasted logic among the different set methods. Could we move some of that into a helper function to reduce that?

pyuvdata/uvdata/uvdata.py
```diff
@@ -3437,6 +3443,231 @@ def get_ENU_antpos(self, center=False, pick_data_ants=False):
...
+        """
+        Set the nsamples array to some values provided by the user.
+
+        Parameters
+        ----------
+        nsamples : ndarray of float
+            The nsamples to overwrite into the nsample_array. Must be the same
+            shape as the target indices.
+        key1, key2, key3 : int or tuple of ints
+            Identifier of which nsamples to get, can be passed as 1, 2, or 3
```
bhazelton, line 3634: get -> set

docs/tutorial.rst
```diff
@@ -411,6 +411,23 @@ g) Quick access to file attributes of a UV* object (UVData, UVCal, UVBeam)
   >>> # Load object to instance name "uv" and will remain in interpreter
   pyuvdata_inspect.py -i <uv*_file>
 
+h) Update data arrays in place for UVData
```
bhazelton, line 414: I think maybe this should be moved to between the existing d & e items, it should certainly come before the existing g item. 

Maybe we should also update the top-level text under this whole section ("UVData: Quick data access"), it's pretty terse at the moment.

bhazelton:
I think we also need to update the changelog on this PR.

pyuvdata/uvdata/tests/test_uvdata.py
plaplant: Done in 0341185

pyuvdata/uvdata/uvdata.py
plaplant: Done in 31c16d5

docs/tutorial.rst
plaplant: Done in bb9c09e

plaplant:
@bhazelton thanks for reviewing this! I think I've addressed all of the comments, but let me know if there's anything else that should be done.

bhazelton (approved):
Looks good, thanks!

PR #1058, opened 2021-09-27
Warn rather than error if the integration time is missing in uvfits

### LINKED ISSUE #1057: Integration time Error with OVRO-LWA data
I have OVRO-LWA uvfits files which were converted from ms and am getting an error when I try to read them in. 

The data has a single integration time of 10s, but when I attempt to use `read_uvfits`, I get this error: "ValueError: integration time not specified and only one time present". I tried to read it in with `run_check` as False, but it still gave the same error.

bhazelton:
## Description

Warn rather than error if the integration time is missing in uvfits. This allows users to read the file into UVData if they set run_check=False.

## Motivation and Context

fixes #1057

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/uvdata/tests/test_uvfits.py
+++ b/pyuvdata/uvdata/tests/test_uvfits.py
@@ -551,21 +551,26 @@ def test_readwriteread_error_single_time(tmp_path, casa_uvfits):
         hdulist = fits.HDUList(hdus=[vis_hdu, ant_hdu])
         hdulist.writeto(write_file2, overwrite=True)
 
-    with pytest.raises(ValueError) as cm:
+    with pytest.raises(
+        ValueError, match="Required UVParameter _integration_time has not been set"
+    ):
         with uvtest.check_warnings(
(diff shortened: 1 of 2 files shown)


mkolopanis (approved):
Seems like an easy fix. 

This has kind of sparked an idea in my mind about competing ideologies for UVData probably worth discussing in a wider group. We seem to have competing ideas about being brittle and flexible both at the same time. Do we want to be as flexible as possible _reading_ data but issue an abundance of warnings to users about what required information is missing from their data? Then we probably want to be as brittle as possible when _writing_ because we want the files to conform to the standards we support writing? Maybe it's not competing ideas and this is just turning into a rant? But either way, this early in the morning, I wasn't sure if it was well defined somewhere and at least my questions will live on in perpetuity now.

mkolopanis:
was there anyone looking to do a live test of this branch before we merge just to verify it fixes their issue?

bhazelton:
I don't think that's necessary

bhazelton:
I like your rant and I think I agree with you that that should be our approach. Maybe we should write it down somewhere more formally.

PR #1038, opened 2021-07-21
Update to version 2.2.2 including a small bug fix and dependency tweak

bhazelton:
## Description

- Fix a bug in `UVBeam.efield_to_power` where a the `data_array` remained complex rather than real because the tolerance was too low in the numpy `real_if_close` call.
- Added an explicit pyERFA dependency (rather than the implicit requirement because we depend on astropy).
- Updated the version number

## Motivation and Context

The conda build for version 2.2.1 broke on all the Linux builds because of the `UVBeam.efield_to_power` bug. This fixes that problem and sets up a new release so we can get a Conda package out.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation change (documentation changes only)
- [x] Version change
- [ ] Build or continuous integration change

## Checklist:

- [x] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/.github/CONTRIBUTING.md).
- [x] My code follows the code style of this project.

Bug fix checklist:
- [ ] My fix includes a new test that breaks as a result of the bug (if possible).
- [x] All new and existing tests pass.
- [x] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/CHANGELOG.md).

Version change checklist:
- [x] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/CHANGELOG.md) to put all the unreleased changes under the new version (leaving the unreleased section empty).
- [x] I have noted any dependency changes since the last version and will update the conda package build accordingly.
**pyerfa>=2.0 has been added as an explicit dependency**

PR #1066, opened 2021-10-01
Increase the tolerance on uvbeam real_if_close call

bhazelton:
## Description

Increase the tolerance in the numpy `real_if_close` call in `UVBeam.efield_to_power` because in some tests on the conda package the beam remains complex rather than real because the tolerance was too low.

## Motivation and Context

I tried to fix this problem in #1038 but errors are happening again with the v2.2.2 conda package so I think the tolerance needs to be even higher.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:


plaplant (approved):
Looks good, thanks!

PR #1082, opened 2021-11-09
Fix expected shapes for data-like arrays on metadata only objects

bhazelton:
## Description

- Changed `_set_future_array_shapes` method to always update the form of data-like arrays, regardless of whether the object is a metadata only object.
- Explicitly put `antenna_diameters` into measurement sets as doubles.

## Motivation and Context

The form for the data-like arrays were not being updated for the future shapes if the UVData object was a metadata only object. This is a bug and caused problems on pyuvsim, where the object is first created and set to use future shapes and then the metadata and data are added later. The result was that the check errored because the forms hadn't been properly updated for the data-like arrays.

I also found another small bug when adding the option to write measurement sets to pyuvsim. The antenna diameters in a file I was working with had a dtype of `">f4"` and this caused an error in adding the diameters to the measurement set.  @kartographer suggested forcing them to be doubles and that fixed the error, so I added that fix (and a test) to this PR.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/uvdata/tests/conftest.py
+++ b/pyuvdata/uvdata/tests/conftest.py
@@ -1,0 +1,39 @@
+# -*- mode: python; coding: utf-8 -*-
+# Copyright (c) 2021 Radio Astronomy Software Group
+# Licensed under the 2-clause BSD License
+
+"""Testing fixtures for UVData."""
+import os
+
+import pytest
+
+from pyuvdata.data import DATA_PATH
+from pyuvdata import UVData
+
+
+@pytest.fixture(scope="session")
+def hera_uvh5_main():
+    # read in test file for the resampling in time functions
+    uv_object = UVData()
+    testfile = os.path.join(DATA_PATH, "zen.2458661.23480.HH.uvh5")
+    uv_object.read(testfile)
+
+    yield uv_object
+
+    # cleanup
+    del uv_object
+
+    return
+
+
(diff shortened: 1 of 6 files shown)


pyuvdata/uvdata/tests/conftest.py
```diff
@@ -0,0 +1,39 @@
...
+    uv_object = UVData()
+    testfile = os.path.join(DATA_PATH, "zen.2458661.23480.HH.uvh5")
+    uv_object.read(testfile)
+
+    yield uv_object
+
+    # cleanup
+    del uv_object
+
+    return
```
mkolopanis, line 26: Is this fixture defined somewhere else currently? I can't help but think this feels familiar :thinking:

mkolopanis (approved):
Simple. This looks good thank you. I have a question which will not block this for me so looks good to go to me.

pyuvdata/uvdata/tests/conftest.py
bhazelton: yes, I thought I moved it here so I could use it across multiple files, but maybe I just copied it. I'll check.

bhazelton: Yep, that's what happened. I just pushed up the removal from test_uvdata.py. Thanks for catching that!

mkolopanis (approved):

PR #1086, opened 2021-11-15
Uvfits frame

### LINKED ISSUE #1085: Make UVFITS more frame assumption more (less?) flexible
New MWAX files written by [birli](https://github.com/MWATelescope/Birli.git) do not contain a `FRAME` keyword and the logic in uvfits does not correctly guess a frame.

Moreover these files have exposed a fail-able entry in the lines linked below. Mainly where `COMMENT` exist but none of our matching phrases (only one for MWA) exist in the comments.

This can easily be fixed by looking for the `birli` software field in the comments.

```python
In [14]: with fits.open("1320409688.uvfits", memmap=True) as hdu:
    ...:     extra = utils._get_fits_extra_keywords(hdu[0].header.copy(), keywords_to_skip=["DATE-OBS"])


In [15]: extra
Out[15]: 
{'GCOUNT': 495360,
 'COMMENT': "  FITS (Flexible Image Transport System) format is defined in 'Astronomy\n  and Astrophysics', volume 376, page 359; bibcode: 2001A&A...376..359H\nCreated by birli v0.1.9",
 'OBSRA': 333.6069,
 'OBSDEC': -17.0288,
 'EPOCH': 2000.0,
 'OBJECT': 'high_3C444_2021A_2459526_RADec333.6,-17.0_Ch143',
 'TELESCOP': 'MWA',
 'INSTRUME': 'MWA',
 'SOFTWARE': 'birli',
 'GITLABEL': 'v0.1.9'}

```

https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/b382be644848abe315df035d1cec23f694e65427/pyuvdata/uvdata/uvfits.py#L624-L642

mkolopanis:
Adds an assumption clause to the FRAME reading in UVFITS for [birli](https://github.com/MWATelescope/Birli.git) software package

## Motivation and Context

closes #1085 
## Types of changes

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,11 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
-- Support for reading MWAX/birli fits files
+- Support for reading MWAX/birli uvfits files.
 - Flexible spectral windows to `read_mwa_corr_fits`.
 
 ### Changed
+- Assumes uvfits files are in ITRF frame unless explicitly stated otherwise. Consistent with AIPS 117.
 - Improved readability, functionality, and memory usage in `read_mwa_corr_fits`.
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- Support for reading MWAX/birli fits files
 - Flexible spectral windows to `read_mwa_corr_fits`.
 
 ### Changed
--- a/pyuvdata/uvdata/uvfits.py
+++ b/pyuvdata/uvdata/uvfits.py
@@ -635,6 +635,19 @@ class UVFITS(UVData):
                             "to be a Cotter file, setting to ITRF."
                         )
                         xyz_telescope_frame = "ITRF"
+
+                    elif self.extra_keywords.get("SOFTWARE", None).lower() == "birli":
+                        # MWAX pipeline has a package named birli used to write
+                        # uvfits files.
+                        warnings.warn(
+                            "Required Antenna frame keyword not set, but this appears "
+                            "to be a MWAX file, setting to ITRF."
+                        )
+                        xyz_telescope_frame = "ITRF"
+
+                    else:
+                        xyz_telescope_frame = "????"
+
                 else:
                     warnings.warn(
                         "Required Antenna frame keyword not set, setting to ????"
--- a/pyuvdata/uvdata/uvfits.py
+++ b/pyuvdata/uvdata/uvfits.py
@@ -624,7 +624,24 @@ class UVFITS(UVData):
             if "FRAME" in ant_hdu.header.keys():
                 xyz_telescope_frame = ant_hdu.header["FRAME"]
             else:
-                if "COMMENT" in self.extra_keywords:
+                # want to make a case-insensitive comparison so look for existance
+                # first before comparing. This may need to change later if
+                # there are other explicit softwares we want to support assuming
+                # a frame for.
+                if (
+                    "SOFTWARE" in self.extra_keywords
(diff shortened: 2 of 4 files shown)


mkolopanis:
I can't add an official file from the mwax correlator file just yet since they aren't public domain data, but I have verified I can read them with this branch.

pyuvdata/uvdata/uvfits.py
```diff
@@ -624,7 +624,24 @@ def read_uvfits(
             if "FRAME" in ant_hdu.header.keys():
                 xyz_telescope_frame = ant_hdu.header["FRAME"]
             else:
-                if "COMMENT" in self.extra_keywords:
+                # want to make a case-insensitive comparison so look for existance
```
kartographer, line 627: Can I suggest an alternative idea here? Rather than growing out this if/else decision tree, it seems like perhaps the default should just be ITRF, and if `FRAME` isn't set, then a message/warning issued indicating the use of said default value (otherwise one could imagine it growing larger and larger with time).

mkolopanis: That's definitely a reasonable solution for this too. Just reading the older code, I couldn't tell if we were trying to be flexible or rigid so I went with rigid.

bhazelton: I think that's reasonable. This particular case is one that will hopefully go away -- it's caused by a new piece of software that @mkolopanis has already made an issue on.

mkolopanis: ITRF is the only actually supported one too. We could default the unknown case to "?????" as well. It is accepted as the "ignorant" case according to AIPS 117. But the problem I can see then is if LAT, LON, and ALT are not set the telescope location will never be set.

bhazelton: Yeah, that causes lots of issues. So what we did in the past was default to "????" but then for formats we knew were supposed to be in ITRF we set it to that to avoid the errors, leading to the current situation. I think it's ok to just set it to ITRF with a warning message.

mkolopanis: I've changed to this approach in https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1086/commits/3d93f4d7f51b34cc0ec9b0f135546475f1f5b3a8

mkolopanis:
Codecov's posted report is uh going crazy. I'm pretty sure it should just be a small delta because we took a net loss in the number of lines in the repo.

kartographer:
> Codecov's posted report is uh going crazy. I'm pretty sure it should just be a small delta because we took a net loss in the number of lines in the repo.

From the report, it almost looks as though it skipped over all of the tests related to optional imports (i.e., casacore, miriad, astroquery, etc.), although _why_ it's doing that, I have no idea...

mkolopanis:
It looks like it reported back immediately after the first thing finished (probably a minimum dependency run) but then never updated after everything was done.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- Support for reading MWAX/birli fits files
```
bhazelton, line 8: This is uvfits not just fits right? Pyxie's reader is for the raw fits files.

bhazelton (commented):
Looks good. Just one clarification in the changelog.

CHANGELOG.md
mkolopanis: it is. I've updated this in https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1086/commits/2905273e3c724fdb40789345bb91c537dd78b496 and added a `changed` entry to reflect what uvfits reader does now.

bhazelton (approved):
Looks great, thanks!

bhazelton (approved):

PR #1087, opened 2021-11-15
Write scan number for MS files

e-koch:
Adds setting scan numbers when writing MS files.

## Description

 For data with multi-phase centers, the scan numbers in the MS should bet set (somehow) to match the usual format in CASA. The SMA will need this for converting MIR to MS (@kartographer).

I've added grouping by phase centers, incrementally increasing whenever the data switches between phase centers (e.g., between the target and gain calibrator) based on `UVData.phase_center_id_array`. I _think_ this is a good grouping definition when scans are not already defined, for example in MIR data.

To-dos and questions:
* [x] Is this grouping on fields from `UVData.phase_center_id_array` reasonable?
* [x] Read in scan numbers from an MS. Add a new attribute `UVData.scan_numbers_array` that is defined only when reading in a MS.
* [x] Add a read and write test for scan numbers. Checks for the scan numbers added to round trip tests with reading/writing a MS; added read/write MS test for multi-phase centers using the CARMA MIRIAD 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,13 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
-- Added a `filename` attribute to UVCal, UVBeam and UVFlag objects.
-- Support for reading MWAX/birli uvfits files.
-- Flexible spectral windows to `read_mwa_corr_fits`.
 - Reading writing of scan numbers for MS files as `UVData.scan_number_array`.
 - Grouping of contiguous integrations for a phase center into "scan numbers" in `UVData._set_scan_numbers`.
 This grouping defines `UVData.scan_number_array` when not originally present in the data (e.g. reading in
 a non-MS file) and is used when writing to an MS file.
+- Added a `filename` attribute to UVCal, UVBeam and UVFlag objects.
+- Support for reading MWAX/birli uvfits files.
+- Flexible spectral windows to `read_mwa_corr_fits`.
 
 ### Changed
 - Assumes uvfits files are in ITRF frame unless explicitly stated otherwise. Consistent with AIPS 117.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 - Added a `filename` attribute to UVCal, UVBeam and UVFlag objects.
 - Support for reading MWAX/birli uvfits files.
 - Flexible spectral windows to `read_mwa_corr_fits`.
+- Reading writing of scan numbers for MS files as `UVData.scan_number_array`.
+- Grouping of contiguous integrations for a phase center into "scan numbers" in `UVData._set_scan_numbers`.
+This grouping defines `UVData.scan_number_array` when not originally present in the data (e.g. reading in
+a non-MS file) and is used when writing to an MS file.
 
 ### Changed
 - Assumes uvfits files are in ITRF frame unless explicitly stated otherwise. Consistent with AIPS 117.
--- a/pyuvdata/uvdata/ms.py
+++ b/pyuvdata/uvdata/ms.py
@@ -1170,7 +1170,7 @@ class MS(UVData):
         if self.scan_number_array is None:
             self._set_scan_numbers()
 
-        if self.multi_phase_center:
+        if self.Nspws > 1:
             scan_array_tiled = np.repeat(self.scan_number_array, self.Nspws)
             ms.putcol("SCAN_NUMBER", scan_array_tiled)
         else:
--- a/pyuvdata/uvdata/ms.py
+++ b/pyuvdata/uvdata/ms.py
@@ -11,7 +11,6 @@ import numpy as np
 import os
 import warnings
 from astropy.time import Time
-from scipy import ndimage as nd
 
 from .uvdata import UVData
 from .. import utils as uvutils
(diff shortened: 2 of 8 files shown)


e-koch:
I've added `UVData.scan_number_array` to track the scan numbers when reading in a MS file. But because it is only defined for that case, there were several test failures comparing to `UVData` objects created from different file types.

For those tests, I added `scan_number_array` to the allowed exceptions list, or defined a `scan_number_array` for the other object when the test did something similar for the `filename`. I'm unsure which is the better approach.

e-koch:
@bhazelton @kartographer -- This is ready for a review!

The doctest failure is unclear to me. I don't think I touched the code affecting those outputs in the docs.

e-koch:
@kartographer -- Rebased.

pyuvdata/uvdata/ms.py
```diff
@@ -1165,6 +1166,40 @@ def write_ms(
...
             ms.putcol("FIELD_ID", np.repeat(field_id_array, self.Nspws))
 
+            # If scan numbers are not already defined from reading an MS,
+            # group integrations (rows) into scan numbers.
+            if self.scan_number_array is None:
+
+                slice_list = []
+                for idx in range(self.Nphase):
+                    sou_id = self.phase_center_catalog[sou_list[idx]]["cat_id"]
+                    slice_list.extend(
```
kartographer, line 1176: Took me a second to parse what this piece of code was doing, but this is spiffy! I actually wonder if this should actually be a function with the `utils` module (or maybe even the `UVData` object? Would be curious what @bhazelton might say here), since I could see this code being used outside of this specific module (e.g., I might similarly want to use in in `mir`).

pyuvdata/uvdata/tests/test_miriad.py
```diff
@@ -262,7 +263,7 @@ def test_read_carma_miriad_write_ms(tmp_path):
     uv_in.history = uv_out.history
 
     # Final equality check
-    assert uv_in == uv_out
+    assert uv_in.__eq__(uv_out, allowed_failures=allowed_failures_with_ms)
```
kartographer, line 266: If you were to add the scan number to `utils`, I think you may be able to drop the added `allowed-failures` in the equality call. Not a strict requirement, but perhaps is a nicety.

kartographer (changes requested):
Overall I think this looks good @e-koch -- thanks for putting this together! Just have a couple of small requested changes (and there's one comment here that I'd ask @bhazelton to chime in on) that should hopefully be quick and easy to address.

pyuvdata/uvdata/ms.py
bhazelton: If it's useful outside of this context we should definitely pull it out. Whether as a utility function or a private method on UVData depends on whether it's likely to be useful for other objects and how many other object attributes contribute to it. I think you're best placed to decide between those @kartographer.

kartographer: In that case, I think I might vote to make it a private method under `UVData` -- even if we add a scan number attribute to `UVFlag`, `UVCal`, etc., I expect those values to be imported from a `UVData` object rather than being derived in the way they are here.

pyuvdata/uvdata/tests/test_miriad.py
e-koch: Done! I've reverted most of these extra `allowed_failures` to be only for `filename` as they originally were.

pyuvdata/uvdata/ms.py
e-koch: Thanks for the input! I've moved this code to `UVData._set_scan_numbers`.

pyuvdata/uvdata/ms.py
```diff
@@ -1165,6 +1166,40 @@ def write_ms(
...
+                # Incrementally increase the scan number with each group in
+                # slice_list_ord
+                scan_array = np.zeros_like(self.phase_center_id_array)
+                for ii, slice_scan in enumerate(slice_list_ord):
+                    scan_array[slice_scan] = ii + 1
+
+                scan_array_tiled = np.repeat(scan_array, self.Nspws)
+
+            else:
+                scan_array_tiled = np.repeat(self.scan_number_array, self.Nspws)
```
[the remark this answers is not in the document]
e-koch: I'll check this after the next set of tests run, but I think it will be covered in an existing test.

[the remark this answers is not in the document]
kartographer: Looks like it's covered now!

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 - Added a `filename` attribute to UVCal, UVBeam and UVFlag objects.
 - Support for reading MWAX/birli uvfits files.
 - Flexible spectral windows to `read_mwa_corr_fits`.
+- Reading writing of scan numbers for MS files as `UVData.scan_number_array`.
+- Grouping of contiguous integrations for a phase center into "scan numbers" in `UVData._set_scan_numbers`.
+This grouping defines `UVData.scan_number_array` when not originally present in the data (e.g. reading in
+a non-MS file) and is used when writing to an MS file.
```
bhazelton, line 14: These should go to the top of this section. The changelog is in reverse chronological order.

pyuvdata/uvdata/ms.py
```diff
@@ -1165,6 +1165,17 @@ def write_ms(
 
             ms.putcol("FIELD_ID", np.repeat(field_id_array, self.Nspws))
 
+        # If scan numbers are not already defined from reading an MS,
+        # group integrations (rows) into scan numbers.
+        if self.scan_number_array is None:
+            self._set_scan_numbers()
+
+        if self.multi_phase_center:
```
kartographer, line 1173: I'm a little confused as to what this block of code is doing -- why the dependence on `self.multi_phase_center`? If anything, I would have expected it to be if `self.flex_spw` or `self.Nspws != 1` (or even forgoing the if/then statement all together, since the code below should work if you only have 1 spectral window).

pyuvdata/uvdata/uvdata.py
```diff
@@ -707,6 +720,55 @@ def _set_flex_spw(self):
         # Now make sure that chan_width is set to be an array
         self._channel_width.form = ("Nfreqs",)
 
+    def _set_scan_numbers(self, override=False):
```
kartographer, line 723: 👍

kartographer (commented):
Thanks @e-koch -- I just had one question, after which (and fixing the CHANGELOG) I think you should be good to go.

CHANGELOG.md
e-koch: Fixed! Thank you!

pyuvdata/uvdata/ms.py
e-koch: Thanks for catching this! Agreed, this was wrong. I think it was leftover from when this code was in the above `if self.multi_phase_center` block above.

I've swapped to `self.Nspws > 1`. I could drop it if preferred, though I figured skipping a redundant `np.repeat` call is slightly more efficient. What do you think?

kartographer: I'm reasonably content either way, just as long as `self.multi_phase_center` isn't in there anymore 😉  Let's keep it as `self.Nspws > 1`.

kartographer (approved):

kartographer:
Looks good, thanks @e-koch!

PR #1088, opened 2021-11-17
Handle `None` values in `extra_keywords` in UVH5 files

### LINKED ISSUE #1081: Handle `None` values in `extra_keywords` better when writing UVH5 files
If the user has any keys in the `extra_keywords` dict that have a key of `None`, writing UVH5 files raises an error that doesn't make it obvious what is going wrong. We should either handle values of `None` in a better way, or give the user a better error message and a hint about how to fix it.

plaplant:
## Description

This PR adds the ability to round-trip `None` values in the `extra_keywords` dictionary for UVH5 files.

## Motivation and Context

In detail, this is done by creating an [empty dataset](https://docs.h5py.org/en/stable/high/dataset.html#creating-and-reading-empty-or-null-datasets-and-attributes) in the HDF5 file with a name corresponding to the key of the `extra_keywords`. To me, this approach is more precise than, e.g., saving a string like `"None"` because it's clear that this is a null-like data object, but I am open to hearing what others think.

Fixes #1081.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ All notable changes to this project will be documented in this file.
 - Improved readability, functionality, and memory usage in `read_mwa_corr_fits`.
 
 ### Fixed
+- A bug that prevented `extra_keywords` keys with a value of `None` from being
+  saved to UVH5 files.
 - A bug that resulted in the wrong expected shapes for data-like arrays when metadata
 only UVData objects were set to use future array shapes.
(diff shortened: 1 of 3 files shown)


mkolopanis (commented):
A good use of the empty dataset from hdf5. I'd say probably worth a changelog blip and then good to go?

mkolopanis (approved):

PR #1090, opened 2021-11-18
Mwax corr fits reader

### LINKED ISSUE #1089: Update mwa_corr_fits reader to handle MWAX files
Currently `mwa_corr_fits.read` only handles MWA legacy files. The function should be updated to also read in files from the new MWAX correlator.

PyxieLouStar:
Add capability to read files from the new MWAX correlator.

## Motivation and Context

closes #1089 

The MWA has a brand new correlator which outputs raw data as fits files with some different formatting and metadata than the legacy correlator. This PR adds the capability to read these raw data files.
## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:


pyuvdata/uvdata/mwa_corr_fits.py
```diff
@@ -1653,9 +1743,16 @@ def read_mwa_corr_fits(
...
+            if not mwax:
+                self.data_array *= self.extra_keywords["SCALEFAC"]
 
             # cable delay corrections
+            if correct_cable_len is None:
+                correct_cable_len = True
+                warnings.warn(
+                    "cable length correction is defaulted to True. To read in files \
+                    without applying the correction set correct_cable_len=False."
+                )
```
bhazelton, line 1755: I think this is good, but I think we want to identify that it's a change from before and a version where we'll remove this warning (v2.4?).

pyuvdata/uvdata/tests/test_mwa_corr_fits.py
```diff
@@ -102,8 +104,67 @@ def test_read_mwa_write_uvfits(tmp_path):
...
+        )
+
+    testfile = str(tmp_path / "outtest_MWAXcorr.uvfits")
+    mwax_uv.write_uvfits(testfile, spoof_nonessential=True)
+    uvfits_uv.read_uvfits(testfile)
+
+    assert mwax_uv == uvfits_uv
+
+
 @pytest.mark.filterwarnings("ignore:telescope_location is not set. ")
```
bhazelton, line 150: why is the telescope location not being set in most of these tests? Is that always the case for these files or something weird about the test files?

bhazelton (commented):
looks good, I just have a few comments

pyuvdata/uvdata/mwa_corr_fits.py
```diff
@@ -1375,13 +1444,18 @@ def read_mwa_corr_fits(
             self.instrument = meta_hdr["TELESCOP"]
             self.telescope_name = meta_hdr.pop("TELESCOP")
             self.object_name = meta_hdr.pop("FILENAME")
+            self.Nants_telescope = int(meta_hdr.pop("INSTRUME")[0:3])
```
bhazelton, line 1447: I could imagine this failing at some point because MWA supports the option to plug in experimental antennas into receivers (labelling them with numbers that end in 9 or 0) which would extend the number of antennas in the positions beyond 128 or 256 (even though only 128 or 256 can be plugged in for any particular observation). I'm ok with leaving it like this but wanted to flag that we might possibly see a problem at some point. Another approach is to set it to the length of the numbers or positions.

pyuvdata/uvdata/mwa_corr_fits.py
```diff
@@ -492,19 +506,26 @@ def _read_fits_file(
                     hdu.data[:, 0::2] + 1j * hdu.data[:, 1::2]
                 )
                 # fill nsample and flag arrays
+                # think about using the mwax weights array in the future
```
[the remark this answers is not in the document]
PyxieLouStar: I believe it is the former. This is from the [docs](https://wiki.mwatelescope.org/display/MP/MWAX+Visibility+File+Format#MWAXVisibilityFileFormat-FITSExtensionWeightsImgHDU):
```
Weights are provided in the gpubox files after each visibility HDU containing one timestep/integration. The weights and how they are determined is discussed below:

    Each visibility has a multiplicative weight applied, based on a data occupancy metric that takes account of any input data blocks that are missing due to lost UDP packets or RFI excision (a potential future enhancement).  The centre (DC) ultrafine channel is excluded when averaging and the centre output channel values are re-scaled accordingly.  Note that only 200 Hz of bandwidth is lost in this process, rather than a complete output channel. 
    As part of the M&C system, the application of weights can be turned on or off per observation based on the science case/needs. With weights not applied, the data will be averaged in the correlator without taking into account the weights. Either way the weights are supplied in each alternate ImgHDU for your information.
```

pyuvdata/uvdata/mwa_corr_fits.py
PyxieLouStar: To make sure I'm understanding correctly, are you saying that `Nants_telescope` should reflect the number of antenna positions (which might not all be plugged in) rather than the number of antennas that are plugged in for that observation?

pyuvdata/uvdata/tests/test_mwa_corr_fits.py
PyxieLouStar: The telescope location is set by `set_telescope_params`, which generates this warning before setting the telescope location

pyuvdata/uvdata/mwa_corr_fits.py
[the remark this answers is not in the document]
bhazelton: great. Then I think we should just pipe this into the n_sample array. I'm curious about the details of the center channel, I don't quite get that comment. I guess if the application of weights is turned off we should set the n_samples to 1 because everything is included?

pyuvdata/uvdata/tests/test_mwa_corr_fits.py
bhazelton: hmm. I'm not sure we want this kind of warning under the normal course of reading files. I'll take a look at that method and see if there's a way to modify it.

pyuvdata/uvdata/mwa_corr_fits.py
bhazelton: Yes, I think so. If the metafits file lists them I think we want to keep them and then Nants_telescope needs to be the length of the antenna_names/antenna_numbers/antenna_positions arrays. So I guess it's a question of what is included in the metafits. If that is always limited to the number that can be plugged in then it's not an issue.

pyuvdata/uvdata/mwa_corr_fits.py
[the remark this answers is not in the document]
PyxieLouStar: I think that makes sense.

pyuvdata/uvdata/tests/test_mwa_corr_fits.py
bhazelton: Looking at your code, it looks like the metafits specifies the ENU positions, does it also specify the center of the array? It seems like it should.

PyxieLouStar: If this isn't the correct usage for the `set_telescope_params` function, I can set the telescope location another way.

PyxieLouStar: Oops, your comment came in while I was writing. Yes, I can get the positions from the file. I think that somewhere in the early process I got the idea that using `set_telescope_params` was a preferred method.

bhazelton: It _should_ be the same. but I worry that if there's a slight difference, we want to make sure that we use the value that the antenna positions are reported with respect to. So I'd favor using the one in the file.

pyuvdata/uvdata/mwa_corr_fits.py
PyxieLouStar: The metafits file only has information for the antennas that are included in that observation

pyuvdata/uvdata/tests/test_mwa_corr_fits.py
PyxieLouStar: Ok, sounds good.

pyuvdata/uvdata/mwa_corr_fits.py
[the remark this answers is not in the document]
PyxieLouStar: I wonder actually if we would want to keep the weights even if the application is turned off, in case we want to applyy them?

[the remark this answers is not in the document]
bhazelton: I guess I'm not sure, it depends on what is actually being done. I think we should parse this together.

pyuvdata/uvdata/mwa_corr_fits.py
bhazelton: ok, then I guess this won't be an issue and we can leave it alone. If it becomes an issue in the future we can fix it then.

pyuvdata/uvdata/tests/test_mwa_corr_fits.py
PyxieLouStar: Oooh actually those values aren't in the metafits file. When I initially wrote the reader I hard-coded the location values in (using the values from cotter). Then I switched to using the `set_telescope_params` function.

bhazelton: Ok, then I think we should modify that method to accept a boolean that turns off that warning. I don't think we want to warn people when there's nothing to worry about.

pyuvdata/uvdata/mwa_corr_fits.py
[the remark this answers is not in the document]
PyxieLouStar: Since the proper weights values aren't being provided from MWAX yet, we'll wait to implement them in the reader.

PyxieLouStar:
I'm getting a warning from `antnums_to_baseline` because the files have antenna numbers which are greater than 256. 
`UserWarning: antnums_to_baseline: found > 256 antennas, using 2048 baseline indexing. Beware compatibility with CASA etc`

PyxieLouStar:
> I'm getting a warning from `antnums_to_baseline` because the files have antenna numbers which are greater than 256. `UserWarning: antnums_to_baseline: found > 256 antennas, using 2048 baseline indexing. Beware compatibility with CASA etc`

This is happening because the mwax files have antenna numbers greater than 256, which generates a warning when the uvfits file is read in the turnaround test. I edited the warning to be more clear.

bhazelton (approved):
Looks good. Thanks @PyxieLouStar!

PR #1069, opened 2021-10-05
Putting in fix for astroquery/JPL-Horizons related error

kartographer:
## Description

A small change to the call to the JPL-Horizons online calculator (accessed through `astroquery`), setting `extra_precision=False`. The net result of which is that the maximum accuracy of the ephems will be precision-limited to 0.05 arcsec.

## Motivation and Context

This is a temporary fix to an issue that has risen when using MacOS-10.15 to make calls to JPL-Horizons using `astroquery`. Once the underlying issue is resolved, then presumably this fix can be revered.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/.github/CONTRIBUTING.md).
- [x] My code follows the code style of this project.

Bug fix checklist:
- [x] My fix includes a new test that breaks as a result of the bug (if possible). (N.B., bug was covered w/ existing test, no tests added)
- [x] All new and existing tests pass.
- [x] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/CHANGELOG.md).

PR #1091, opened 2021-11-18
fix doctests by requiring recent astroquery, updating tutorial output

bhazelton:
## Description

Update our test yaml to require astroquery>=0.4.4 and update the tutorial output to match what we get with that version of astroquery.

## Motivation and Context

A new astroquery version (0.4.4) just came out which fixes the problem addressed in #1069. We want to leave the try/except handling introduced in that PR because it is more robust and allows us to support more versions of astroquery. But in order for the doctests to pass we need some version certainty.

## Types of changes

## Checklist:

### DIFF
--- a/ci/pyuvdata_tests.yml
+++ b/ci/pyuvdata_tests.yml
@@ -4,7 +4,7 @@ channels:
   - defaults
 dependencies:
   - astropy>=4.2.1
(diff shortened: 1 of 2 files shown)


mkolopanis (approved):
Works for me :rocket:

PR #1101, opened 2021-12-09
MS-related bug fixes

### LINKED ISSUE #1063: Consider changing the default assignment of MS antenna names
The current default uses the "STATION" field if if has variable values and the "ANTENNA" field otherwise. We might always want to default to the "ANTENNA" field but that needs more evaluation with existing data files.

### LINKED ISSUE #1083: Round-trip `vis_units` and `x_orientation` through write_ms/read_ms
Currently `vis_units` is set to `'Jy'` by default, there seems to be some interaction with the data_column parameter on `read_ms` that I don't understand.

`x_orientation` is just lost.

kartographer:
Closes a couple of outstanding issues related to CASA MS read/write

## Description

- Fixed a bug by which `UVData.x_orientation` and `UVData.vis_units` were not being propogated by the `write_ms` method.
- Changed the default scheme by which antenna names are imported in `UVData.read_ms`.

## Motivation and Context

This PR addresses two open issues, the first of which being some missing meta-information that was not being written when writing out data sets in measurement set format. As part of that, a warning message has been added when writing out data sets where the units are the visibilities are something other than Jy, since 1) CASA procedures will sometimes by default assume that the data are given in Jy, and 2) CASA has no knowledge of what `uncalib` and `K Jy` units are.

The second issue addressed is a change in how antenna names are imported from measurement sets. This had been intended to be included in #995, but it caused an error that I didn't understand at the time. Turns out it was just related to usage `importuvfits`, which was an easy thing to capture for looking at the file history.

Closes #1083.
Closes #1063.

## Types of changes

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/uvdata/ms.py
+++ b/pyuvdata/uvdata/ms.py
@@ -1958,11 +1958,11 @@ class MS(UVData):
         # importuvfits measurement sets store antenna names in the STATION column.
         # cotter measurement sets store antenna names in the NAME column, which is
         # inline with the MS definition doc. In that case all the station names are
-        # the same.
-        # TODO: Potentially flip which is the default -- use ant_names if filled and
-        # unique. This does seem to fail though on "day2_TDEM0003_10s_norx_1src_1spw".
-        # See the test "test_read_ms_read_uvfits".
-        if len(station_names) != len(np.unique(station_names)):
+        # the same. Default to using what the MS definition doc specifies, unless
+        # we read importuvfits in the history, of the antenna column is not filled
+        if ("importuvfits" not in self.history) and (
+            len(ant_names) == len(np.unique(ant_names)) and ("" not in ant_names)
+        ):
             self.antenna_names = ant_names
         else:
             self.antenna_names = station_names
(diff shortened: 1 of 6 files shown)


pyuvdata/uvdata/ms.py
```diff
@@ -1943,11 +1964,11 @@ def read_ms(
         # importuvfits measurement sets store antenna names in the STATION column.
         # cotter measurement sets store antenna names in the NAME column, which is
         # inline with the MS definition doc. In that case all the station names are
-        # the same.
-        # TODO: Potentially flip which is the default -- use ant_names if filled and
-        # unique. This does seem to fail though on "day2_TDEM0003_10s_norx_1src_1spw".
-        # See the test "test_read_ms_read_uvfits".
-        if len(station_names) != len(np.unique(station_names)):
+        # the same. Default to using what the MS definition doc specifies, unless
+        # we read importuvfits in the history, of the antenna column is not filled
```
bhazelton, line 1968: I think there's a typo in this sentence. Maybe "history, of the antenna column" -> "history, or if the antenna column"?

bhazelton (commented):
Looks good, just one typo in a comment.

pyuvdata/uvdata/ms.py
kartographer: Ah, good catch -- fixed in 55fda3c.

bhazelton (approved):
Looks good, thanks @kartographer!

PR #1106, opened 2021-12-10
Add support for baseline coordinate suffixes in uvfits files

### LINKED ISSUE #1104: Errors in reading GMRT UVFITS data
I'm getting the following error in reading UVFITS data from GMRT (using python 3.8.5):
`UnboundLocalError: local variable 'xyz_telescope_frame' referenced before assignment` 

In `uvfits.py`, `xyz_telescope_frame` needs to be defined and initialized to a default value outside the if/else block before it is used (before line 624).

I corrected this for my local file. Then I get the following error:

```~/.local/lib/python3.8/site-packages/pyuvdata/uvdata/uvfits.py in _get_parameter_data(self, vis_hdu, run_check_acceptability, background_lsts)
    142                 np.stack(
    143                     (
--> 144                         vis_hdu.data.par("UU"),
    145                         vis_hdu.data.par("VV"),
    146                         vis_hdu.data.par("WW"),

~/.local/lib/python3.8/site-packages/astropy/io/fits/hdu/groups.py in par(self, parname)
    237             result = self.field(parname)
    238         else:
--> 239             indx = self._unique[parname.upper()]
    240             if len(indx) == 1:
    241                 result = self.field(indx[0])

KeyError: 'UU'
```

The file has keywords 'UU---SIN', 'VV---SIN', 'WW---SIN'
According to the AIPS convention (http://parac.eu/AIPSMEM117.pdf), the keywords are usually 'UU, VV, WW' but may be followed by modifiers '---SIN' (default) or '---NCP'.

Could you add these to the uvfits reader?

bhazelton:
## Description

Add support for the `---SIN` and `---NCP` suffixes that can optionally appear on the uvw coordinate parameter names in uvfits files. Note that the absence of a suffix indicates the same baseline coordinate system as the `---SIN` suffix. The `---NCP` suffix indicates a different baseline coordinate system that does not agree with our conventions. For files of that type, the uvws are rotated to match our convention.

## Motivation and Context

fixes #1104

## Types of changes

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/uvdata/uvfits.py
+++ b/pyuvdata/uvdata/uvfits.py
@@ -154,12 +154,6 @@ class UVFITS(UVData):
             and "WW---NCP" in vis_hdu.data.parnames
         ):
             uvw_names = ["UU---NCP", "VV---NCP", "WW---NCP"]
-            warnings.warn(
-                "The baseline coordinates (uvws) in this file are specified in the "
-                "---NCP coordinate system, which is does not agree with our baseline "
-                "coordinate conventions. The values are not being modified to align "
-                "with our convention."
-            )
         else:
             raise ValueError(
                 "There is no consistent set of baseline coordinates in this file. "
@@ -817,6 +811,13 @@ class UVFITS(UVData):
 
             # Calculate the apparent coordinate values
             self._set_app_coords_helper()
+
+            # fix up the uvws if in the NCP baseline coordinate frame.
+            # Must be done here because it requires the phase_center_app_dec
+            if "UU---NCP" in vis_hdu.data.parnames:
+                self.uvw_array = uvutils._rotate_one_axis(
+                    self.uvw_array[:, :, None], self.phase_center_app_dec - np.pi / 2, 0
+                )[:, :, 0]
 
             if not read_data:
                 # don't read in the data. This means the object is a metadata
--- a/pyuvdata/uvdata/uvfits.py
+++ b/pyuvdata/uvdata/uvfits.py
@@ -134,6 +134,40 @@ class UVFITS(UVData):
         # initialize internal variables based on the antenna lists
         self.Nants_data = int(np.union1d(self.ant_1_array, self.ant_2_array).size)
 
+        # check for suffixes in the baseline coordinate names indicating the
+        # baseline coordinate system
+        if (
+            "UU" in vis_hdu.data.parnames
+            and "VV" in vis_hdu.data.parnames
+            and "WW" in vis_hdu.data.parnames
+        ):
+            uvw_names = ["UU", "VV", "WW"]
+        elif (
+            "UU---SIN" in vis_hdu.data.parnames
+            and "VV---SIN" in vis_hdu.data.parnames
+            and "WW---SIN" in vis_hdu.data.parnames
+        ):
+            uvw_names = ["UU---SIN", "VV---SIN", "WW---SIN"]
+        elif (
+            "UU---NCP" in vis_hdu.data.parnames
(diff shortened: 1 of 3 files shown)


pyuvdata/uvdata/uvfits.py
```diff
@@ -134,16 +134,50 @@ def _get_parameter_data(
...
+            and "VV---NCP" in vis_hdu.data.parnames
+            and "WW---NCP" in vis_hdu.data.parnames
+        ):
+            uvw_names = ["UU---NCP", "VV---NCP", "WW---NCP"]
+            warnings.warn(
+                "The baseline coordinates (uvws) in this file are specified in the "
+                "---NCP coordinate system, which is does not agree with our baseline "
+                "coordinate conventions. The values are not being modified to align "
+                "with our convention."
+            )
```
mkolopanis, line 162: I've been reading up on this a little bit, but am still confused. NCP _should_ just be an euler rotation (or 2? 3?) away from SIN projection right? Can we rotate the NCPs to the frame we expect using other metadata? @kartographer do you have any ideas? Or do I have the wrong idea about this frame in my head.

mkolopanis (commented):
I think this is looking pretty good. I do wonder why we don't try to rotate the NCP back. Is there something I'm missing about it? more of the question is in my other comment.

pyuvdata/uvdata/uvfits.py
kartographer: @mkolopanis -- I believe you are correct, and we should be able to rotate the uvws to the right position with the source information alone (namely the apparent dec).

bhazelton: Ok. Do you know what the rotation should be? I don't quite have it in my head.

bhazelton: (To answer your question about why not rotate it @mkolopanis, I was being lazy on the assumption that it will rarely come up. But I think you're right, we should probably fix them up.)

kartographer: @bhazelton -- I think the line of code you need is just the following:
```
new_uvw = uvutils._rotate_one_axis(old_uvw[:, :, None], phase_center_app_dec - np.pi/2, 0)[:, :, 0]
```
This will "drop" the w-axis by rotating around the u-axis, which should give you the uvw's in the more typical form that we'd expect.

bhazelton: Ok, I've added that and I have a test that just rotates back when checking for equality. Is there a better way to test that this rotation does exactly what we want?

I'm also not sure if what I'm doing properly handles cases where there are multiple phase centers. The `phase_center_app_dec` attribute is length `Nblts` so maybe it does?

kartographer: With respect to multiple phase centers, yes, that's basically handled by the fact that `phase_center_app_dec` is calculated on a per-baseline-time basis, so no worries there.

As for tests, the _only_ think I could think of would be maybe to construct a east-west interferometer/array and verify that for some given position in the sky, the NCP-projected coordinates actually have zero in the NCP-w term? I don't know how practical that is, but it's the only additional thing that comes to mind.

bhazelton: As discussed on the telecon, I've added a warning back so the user will be aware that a rotation has happened.

mkolopanis:
why are the circle ci jobs not reporting now? I see they were actually execute on circleci's website?

bhazelton:
I don't know! I tried closing and re-opening the PR to get them to re-trigger but it didn't work.

mkolopanis (approved):

Issue #1102, opened 2021-12-09
Frequency concatenation yields wrong data


wfarah:
Using `pyuvdata==2.2.4`

I’m trying to concatenate `uvh5` files in frequency, and I’m simply using `uv = uv1 + uv2 + uv3 + ...`  then writing out `.ms` file and parsing it to `casa`. This worked perfectly well all the time I used it, except for now where I started seeing some weirdness. The phase of the autocorrelation for some of the subbands appeared to be non-zero, which is not expected.
What I’m doing is the following:
```
uv1.read("subband1.uvh5")
uv2.read("subband2.uvh5")
...
uv = uv1 + uv2 + ...
uv.write_ms()
uv.write_hdf5()
```

I then read the `ant_1_names` and `ant_2_names` from all the `uv` files, and I could see that the first integrations in the concatenated `uvfiles` are not the autocorrelations, unlike the subbands:
```
subband1.uvh5
[ 3  5  7 ... 33 33 35] #ant_1_names
[ 3  5  7 ... 35 38 38] #ant_2_names

subband2.uvh5
[ 3  5  7 ... 33 33 35]
[ 3  5  7 ... 35 38 38]

...

concatenated.uvh5
[ 3  3  3 ... 33 33 35] #something is wrong here
[ 3  5  7 ... 35 38 38]
```

These are some cross/auto correlations respectively, viewed in casa. I also extracted the visibilities straight from the uvh5 files, and I see the same thing
<img width="1413" alt="Screen Shot 2021-12-09 at 1 49 35 PM" src="https://user-images.githubusercontent.com/10710289/145481221-6c194b2c-3f18-4a89-9317-59fd8b191ae2.png">
<img width="1424" alt="Screen Shot 2021-12-09 at 1 49 47 PM" src="https://user-images.githubusercontent.com/10710289/145481250-91be83c9-a879-424b-88fb-66b6ae01893c.png">

wfarah:
Data is available on google drive:

https://drive.google.com/drive/folders/1BzNo8fa2yeS8kbO7MoNSPY8pY1FS_j8U

david-macmahon:
Does the baseline ordering vary between input datasets or are the `time_array`s from each input dataset not identical?  AFAIK, these input datasets should have all the same metadata *except* for the `freq_array`s (which should even be contiguous in frequency).  But it's not my data so I can't say for sure whether that's all true.

kartographer:
[...]

PR #1110, opened 2021-12-12
Adding a check to verify autos are real-only

### LINKED ISSUE #1024: Add check that autos are real
It might be nice to include a check that the autos (only for same pols) are identically real and have no imaginary component.

kartographer:
_A PR a day keeps the doctor away..._
## Description

- Added a new method to `UVData` called `_fix_autos`, which will force the auto-correlation data in `data_array` to be real-only.
- Added a new feature to `UVData.check`, which will see whether or not auto-correlation data have non-zero imaginary components in them. If they detected, depending on arguments supplied, `check` can either raise an error or attempt to fix the auto-correlations (via a call to `_fix_autos`).
- Did some minor test fixture cleanup (mostly with the SMA MIR test data set).

## Motivation and Context

As originally motivated in #1024, and demonstrated in #1102, checking that the autos are real-only is a good way to help verify the health of any given `UVData` object, to make sure that something weird hasn't happened in the handling of the data set. However, since this check is potentially inspecting lots of values in `data_array`, it can be somewhat computationally expensive to run, which seems somewhat antithetical to how `UVData.check` _should_ work -- quoting @bhazelton here from a comment in #1024:

> It'll almost always pass so we want to make sure it's lightweight, but catching this early on a dataset is very useful.
 
To that end, I've set up this new functionality with the following defaults:

- On **read**, if non-zero imaginary components are found in the autocorrelation data, `check` will raise a warning, and will subsequently fix the issue by making those values real-only (using `np.abs` on the autocorrelation data).
- On **write**, if non-zero imaginary components are found in the autocorrelation data, `check` will raise an error.
- On all other operations, no check of the autos is done.

I think the above is a reasonable balance between ease-of-use, speed, and ensuring the integrity of `UVData` objects. FWIW, The addition of `_fix_autos` was motivated in part by the fact that some of the test files so have non-zero imaginary components in their auto-correlation data -- something that I believe is a hangover from the old `phase` method (something I noticed while working on #979).

Closes #1024
## Types of changes

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- Added new functionality to `UVData.check` to verify that auto-correlations are real-only,
+along with an option to force them to be real-only if non-zero imaginary components are detected.
+
 ## [2.2.6] - 2022-01-12
 
 ### Added
--- a/pyuvdata/tests/test_utils.py
+++ b/pyuvdata/tests/test_utils.py
@@ -2671,7 +2671,7 @@ def test_mean_weights_and_weights_square():
     assert np.allclose(wso, np.sum(1.0 / (np.arange(data.shape[1]) + 1) ** 2))
 
     # Zero weights
-    w = np.ones_like(w)
+    w = np.ones_like(data)
     w[0, :] = 0
     w[:, 0] = 0
     out, wo = uvutils.mean_collapse(data, weights=w, axis=0, return_weights=True)
@@ -2810,6 +2810,7 @@ def test_and_collapse_errors():
     pytest.raises(ValueError, uvutils.and_collapse, data)
 
 
+@pytest.mark.filterwarnings("ignore:Fixing auto-correlations to be be real-only,")
 @pytest.mark.filterwarnings("ignore:The uvw_array does not match the expected values")
 def test_uvcalibrate_apply_gains_oldfiles():
     # read data
@@ -2863,6 +2864,7 @@ def test_uvcalibrate_apply_gains_oldfiles():
 
 
 @pytest.mark.filterwarnings("ignore:When converting a delay-style cal to future array")
+@pytest.mark.filterwarnings("ignore:Fixing auto-correlations to be be real-only,")
 @pytest.mark.filterwarnings("ignore:The uvw_array does not match the expected values")
 @pytest.mark.parametrize("uvd_future_shapes", [True, False])
 @pytest.mark.parametrize("uvc_future_shapes", [True, False])
@@ -2918,6 +2920,7 @@ def test_uvcalibrate_delay_oldfiles(uvd_future_shapes, uvc_future_shapes):
 
 @pytest.mark.parametrize("uvc_future_shapes", [True, False])
 @pytest.mark.parametrize("uvd_future_shapes", [True, False])
+@pytest.mark.filterwarnings("ignore:Fixing auto-correlations to be be real-only,")
 @pytest.mark.parametrize("flip_gain_conj", [False, True])
 @pytest.mark.parametrize("gain_convention", ["divide", "multiply"])
 def test_uvcalibrate(
@@ -3401,6 +3404,7 @@ def test_uvcalibrate_delay_multispw(uvcalibrate_data):
         uvutils.uvcalibrate(uvd, uvc, inplace=False)
 
 
+@pytest.mark.filterwarnings("ignore:Fixing auto-correlations to be be real-only,")
 @pytest.mark.filterwarnings("ignore:The uvw_array does not match the expected values")
 @pytest.mark.parametrize("future_shapes", [True, False])
 def test_apply_uvflag(future_shapes):
--- a/pyuvdata/tests/test_utils.py
+++ b/pyuvdata/tests/test_utils.py
@@ -3366,6 +3366,7 @@ def test_uvcalibrate_wideband_gain(uvcalibrate_data):
         uvutils.uvcalibrate(uvd, uvc, inplace=False)
 
 
+@pytest.mark.filterwarnings("ignore:Fixing auto-correlations to be be real-only")
 @pytest.mark.filterwarnings("ignore:The uvw_array does not match the expected values")
 @pytest.mark.filterwarnings("ignore:When converting a delay-style cal to future array")
 @pytest.mark.filterwarnings("ignore:Nfreqs will be required to be 1 for wide_band cals")
--- a/pyuvdata/uvdata/fhd.py
+++ b/pyuvdata/uvdata/fhd.py
@@ -372,11 +372,11 @@ class FHD(UVData):
             Option to raise an error rather than a warning if the check that
             uvws match antenna positions does not pass.
         check_autos : bool
-            Check whether any auto-correlations have imaginary values in them (which
-            should not mathematically exist). Default is True.
+            Check whether any auto-correlations have non-zero imaginary values in
+            data_array (which should not mathematically exist). Default is True.
         fix_autos : bool
             If auto-correlations with imaginary values are found, fix those values so
-            that they are real-only. Default is False.
+            that they are real-only in data_array. Default is False.
 
         Raises
         ------
--- a/pyuvdata/uvdata/fhd.py
+++ b/pyuvdata/uvdata/fhd.py
@@ -331,6 +331,8 @@ class FHD(UVData):
         check_extra=True,
         run_check_acceptability=True,
         strict_uvw_antpos_check=False,
+        check_autos=True,
+        fix_autos=True,
     ):
         """
         Read in data from a list of FHD files.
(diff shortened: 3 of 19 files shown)


bhazelton:
I like the approach here and it's very timely! The biggest question I have is do we want to default to making the autos be real? I worry that non-real autos is a big enough problem that we should maybe default to erroring rather than warning.

kartographer:
So don't have particularly strong feelings about warn vs error, especially given that the remedy (calling `_fix_autos` by way of setting `fix_autos=True`) is fairly straight forward and already called out in the error message itself. I generally favor being more tolerant of importing imperfect data versus writing out imperfect data, but that's a nicety, not a requirement. 

However, given that the old `phase` method used to leave behind some (small) imaginary values in the autos, I think it would be prudent to at least have a window during which we are issuing a warning rather than a full-blown error, just to give users some heads up on which of their datasets may be impacted by this new check. That includes a few of our own test datasets (which may warrant a separate issue being filed), as well as those I see passing through the HERA pipeline (or whatever the Azure Pipeline tests are running).

bhazelton:
@jsdillon This PR is adding a check for non-real autos, which is an issue that comes up from time to time and generally means there's a bug in upstream code (sometimes including the correlator). This new check seems to be breaking some tests on hera_cal, presumably because you have some old data files with these issues.

It is possible those problems originated with pyuvdata, we recently realized that we sometimes had some numerical precision issues in the old phasing code that caused the autos to have very small (< one part in 10^10) imaginary parts. If that's the source I apologize! In this new checking code, there is an option to just set the imaginary parts of the autos to zero, which you might want to use, depending on the source of the problems in your files.

jsdillon:
This is going to be tricky to fix in `hera_cal`, where it appears to be the case that the test files do have small imaginary parts in the autos (<1e-10) but also I'm finding that calibrating autos can also introduce very small imaginary components because the calibration involves multiplication by each gain individually (so there's roundoff errors). That may be fixable, I'm not sure.

In the meantime, this PR needs a `fix_autos` kwarg for this function: https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/403956f07c92d024360046dc9bc7d4e2038621a7/pyuvdata/uvdata/uvh5.py#L1471

jsdillon:
Ok, I've done some digging and opened a draft PR which fixes 9 of the 12 test failures when running under this branch: https://github.com/HERA-Team/hera_cal/pull/749

The rest are in `hera_cal.vis_clean` and so I'm passing this off to @aewallwi to figure out how he'd like to adapt his code.

bhazelton:
Thanks @jsdillon! Sorry to make work for you.

kartographer:
Thanks @jsdillon -- I've gone ahead and added the `fix_autos` keyword to the `UVData.write_uvh5_part` method in 0369ee1.

jsdillon:
Apologies for the delay, but I think we have a PR that fixes the hera_cal issue. Does anyone here want to review it?

kartographer:
@jsdillon -- FWIW, I reran the above hera_cal checks, and it seems like there are three tests that are still failing. On first glance though, it seems like the failing tests themselves are actually plugging random complex data into `data_array`, which may be inadvertently injecting complex values into the autos (and hence triggering the error).

aewallwi:
Some thoughts on this issue:

While they are only real, auto-correlations are complex data. Radio astronomers routinely perform completely valid complex arithmetic on auto-correlations which, in every case,  introduce small non-real components from numerical precision errors. Examples include applying calibration solutions, interpolating flagged channels etc... 

By requiring that auto-correlations to be exactly zero, any arithmetic operation on visibilities must always manually set auto-correlation amplitudes to zero after that operation, even when this is not actually necessary. 

In order to avoid lots of downstream manual setting of autocorrelations to zero, would it make sense to define the real autos requirement to be that the imag part of the autos is below some fraction of the real part rather then the imag part being precisely equal to zero?

bhazelton:
I think that's probably a reasonable approach. I do think we need a check on this as this has been used to identify bugs in both correlators and downstream processing on multiple occasions now. The question then becomes what should the tolerance be?

aewallwi:
> I think that's probably a reasonable approach. I do think we need a check on this as this has been used to identify bugs in both correlators and downstream processing on multiple occasions now. The question then becomes what should the tolerance be?

Maybe we can figure out what that should be through monte carlo simulations of some arithmatic operations?

bhazelton:
If you want to do that it'd be interesting input. I think it should also be somewhat physically motivated in terms of how much phase angle it is. In some other parts of pyuvdata we use 1 milliarcsecond, but those are usually for physical angles not phase...

aewallwi:
> If you want to do that it'd be interesting input. I think it should also be somewhat physically motivated in terms of how much phase angle it is. In some other parts of pyuvdata we use 1 milliarcsecond, but those are usually for physical angles not phase...

The other option is to just pick a value that we know will never be scientifically relevant.

bhazelton:
I guess the problem with the phase argument is that it is related to the absolute value of the visibility, which we don't want here. A value that won't be scientifically relevant was where I was leading with the "physically motivated" comment before.

kartographer:
Hi @aewallwi -- the choice of checking only that the imaginary component was zero was based on three observations/concerns:

1.  In just about every (local) test I ran, doing a relative check between real and imaginary values was considerably slower. Admittedly, my testing wasn't exhaustive, but the difference I was generally seeing was about an order-of-magnitude or worse (`np.real_if_close` is faster than this, but also has some peculiar behavior  that actually makes it a less-than-ideal choice). Because of this, I think I'd still keep the first `iscomplex` check, and put additional logic under/inside it to handle the `isclose` case (otherwise this check starts feeling about as "lightweight" as a deep-fried Twinkie).
2. As perhaps the above discussion already demonstrates, I was a bit concerned that any choice of tolerance would be arbitrary, particularly since the induced errors likely depend on the floating-point precision of the arithmetic in question (not necessarily something intrinsic to the instrument). A Monte Carlo simulation could be quite helpful in that regard.
3. I was a little worried about what I can only think to call "floating point creep", where multiple subsequent calls to some function inducing complex values results in an error, that's not immediately traceable to any single operation. As you point out, needing to do this cleanup manually after each operation is a bit annoying, which was part of the motivation for breaking out `UVData._fix_autos` as a separate function, since it at least make "cleanup" of this issue a bit more straightforward (although it's not _currently_ in the public API, though that could change).

The only dog I have in this fight is the first item above - I'd just like it to be fast. But I'm happy to make the code a bit more tolerant, if that's what folks desire. Though FWIW, I don't think the tolerance issue is what's causing the current fails on the hera_cal pipeline.

david-macmahon:
Just another example of non-zero imaginary components of auto-correlations...

xGPU computes(-ed?) auto correlations the same as cross correlations because it is (was?) faster to do them the same than to have a separate real-only computation.  When configured to use floating point math, the GPUs used in the early days of xGPU had a fused-multiply-add (FMA) instruction but not a corresponding fused-multiply-subtract operation, so some terms of the imaginary component of auto correlations that cancel out in theory did not cancel out completely in practice due to the difference in precision of the fused-multiply-add vs the non-fused-multiply-subtract.  Not sure if that's still the case with today's GPUs (and most folks use xGPU's integer math now which is not afflicted this way), but thought I'd add it here for historical context/posterity.

aewallwi:
> Hi @aewallwi -- the choice of checking only that the imaginary component was zero was based on three observations/concerns:
> 
> 1. In just about every (local) test I ran, doing a relative check between real and imaginary values was considerably slower. Admittedly, my testing wasn't exhaustive, but the difference I was generally seeing was about an order-of-magnitude or worse (`np.real_if_close` is faster than this, but also has some peculiar behavior  that actually makes it a less-than-ideal choice). Because of this, I think I'd still keep the first `iscomplex` check, and put additional logic under/inside it to handle the `isclose` case (otherwise this check starts feeling about as "lightweight" as a deep-fried Twinkie).
> 2. As perhaps the above discussion already demonstrates, I was a bit concerned that any choice of tolerance would be arbitrary, particularly since the induced errors likely depend on the floating-point precision of the arithmetic in question (not necessarily something intrinsic to the instrument). A Monte Carlo simulation could be quite helpful in that regard.
> 3. I was a little worried about what I can only think to call "floating point creep", where multiple subsequent calls to some function inducing complex values results in an error, that's not immediately traceable to any single operation. As you point out, needing to do this cleanup manually after each operation is a bit annoying, which was part of the motivation for breaking out `UVData._fix_autos` as a separate function, since it at least make "cleanup" of this issue a bit more straightforward (although it's not _currently_ in the public API, though that could change).
> 
> The only dog I have in this fight is the first item above - I'd just like it to be fast. But I'm happy to make the code a bit more tolerant, if that's what folks desire. Though FWIW, I don't think the tolerance issue is what's causing the current fails on the hera_cal pipeline.

I agree that we should prioritize speed. I think the `fix_autos` option is concise enough that it will let downstream code avoid clunky manual setting of the autos. Thanks for clarifying this! I'm happy with this to proceed.

bhazelton (approved):
This looks good and I think the discussion has converged.

PR #1117, opened 2021-12-19
Improved handling of MIR Data + flexible polarization for UVData

### LINKED ISSUE #1076: Adding flexible polarization handling for spectral windows
The Submillimeter Array has the capability to separately tune the two polarizations we are sensible to (i.e., V & H), which is typically used to provide additional spectral coverage of the interferometer. As such, we effectively have different spectral windows which do not have the same polarization.

Both the SMA MIR and CASA MS formats support the ability to support such an operating mode, but `UVData` objects do not yet as of the writing of this issue. One workaround would be to label everything effectively Stokes I, although that results in some loss of information about the data which is less-than-desirable. One alternative could be the addition of an optional attribute, `flex_spw_polarization` (which would only be used when `flex_spw=True`), an array on `ints` with length `Nspws`, which would allow for the assignment of polarization on a per-window basis.

kartographer:
## Description

- Added a new feature, which I've termed "flexible polarization handling", has been added to the `UVData` object. This new feature leverages a new, optional attribute called `flex_spw_polarization_array`, which enables one to encode a different polarization type for each spectral window. When this value is set to something other than `None`, `polarization_array` is set to be equal to `[0]`.
- Added New methods `_make_flex_pol` and `remove_flex_pol` have been added to allow one to convert back/forth from "flex-pol" to "regular" `UVData` objects, when needed.
- Overhauled the `mir` and `mir_parser` objects, allowing them to process much faster and enabling one to read in multiple polarization types simultaneously (currently only one polarization type supported on each read), and enabling first-order flux calibration (moving `vis_units` from "uncalib" to "Jy").

## Motivation and Context

After some discussion within the SMA user community, we determined that some changes were needed with the way that pyuvdata handles MIR data in order to enable "full" handling of SMA data for science users (i.e., moving MIR support from "beta" to v1.0). This PR covers approximately one-third of the desired functionality, which we expect should be sufficient for end-to-end handling of continuum data (a subsequent PR will deal with what's required for spectral line observations). 

**Changes to `mir` and `mir_parser`**

Beyond a reorganization of internal functions, the most significant addition to `mir_parser` is the addition of the `_apply_tsys` method, which will convert visibilities (normally recorded in correlator coefficients) from uncalibrated units to Jy units. As currently written, `_apply_tsys` will be called when working with the "normal"/floating-point visibility values, but not the "raw"/compressed/integer values. `mir` has had a significant rewrite to improve both the speed at which a `MirParser` object is converted into a `UVData` object, and to enable more than one polarization mode to be passed at any given time. Additionally, `mir` has been modified so that one can instantiate a `UVData` object from _either_ a file name or a `MirParser` object.

**Changes to `UVData`**

The most significant addition to `UVData` has been the addition of "flexible-polarization" handling, which allows for different spectral windows to have different polarizations (currently only one polarization allowed per-window). The primary benefit of allowing this is that one can more efficiently store data when different windows sparsely populate the polarization axis (as is the case for SMA, which can separately tune its X- and Y-polarization receivers). These per-window polarization values are stored in a new, optional attribute called `flex_spw_polarization_array` (which would also nominally be added to the UVH5 format). I have attempted to implement flex-pol in such a way that it ought to be seamless for most users to use -- the only current method that might require a user to "remove" this new feature is the `__add__` method, which does not nominally allow for "flex-pol" and "regular" data sets to be added together.

Closes #1076

## Types of changes

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/uvdata/mir.py
+++ b/pyuvdata/uvdata/mir.py
@@ -289,6 +289,7 @@ class Mir(UVData):
         # Load up the visibilities into the MirParser object.
         vis_data = np.zeros((Nblts, Npols, Nfreqs), dtype=np.complex64)
         mir_data.load_data(load_vis=True)
+        mir_data._apply_tsys()
 
         for sp_rec, window, vis_rec in zip(
             mir_data.sp_data, spdx_list, mir_data.vis_data
@@ -436,8 +437,7 @@ class Mir(UVData):
         # standard a2-a1 that uvdata expects
         self.uvw_array = (-1.0) * uvw_array
 
-        # todo: Raw data is in correlation coefficients, we may want to convert to Jy.
-        self.vis_units = "uncalib"
+        self.vis_units = "Jy"
 
         sou_list = mir_data.codes_data[mir_data.codes_data["v_name"] == b"source"]
         isource = np.unique(mir_data.in_data["isource"])
--- a/pyuvdata/uvdata/mir.py
+++ b/pyuvdata/uvdata/mir.py
@@ -74,12 +74,12 @@ class Mir(UVData):
 
         if irec is not None:
             mir_data.use_bl *= np.isin(mir_data.bl_read["irec"], irec)
-            if not np.any(mir_data.use_in):
+            if not np.any(mir_data.use_bl):
                 raise ValueError("No valid receivers selected!")
 
         if isb is not None:
             mir_data.use_bl *= np.isin(mir_data.bl_read["isb"], isb)
-            if not np.any(mir_data.use_in):
+            if not np.any(mir_data.use_bl):
                 raise ValueError("No valid sidebands selected!")
 
         if corrchunk is not None:
@@ -90,9 +90,6 @@ class Mir(UVData):
             mir_data.use_sp *= mir_data.sp_read["corrchunk"] != 0
 
         mir_data._update_filter()
-
-        if len(mir_data.in_data) == 0:
-            raise IndexError("No valid records matching those selections!")
 
         self._init_from_mir_parser(mir_data)
 
@@ -273,15 +270,18 @@ class Mir(UVData):
             Nfreqs += spw_dict[key]["nchan"]
 
         # Initialize some arrays that we'll be appending to
-        flex_spw_id_array = np.zeros(Nfreqs, dtype=np.int64)
+        flex_spw_id_array = np.zeros(Nfreqs, dtype=int)
         channel_width = np.zeros(Nfreqs, dtype=np.float64)
         freq_array = np.zeros(Nfreqs, dtype=np.float64)
-        flex_pol = np.zeros(Nspws, dtype=np.int64)
+        flex_pol = np.zeros(Nspws, dtype=int)
         for idx, key in enumerate(spw_array):
             flex_spw_id_array[spw_dict[key]["ch_slice"]] = spw_dict[key]["spw_id_array"]
             channel_width[spw_dict[key]["ch_slice"]] = spw_dict[key]["channel_width"]
             freq_array[spw_dict[key]["ch_slice"]] = spw_dict[key]["freq_array"]
             flex_pol[idx] = spw_dict[key]["pol_state"] if pol_split_tuning else 0.0
+
+        if pol_split_tuning:
+            flex_pol = None
 
         for key in spdx_dict:
             spdx_dict[key]["ch_slice"] = spw_dict[spdx_dict[key]["spw_id"]]["ch_slice"]
@@ -429,7 +429,7 @@ class Mir(UVData):
         self.time_array = Time(mjd_array, scale="tt", format="mjd").utc.jd
 
         self.polarization_array = polarization_array
-        self.spw_array = spw_array
+        self.spw_array = np.array(spw_array, dtype=int)
         self.telescope_name = "SMA"
 
         # Need to flip the sign convention here on uvw, since we use a1-a2 versus the
@@ -460,7 +460,7 @@ class Mir(UVData):
                 cat_epoch=source_epoch,
                 cat_frame="fk5",
                 info_source="file",
-                cat_id=sou_id,
+                cat_id=int(sou_id),
             )
 
         # Regenerate the sou_id_array thats native to MIR into a zero-indexed per-blt
--- a/pyuvdata/uvdata/mir.py
+++ b/pyuvdata/uvdata/mir.py
@@ -33,11 +33,11 @@ class Mir(UVData):
         isb=None,
         corrchunk=None,
         pseudo_cont=False,
-        flex_spw=True,
         run_check=True,
         check_extra=True,
         run_check_acceptability=True,
         strict_uvw_antpos_check=False,
+        allow_flex_pol=True,
     ):
         """
         Read in data from an SMA MIR file, and map to the UVData model.
@@ -63,6 +63,23 @@ class Mir(UVData):
             Read in only pseudo-continuuum values. Default is false.
         flex_spw : boolean
             Allow for support of multiple spectral windows. Default is true.
+        run_check : bool
+            Option to check for the existence and proper shapes of parameters
+            before writing the file.
+        check_extra : bool
+            Option to check optional parameters as well as required ones.
+        run_check_acceptability : bool
+            Option to check acceptable range of the values of parameters before
+            writing the file.
+        strict_uvw_antpos_check : bool
+            Option to raise an error rather than a warning if the check that
+            uvws match antenna positions does not pass.
+        allow_flex_pol : bool
+            If only one polarization per spectral window is read (and the polarization
+            differs from window to window), allow for the `UVData` object to use
+            "flexible polarization", which compresses the polarization-axis of various
+            attributes to be of length 1, sets the `flex_spw_polarization_array`
+            attribute to define the polarization per spectral window. Default is True.
         """
         # Use the mir_parser to read in metadata, which can be used to select data.
         mir_data = mir_parser.MirParser(filepath)
@@ -91,7 +108,7 @@ class Mir(UVData):
 
         mir_data._update_filter()
 
-        self._init_from_mir_parser(mir_data)
+        self._init_from_mir_parser(mir_data, allow_flex_pol=allow_flex_pol)
 
         if run_check:
             self.check(
@@ -101,7 +118,7 @@ class Mir(UVData):
                 allow_flip_conj=True,
             )
 
-    def _init_from_mir_parser(self, mir_data):
+    def _init_from_mir_parser(self, mir_data, allow_flex_pol=True):
         """
         Convert a MirParser object into a UVData object.
 
@@ -109,6 +126,12 @@ class Mir(UVData):
         ----------
         mir_data : MirParser object
             MIR dataset to be converted into a UVData object.
+        allow_flex_pol : bool
+            If only one polarization per spectral window is read (and the polarization
+            differs from window to window), allow for the `UVData` object to use
+            "flexible polarization", which compresses the polarization-axis of various
+            attributes to be of length 1, sets the `flex_spw_polarization_array`
+            attribute to define the polarization per spectral window. Default is True.
         """
         # By default, we will want to assume that MIR datasets are phased, multi-spw,
         # and multi phase center. At present, there is no advantage to allowing these
@@ -175,11 +198,9 @@ class Mir(UVData):
                 code["code"].decode("UTF-8").lower()
             ]
 
-        if pol_split_tuning:
-            # Encode polarization array to be pseudo Stokes I, since that is effectively
-            # how these values are going to be treated.
+        if pol_split_tuning and allow_flex_pol:
             Npols = 1
-            polarization_array = np.array([1])
+            polarization_array = np.array([0])
         else:
             Npols = len(set(pol_dict.values()))
             polarization_array = np.zeros(Npols, dtype=int)
@@ -255,7 +276,9 @@ class Mir(UVData):
 
             spdx_dict[spdx] = {
                 "spw_id": spw_id,
-                "pol_idx": 0 if pol_split_tuning else pol_dict[spdx[2]],
+                "pol_idx": (
+                    0 if (pol_split_tuning and allow_flex_pol) else pol_dict[spdx[2]]
+                ),
             }
 
             spw_dict[spw_id] = {
--- a/pyuvdata/uvdata/mir.py
+++ b/pyuvdata/uvdata/mir.py
@@ -167,26 +167,25 @@ class Mir(UVData):
                 loa_freq = np.median(mir_data.sp_data["gunnLO"][sel_mask])
                 lob_freq = np.median(mir_data.sp_data["gunnLO"][~sel_mask])
                 pol_split_tuning = not np.isclose(loa_freq, lob_freq)
-                pol_dict = {0: 0, 1: int(not pol_split_tuning)}
+                pol_dict = {0: 0, 1: 1}
 
         pol_code_dict = {}
         for code in mir_data.codes_read[mir_data.codes_read["v_name"] == b"pol"]:
-            pol_code_dict[code["icode"]] = code["code"].decode("UTF-8").lower()
+            pol_code_dict[code["icode"]] = uvutils.POL_STR2NUM_DICT[
+                code["code"].decode("UTF-8").lower()
+            ]
 
         if pol_split_tuning:
             # Encode polarization array to be pseudo Stokes I, since that is effectively
             # how these values are going to be treated.
             Npols = 1
             polarization_array = np.array([1])
-            # TODO: Add split tuning handing here
         else:
             Npols = len(set(pol_dict.values()))
             polarization_array = np.zeros(Npols, dtype=int)
 
             for key in pol_dict.keys():
-                polarization_array[pol_dict[key]] = uvutils.POL_STR2NUM_DICT[
-                    pol_code_dict[key]
-                ]
+                polarization_array[pol_dict[key]] = pol_code_dict[key]
 
         blt_list = [
             (intid, ant1, ant2)
@@ -235,7 +234,7 @@ class Mir(UVData):
 
             spw_id = 255 if (spdx[0] == 0) else spdx[0]
             spw_id *= (-1) ** (1 + spdx[1])
-            spw_id += 512 if pol_split_tuning else 0
+            spw_id += 512 if (pol_split_tuning and spdx[2] == 1) else 0
 
             # Populate the channel width array
             channel_width = abs(spw_fres) + np.zeros(spw_nchan, dtype=np.float64)
@@ -256,7 +255,7 @@ class Mir(UVData):
 
             spdx_dict[spdx] = {
                 "spw_id": spw_id,
-                "pol_idx": pol_dict[spdx[2]],
+                "pol_idx": 0 if pol_split_tuning else pol_dict[spdx[2]],
             }
 
             spw_dict[spw_id] = {
@@ -266,7 +265,7 @@ class Mir(UVData):
                 "channel_width": channel_width,
                 "spw_id_array": spw_id_array,
                 "freq_array": freq_array,
-                "pol_state": spdx[2],
+                "pol_state": pol_code_dict[spdx[2]],
             }
 
         spw_array = sorted(spw_dict.keys())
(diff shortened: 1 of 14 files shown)


pyuvdata/uvdata/tests/test_uvh5.py
```diff
@@ -62,6 +62,24 @@ def uv_partial_write(casa_uvfits, tmp_path):
...
+
+
+@pytest.fixture(scope="function")
+def sma_mir(sma_mir_main):
+    # read in test file for the resampling in time functions
+    uv_object = sma_mir_main.copy()
+
+    yield uv_object
+
+
```
bhazelton, line 82: I don't see where these fixtures are used...

bhazelton (commented):
I think that the UVData part of this looks pretty reasonable. I don't think I'm competent to review the MIR parser parts of this...

pyuvdata/uvdata/mir.py
```diff
@@ -33,7 +33,11 @@ def read_mir(
...
         pseudo_cont=False,
-        flex_spw=True,
+        run_check=True,
+        check_extra=True,
+        run_check_acceptability=True,
+        strict_uvw_antpos_check=False,
+        allow_flex_pol=True,
     ):
         """
         Read in data from an SMA MIR file, and map to the UVData model.
```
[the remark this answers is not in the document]
kartographer: The steady drumbeat of progress! Fixed in a450660.

pyuvdata/uvdata/mir.py
```diff
@@ -49,20 +53,86 @@ def read_mir(
             The file path to the MIR folder to read from.
         isource : list of int
             Source code(s) for MIR dataset
-        irec : int
+        irec : array-like of int
```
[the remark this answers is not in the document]
kartographer: Fixed in a450660.

pyuvdata/uvdata/mir.py
```diff
@@ -49,20 +53,86 @@ def read_mir(
...
             Receiver code for MIR dataset
-        isb : list of int
+        isb : array-like of int
             Sideband codes for MIR dataset (0 = LSB, 1 = USB). Default is both sb.
-        corrchunk : list of int
+        corrchunk : array-like of int
```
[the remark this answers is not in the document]
kartographer: Fixed in a450660.

pyuvdata/uvdata/uvdata.py
```diff
@@ -6141,7 +6342,40 @@ def __add__(
...
+                    )
+                }
+                for key in other_flexpol_dict.keys():
+                    try:
+                        if this_flexpol_dict[key] != other_flexpol_dict[key]:
+                            raise ValueError(
+                                "Cannot add a flex-pol UVData objects where the same "
+                                "spectral window contains different polarizations. Use "
+                                "the `remove_flex_pol` method to convert the objects "
+                                "to have a regular polariztion axis."
```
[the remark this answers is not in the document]
kartographer: All those times cheating on spelling tests in elementary school has come back to haunt me. Fixed in a450660.

pyuvdata/uvdata/tests/test_uvh5.py
kartographer: Whoops -- they _were_ used, but I moved a test out of `test_uvh5` and into `test_mir` instead. Fixed in a450660.

kartographer:
Thanks @bhazelton -- I think I've responded to all of your comments. As far as the changes to `MirParser` goes, most of the changes were either just reordering of methods (putting them in "what gets called first" order), or in the case of `test_mir_parser`, changing variable names to make things a bit more consistent (and eventually, easier to combine with #1110). The only really new piece of code is `_apply_tsys` -- maybe @e-koch or @ChrisMoriarty could take a look at it (they're the only other two people to touch this class)? Otherwise, I'd be happy to step through it with whomever might be up for reviewing it.

pyuvdata/uvdata/tests/test_mir.py
```diff
@@ -13,23 +13,53 @@
 
 import pytest
 import numpy as np
+import pyuvdata.tests as uvtest
```
ChrisMoriarty, line 16: Looks like both relative and absolute imports from other pyuvdata modules are used in the same file.  General wisdom says that files that are part of a python package should use relative imports so you can guarantee you don't import some other thing that happens to have the same name.

pyuvdata/uvdata/uvdata.py
```diff
@@ -2360,6 +2380,167 @@ def _check_freq_spacing(self, raise_errors=True):
...
+        # Finally, remove the flex-pol attribute
+        self.flex_spw_polarization_array = None
+
+    def _make_flex_pol(self, raise_error=False, raise_warning=True):
+        """
+        Convert a regular UVData object into one with flex-polarization enabled.
+
+        This is an internal helper function, which is not designed to be called by
+        users, but rather individual read/write functions for the UVData object.
+        This will convert a regulat UVData object into one that uses flexible
```
ChrisMoriarty, line 2447: possible typo regulat?

pyuvdata/uvdata/uvdata.py
```diff
@@ -2360,6 +2380,167 @@ def _check_freq_spacing(self, raise_errors=True):
...
+                    np.zeros_like(self.spw_array) + self.polarization_array[0]
+                )
+                self.polarization_array = np.array([0])
+            return
+
+        flex_pol_idx = np.zeros_like(self.spw_array)
+        for idx, spw in enumerate(self.spw_array):
+            spw_screen = self.flex_spw_id_array == spw
+
+            if self.future_array_shapes:
```
ChrisMoriarty, line 2485: Might add a comment to say why the results of np.all in both cases have to be flipped with ~

pyuvdata/uvdata/uvdata.py
```diff
@@ -2360,6 +2380,167 @@ def _check_freq_spacing(self, raise_errors=True):
...
+        # Otherwise, move through all of the data params
+        for name, param in zip(self._data_params, self.data_like_parameters):
+            # We need to construct arrays with the appropriate shape
+            new_shape = [self.Nblts, 1, self.Nfreqs, self.Npols]
+
+            # Pop the spw-axis if using future array shapes
+            if self.future_array_shapes:
+                del new_shape[1]
+
+            # Use fill here, since we want zeros if we are working with nsample_array
```
ChrisMoriarty, line 2416: I might be getting pedantic, but it says "fill", but actually uses "full"

pyuvdata/uvdata/uvdata.py
```diff
@@ -2360,6 +2380,167 @@ def _check_freq_spacing(self, raise_errors=True):
...
+        # right shape for the new values, modulo future_array_shapes
+        new_shape = [self.Nblts, 1, self.Nfreqs, 1]
+
+        # Pop the spw-axis if using future array shapes
+        if self.future_array_shapes:
+            del new_shape[1]
+
+        # Now go through one-by-one with data-like parameters and update
+        for name, param in zip(self._data_params, self.data_like_parameters):
+            # Use fill here, since we want zeros if we are working with nsample_array
```
ChrisMoriarty, line 2525: again says "fill" but uses empty()

ChrisMoriarty (commented):
Just a cursory look over the PR, looking for typos, formatting and such.

pyuvdata/uvdata/tests/test_mir.py
kartographer: Fixed in 8f8a8af4.

pyuvdata/uvdata/uvdata.py
kartographer: Well that's not right. Fixed in 8f8a8af4.

pyuvdata/uvdata/uvdata.py
kartographer: Added in 8f8a8af4.

pyuvdata/uvdata/uvdata.py
kartographer: I blame the good folks at numpy. Fixed in 8f8a8af4.

pyuvdata/uvdata/uvdata.py
kartographer: Ah, good catch -- old code used "fill", until I realized I was writing to every single value in the array, hence why I switched to empty.

kartographer:
Thanks @ChrisMoriarty -- updates pushed.

pyuvdata/uvdata/mir.py
```diff
@@ -71,166 +141,242 @@ def read_mir(
...
-        for idx in range(len(corrchunk)):
-            data_mask = np.logical_and(
-                mir_data.sp_data["corrchunk"] == corrchunk[idx],
-                mir_data.bl_data["isb"][sp_bl_maparr] == corrchunk_sb[idx],
+        # Create a dict with the ordering of the pols
+        pol_dict = {key: idx for idx, key in enumerate(np.unique(pol_arr))}
+
+        pol_split_tuning = False
+        if len(pol_dict) == 2:
+            # If dual-pol, then we need to check if the tunings are spilt, because
```
e-koch, line 194: Typo in split.

pyuvdata/uvdata/mir.py
```diff
@@ -71,166 +141,242 @@ def read_mir(
...
+                code["code"].decode("UTF-8").lower()
+            ]
+
+        if pol_split_tuning and allow_flex_pol:
+            # If we have a split tuning that, that means we can take advantage of
+            # the flex-pol feature within UVData
+            Npols = 1
+            polarization_array = np.array([0])
+        else:
+            # Otherwise, calculate dimentions and arrays for the polarization axis
```
e-koch, line 226: Typo in dimensions.

pyuvdata/uvdata/mir.py
```diff
@@ -274,49 +417,104 @@ def read_mir(
...
-        self.integration_time = mir_data.sp_data["integ"][:: len(corrchunk)].astype(
-            float
-        )
+        # Let's try to chalk up all the per-sp stuff here, eh?
+        sp_to_blt = ["igq", "ipq", "flags", "vradcat"]
+        sp_temp_dict = {}
+        suppress_warning = False
+        for sp_rec in mir_data.sp_data:
+            temp_dict = {item: sp_rec[item] for item in sp_to_blt}
+            try:
```
e-koch, line 430: @kartographer -- Can you add a short explanation here for which `dict` throws the keyerror? Lines 440/442 make this a bit confusing.

pyuvdata/uvdata/mir.py
```diff
@@ -274,49 +417,104 @@ def read_mir(
...
+                if blt_temp_dict[blhid_blt_order[bl_rec["blhid"]]] != temp_dict:
+                    # Again, if we reach this point, only raise a warning one time
+                    if not suppress_warning:
+                        warnings.warn(
+                            "Per-baseline metadata differ. Defaulting to using "
+                            "the last value in the data set."
+                        )
+                        suppress_warning = True
+                    blt_temp_dict[blhid_blt_order[bl_rec["blhid"]]] = temp_dict
```
e-koch, line 464: Same as above.

pyuvdata/uvdata/mir.py
```diff
@@ -369,43 +566,11 @@ def read_mir(
...
-            data_array = np.reshape(
-                np.concatenate(mir_data.vis_data),
-                (self.Nblts, 1, self.Nfreqs, self.Npols),
-            )
-
-        # Don't need the data anymore, so drop it
-        mir_data.unload_data()
-
-        self.data_array = data_array
+        self.data_array = np.transpose(vis_data, (0, 2, 1))[:, np.newaxis, :, :]
```
e-koch, line 573: Which axes need to be transposed here?

pyuvdata/uvdata/mir_parser.py
```diff
@@ -925,3 +728,359 @@ def read_vis_data(filepath, in_start_dict):
...
+        jypk : float
+            Forward gain of the antenna (in units of Jy/K), which is multiplied against
+            the system temperatures in order to produce values in units of Jy
+            (technically this is the SEFD, which when multiplied against correlator
+            coefficients produces visibilities in units of Jy). Default is 130.0, which
+            is the estiamted value for SMA.
+        """
+        # Create a dictionary here to map antenna pair + integration time step with
+        # a sqrt(tsys) value. Note that the last index here is the receiver number,
+        # which techically has a diffenret keyword under which the system temperatures
```
e-koch, line 911: Typo in different

pyuvdata/uvdata/ms.py
```diff
@@ -382,7 +394,7 @@ def _write_ms_field(self, filepath):
 
         field_table = tables.table(filepath + "::FIELD", ack=False, readonly=False)
         time_val = (
-            Time(np.median(self.time_array), format="jd", scale="utc").tai.mjd * 86400.0
+            Time(np.median(self.time_array), format="jd", scale="utc").mjd * 86400.0
```
e-koch, line 397: Very minor, but I think this could be tied to an astropy unit conversion: `Time().mjd.to(u.s).value`?

pyuvdata/uvdata/ms.py
```diff
@@ -499,7 +511,7 @@ def _write_ms_source(self, filepath):
         )
 
         time_val = (
-            Time(np.median(self.time_array), format="jd", scale="utc").tai.mjd * 86400.0
+            Time(np.median(self.time_array), format="jd", scale="utc").mjd * 86400.0
```
e-koch, line 514: As above: `Time().mjd.to(u.s).value`?

pyuvdata/uvdata/ms.py
```diff
@@ -576,7 +588,7 @@ def _write_ms_pointing(self, filepath):
         # per-timestamp, per-antenna entry
         times = np.asarray(
             [
-                Time(t, format="jd", scale="utc").tai.mjd * 3600.0 * 24.0
+                Time(t, format="jd", scale="utc").mjd * 3600.0 * 24.0
```
e-koch, line 591: `Time().mjd.to(u.s).value` or `86400.0` for consistency.

pyuvdata/uvdata/ms.py
```diff
@@ -1107,7 +1142,10 @@ def write_ms(
         # this date conversion as few times as possible. Note that the default for MS
         # is MJD UTC seconds, versus JD UTC days for UVData.
         time_array, time_ind = np.unique(self.time_array, return_inverse=True)
-        time_array = (Time(time_array, format="jd").mjd * 3600.0 * 24.0)[time_ind]
+        # TODO: Verify this should actually be UTC, and not some other scale
+        time_array = (Time(time_array, format="jd", scale="utc").mjd * 3600.0 * 24.0)[
```
e-koch, line 1146: `Time().mjd.to(u.s).value` or `86400` for consistency.

pyuvdata/uvdata/tests/test_mir_parser.py
```diff
@@ -74,3 +70,334 @@ def test_mir_parser_index_linked(mir_data_object):
...
+# sure that we haven't broken the ability of the reader to handle the data. Since
+# this file is the basis for the above checks, we've put this here rather than in
+# test_mir_parser.py
+
+
+def test_mir_remember_me_record_lengths(mir_data):
+    """
+    Mir record length checker
+
+    Make sure the test file containts the right number of records
```
e-koch, line 115: Typo in contains

pyuvdata/uvdata/uvdata.py
```diff
@@ -11173,6 +11510,14 @@ def read(
             combining multiple files, which would otherwise result in an error being
             raised because of attributes not matching. Doing so effectively adopts the
             name found in the first file read in. Default is False.
+        allow_flex_pol : bool
+            If only one polarization per spectral window is read (and the polarization
+            differs from window to window), allow for the `UVData` object to use
+            "flexible polarization", which compresses the polarization-axis of various
+            attributes to be of length 1, sets the `flex_spw_polarization_array`
+            attribute to define the polarization per spectral window. Only applicable
+            for MIR and MS filetypes, otherwise this argument is ignord. Default is
```
e-koch, line 11519: Typo in ignored.

pyuvdata/uvdata/uvdata.py
```diff
@@ -12105,6 +12459,14 @@ def from_file(
...
```
e-koch, line 12468: Typo in ignored.

pyuvdata/uvdata/uvdata.py
```diff
@@ -12246,6 +12609,15 @@ def write_miriad(
             raise ValueError("Cannot write out metadata only objects to a miriad file.")
 
         miriad_obj = self._convert_to_filetype("miriad")
+
+        # If we have a flex-pol dataset, remove it so that we can pass the data to the
+        # writer without issue. We create a copy because otherwise we run the risk of
+        # messing w/ the metadata of the original object, and because the overhead of
+        # doing so is generally smaller than that from removing flex-pol.
+        if self.flex_spw_polarization_array is not None:
+            miriad_obj = miriad_obj.copy()
```
e-koch, line 12618: Does this copy run the risk of a large memory increase?

e-koch (commented):
@kartographer -- I found a few more typos and highlighted a couple of places where clarifying the code would be helpful.

pyuvdata/uvdata/ms.py
kartographer: So as luck has it, the call to `mjd` spits out value rather than a `Quantity`, which means that I can't call the `to` method on it. As such, I've kept this as is for now.

pyuvdata/uvdata/ms.py
kartographer: Ditto to the above.

pyuvdata/uvdata/mir.py
kartographer: Fixed in 5c36be1a.

pyuvdata/uvdata/mir.py
kartographer: Fixed in 5c36be1a.

pyuvdata/uvdata/mir.py
kartographer: I added some additional comments in 5c36be1a. Basically, what's happening here is that I'm scanning through per-spectral window data to see if things are changing on a per-baseline-time basis. The `dict` basically serves as a record for what baseline-times have been checked -- if no key matches, it means that particular baseline-time has not been evaluated before, hence why I plug in the key after the `except`.

pyuvdata/uvdata/mir.py
kartographer: Ditto to above. Address in 5c36be1a.

pyuvdata/uvdata/mir.py
kartographer: I added some comments in 5c36be1a, but the most efficient way to read in the data here has the data organized by (blt, pol, freq), whereas `UVData` expects (blt, 1, freq, pol), hence the flip.

pyuvdata/uvdata/mir_parser.py
kartographer: That's different. Fixed in 5c36be1a.

pyuvdata/uvdata/ms.py
kartographer: Changed to `86400.0` in 5c36be1a.

pyuvdata/uvdata/ms.py
kartographer: Changed to `86400.0` in 5c36be1a.

pyuvdata/uvdata/tests/test_mir_parser.py
kartographer: Fixed in 5c36be1a.

pyuvdata/uvdata/uvdata.py
kartographer: Didn't mean to ignore that. Fixed in 5c36be1a.

pyuvdata/uvdata/uvdata.py
kartographer: Fixed in 5c36be1a.

pyuvdata/uvdata/uvdata.py
kartographer: It will increase memory, although I think the worst case scenario is that it will only do so by 50% (inflating a flex-pol dataset to "normal" size already balloons up the object quite a bit). I debated on whether to do something more sophisticated, but given that MIRIAD will never support this particular mode of writing out data, it seemed like I would be adding a fair bit of code to wrap what was an unusual corner case. The other "easy" alternative was to convert the original object back and forth between flex-pol, although I felt a little uneasy about changing the object like that on a write operation (especially since it means that if an error gets thrown, the object is left in it's supposedly transitory state). I could poke at this more if desired, although I think this path for the data is likely to be seldom used.

kartographer:
Thanks @e-koch for taking a dive into it so quickly! Requested changes pushed, or otherwise addressed in the comments above.

bhazelton (approved):
Looks like all the comments have been addressed. I'll approve but not merge in case you're still waiting on anything @kartographer.

pyuvdata/uvdata/uvdata.py
bhazelton: I agree, I think it's unlikely to be heavily used. We can address it if and when it becomes a problem.

PR #1123, opened 2021-12-30
Fixing conjugation issue in `compress_by_redundancy`

kartographer:
Bugfix for `compress_by_redundancy`.

## Description

- The UVData method `compress_by_redundancy` now checks to see if, after compression, a given baseline lands on the "conjugate" side of the uv-plane (basically u < 0).
- The test `test_redundancy_contract_expand` now uses complex visibility data (with non-zero imaginary components) for testing.  

## Motivation and Context

This PR fixes a bug that caused `compress_by_redundancy` to sometimes return incorrectly conjugated data. The root issue at play was that when averaging data together, the method assumes that the given baseline lands on the u>0 side of the uv-plane, which would give problems if after grouping the "lead" baseline landed on the opposite side (u < 0). Earlier test coverage missed this in part because the tests were using real-only data.

## Types of changes

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 - Added `time_range`, `lsts`, and `lst_range` kwargs from UVH5.write_uvh5_part() to UVData.write_uvh5_part().
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


bhazelton (approved):
Looks good, thanks for figuring out and fixing this @kartographer !

PR #1127, opened 2022-01-13
Add option to `reorder_blts` to sort autos first 

### LINKED ISSUE #824: Support writing auto- and cross-spectra separately in uvh5 files
There have been several instances where users have wanted to look at just auto-correlations from HERA data files. Currently these are being written interspersed with the cross-correlations, and the order in which they are written makes accessing them quite slow. One proposed solution is to write the auto-correlations and cross-correlations to different HDF5 datagroups inside of uvh5 files. This approach would allow for fast access of just auto-correlations, and should be relatively transparent to the user when doing partial I/O.

We'll want to build in backwards compatibility for files that use the current approach of having autos and crosses written together, while supporting these new-style files. The actual implementation may also require additional keywords in the header of uvh5 files. Any significant changes will be noted in the uvh5 memo defining the filetype spec.

### LINKED ISSUE #911: Allow for "autos block" reordering of baseline-time axis in UVData objects
I did some profiling of reading only autocorrelation data from a UVH5 file, and saw that just grouping autocorrelations together as a block provided reading speeds comparable to files that contained only autocorrelations. We should support a keyword ordering of the `UVData.reorder_blts` method that groups autocorrelations together, so that such an ordering is easy for the user to access. Further, these profiling results imply that it may not be necessary to define a new file format (as suggested in #824).

bhazelton:
## Description

Add an option to `reorder_blts` to sort the auto visibilities before the cross visibilities.

## Motivation and Context

closes #824 
closes #911 

## Types of changes

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- An option to `UVData.reorder_blts` to sort the autos before the crosses.
 - Added new functionality to `UVData.check` to verify that auto-correlations are real-only,
 along with an option to force them to be real-only if non-zero imaginary components are detected.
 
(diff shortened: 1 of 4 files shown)


plaplant (approved):
Looks great, thanks @bhazelton!

mkolopanis (approved):
I also approve this approval.

PR #1134, opened 2022-01-31
Make handling of types in Miriad extra_keywords more general

### LINKED ISSUE #1131: Importing `miriad.py` fails on M1 Mac platforms
I've recently switched to using [miniforge3](https://github.com/conda-forge/miniforge) for the new Apple Silicon architecture to use natively-built extensions, and tried to run pyuvdata's tests. Evidently, the `np.float128` type is not available on this platform (at least in the pre-built packages available from conda), and leads to an `AttributeError` when used. It seems we only use it in one particular place in `miriad.py`, when defining a mapping between numpy types and miriad types for writing `extra_keywords`:

https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/fb2e721f0b0dfd0dab642e94c19dfac02fefd7c1/pyuvdata/uvdata/miriad.py#L1857-L1876

It seems like it's not used outside of this dictionary definition, and all tests pass if we remove that line. According to the [numpy docs on extended precision](https://numpy.org/doc/stable/user/basics.types.html#extended-precision), most platforms do not provide higher precision than `np.longdouble` anyway, which is typically 80 bits of precision for x86 machines but is 64 bits for Windows and osx-arm64. I'd vote for either removing that key from the dictionary (I doubt users are trying to write quad-precision numbers to extra keywords entries...), or replacing it with `np.longdouble` for better compatibility.

plaplant:
## Description

When deciding how to write out entries in the `extra_keywords` dictionary to Miriad, we were using a dictionary that contained explicitly numpy types in it. This makes the type checking slightly more flexible, and does not rely on explicit enumeration of various types.

## Motivation and Context

Closes #1131.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/uvdata/miriad.py
+++ b/pyuvdata/uvdata/miriad.py
@@ -1860,20 +1860,6 @@ class Miriad(UVData):
         # vectors.  This is not supported at present!
         # NB: complex numbers *should* be supportable, but are not currently
         # supported due to unexplained errors in _miriad and/or its underlying libraries
-        numpy_types = {
-            np.int8: int,
-            np.int16: int,
-            np.int32: int,
-            np.int64: int,
-            np.uint8: int,
-            np.uint16: int,
-            np.uint32: int,
-            np.uint64: int,
-            np.float16: float,
-            np.float32: float,
-            np.float64: float,
-            np.float128: float,
-        }
         types = {
             str: "a",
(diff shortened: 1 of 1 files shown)


bhazelton (approved):
This looks much better, thanks!

PR #1138, opened 2022-02-04
Use pytest < 7 because of conflict with pytest-cases

bhazelton:
## Description

Pin pytest < 7. This is a temporary fix for pytest-cases not being compatible with pytest 7.

I also added pytest-xdist for the external tests because they are now a limiting factor. Turns out it didn't help much, so I can remove it or leave it 🤷‍♀️ .

## Motivation and Context

The conflict is being tracked here: https://github.com/smarie/python-pytest-cases/issues/251
When it's fixed, we should require a newer version of pytest-cases and remove this pytest pin.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ To compile the binary extension modules such that you can successfully run
 If you want to do development on pyuvdata, in addition to the other dependencies
 you will need the following packages:
 
-* pytest >= 6.2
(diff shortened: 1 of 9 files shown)


mkolopanis (approved):
Works for me. Let's revisit this once that issue is fixed.

PR #1143, opened 2022-02-15
feat: add _get_angular_interpolators function that helps run fast interpolation

steven-murray:
## Description
This refactors the `_interp_az_za_simple` method so that the interpolation functions themselves are obtained by a separate method, `_get_angular_interpolators`. This method returns exactly what is saved into the `saved_interp_functions` dictionary, but instead of just saving them there, it also returns them to the user. In this way, the user can call them directly later on, without having to call `interp()` and doing all the checks that it does. This is useful for simulation where the same kinds of interpolation are performed repeatedly and you only really need to do checks the first time.

Furthermore, a `check_azza_domain` argument was added to the `_interp_az_za_simple` method, which can turn off the check for whether the interpolated points are within the domain of the data. This check becomes the overwhelming dominant bottleneck for simulations that reuse splines. 

## Motivation and Context
The context is the development of `vis_cpu`, which is supposed to be a fast simulator, and has a mode in which it uses `UVBeam.interp`. We've found that using this "direct" interpolation is orders of magnitude slower than using splines, and I tracked it down to the check on the azza domain. 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- The `UVBeam._get_angular_interpolators` generator, which yields the interpolation 
+  functions for the `az_za_simple` method directly, to be used without checks.
+- The `check_azza_domain` parameter to `UVBeam.interp()`, which can turn off checks of the 
+  az/za interpolation domain. This check can be a bottleneck.
 ## [2.2.7] - 2022-02-09
 
 ### Added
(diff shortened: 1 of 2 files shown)


steven-murray:
@bhazelton having written this, I realize I *could* have just left it to save the interpolation functions in the dictionary and let the user access those instead of returning them out of a function. Either way probably requires some documentation on how to use them directly instead of through `interp()` if we'd like them to serve that purpose. 

The `check_azza_domain` parameter is extremely useful, because I was finding it to be a major bottleneck.

steven-murray:
In terms of speedup, for the NF beam interpolating to a Nside=128 healpix array (but converting to az/za before interpolation), for a total of 10 calls it reduces the time from ~3min to 2 seconds.

PR #1144, opened 2022-02-15
feat: add check_azza_domain parameter to interp

steven-murray:
This is a stripped-down version of #1143 that removes the biggest bottleneck.


bhazelton:
I just checked our test for this error message and I found that if I set this new parameter to False it doesn't error when you walk outside the data array. That surprises me, I've definitely seen errors from scipy before when you try to extrapolate. Given that that is the case, I think we need some language in the docstring letting users know that they won't know if that happens.

bhazelton (approved):
Looks good, thanks!

PR #1148, opened 2022-02-22
Fix baseline number calculation and antenna number handling

### LINKED ISSUE #1114: Fix handling of uvfits antenna fields
When reading and writing "ANTENNA1", "ANTENNA2",  "SUBARRAY", and "NOSTA" fields, pyuvdata subtracts and adds 1, respectively. This can result in incorrect antenna numbers in the pyuvdata object or the output file.

### LINKED ISSUE #1147: Fix baseline number calculation
Our definition of baseline numbers does not match the standard definition in most radio astronomy file formats (uvfits, casa ms). This stems from our initial implementation of uvfits reading where we subtracted one from antenna numbers. This needs to be fixed and is tracked in #1114.

We also need to fix our baseline calculation, which currently adds one to the antenna numbers before calculating the baseline, which makes our baseline numbers not match the standard baseline numbers. This will change our internal representation of the baseline number but if we fix this at the same time as #1114 it will not affect how any files are written.

We will need to add warnings to notify users of this change. We hope this will not be too big a problem for most users because it will only affect users who were using explicit baseline numbers rather than calculating them from antenna numbers.

Places to add warnings:
- On file reads
- In any function where baseline numbers can be passed in explicitly if they are used (select, get_*, Miriad's select on read, maybe others).

PyxieLouStar:
## Description

- Remove addition/subtraction of one when reading/writing antenna arrays and antenna numbers in uvfits files.
- Remove addition/subtraction of one in baseline number calculations.

## Motivation and Context

Closes #1114
Closes #1147
Antenna arrays are not indices but rather are numbers associated with unique antennas. Subtracting or adding one to these arrays can result in UVData objects and output files containing the wrong antenna numbers for the antennas in the data. 
Subtracting or adding one to the antenna arrays when calculating baseline numbers from antenna numbers results in baseline numbers that are not consistent with convention.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/uvdata/tests/test_uvdata.py
+++ b/pyuvdata/uvdata/tests/test_uvdata.py
@@ -1560,13 +1560,21 @@ def test_select_antennas(casa_uvfits):
 
     # check for errors associated with antennas not included in data, bad names
     # or providing numbers and names
-    pytest.raises(
-        ValueError, uv_object.select, antenna_nums=np.max(unique_ants) + np.arange(1, 3)
-    )
-    pytest.raises(ValueError, uv_object.select, antenna_names="test1")
-    pytest.raises(
-        ValueError, uv_object.select, antenna_nums=ants_to_keep, antenna_names=ant_names
-    )
+    with pytest.raises(ValueError, match=r"Antenna number \[29 30\] is not present"):
+        uv_object.select(antenna_nums=np.max(unique_ants) + np.arange(1, 3))
+
+    # check the same thing with passing a list
+    with pytest.raises(ValueError, match=r"Antenna number \[29 30\] is not present"):
+        uv_object.select(antenna_nums=(np.max(unique_ants) + np.arange(1, 3)).tolist())
+
+    with pytest.raises(
+        ValueError, match="Antenna name test1 is not present in the antenna_names"
+    ):
+        uv_object.select(antenna_names="test1")
+    with pytest.raises(
+        ValueError, match="Only one of antenna_nums and antenna_names can be provided."
+    ):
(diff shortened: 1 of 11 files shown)


bhazelton:
@jsdillon This PR fixes a mistake we were making in antenna number handling in uvfits files and baseline number calculations. It breaks one test in hera_cal because the antenna numbers are different by one. I'll make a PR for hera_cal to fix the test, but I wanted to give you a heads up.

pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -4762,12 +4817,18 @@ def test_key2inds(casa_uvfits):
     assert np.array_equal(np.array([1]), indp[0])
 
     # Test invalid keys
-    pytest.raises(KeyError, uv._key2inds, "I")  # pol str not in data
-    pytest.raises(KeyError, uv._key2inds, -8)  # pol num not in data
-    pytest.raises(KeyError, uv._key2inds, 6)  # bl num not in data
-    pytest.raises(KeyError, uv._key2inds, (1, 1))  # ant pair not in data
-    pytest.raises(KeyError, uv._key2inds, (1, 1, "rr"))  # ant pair not in data
-    pytest.raises(KeyError, uv._key2inds, (0, 1, "xx"))  # pol not in data
+    with pytest.raises(KeyError, match="Polarization I not found in data."):
```
plaplant, line 4820: Thanks for adding a `match` to all these tests!

plaplant (dismissed):
Thanks so much @PyxieLouStar and @bhazelton! Lots of great changes to make this self-consistent.

plaplant (approved):

Issue #1107, opened 2021-12-10
Fix `__add__` method in UVCal in the case of overlapping and differently sorted data


bhazelton:
see #1102 and #1105 for reference

PR #1152, opened 2022-03-12
Add methods to enable sorting along the various axes in UVCal

bhazelton:
## Description

- Add `reorder_antennas`, `reorder_freqs`, `reorder_times` and `reorder_jones` methods to UVCal to allow for sorting along all the axes of UVCal objects.
- Change the meaning of passing an integer array to the `spw_order` of `UVData.reorder_freqs` to be an index array rather than an array of spw numbers giving the final desired order of spws. This is makes it similar to all the other reorder methods, but it is a breaking change. When we discussed this, we felt that almost no-one was using this yet, but we wanted to check with @kartographer about this belief. I can add some backwards compatibility if needed.

These were based on the following discussion and notes:
**There is a question I'd like to discuss**: Should the integer arrays passed to the various `order` parameters always be index arrays, or should they be arrays specifying the final desired order if the thing being sorted has an integer representation? There's conflicting behavior in the UVData methods on this front -- it's treated as an index array for everything except spw_order. 

We discussed this on 3/17 and leaned towards making them all be index arrays, but didn't have a quorum. We decided on 3/31 that they should all be index arrays.

## Motivation and Context

The similar methods on UVData are broadly useful and these methods enable testing for the bug described in #1107.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ All notable changes to this project will be documented in this file.
 and `reorder_jones`.
 
 ## Changed
+- The `spw_order` parameter for the `UVData.reorder_freqs` method now accepts an index
+array rather than an array of spw numbers, making it match the other reorder methods.
 - Updated the astropy requirement to >= 5.0.4
 - Dropped support for python 3.7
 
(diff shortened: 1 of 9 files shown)


plaplant (approved):
Thanks @bhazelton! I think it's a great idea to pull the frequency reorder out to a general utility function. And it's nice we can now sort UVCal objects the same way we can UVData objects!

PR #1162, opened 2022-04-01
Call methods to properly set up object in `check`

### LINKED ISSUE #1160: UVBeam set-up fails to set up object without explicitly calling `_set_efield()`
When setting up UVBeam object from scratch, setting E-field mode by calling
```
uvb.beam_type = 'efield'
```
fails to create object correctly even if `check()` passes as evidence by trying to dump it to fits and then reading it back in.
Instead, this works:
```
uvb._set_efield()
uvb.pixel_coordinate_system = "az_za"
uvb._set_cs_params()
```
But this is i) not obvious and ii) calls internal functions which is not kosher. 
Full script: 
https://github.com/lusee-night/luseepy/blob/sky_models/simulations/beams/scripts/make_beam.py

bhazelton:
## Description

- Call the `_set_efield`/`_set_power` and `_set_simple`/`_set_phased_array` methods in the `check` method to make sure all the attributes are properly setup on the object.
- Fix some bugs exposed by these additions
- Add a check that power beams that should be real are real (somewhat similar to the check on autos being real in UVData)
- Also fix a bug in the UVData auto checking where `'pQ'` was listed twice and `'pU'` wasn't listed.

## Motivation and Context

This makes it easier for users to fill in an object from scratch. This is more important for UVBeam because there are no really standard file formats.

Fixes #1160 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/uvbeam/uvbeam.py
+++ b/pyuvdata/uvbeam/uvbeam.py
@@ -726,7 +726,39 @@ class UVBeam(UVBase):
         self._gain_array.required = True
         self._coupling_matrix.required = True
 
-    def check(self, check_extra=True, run_check_acceptability=True):
+    def _fix_auto_power(self):
+        """Remove imaginary component of auto polarization power beams."""
+        if self.beam_type != "power" or self.polarization_array is None:
+            warnings.warn(
+                "Cannot use _fix_autos if beam_type is not 'power', or "
+                "polarization_array is None. Leaving data_array untouched."
+            )
+            return
+
+        auto_pol_list = ["xx", "yy", "rr", "ll", "pI", "pQ", "pU", "pV"]
+        pol_screen = np.array(
+            [
+                uvutils.POL_NUM2STR_DICT[pol] in auto_pol_list
+                for pol in self.polarization_array
+            ]
+        )
+
+        # Set any auto pol beams to be real-only by taking the absolute value
+        if np.all(pol_screen):
+            # If we only have auto pol beams the data_array should be float not complex
+            self.data_array = np.abs(self.data_array)
+        elif np.any(pol_screen):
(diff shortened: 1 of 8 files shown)


pyuvdata/uvbeam/uvbeam.py
```diff
@@ -2471,6 +2582,15 @@ def __add__(
...
             )[0]
+
+            if len(pnew_inds) > 0:
+                # if this does not have cross pols but other does promote to complex
+                cross_pols = [-3, -4, -7, -8]
+                if (
+                    np.intersect1d(other.polarization_array, cross_pols).size > 0
+                    and np.intersect1d(initial_pol_array, cross_pols).size == 0
+                ):
+                    this.data_array = np.asarray(this.data_array, dtype=complex)
```
plaplant, line 2593: In testing this on my system, the datatype promotion with `dtype=complex` always resolved to `complex128`, even when the input was `float32`. This would effectively be a quadrupling of the required memory size, which may not be insignificant for large arrays. If we don't think it'll be an issue, we can keep it as is, but we may want to consider matching the data type better if we're expecting users to have single-precision arrays.

plaplant (commented):
Thanks @bhazelton! I just have a few small questions about how the datatype conversions between real and complex happen.

pyuvdata/uvbeam/uvbeam.py
bhazelton: If it's np.float32, promote to np.complex64, otherwise do this.

bhazelton:
@plaplant I think I addressed your comments on this PR, can you re-review when you have a moment?

plaplant (approved):
Thanks for addressing my comments @bhazelton! This looks good to go.

PR #1166, opened 2022-04-22
Improving tolerance to flaky JPL-Horizons tests

kartographer:
Enables pytest to skip certain tests related to JPL-Horizons (caused by failures external to `pyuvdata`). 

## Description

Added code to skip several tests when connection issues (either SSL or just "no route to host" related) prevent querying the JPL Horizons service, which is used via the `astroquery` package for grabbing ephemerides for Solar System bodies.

## Motivation and Context

An infrequent-but-annoyingly-not-rare issue that has popped up in CI testing has been SSL errors that have caused failures in a total of 4 tests. These failures appear stocastic, and unrelated to anything inside of `pyuvdata`, but instead are probably server-side issues either w/ CI services or with JPL Horizons itself. These tests should _mostly_ work, so I've added a bit of wrapping around these tests for force `pytest` to skip them when either an `SSLError` (which is what we were seeing in the test reports) or a `RequestException` (more general connection error that is raised when I locally block any connection to JPL-H) is raised. Since neither of these should come from anything but connection issues, and so I think there's little danger of a problem getting missed by this workaround. In `pytest` reports, these particular failures will show up as 's' -- I think that's reasonanly visable (and of course should be naturally recorded as part of the unit test execution).

## Types of changes

## Checklist:

### DIFF
--- a/pyuvdata/tests/test_utils.py
+++ b/pyuvdata/tests/test_utils.py
@@ -17,7 +17,8 @@ from pyuvdata import UVData, UVFlag, UVCal
 import pyuvdata.utils as uvutils
 import pyuvdata.tests as uvtest
 from pyuvdata.data import DATA_PATH
-
+from ssl import SSLError
+from requests import RequestException
 
 ref_latlonalt = (-26.7 * np.pi / 180.0, 116.7 * np.pi / 180.0, 377.8)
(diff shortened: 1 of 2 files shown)


bhazelton (approved):
This looks great, thank you @kartographer!

PR #1168, opened 2022-05-10
Report the max imaginary part when complex autos are detected

bhazelton:
## Description

Update the warning/error message about complex auto correlations to report the maximum absolute value of the imaginary part so the user can tell if it's significant or likely a rounding error.

## Motivation and Context

The current message doesn't give the needed context to know how big a problem this is.

## Types of changes

- [ ] Version change

## Checklist:

- [ ] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/.github/CONTRIBUTING.md).

### DIFF
--- a/pyuvdata/uvdata/uvdata.py
+++ b/pyuvdata/uvdata/uvdata.py
@@ -2845,15 +2845,29 @@ class UVData(UVBase):
                     else:
                         auto_imag = np.any(auto_imag[:, :, :, pol_screen])
                 if auto_imag:
+                    if np.all(pol_screen):
+                        max_imag = np.max(np.abs(np.imag(self.data_array[autos])))
+                    else:
+                        auto_data = self.data_array[autos]
+                        if self.future_array_shapes:
+                            max_imag = np.max(
+                                np.abs(np.imag(auto_data[:, :, pol_screen]))
+                            )
+                        else:
+                            max_imag = np.max(
+                                np.abs(np.imag(auto_data[:, :, :, pol_screen]))
+                            )
                     if fix_autos:
                         warnings.warn(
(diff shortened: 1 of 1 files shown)


pyuvdata/uvdata/uvdata.py
```diff
@@ -2845,15 +2845,29 @@ def check(
...
+                    if np.all(pol_screen):
+                        max_imag = np.max(np.abs(np.imag(self.data_array[autos])))
+                    else:
+                        auto_data = self.data_array[autos]
+                        if self.future_array_shapes:
+                            max_imag = np.max(
+                                np.abs(np.imag(auto_data[:, :, pol_screen]))
+                            )
+                        else:
+                            max_imag = np.max(
```
kartographer, line 2857: I do like the idea of having some information here about the magnitude of the imaginary component after the "fast" detection, but I wonder if this would be more helpful with some information about the _relative_ magnitude of the imaginary component rather than the absolute magnitude, since it's not uncommon for the auto-correlations to not be scaled to a "normal" flux density scale like Jy. You could imagine this might look something like:
```suggestion
                            temp_data = auto_data[:, :, pol_screen]
                            temp_data = temp_data[temp_data.imag != 0]
                            max_imag = np.max(temp_data.real / temp_data.imag]
```

bhazelton: Sure, I think maybe I'll have it give both the absolute value and the ratio because the absolute value can be of interest for understanding if it's a rounding error level problem. I'd probably be more inclined to report the imaginary/real ratio than the real/imaginary ratio -- do you have a strong opinion?

kartographer (approved):
Looks good, thanks @bhazelton!

PR #1171, opened 2022-05-26
Streamlining the `select` method in `UVData`

### LINKED ISSUE #1130: Wrapping bug in uvd.print_phase_center_info
Where you can get RAs of 350 hours.

### LINKED ISSUE #1159: Parameter scan_number_array is not updated during select
I am reading a ms file and then attempting to select the autocorrelations only with `uvd.select(ant_str="auto", run_check=False)`. Both before and after the select the parameter `scan_number_array` has length equal to the total number of baselines. The select therefore fails if I set `run_check=True`. I can avoid the issue by setting `uvd.scan_number_array=None` initially.

kartographer:
Simplifies the code in `UVData.select` to make it easier to maintain.

## Description

The code in `select` has been generalized such that it no longer operates on individual `UVData` parameters, but instead will look at the attributes of these parameters, namely `UVParameter.form` to decide which selection criteria to apply. The method `UVData. _select_metadata` has been replaced with `UVData._select_by_index`, which will handle both metadata-only and "regular" `UVData` objects, and acts as a drop-in replacement where it is called elsewhere in the code.

Also, two other minor changes: a wrapping bug in `UVData.print_phase_center_ info` has been fixed, and some superfluous code in `_check_flex_spw_contiguous` has been removed (since the function raises an error if there's a problem, there did not seem to be a need for it to return a value if no error was raised).

## Motivation and Context

The changes to `select` where primarily motivated by #1159, which was caused by the addition of a new parameter to `UVData` (namely `_scan_number_array`), which was added in part to help support of MS dataset writing. In the original PR that introduced it, nothing was added to `_select_metadata` that told it to explicitly filter on that parameter. Given that there may be other changes/additions to parameters in the not to distant future, this seemed like a good opportunity to "future-proof" the select command, by using `UVParameter.form` to figure out which -- if any --  axes of a given array of values needed to be selected, without needing to explicitly code for individual parameter names. 

Closes #1159.
Closes #1130.

## Types of changes

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/uvdata/tests/test_uvdata.py
+++ b/pyuvdata/uvdata/tests/test_uvdata.py
@@ -2459,6 +2459,9 @@ def test_select(casa_uvfits, future_shapes):
     # now test selecting along all axes at once
     uv_object = casa_uvfits
 
+    # Set the scan numbers so that we can check to make sure they are selected correctly
+    uv_object._set_scan_numbers()
+
     if future_shapes:
         uv_object.use_future_array_shapes()
 
@@ -6691,6 +6694,10 @@ def test_redundancy_contract_expand_nblts_not_nbls_times_ntimes(method, casa_uvf
         with uvtest.check_warnings(
             UserWarning,
             [
+                "The uvw_array does not match the expected values given the antenna "
+                "positions."
+            ]
+            + [
                 "Index baseline in the redundant group does not have all the "
                 "times, compressed object will be missing those times."
             ]
@@ -9831,8 +9838,8 @@ def test_print_object_full(sma_mir):
     _ = sma_mir._add_phase_center(
         "3c84",
         cat_type="sidereal",
-        cat_lat=sma_mir.phase_center_catalog["3c84"]["cat_lat"],
-        cat_lon=sma_mir.phase_center_catalog["3c84"]["cat_lon"],
+        cat_lat=-1.0,
+        cat_lon=-1.0,
         cat_dist=0.0,
         cat_vrad=0.0,
         cat_pm_ra=0.0,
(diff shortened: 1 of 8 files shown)


pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -6691,6 +6694,10 @@ def test_redundancy_contract_expand_nblts_not_nbls_times_ntimes(method, casa_uvf
         with uvtest.check_warnings(
             UserWarning,
             [
+                "The uvw_array does not match the expected values given the antenna "
+                "positions."
+            ]
```
bhazelton, line 6699: If the problem was that there are now two of these instead of one, you could just multiply the list that matches this below by 2.

pyuvdata/uvdata/uvdata.py
```diff
@@ -8530,68 +8526,80 @@ def _select_metadata(
...
+                        continue
 
-        self.history = self.history + history_update_string
+                    if isinstance(attr.value, np.ndarray):
+                        # If we're working with an ndarray, use take to slice along
+                        # the axis that we want to grab from.
+                        attr.value = attr.value.take(ind_arr, axis=sel_axis)
+                    elif isinstance(attr.value, list):
+                        # If this is a list, it _should_ always have 1-dimension.
+                        assert sel_axis == 0
```
bhazelton, line 8563: Let's add an error message here tells the user this is a bug and they should make an issue.

bhazelton (commented):
This is very nice, thanks @kartographer !

pyuvdata/uvdata/tests/test_uvdata.py
kartographer: Ahh - for some reason I thought it was order sensitive. Fixed in 5279a6f.

pyuvdata/uvdata/uvdata.py
kartographer: Added in 5279a6f.

bhazelton (dismissed):
Looks good!

bhazelton (approved):

PR #1142, opened 2022-02-11
Improve uvh5 storage of new phasing info

bhazelton:
## Description

- Always write out uvh5 version 1.0 or greater files (with future array shapes) with `UVData.write_uvh5`.
- Improve the way the new phasing metadata is stored in uvh5 files
- Define a new uvh5 version (1.1) that uses the future phasing info (the phase_center_catalog) rather than the older phasing info.
- update the uvh5 memo with the new versioning information
- change "unphased" to "unprojected" throughout (with deprecation)
- change the way the phase_center_catalog is keyed (from name to ID number). This allows for mosaic observations where different phase centers have the same name (common in CASA and Miriad and MIR). 

TODO: add pdf to PR once the tex file is settled on.

## Motivation and Context

The current way to store the new phasing info is kludgy and undocumented. We intend to deprecate the old style phasing info, so need good support for the new style.
closes #1137

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation change (documentation changes only)
- [ ] Version change
- [ ] Build or continuous integration change

## Checklist:

- [x] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/.github/CONTRIBUTING.md).
- [x] My code follows the code style of this project.

New feature checklist:
- [x] I have added or updated the docstrings associated with my feature using the [numpy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ ] I have updated the tutorial to highlight my new feature (if appropriate).
- [x] I have added tests to cover my new feature.
- [x] All new and existing tests pass.
- [x] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/CHANGELOG.md).

PR #1172, opened 2022-05-27
Adding a warning to `fix_phase`

### LINKED ISSUE #1136: Make `fix_phase` result in a warning when phases are "fixed"
As part of closing out #929, a method was added called `fix_phase`, whose purpose was to fix data that was projected using the older (and less accurate) version of `phase`. In order to make that change seamless, `fix_phase` was incorporated into the readers for a few objects to make this fix happen automatically by default (though controlled w/ the switch `fix_old_phase`).  However, this fix is applied silently, which can result in some mysterious behaviors when dealing with data. The proposed solution would then be to add a `UserWarning`,  when this fix happens, so that folks aren't caught unaware that something about the data is being changed (even if we might want it changed).

kartographer:
Adds a warning to `fix_phase`.

## Description

Two new warning messages have been added for when users call the `fix_phase` method -- one of which will be shown based on user-provided arguments.

## Motivation and Context

This warning is intended to alert users when `fix_phase` is being called, particularly as part of other method calls like `read`. The severity of the warning depends on what arguments are supplied by the user.

Closes #1136.
## Types of changes

N.b., this isn't exactly a _bug fix_, per se, but it seems a better choice than my second best fit "documentation change".

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ real-only if non-zero imaginary components are detected.
 and `reorder_jones`.
 
 ### Changed
+- `UVData.fix_phase` now raises a warning when called.
 - Changed `UVData.write_uvfits` to allow for one to write out datasets in UVFITS format
 without "spoofing" (via setting `spoof_nonessential=True`) UVFITS-specific values.
 - Methods that ensure that attributes are set up properly on `UVBeam` are now called in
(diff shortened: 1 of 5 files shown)


kartographer:
> I think I'd make this change on #1142, there's a bunch of changes to the code that's being changed here on that branch. I'm happy to do it, but I'll wait until you (@kartographer) have had a chance to look at that branch to make sure the changes there make sense.

Thanks @bhazelton -- I'll take a look at it soon, although I _think_ the changes in this branch are orthogonal. In #1142, there's a warning if it looks like the data were phased with the old method but `fix_phase` is not called, whereas this branch raises a warning when `fix_phase` _is_ called. Both deserve their own warnings, I think :)

bhazelton:
Ok, then it may make sense to do it separately. It looked to me on a quick glance like you were modifying things that are just deleted in that branch, but I might be wrong.

bhazelton (approved):

PR #1175, opened 2022-05-31
Fixing add for flex-spw datasets

kartographer:
Fixes a bug when adding flex-spw datasets.

## Description

Changes `UVData.__add__` to account for spectral window number in flex-spw objects when deciding whether or not frequency channels between objects overlap.

## Motivation and Context

Fixes an issue where the `__add__` method would not correctly combine `UVData` objects. When `flex_spw=True` (which is the case for all multi-spectral window datasets), different spectral windows are allowed to have overlapping frequency values within a single object. The aforementioned bug caused a crash when using `__add__` where two such flex-spw objects had overlapping frequencies in different spectral windows.

## Types of changes

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ array rather than an array of spw numbers, making it match the other reorder met
 - Dropped support for python 3.7
 
 ### Fixed
+- A bug in `UVData.__add__` where flexible spectral window datasets were not combined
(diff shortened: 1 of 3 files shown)


bhazelton (dismissed):
Looks good, thanks!!

bhazelton (approved):

PR #1177, opened 2022-06-01
"Spoofless" writing of UVFITS files

### LINKED ISSUE #1068: Better handling for DUT1, GST0, RDATE, EARTH_OMEGA and TIMESYS for uvfits
These are required to write uvfits files. There's currently an approach which requires setting `spoof_nonessential=True` in `write_uvfits`, but I wonder if we can actually just calculate these using astropy so that users don't have to worry about it.

kartographer:
Adds the ability to write UVFITS files without needing to "spoof" values.

## Description

Added some code to automatically calculate several of the optional `UVParameters` when writing to UVFITS format. Additionally, "antenna_positions" has been removed from `UVFITS.uvfits_required_extra` since it is now a required paramter for `UVData` (and is caught by `check` if not set).

## Motivation and Context

Removes the requirement to set `spoof_nonessential=True` when calling `UVData.write_uvfits`, which while not a major problem, is a minor annoyance that pops up every time one needs to write out to the UVFITS format. Note that I've kept the keyword and behavior in place for now in order to ensure backwards compatibility, although this too could be removed at some point in the future (as a breaking change).

Closes #1068.

## Types of changes

## Checklist:

New feature checklist:


bhazelton:
This is awesome, thank you!

I think we should add a deprecation warning if `spoof_nonessential` is set to let people know it's no longer required. And I would lean towards making it do nothing if it is set -- basically to always set them to the right values rather than the spoofed values. It's true that setting them to the right values instead of the spoofed ones is a change, but it's a change towards better accuracy.

bhazelton:
This is awesome, thank you!

I think we should add a deprecation warning if `spoof_nonessential` is set to let people know it's no longer required. And I would lean towards making it do nothing if it is set -- basically to always set them to the right values rather than the spoofed values. It's true that setting them to the right values instead of the spoofed ones is a change, but it's a change towards better accuracy.

pyuvdata/uvdata/uvfits.py
```diff
@@ -954,6 +937,15 @@ def write_uvfits(
                 fix_autos=fix_autos,
             )
 
+        if spoof_nonessential:
+            warnings.warn(
+                (
+                    "The spoof_nonessential parameter is deprecated and will be "
+                    "removed in a future release."
```
bhazelton, line 944: Maybe add something to let users know that all the uvfits metadata are now properly defined and don't need to be spoofed?

kartographer: Done in 587adde5.

bhazelton (approved):
Looks great, thanks!

PR #1180, opened 2022-06-02
Add Azure pipelines CI back

bhazelton:
## Description

Add the Azure pipelines CI for external tests back.

The main problem that was happening was that we listed both condo-forge and defaults in the channels. For some reason, conda was installing gcc from defaults instead of condo-forge and we then ended up with a conflict between the installed gcc and the version that was needed for scipy (which was installed from conda-forge).

This PR removes "defaults" as a channel from all our conda yamls to fix this problem. It also makes other changes to the old Azure Pipelines yaml to use mamba, following what is done in the circleci yaml. And it uses the conda yamls provided by the external repositories rather than ones we maintain.

I think we should probably remove the external tests that we are now running through github actions, we don't really need to do them twice.

## Motivation and Context

This keeps around diverse CI options with the hope that we can get them back in in the future.

## Types of changes

- [ ] Version change

## Checklist:


ci/azure-piplines.yml
```diff
@@ -0,0 +1,86 @@
...
+  variables:
+    PYTHON: "3.9"
+    ENV_NAME: hera_cal_tests
+    WITH_SUDO: true
+
+  steps:
+  - bash: |
+      cd ../
+      git clone https://github.com/HERA-Team/hera_cal.git
+      cd s
```
mkolopanis, line 14: Is the directory really named `s` on azure for the current project?

bhazelton: yes! I assumed it was going to be pyuvdata, but that errored so I dug up the documentation and it's `s`. Totally crazy.

PR #1182, opened 2022-06-06
change setup-miniconda to use mamba instead

mkolopanis:
use mamba in github Action building
and circleci (this was _way_ too complicate to enable)


.circleci/config.yml
```diff
@@ -22,30 +25,37 @@ jobs:
       - restore_cache:
           key: deps-{{ .Branch }}-{{ checksum "ci/pyuvdata_tests.yml" }}
       - run:
-          name: Install miniconda
-          command: |
-            wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O $HOME/miniconda.sh;
-            bash $HOME/miniconda.sh -b -p $HOME/miniconda
+          name: Instal Micromamba
```
bhazelton, line 28: Here and elsewhere in this file: "Instal" -> "Install"

.circleci/config.yml
```diff
@@ -22,30 +25,37 @@ jobs:
...
+          command: wget -qO- https://micro.mamba.pm/api/micromamba/linux-64/latest | sudo tar -C / -xvj bin/micromamba
       - run:
           name: configure conda environment
           command: |
-            export PATH="$HOME/miniconda/bin:$PATH"
-            ./ci/install_conda.sh
+            micromamba info
+            # need these to add gxx and gcc to build novas and cython
+            micromamba create --name=${ENV_NAME}  python=$PYTHON gxx gcc -f ci/${ENV_NAME}.yml -yq
```
bhazelton, line 35: Does this mean we aren't using "install_conda.sh" in CircleCI any more?

ci/install_conda.sh
```diff
@@ -1,24 +1,11 @@
...
-conda create --name=${ENV_NAME}  python=$PYTHON --quiet
-conda env update -n ${ENV_NAME} -f ci/${ENV_NAME}.yml
-source activate ${ENV_NAME}
-
-conda list -n ${ENV_NAME}
+micromamba info
+# need these to add gxx and gcc to build novas and cython
+micromamba create --name=${ENV_NAME}  python=$PYTHON gxx gcc -f ci/${ENV_NAME}.yml -yq
+micromamba activate ${ENV_NAME}
+micromamba list -n ${ENV_NAME}
```
bhazelton, line 8: Are we using this anymore?

bhazelton (commented):
Looks good. I found a typo and I have a question about the "install conda" script.

.circleci/config.yml
mkolopanis: That's correct we're not. I can put some of this back in there if we'd like but ended up pulling it out while trying to debug.

mkolopanis:
I've deleted the `install_conda.sh` script, if we end up bringing back azure, I can deal with re-implementing the logic there.

bhazelton (approved):
Looks good, thank you!

PR #1191, opened 2022-07-12
Enable normalizing cross-correlations by autos

kartographer:
## Description

Adds a new method to `UVData` called `normalize_by_autos`. 

## Motivation and Context

This adds the ability to "normalize" the cross-correlations by the geometric mean of the autos for each baseline, which allows one to convert between arbitarily scaled data and correlation coefficients (the latter of which is used by SMA). Correlation coefficients are sometimes used when storing visibilities, as if you have information about the system temperature and the forward gain of your antenna (typically in Jy/K), you can make an easy first-order conversion of the data into flux density (Jy) units.

No issue to link to, although it was an informally requested feature for both ATA and SMA (in the latter case, for potentially undoing this normalization inside the `UVData` object).

## Types of changes

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- Added the `normalize_by_autos` method to `UVData`, which allows one to normalize
+cross-correlations (i.e., covert values from arbitrary scale to correlation coefficients)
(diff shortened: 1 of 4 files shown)


bhazelton (commented):
Looks good, pretty straight forward. We should definitely add something to the tutorial so users can know it exists.

bhazelton (approved):

PR #1197, opened 2022-08-02
Add support for antenna numbers > 2047

### LINKED ISSUE #1193: Add support for antenna numbers greater than 2047
The MWA has tile id numbers ranging from 0 to 32767, inclusive. Numbers greater than 2047 are not currently supported in pyuvdata.

PyxieLouStar:
## Description
Update baseline-to-antnum and antnum-to-baseline calculations to handle antenna numbers > 2047.

## Motivation and Context

Currently antenna numbers > 2047 are not supported. In order to handle these numbers, a new calculation for baseline numbers is required. Additionally, the corresponding conversion from baseline numbers back to antenna numbers is necessary.

Closes #1193 
## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -791,9 +791,10 @@ def antnums_to_baseline(ant1, ant2, Nants_telescope, attempt256=False):
         number of antennas
     attempt256 : bool
         Option to try to use the older 256 standard used in
-        many uvfits files. If there are more than 256 antennas, the 2048
-        standard will be used unless there are more than 2048 antennas.
-        In that case, the 2147483648 standard will be used. Default is False.
+        many uvfits files. If there are antenna numbers >= 256, the 2048
+        standard will be used unless there are antenna numbers >= 2048
+        or Nants_telescope > 2048. In that case, the 2147483648 standard
+        will be used. Default is False.
 
     Returns
     -------
@@ -817,11 +818,16 @@ def antnums_to_baseline(ant1, ant2, Nants_telescope, attempt256=False):
             "with antenna numbers less than zero."
         )
 
+    nants_less2048 = True
+    if Nants_telescope is not None and Nants_telescope > 2048:
+        nants_less2048 = False
+
     return_array = isinstance(ant1, (np.ndarray, list, tuple))
     baseline = _utils.antnums_to_baseline(
         np.ascontiguousarray(ant1, dtype=np.int64),
         np.ascontiguousarray(ant2, dtype=np.int64),
         attempt256=attempt256,
+        nants_less2048=nants_less2048,
     )
     if return_array:
         return baseline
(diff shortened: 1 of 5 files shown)


pyuvdata/utils.py
```diff
@@ -791,24 +791,25 @@ def antnums_to_baseline(ant1, ant2, Nants_telescope, attempt256=False):
         number of antennas
     attempt256 : bool
         Option to try to use the older 256 standard used in
-        many uvfits files (will use 2048 standard if there are more
-        than 256 antennas). Default is False.
+        many uvfits files. If there are more than 256 antennas, the 2048
+        standard will be used unless there are more than 2048 antennas.
+        In that case, the 2147483648 standard will be used. Default is False.
```
bhazelton, line 796: I think this is actually triggered by antenna numbers higher than 255/2047, not by more than 256/2048 antennas.

pyuvdata/utils.py
```diff
@@ -791,24 +791,25 @@ def antnums_to_baseline(ant1, ant2, Nants_telescope, attempt256=False):
...
 
     Returns
     -------
     int or array of int
         baseline number corresponding to the two antenna numbers.
 
     """
-    if Nants_telescope is not None and Nants_telescope > 2048:
+    if Nants_telescope is not None and Nants_telescope > 2147483648:
```
bhazelton, line 804: hmm, do we need this test given that we have the next one (for antenna number)? Maybe we do in case that we are only passing some of the antennas. But if that's what we are worried about then I think we should update the code below to use the largest standard if Nants>2048 even if the antenna numbers that are passed are smaller than the limit.

pyuvdata/utils.pyx
```diff
@@ -162,18 +199,29 @@ cpdef numpy.ndarray[dtype=numpy.int64_t] antnums_to_baseline(
     if less256:
       _antnum_to_bl_256(ant1, ant2, _bl, nbls)
 
+    elif less2048:
+        message = (
+          "antnums_to_baseline: found antenna numbers > 255, using "
+          "2048 baseline indexing. Beware compatibility "
+          "with CASA etc"
+        )
```
bhazelton, line 207: I don't think we need this message or the similar one below anymore. To get to CASA, users can use the `write_ms` method and uvfits and other formats have support for more than 256 antennas.

bhazelton (commented):
Looks good, I just have a few questions.

pyuvdata/utils.pyx
bhazelton: unless this warning is issued because the user set the `attempt256` flag.

pyuvdata/utils.py
PyxieLouStar: I think that makes sense. I've updated the code to reflect we are checking both Nants>2048 and antenna numbers > 2047, and updated the docstring as well.

bhazelton (approved):
Looks good!

PR #1199, opened 2022-08-10
Telescope frame

aelanman:
## Description

Introduces a new parameter `telescope_frame` to `UVData`, which can be either "ITRS" (default) or "MCMF" (Moon-Centered Moon-Fixed). This new parameter indicates the coordinate frame the `telescope_position` is defined in. Several utility functions and UVData methods are configured to use this parameter:
 - ENU_from_ECEF
 - ECEF_from_ENU
 - LatLonAlt_from_XYZ
 - XYZ_from_LatLonAlt
 - get_lst_for_time

The package `lunarsky` is needed to use the MCMF frame. It's been added as an optional dependency.

## Motivation and Context

`pyuvsim` relies on `pyuvdata` for computing uvw coordinates and correctly transforming antenna positions. While `pyradiosky` and `pyuvsim` both support `lunarsky` coordinate frame definitions, so far `pyuvdata` has not included this. Once this is merged, a corresponding small change to `pyuvsim` should fix https://github.com/RadioAstronomySoftwareGroup/pyuvsim/issues/408.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:


bhazelton:
I would think that the LST should be set to the RA at zenith.

aelanman:
@bhazelton My changes to the tests should cover the new cases, but I'm not sure how to change the circleci config to install `lunarsky` and run them.

pyuvdata/astropy_interface.py
```diff
@@ -0,0 +1,52 @@
...
+    LUNAR_RADIUS = MoonLocation._lunar_radius
+    hasmoon = True
+except ImportError:
+    from astropy.coordinates import SkyCoord
+    from astropy.time import Time
+
+    LUNAR_RADIUS = None
+    hasmoon = False
+
+    class MoonLocation:
```
jpober, line 35: Are these mock classes a standard way to handle optional dependencies?  I honestly don't know.

pyuvdata/astropy_interface.py
```diff
@@ -0,0 +1,52 @@
...
+
+The lunarsky classes are entirely compatible with normal astropy behavior, even
+if observing locations on the Moon are not used.
+"""
+
+try:
+    from lunarsky import MCMF, LunarTopo, MoonLocation, SkyCoord, Time
+
+    LUNAR_RADIUS = MoonLocation._lunar_radius
+    hasmoon = True
```
jpober, line 27: Any concern over this codecov warning?

pyuvdata/utils.py
```diff
@@ -1216,6 +1220,15 @@ def LatLonAlt_from_XYZ(xyz, check_acceptability=True):
         altitude, numpy array (if Npts > 1) or value (if Npts = 1) in meters
 
     """
+    frame = frame.upper()
+    if not hasmoon and frame == "MCMF":
+        raise ValueError("Need to install `lunarsky` package to work with MCMF frame.")
+
+    if frame == "ITRS":
+        accept_bounds = (6.35e6, 6.39e6)
+    elif frame == "MCMF":
```
jpober, line 1229: Maybe this codecov warning is worth addressing?

jpober (commented):
I've given things a first look and while it all makes sense to me, I'm not comfortable being the only reviewer before it gets merged.  Looks like a lot of code coverage warnings, some seemingly more significant than others.  But I'm certainly eager to see this done so that we can get pyuvsim working

pyuvdata/astropy_interface.py
aelanman: It seems that circleci isn't properly installing `lunarsky`, so the test cases with lunarsky aren't being run. We should not merge this until that's resolved.

bhazelton:
@aelanman sorry for the delay. I just added `lunarsky` to some of the test environments. I need to make some more updates for it but I needed to know if there was a minimum version of lunarsky to use? Also, it looks like you need to rebase the branch.

aelanman:
@bhazelton Thanks for fixing that! I've made some changes to cover the remaining missing lines, and rebased off `main`. For lunarsky, I would recommend a minimum version of `0.1.2`, which is the third release.

bhazelton:
@aelanman we discussed this in last week's telecon and we're broadly supportive but we don't really like the approach of replacing the astropy SkyCoord with the lunarsky one if lunarsky is installed. We'd prefer to import the lunarsky one as a different name and only use it if needed (i.e. if the telescope is on the moon). Similarly we'd like to avoid using the mock classes in case lunarsky is not installed. I'm happy to tackle these changes if you're too busy on other projects.

aelanman:
@bhazelton I've made some changes which ensure SkyCoord and Time are imported with separate names for astropy and lunarsky, and the lunarsky versions (LSkyCoord and LTime) are only used when a lunarsky-related class is used. It also removes the dummy classes. Fortunately, the changes here used lunarsky sparingly and mostly defined new XYZ/LatLonAlt transformations by using basic erfa functions. There was a slight hit in coverage; I'll have to fix that later.

The main reason for replacing the class globally is that it can be hard to predict where each form will be needed. The lunarsky classes inherit from their astropy equivalents, but just add support for the new `MoonLocation` class, etc. In principle, they should be identical (including passing tests of "isinstance"). However, it can be confusing if you look at the, say, "SkyCoord" instance and its class is `lunarsky.sky_coord.SkyCoord`.

pyuvdata/utils.py
```diff
@@ -3623,6 +3723,10 @@ def get_lst_for_time(
         Library used for running the LST calculations. Allowed options are 'erfa'
         (which uses the pyERFA), 'novas' (which uses the python-novas library),
         and 'astropy' (which uses the astropy utilities). Default is erfa.
+    frame : str
+        Frame of reference ofr latitude/longitude/altitude.
```
bhazelton, line 3727: ofr -> for (?)

bhazelton (commented):
I found one tiny typo, but otherwise I think this looks great! Thanks!!

pyuvdata/utils.py
aelanman: Fixed! Thanks for spotting this

PR #1205, opened 2022-08-25
Partial Reading for Beamfits

### LINKED ISSUE #1201: beamfits format partial read
It would be super if we could do partial reads, at least for beamfits. The use case I have in mind is our current attempt at H4C simulations, where the beam file has a whole bunch of frequencies in it, but we only simulate one frequency at a time. To run as many frequencies as possible on a cluster, we run each on one CPU via a slurm task, and we try to minimize the requested RAM. However, just reading in the beam file takes quite a bit of RAM because it reads in all the frequencies. To do the cubic spline interpolation down to a single frequency, we'd need at most I guess 4 frequencies, which would reduce RAM considerably. This is not possible currently, it seems.

steven-murray:
## Description

I added three parameters to the `read_beamfits` method: `freq_range`, `az_range` and `za_range`. Each should be a 2-tuple of floats specifying the range to read in. This is meant to reduce peak memory usage. It is enabled by the FITS memmap feature.

## Motivation and Context

Closes #1201

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,10 @@ All notable changes to this project will be documented in this file.
 ### Added
 - Support for partial reads on beamfits files. This can be done on frequency, azimuth
   or zenith angle (if beam is in az/za coordinates).
-
+- Logging statements (INFO level) in `UVdata.check`.
+### Changed
+- Performance improvement when doing a `select` using the `bls` parameter (~70%
+  improvement in the `_select_preprocess` function)
 ## [2.2.9] - 2022-8-23
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 - Support for partial reads on beamfits files. This can be done on frequency, azimuth
-  or zenith angle (if beam is in az/za coordinates). 
+  or zenith angle (if beam is in az/za coordinates).
 
 ## [2.2.9] - 2022-8-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file.
   or zenith angle (if beam is in az/za coordinates).
 - Logging statements (INFO level) in `UVdata.check`.
 ### Changed
-- Performance improvement when doing a `select` using the `bls` parameter (~70%
+- Performance improvement when doing a `UVData.select` using the `bls` parameter (~70%
   improvement in the `_select_preprocess` function)
 ## [2.2.9] - 2022-8-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- Support for partial reads on beamfits files. This can be done on frequency, azimuth
+  or zenith angle (if beam is in az/za coordinates). 
+
 ## [2.2.9] - 2022-8-23
 
 ### Added
--- a/docs/uvbeam_tutorial.rst
+++ b/docs/uvbeam_tutorial.rst
@@ -195,7 +195,9 @@ d) Writing a regularly gridded beam FITS file
 **********************************************
 When reading a beam FITS file, you also have the option of selecting frequencies and
 az/za values at the read step -- i.e. so that memory is never allocated for data outside
-these ranges. Use the ``freq_range`` and ``za_range`` parameters to achieve this.
+these ranges. Use the ``freq_range``, ``za_range`` and ``az_range`` parameters to
+achieve this. The ``freq_range`` parameter will be effective for both HEALpix beamfits
+files and az/za grid.
 
 .. code-block:: python
 
--- a/docs/uvbeam_tutorial.rst
+++ b/docs/uvbeam_tutorial.rst
@@ -193,6 +193,10 @@ c) Reading in the MWA full embedded element beam
 
 d) Writing a regularly gridded beam FITS file
 **********************************************
+When reading a beam FITS file, you also have the option of selecting frequencies and
+az/za values at the read step -- i.e. so that memory is never allocated for data outside
+these ranges. Use the ``freq_range`` and ``za_range`` parameters to achieve this.
+
 .. code-block:: python
 
   >>> import os
@@ -200,7 +204,7 @@ d) Writing a regularly gridded beam FITS file
   >>> from pyuvdata.data import DATA_PATH
   >>> beam = UVBeam()
   >>> settings_file = os.path.join(DATA_PATH, 'NicCSTbeams/NicCSTbeams.yaml')
-  >>> beam.read(settings_file, beam_type='power')
+  >>> beam.read(settings_file, beam_type='power', freq_range=(1e8, 1.5e8), za_range=(0, 90.0))
   >>> write_file = os.path.join('.', 'tutorial.fits')
   >>> beam.write_beamfits(write_file, clobber=True)
 
--- a/pyuvdata/uvbeam/beamfits.py
+++ b/pyuvdata/uvbeam/beamfits.py
@@ -56,7 +56,7 @@ class BeamFITS(UVBeam):
     formats because it needs to support multiple dimensions (e.g. polarization,
     frequency, efield vectors).
     """
-    @profile
+
     def read_beamfits(
         self,
         filename,
@@ -99,8 +99,9 @@ class BeamFITS(UVBeam):
             Default is to read in all azimuths. Restricting the azimuth reduces peak
             memory usage.
         za_range : tuple of float in deg
-            The zenith angle range to read in, if the beam is specified in za/za coordinates.
-            Default is to read in all za. Restricting the za reduces peak memory.
+            The zenith angle range to read in, if the beam is specified in za/za
+            coordinates. Default is to read in all za. Restricting the za reduces peak
+            memory.
         """
         # update filename attribute
         basename = os.path.basename(filename)
@@ -194,7 +195,7 @@ class BeamFITS(UVBeam):
                     raise ValueError(
                         'Units of first axis array are not "deg" or "rad".'
                     )
-                
+
                 if az_range is not None:
                     azmin = np.where(self.axis1_array >= np.deg2rad(az_range[0]))[0][0]
                     azmax = np.where(self.axis1_array <= np.deg2rad(az_range[1]))[0][-1]
@@ -202,7 +203,7 @@ class BeamFITS(UVBeam):
                     print(azmax, azmin)
                     az_mask = slice(azmin, azmax)
                     self.axis1_array = self.axis1_array[az_mask]
-                
+
                 axis2_units = primary_header.pop(
                     "CUNIT" + str(ax_nums["img_ax2"]), "deg"
                 )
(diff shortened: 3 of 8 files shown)


steven-murray:
@bhazelton and @mkolopanis : this is essentially working (I will write some tests soon). My non-unit-test way of testing this was to run the following script:

```python
import os
import numpy as np
from pyuvdata import UVBeam
from pyuvdata.data import DATA_PATH


filenames = ["HERA_NicCST_150MHz.txt", "HERA_NicCST_123MHz.txt"]
cst_folder = "NicCSTbeams"
cst_files = [os.path.join(DATA_PATH, cst_folder, f) for f in filenames] * 5

def make_cst_beam(beam_type):
    """Make the default CST testing beam."""
    extra_keywords = {
        "software": "CST 2016",
        "sim_type": "E-farfield",
        "layout": "1 antenna",
        "port_num": 1,
    }

    beam = UVBeam()
    beam.read_cst_beam(
        cst_files,
        beam_type=beam_type,
        frequency=np.linspace(120e6, 170e6, 10),
        telescope_name="HERA",
        feed_name="Dipole",
        feed_version="1.0",
        feed_pol=["x"],
        model_name="Dipole - Rigging height 4.9 m",
        model_version="1.0",
        x_orientation="east",
        reference_impedance=100,
        history=(
            "Derived from https://github.com/Nicolas-Fagnoni/Simulations."
            "\nOnly 2 files included to keep test data volume low."
        ),
        extra_keywords=extra_keywords,
    )
    return beam


if __name__=="__main__":
    from pathlib import Path

    fname = Path("tempbeam.fits")

    if not fname.exists():
        beam = make_cst_beam('efield')
        beam.write_beamfits("tempbeam.fits")
        del beam
    
    uvb= UVBeam()
    uvb.read_beamfits('tempbeam.fits',  za_range=(0, 95), freq_range=(120e6, 155e6))
```

(Notice that the `make_cst_beam` function is just out of the `conftest.py`). Then I switched the last line to pass different combinations of `za_range` and `freq_range`. I ran the script through `memory-profiler`. I'm attaching the (redacted) outputs:

[freq_selector.txt](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/9428518/freq_selector.txt)
[full_selector.txt](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/9428519/full_selector.txt)
[no_selector.txt](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/9428520/no_selector.txt)
[za_selector.txt](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/9428521/za_selector.txt)



The things to notice are:

1. If you specify the ranges such that everything gets read in, there's no difference to the memory compared to not specifying any ranges.
2. The reason the downselect statements don't take any memory is because `data` is a memmap array, and we're doing simple indexing. This relies on the regular-spaced frequencies/az/za that beamfits requires.
3. There is something a little weird (not particular to this PR) -- for the full array, the size should be about 39 MB, however it seems to be using about double that memory when creating the complex array. I presume this is because its loading the memory into `data`, and *then* creating the complex array, then deleting `data`. So the peak memory is ~twice the original data. I feel sure this is something that can be gotten around...

steven-murray:
@mkolopanis and @bhazelton I believe this PR is ready.

docs/uvbeam_tutorial.rst
```diff
@@ -193,14 +193,18 @@ c) Reading in the MWA full embedded element beam
 
 d) Writing a regularly gridded beam FITS file
 **********************************************
+When reading a beam FITS file, you also have the option of selecting frequencies and
+az/za values at the read step -- i.e. so that memory is never allocated for data outside
+these ranges. Use the ``freq_range`` and ``za_range`` parameters to achieve this.
```
bhazelton, line 198: maybe also list ``az_range`` here? And I think the ``freq_range`` keyword will also work for HEALPix beam fits files, right?

pyuvdata/uvbeam/beamfits.py
```diff
@@ -193,8 +215,87 @@ def read_beamfits(
...
+                fmin = np.where(self.freq_array[0] >= freq_range[0])[0][0]
+                fmax = np.where(self.freq_array[0] <= freq_range[1])[0][-1] + 1
+                freq_mask = slice(fmin, fmax)
+                self.Nfreqs = fmax - fmin
+                self.freq_array = self.freq_array[:, freq_mask]
+
+            if az_range is not None:
+                data = data[..., az_mask]
+            if za_range is not None:
+                data = data[..., za_mask, :]
```
bhazelton, line 295: I can't quite tell the context here, will this run for healpix files? If so, we need to error if these keywords are set for healpix files.

pyuvdata/uvbeam/tests/test_beamfits.py
```diff
@@ -774,3 +774,35 @@ def test_multi_files(cst_efield_2freq, tmp_path):
...
+
+
+@pytest.fixture(scope="module")
+def twofreq(cst_efield_2freq_main, tmp_path_factory):
+    testfile = str(tmp_path_factory.mktemp("beams") / "twofreq.fits")
+    cst_efield_2freq_main.write_beamfits(testfile)
+    return testfile
+
+
+def test_partial_read_freq(twofreq):
```
bhazelton, line 786: Can we also test healpix beam files? I think that the frequency selection will just work and I also want to make sure there aren't places where the az/za ranges come in when they shouldn't.

pyuvdata/uvbeam/uvbeam.py
```diff
@@ -3820,6 +3846,18 @@ def read(
...
+            is to read in all frequencies. Restricting the frequencies reduces peak
+            memory usage.
+        az_range : tuple of float in deg
+            The azimuth range to read in, if the beam is specified in az/za coordinates.
+            Default is to read in all azimuths. Restricting the azimuth reduces peak
+            memory usage.
+        za_range : tuple of float in deg
+            The zenith angle range to read in, if the beam is specified in za/za
+            coordinates. Default is to read in all za. Restricting the za reduces peak
+            memory.
```
bhazelton, line 3860: Please specify when these will be used (only for beam fits files, az/za only for az/za beam fits files).

bhazelton: I am wondering whether we should implement these same parameters for the `select` method to keep the interfaces similar (as we do in UVData) or whether that doesn't matter.

pyuvdata/uvdata/tests/test_uvdata.py
```diff
@@ -1729,14 +1729,10 @@ def test_select_bls(casa_uvfits):
 
     with pytest.raises(ValueError) as cm:
         uv_object.select(bls=(5, 1))
-    assert str(cm.value).startswith(
-        "Antenna number 5 is not present in the " "ant_1_array or ant_2_array"
-    )
+    assert str(cm.value).startswith("Antenna pair (5, 1) does not have any data")
```
bhazelton, line 1732: I actually prefer to use the `match` parameter to pytest raises like this:
```
with pytest.raises(ValueError, match="Antenna pair (5, 1) does not have any data"):
         uv_object.select(bls=(5, 1))
```

I've been slowly converting all of the tests that use this form to use that one, I'd be happy for help!

pyuvdata/uvdata/uvdata.py
```diff
@@ -8271,36 +8289,42 @@ def _select_preprocess(
...
-                    bls_blt_inds = np.append(bls_blt_inds, list(wh2))
-                    if len(bl) == 3:
-                        # find conjugate polarization
-                        bl_pols.add(uvutils.conj_pol(bl[2]))
                 else:
-                    raise ValueError(
-                        "Antenna pair {p} does not have any data "
-                        "associated with it.".format(p=bl)
-                    )
+                    wh2 = np.where(
```
bhazelton, line 8312: This is great, thanks for figuring it out!

pyuvdata/uvdata/uvdata.py
```diff
@@ -23,6 +23,9 @@
 from ..uvbase import UVBase
 
 __all__ = ["UVData"]
+import logging
+
+logger = logging.getLogger(__name__)
```
bhazelton, line 28: Do we need to keep all these logging lines in?

bhazelton (commented):
This looks great, thanks @steven-murray and sorry for the delay in reviewing it. I just had a couple comments.

docs/uvbeam_tutorial.rst
steven-murray: Yes, added. Thank you.

pyuvdata/uvbeam/beamfits.py
steven-murray: It will *not* run for healpix! I'll add an error -- good catch!

pyuvdata/uvbeam/uvbeam.py
steven-murray: I removed the `freq_range` entry because it was duplicate. I added some info to the existing entry. Also added that the az/za works for beamfits az/za grid files. I don't know if we should add to `select`...

pyuvdata/uvdata/tests/test_uvdata.py
steven-murray: Converted all of them in this test :-)

pyuvdata/uvdata/uvdata.py
steven-murray: No. I like having some logging -- and I think especially at the INFO/DEBUG level it's not hurting anyone, because it's not printed by default anyway. But some of these are probably egregious (put in because I was looking for something particular). I've taken out all the ones I think are unnecessary. Happy to hear your opinion on others...

pyuvdata/uvdata/uvdata.py
```diff
@@ -8271,36 +8285,42 @@ def _select_preprocess(
...
+                # if not (bl[0] in self.ant_1_array or bl[0] in self.ant_2_array):
+                #     raise ValueError(
+                #         "Antenna number {a} is not present in the "
+                #         "ant_1_array or ant_2_array".format(a=bl[0])
+                #     )
+                # if not (bl[1] in self.ant_1_array or bl[1] in self.ant_2_array):
+                #     raise ValueError(
+                #         "Antenna number {a} is not present in the "
+                #         "ant_1_array or ant_2_array".format(a=bl[1])
+                #     )
```
bhazelton, line 8299: Let's just remove this commented code.

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- Support for partial reads on beamfits files. This can be done on frequency, azimuth
+  or zenith angle (if beam is in az/za coordinates).
+- Logging statements (INFO level) in `UVdata.check`.
+### Changed
+- Performance improvement when doing a `select` using the `bls` parameter (~70%
+  improvement in the `_select_preprocess` function)
```
bhazelton, line 12: maybe add something to make it clear this is on the UVData object?

pyuvdata/uvdata/uvdata.py
steven-murray: Done.

CHANGELOG.md
steven-murray: Done.

bhazelton: I think you forgot to push after you did it?

bhazelton (approved):
Looks good!

PR #1194, opened 2022-07-28
Add support for future array shapes in UVBeam

bhazelton:
## Description

Add support for future array shapes in UVBeam

Also added testing for a number of uncovered lines of code related to phased_array beams. In the process, I found and fixed a number of bugs related to the handling of phased array beams in various methods. This brings `uvbeam.py` to 100% test coverage!

## Motivation and Context

closes #914

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation change (documentation changes only)
- [ ] Version change
- [ ] Build or continuous integration change

## Checklist:

- [x] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/.github/CONTRIBUTING.md).
- [x] My code follows the code style of this project.

New feature checklist:
- [x] I have added or updated the docstrings associated with my feature using the [numpy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [x] I have updated the tutorial to highlight my new feature (if appropriate).
- [x] I have added tests to cover my new feature.
- [x] All new and existing tests pass.
- [x] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/CHANGELOG.md).

PR #1208, opened 2022-10-06
UVBeam cython extension

mkolopanis:
## Types of changes

- [ ] Version change
- [x] super secret rewrite again

## Checklist:

see attached pdf for timing difference. 
Memory footprint is **greatly** reduced however memory profiler did not seem to capture the difference. 
[uvbeam_to_healpix.pdf](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/9728762/uvbeam_to_healpix.pdf)


bhazelton:
I tested this rebased on top of #1194 and the tests passed and did not run out of memory. So I'm eager to get it in.

pyuvdata/uvbeam/uvbeam.pyx
```diff
@@ -53,18 +53,30 @@ cpdef numpy.ndarray[dtype=numpy.npy_bool] find_healpix_indices(
...
+          #  since the origin of the coordinate system is a Pole
           theta_g = theta_grid[itheta]
+
+          if theta_g < 0:
+            theta_g = fabs(theta_g)
+
           dtheta = fabs(theta_h - theta_g)
-          if dtheta > PI:
+          while dtheta > PI:
               dtheta -= 2 * PI
```
bhazelton, line 65: I think this should maybe be `dtheta -= PI` because theta only runs 0 -> pi.

bhazelton: @plaplant Matt wanted your opinion on this.

plaplant: Yes, I was just chatting with Matt about this. I think we need to change the logic slightly, because while it's true that positive and negative values of theta map to the same latitudes, they do not map to the same values of phi. In particular, a negative value (-theta, phi) maps to (+theta, phi + pi). See [the wiki page on spherical coordinates](https://en.wikipedia.org/wiki/Spherical_coordinate_system#Unique_coordinates).

I think the approach here should be: (1) normalize the value of dtheta to map to 0 <= dtheta <= pi, adding phase shifts to phi as necessary, then (2) normalize the value of dphi to map to 0 <= dphi <= 2pi. That should make it so our distance tests are working correctly for arbitrary input. I can try to implement this (though probably won't happen today).

bhazelton (approved):
This looks good to me now. Thanks @mkolopanis and @plaplant!

PR #1212, opened 2022-10-11
Support interdependent UVParameters

aelanman:
## Description

This adds a keyword "dependencies" to the declaration of UVParameter and its subclasses, which should be a dictionary. The keys are names of other UVParameters in the UVBase-derived object the UVParameter is kept on, and the values are "local" names.

In short -- If a UVParameter "_param" on a UVBase-derived object "base" is initialized with a dependencies dictionary {key: val}, then it will have `_param.val == base.key`.

## Motivation and Context

The `telescope_location` parameter is interpreted differently depending on the value of the `telescope_frame` attribute. This provides a way for the `telescope_location` parameter to be aware of the telescope_frame.

More specifically, this is needed because `telescope_location_lat_lon_alt` automatically converts between XYZ and LatLonAlt, but explicitly assumes the ITRS frame. The `...lat_lon_alt` attribute is instantiated before the `telescope_frame` might be defined, so it needs to be aware of the current value of `telescope_frame`.

It's not an open issue, but this makes the telescope_location_lat_lon_alt values to be sensible for lunar coordinates (it was previously giving deeply negative heights and incorrect lat/lon). It also adjusts the acceptable_range for telescope_position for lunar coordinates.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/parameter.py
+++ b/pyuvdata/parameter.py
@@ -125,6 +125,7 @@ class UVParameter(object):
         expected_type=int,
         acceptable_vals=None,
         acceptable_range=None,
+        dependencies=None,
         tols=(1e-05, 1e-08),
         strict_type_check=False,
     ):
@@ -137,6 +138,7 @@ class UVParameter(object):
         self.value = value
         self.description = description
         self.form = form
+        self.dependencies = dependencies
         if self.form == "str":
             self.expected_type = str
             self.strict_type = True
@@ -523,9 +525,13 @@ class LocationParameter(UVParameter):
         value=None,
         spoof_val=None,
         description="",
+        dependencies=None,
+        frame="itrs",
         acceptable_range=(6.35e6, 6.39e6),
         tols=1e-3,
     ):
+        if frame == "mcmf":
+            acceptable_range = (1717100.0, 1757100.0)
         super(LocationParameter, self).__init__(
             name,
             required=required,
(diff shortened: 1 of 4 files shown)


aelanman:
@bhazelton In trying to update pyuvsim for the latest changes in pyuvdata (r.e. lunarsky), I found some more issues that ended up being a little more fundamental. I've started this draft PR to address it, but I'm not sure if this is the best way forward.

An alternative approach might be to just add the attribute `frame` to `LocationParameter` and just let that be how we specify telescope frames in UVData. I did that to begin with, but then I had to add this other feature to ensure the `LocationParameter.frame` is consistent with `UVData.telescope_frame` at all times.

Please let me know what you think! If this seems appropriate, I can add it to the tutorial and changelog and mark it ready for review (and also set a new minimum pyuvdata version in pyuvsim). If not, we can decide on another approach.

bhazelton:
@aelanman we talked about this in the last pyuvdata telecon and we really prefer the approach of adding the attribute `frame` to LocationParameter objects. I think we'd then remove the `telescope_frame` from UVData objects. Let me know if you want me to help tackle that approach.

aelanman:
That makes sense to me! A lot easier to handle. I can get started on it tonight. I'll let you know if I have any trouble.

PR #1189, opened 2022-07-04
Unpin setuptools_scm

plaplant:
## Description

Related to #1187, we had pinned the version of `setuptools_scm` in #1188. The underlying issue has been resolved upstream, so we can relax the requirement. We just have to dodge versions `7.0.{0,1,2}`.

## Motivation and Context

Closes #1187.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation change (documentation changes only)
- [ ] Version change
- [x] Build or continuous integration change

## Checklist:

- [x] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/.github/CONTRIBUTING.md).
- [x] My code follows the code style of this project.

PR #1214, opened 2022-10-13
Fix how setuptools_scm requirements are specified

plaplant:
## Description

This PR fixes the `setuptools_scm` version specification from #1189. Evidently only conda understands the `|` operator as "OR", and pip/setuptools does not. Instead, we need to explicitly exclude versions, and add them together with `,` (which _does_ act as "AND").

## Types of changes

- [ ] Version change
- [x] Packaging change

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
 requires = ["setuptools>=30.3.0",
             "wheel",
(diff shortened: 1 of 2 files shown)


bhazelton (approved):
Thanks for figuring this out!

PR #1218, opened 2022-10-25
Add LocationParameter.frame

aelanman:
## Description

Removes the telescope_frame UVParameter and instead makes `frame` an attribute of the `LocationParameter` class. Default behavior is still to assume everything is ITRS unless otherwise specified.

## Motivation and Context

Coordinate transformations need to be aware of the frame XYZ coordinates are defined in. Several internal automated transformations (like XYZ to spherical) implicitly assumed ITRS coordinates. To provide support for lunar surface positions, these changes make the LocationParameter more general.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

Breaking change checklist:

Version change checklist:

### DIFF
--- a/pyuvdata/parameter.py
+++ b/pyuvdata/parameter.py
@@ -680,6 +680,8 @@ class LocationParameter(UVParameter):
         This is not an attribute of required UVParameters.
     description : str
         Description of the data or metadata in the object.
+    frame : str, optional
+        Coordinate frame. Valid options are "itrs" (default) or "mcmf".
     acceptable_vals : list, optional
         List giving allowed values for elements of value.
     acceptable_range: 2-tuple, optional
@@ -711,6 +713,8 @@ class LocationParameter(UVParameter):
        Always set to 3.
     description : str
         Description of the data or metadata in the object.
+    frame : str, optional
+        Coordinate frame. Valid options are "itrs" (default) or "mcmf".
     expected_type
         Always set to float.
     acceptable_vals : list, optional
@@ -735,11 +739,13 @@ class LocationParameter(UVParameter):
         spoof_val=None,
         description="",
         frame="itrs",
-        acceptable_range=(6.35e6, 6.39e6),
+        acceptable_range=None,
         tols=1e-3,
     ):
-        if frame == "mcmf":
-            acceptable_range = (1717100.0, 1757100.0)
+        if acceptable_range is None:
+            acceptable_range = (
+                (6.35e6, 6.39e6) if frame == "itrs" else (1717100.0, 1757100.0)
+            )
         super(LocationParameter, self).__init__(
             name,
             required=required,
--- a/pyuvdata/parameter.py
+++ b/pyuvdata/parameter.py
@@ -734,9 +734,12 @@ class LocationParameter(UVParameter):
         value=None,
         spoof_val=None,
         description="",
+        frame="itrs",
         acceptable_range=(6.35e6, 6.39e6),
         tols=1e-3,
     ):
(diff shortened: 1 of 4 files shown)


pyuvdata/parameter.py
```diff
@@ -734,9 +734,12 @@ def __init__(
         value=None,
         spoof_val=None,
         description="",
+        frame="itrs",
```
bhazelton, line 737: This needs to be added to both the `Parameters` and `Attributes` section of the docstring.

pyuvdata/parameter.py
```diff
@@ -734,9 +734,12 @@ def __init__(
...
         acceptable_range=(6.35e6, 6.39e6),
         tols=1e-3,
     ):
+        if frame == "mcmf":
+            acceptable_range = (1717100.0, 1757100.0)
```
bhazelton, line 742: I'd prefer to set the default to `None` and then if it is `None`, set it to the right value based on the frame (the existing default for `itrs` and this for `mcmd`. Otherwise the user cannot override it.

bhazelton (commented):
This is close, just a few comments. It will need to be rebased once #1220 is merged to get CI to pass.

pyuvdata/parameter.py
aelanman: done!

pyuvdata/parameter.py
aelanman: done as suggested!

pyuvdata/parameter.py
```diff
@@ -711,6 +713,8 @@ class LocationParameter(UVParameter):
        Always set to 3.
     description : str
         Description of the data or metadata in the object.
+    frame : str, optional
+        Coordinate frame. Valid options are "itrs" (default) or "mcmf".
```
bhazelton, line 717: Is this restriction (to "itrs" or "mcmf") enforced anywhere?

bhazelton: It looks like it's implicitly true because of the way you're setting the `acceptable_range`, but I don't see it being tested explicitly. Another approach would be to just not set the acceptable range if it's not "itrs" or "mcmf".

bhazelton:
I can't comment on the line because it's not changed, but it looks like we somehow lost coverage for line 825 of parameter.py

aelanman:
@bhazelton Currently, acceptable_range for a location parameter can only be set to None if the frame is also set to None. The default frame is `itrs` which causes the acceptable_range to default to the range near the Earth radius.

I think the behavior we want is, ultimately, for the acceptable_range to default according to the frame, unless the user specifically sets the keyword. Unfortunately, the case of `acceptable_range = None` is interpreted as both the user "deactivating" the acceptability check and the user not setting it.

To avoid this, I'm setting the default value to -1 in the function definition as a way to check if the user is setting the keyword. If it's -1, the constructor will set a default range according to the frame. If not, it will pass along what the user set.

So the behavior should be:
1. param = LocationParameter(..., acceptable_range=None) --> param.acceptable_range = None
2. param = LocationParameter(..., frame = 'mcmf' ) --> param.acceptable_range = (1717100.0, 1757100.0)
3. param = LocationParameter(... ) --> param.acceptable_range = (6.35e6, 6.39e6)
4. param = LocationParameter(..., frame = 'itrs' ) --> param.acceptable_range = (6.35e6, 6.39e6)
5. param = LocationParameter(..., frame = 'other' ) --> param.acceptable_range = None

bhazelton (approved):
Looks good, thanks for pushing this through!

PR #1229, opened 2022-12-05
fix: allow swapping of theta/phi in read_cst_beam

steven-murray:
## Description

Allows CST beam files to swap theta/phi convension.

## Motivation and Context

I have files from CST that don't work unless this is implemented.

## Types of changes

- [ ] Version change

## Checklist:

- [ ] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/.github/CONTRIBUTING.md).

Bug fix checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- Bugs in `UVBeam.read_cst_beam()` related to non-standard CST output text files.
+  Fix adds new parameter `swap_thetaphi_conventions` that swaps theta and phi, along
+  with a check that will suggest setting this parameter if reading fails in a specific
+  way. It also fixes a potential bug in which the input phi can be from -180 to 180
+  instead of 0 to 360. Does not change any behaviour for existing "normal" files.
 ## [2.2.11] - 2022-11-30
 
 ### Added
--- a/pyuvdata/uvbeam/cst_beam.py
+++ b/pyuvdata/uvbeam/cst_beam.py
@@ -73,6 +73,7 @@ class CSTBeam(UVBeam):
         run_check_acceptability=True,
         check_auto_power=True,
         fix_auto_power=True,
+        swap_thetaphi_conventions=False,
     ):
         """
         Read in data from a cst file.
@@ -135,7 +136,9 @@ class CSTBeam(UVBeam):
         fix_auto_power : bool
             For power beams, if auto polarization beams with imaginary values are found,
             fix those values so that they are real-only in data_array.
-
+        swap_thetaphi_conventions : bool
+            Whether to swap the standard assumed convention on theta/phi. By default,
+            phi is assumed to be synonymous with azimuth, and theta with zenith angle.
         """
         # update filename attribute
         basename = os.path.basename(filename)
@@ -218,27 +221,42 @@ class CSTBeam(UVBeam):
             column_names.append("".join(column_name.lower().split(" ")))
             units.append(unit.lower().strip())
 
+        if swap_thetaphi_conventions:
+            column_names = [
+                cn.replace("theta", "phi").replace("phi", "theta")
+                for cn in column_names
+            ]
+
         data = np.loadtxt(filename, skiprows=2)
 
-        theta_col = np.where(np.array(column_names) == "theta")[0][0]
-        phi_col = np.where(np.array(column_names) == "phi")[0][0]
+        theta_col = column_names.index("theta")
+        phi_col = column_names.index("phi")
 
+        theta_data = data[:, theta_col]
+        phi_data = data[:, phi_col]
         if "deg" in units[theta_col]:
-            theta_data = np.radians(data[:, theta_col])
-        else:
-            theta_data = data[:, theta_col]
+            theta_data = np.radians(theta_data)
         if "deg" in units[phi_col]:
-            phi_data = np.radians(data[:, phi_col])
-        else:
-            phi_data = data[:, phi_col]
+            phi_data = np.radians(phi_data)
 
+        if theta_data.max() - theta_data.min() > np.pi * 1.00001:
+            raise ValueError(
+                "It seems that the input file uses a swapped convention for theta/phi. "
+                "Try setting 'swap_thetaphi_conventions' to True."
+            )
+
+        phi_data %= (
+            2 * np.pi
+        )  # this begs the question of how the branch cut is defined...
         theta_axis = np.sort(np.unique(theta_data))
         phi_axis = np.sort(np.unique(phi_data))
         if not theta_axis.size * phi_axis.size == theta_data.size:
             raise ValueError("Data does not appear to be on a grid")
 
-        theta_data = theta_data.reshape((theta_axis.size, phi_axis.size), order="F")
-        phi_data = phi_data.reshape((theta_axis.size, phi_axis.size), order="F")
+        fast_moving = "phi" if phi_data[1] > phi_data[0] else "theta"
+        order = "F" if fast_moving == "theta" else "C"
+        theta_data = theta_data.reshape((theta_axis.size, phi_axis.size), order=order)
+        phi_data = phi_data.reshape((theta_axis.size, phi_axis.size), order=order)
 
         if not uvutils._test_array_constant_spacing(theta_axis, self._axis2_array.tols):
             raise ValueError(


pyuvdata/uvbeam/cst_beam.py
```diff
@@ -220,36 +226,127 @@ def read_cst_beam(
...
+                if (idx + 1) < n and (za, az) == tupled[idx + 1]:
+                    # best case scenario is everything is in grid already.
+                    # For most beams, the large majority of iterations will hit this.
+                    idx = idx + 1
+                    new_data[i] = data[idx]
+                else:
+                    try:
+                        # Find the index of these coordinates
+                        # There may be multiple indices with the same coords, because
+                        # of poles -- here we just take the first one.
```
bhazelton, line 301: I'm confused about this. At poles you can have multiple points with the same za but not the same az, right?

pyuvdata/uvbeam/cst_beam.py
```diff
@@ -220,36 +226,127 @@ def read_cst_beam(
...
+                        idx = tupled.index((za, az))
+                        new_data[i] = data[idx]
+                        if np.isclose(new_data[i, theta_col], -np.pi):
+                            print(data[idx], za, az)
```
bhazelton, line 305: What is this print for?

pyuvdata/uvbeam/cst_beam.py
```diff
@@ -220,36 +226,127 @@ def read_cst_beam(
...
+                                    & np.isclose(phi_data, (az + np.pi) % (2 * np.pi))
+                                )[0]
+                                if len(idx) == 0:
+                                    raise ValueError(
+                                        f"No coordinate data for za={za} with az={az}"
+                                    ) from err
+                                idx = idx[0]
+
+                                for icol, name in enumerate(column_names):
+                                    if "phase" in name:  # flip the phase.
```
bhazelton, line 328: Can you expand the comment to explain why we're flipping the phase? Will help with understanding the code in the future.

pyuvdata/uvbeam/tests/test_cst_beam.py
```diff
@@ -49,6 +49,13 @@ def cst_efield_2freq_mod(cst_efield_2freq_main):
     del beam
 
 
+def test_swapped_cst_beam():
+    beam = UVBeam()
+    beam.read_cst_beam(
+        os.path.join(DATA_PATH, cst_folder, "CST_f50_Eloy.txt"),
+    )
```
bhazelton, line 56: Should we check anything about this beam object? Maybe that the value in a particular spot is what we expect it to be or that the peak is at zenith? So we're sure it was actually read in properly.

pyuvdata/uvbeam/uvbeam.py
```diff
@@ -3986,7 +3988,9 @@ def read_cst_beam(
         fix_auto_power : bool
             For power beams, if auto polarization beams with imaginary values are found,
             fix those values so that they are real-only in data_array.
-
+        swap_thetaphi_conventions : bool
+            Whether to swap the standard assumed convention on theta/phi. By default,
+            phi is assumed to be synonymous with azimuth, and theta with zenith angle.
```
bhazelton, line 3993: This needs to be changed to match the actual new parameter. Also, there should be a blank line before the end of the docstring.

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- Bugs in `UVBeam.read_cst_beam()` related to non-standard CST output text files.
+  Fix adds new parameter `swap_thetaphi_conventions` that swaps theta and phi, along
+  with a check that will suggest setting this parameter if reading fails in a specific
+  way. It also fixes a potential bug in which the input phi can be from -180 to 180
+  instead of 0 to 360. Does not change any behaviour for existing "normal" files.
```
bhazelton, line 11: This is out of date, it needs to be changed to match the actual new parameter.

pyuvdata/uvbeam/uvbeam.py
```diff
@@ -4018,11 +4022,12 @@ def read_cst_beam(
                 "model_name": model_name,
                 "model_version": model_version,
                 "history": history,
+                "thetaphi_to_azza_fnc": thetaphi_to_azza_fnc,
             }
             if "ref_imp" in settings_dict:
                 overriding_keywords["reference_impedance"] = reference_impedance
             if "x_orientation" in settings_dict:
-                overriding_keywords["x_orientation"] = reference_impedance
+                overriding_keywords["x_orientation"] = x_orientation
```
bhazelton, line 4030: oof, thanks for finding and fixing this!

bhazelton (commented):
I think this approach is reasonable, it's certainly very flexible but it does put more onus on the user to write the correct function. If we identify common formats, it might be nice to add common functions that might need to be passed to `theta phi_to_azza_fnc` to the repo so people don't have to keep reinventing them (possibly with bugs).

pyuvdata/uvbeam/cst_beam.py
steven-murray: I added a much larger discussion of this in the comments -- does it make more sense now? To be even more clear, with an example from my current situation: in the CST files I got from Eloy, the convention is that theta runs from -180 to 179 and phi goes from -90 to 90 (inclusive). This means that the phi=-90 meridian, and the phi=90 meridian, share all their points. In the conversion function, I would take, eg. (theta, phi)=(-90, -90) and (theta, phi)=(90, 90), and map BOTH of them to (az,za)=(270, 90). However, I only need a single point at (270, 90) in my az/za grid, and both of these should be identical, so I just take the first one. 

Of course, the number of "redundant" points here is 179, which is the same number of redundant points in the az/za grid (but in that case they're all at zenith).

pyuvdata/uvbeam/cst_beam.py
steven-murray: woops, removed

pyuvdata/uvbeam/cst_beam.py
steven-murray: I actually don't really know the technical reasons, but I know it's true 😆 . I made the comment larger above here.

pyuvdata/uvbeam/tests/test_cst_beam.py
steven-murray: I updated it to check whether the amplitudes at zenith/anti-zenith are all the same.

pyuvdata/uvbeam/uvbeam.py
steven-murray: Done, and added the parameter to all the other `read` docs

CHANGELOG.md
steven-murray: Done.

pyuvdata/uvbeam/uvbeam.py
steven-murray: 👍🏻

pyuvdata/uvbeam/cst_beam.py
```diff
@@ -220,36 +226,127 @@ def read_cst_beam(
...
+                "Check your thetaphi_to_azza_fnc."
             )
 
-        if not uvutils._test_array_constant_spacing(phi_axis, self._axis1_array.tols):
+        if phi_data.max() > phi_data.min() + 2 * np.pi:
             raise ValueError(
-                "Data does not appear to be regularly gridded in azimuth angle"
+                "The azimuthal angle is outside (0, 2*pi): "
+                f"{phi_data.min(), phi_data.max()}. "
+                "Check your thetaphi_to_azza_fnc."
```
bhazelton, line 253: I think we need to reword this. It doesn't look like this requires that it is in the interval [0, 2pi], it requires that it only spans 2pi, right?

pyuvdata/uvbeam/cst_beam.py
```diff
@@ -135,7 +137,11 @@ def read_cst_beam(
         fix_auto_power : bool
             For power beams, if auto polarization beams with imaginary values are found,
             fix those values so that they are real-only in data_array.
-
+        thetaphi_to_azza_fnc : callable
+            Since CST can output theta/phi coordinates in a number of conventions, you
+            can provide a function to convert from the input theta/phi to azimuth/
+            zenith-angle, as required by UVBeam. The signature of the function should
+            be ``az, za = fnc(theta, phi)``. Theta and Phi will be in radians.
```
bhazelton, line 144: There should be a blank line at the end of this docstring. See examples here: https://numpydoc.readthedocs.io/en/latest/format.html

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- Bugs in `UVBeam.read_cst_beam()` related to non-standard CST output text files.
+  Fix adds new parameter `thetaphi_to_azza_fnc` that allows the user to input a function
+  converting theta/phi in the CST file to az/za values for pyuvbeam. It also adds code
+  to re-order the input coordinates to be regular in az/za.
```
bhazelton, line 10: Let's keep the indenting consistent with prior lines.

pyuvdata/uvbeam/cst_beam.py
bhazelton: I don't think you committed the changes to the comments mentioned here and below (on the "flip the phase" comment). I don't see any new comments in your latest commit.

pyuvdata/uvbeam/cst_beam.py
bhazelton: I don't think you committed this change.

pyuvdata/uvbeam/cst_beam.py
bhazelton: I don't think you committed this change.

pyuvdata/uvbeam/cst_beam.py
bhazelton: The part that confuses me is `The azimuthal angle is outside (0, 2*pi)`. I read that to say that it is required to be within the interval [0, 2pi].

steven-murray: Got it, yup missed this comment originally. I've fixed it now.

bhazelton (commented):
This is looking good to me now, but CI is erroring badly on the typing of the new parameter.

PR #1242, opened 2022-12-21
Move warnings tests from circleci to GH actions to fix codecov reporting, other CI improvements

bhazelton:
## Description

Move the warnings test from circleci to github actions.

Also limited the publishing CI to only run on PRs and releases (dropping all pushes) and added coherency groups so github actions can cancel runs when they are superseded by another push.

## Motivation and Context

Currently, if the warnings test fails, Codecov does not report coverage information on a PR. This is annoying because the warning test isn't actually required to pass for our PRs but the codecov patch is required. This is known behavior for codecov with CircleCI -- it only reports if all the CI jobs pass. This has caused problems several times on recent PRs.

## Types of changes

- [ ] Version change

## Checklist:

### DIFF
--- a/.github/workflows/warning_test.yaml
+++ b/.github/workflows/warning_test.yaml
@@ -1,5 +1,10 @@
 name: Warning Tests
+
 on: [push, pull_request]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
 
 jobs:
   warning_test:
--- a/.github/workflows/warning_test.yaml
+++ b/.github/workflows/warning_test.yaml
@@ -29,7 +29,7 @@ jobs:
           miniforge-version: latest
           use-mamba: true
           python-version: ${{ env.PYTHON }}
-          environment-file: ci/${{ env.ENV_NAME }}.yaml
+          environment-file: ci/${{ env.ENV_NAME }}.yml
           activate-environment: ${{ env.ENV_NAME }}
 
       - name: Conda Info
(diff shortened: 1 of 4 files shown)


.github/workflows/warning_test.yaml
```diff
@@ -44,6 +44,7 @@ jobs:
       - name: Install
         run: |
           CFLAGS="-DCYTHON_TRACE=1 -DCYTHON_TRACE_NOGIL=1" pip install .
+          python setup.py build_ext --inplace
```
mkolopanis, line 47: this is interesting. We haven't needed this anywhere else. but obviously without it you had an issues. 
I'm willing to bet the real issue is the namespace degeneracy between the `pyuvdata` package and the directory with the same name. So in the next step pytest attempts to run tests from the directory.

I personally would prefer to remove this line, but add something in the next step that makes an empty directory and `cd`s into it.

mkolopanis (commented):
This look good. Annoying that circleCI failures blocks coverage but I think I get why they did that too.

I would personally like to handle the cython building a little differently as noted below but I'm open to other options.

.github/workflows/warning_test.yaml
bhazelton: We can do it that way, that's the way it's done in other checks. The advantage of this is that I think it may cover more lines in the __init__.py where it figures out the version number. We currently have those lines marked with a `pragma: no cover` so it's not showing as more coverage, but I think they get hit with this approach. I can try removing the pragma to see.

bhazelton: This line is what I always have to run on my own machine to be able to run tests from top level pyuvdata folder.

jpober (approved):

PR #1244, opened 2023-01-03
Fix some handling for old phase attributes

bhazelton:
## Description

This is a follow-on to #1170 to fix a couple of things that were left out and were breaking hera_cal.

The main thing this fixes is that if a new object is created from scratch with the old phase attributes, the old attributes were not getting correctly converted into the phase_center_catalog. In hera_cal there's a method to make a new object from scratch (sort of) and then write it out and it errored in the write (in the call to `check` before any actual writing). I decided to put the conversion from the old to new attributes in the `check` method so that it would always happen before a write. There could still be errors though if someone makes an object and doesn't call check before calling some of the other methods on the object, hopefully that is uncommon.

It turned out to be a little tricky to handle the interaction between `__getattribute__` and the detection of the old phase attributes because the handling for the old attributes in `__getattribute__` make `hasattr` return `True` for the old phase attributes. So I used the presence of a sensible `phase_center_catalog` to decide whether or not to check for the old attributes. A consequence is that if there is a sensible `phase_center_catalog`, setting any of the old attributes will not change any behavior. There will be a warning if a user does that, however.

I also added a deprecation warning when setting the old attributes (using `__setattr__`) which @kartographer suggested on #1170 but I forgot to do.

## Motivation and Context

Fixes errors in hera_cal.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ they are converted between types using UVData and UVCalobjects.
 to set several of these new parameters.
 
 ### Fixed
+- Fixed a bug where objects created from scratch with the old phase attributes weren't
+properly getting converted to the new `phase_center_catalog` representation. That
+conversion now happens in the check so will happen before any `write`. There could still
+be errors if some methods are called before a check is done (or if the check is turned
+off).
 - Fix a bug in how `frame_pa` was calculated in `phase` and `_set_app_coords_helper` for
 multi_phase_center objects (it was using the old `phase_center_frame` attribute).
 - Fix a bug where trying to select lsts or lst_ranges on read didn't work for some file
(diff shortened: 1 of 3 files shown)


mkolopanis (commented):
I don't think check is a bad place for this to happen, but maybe we could pull out the logic into a separate "private" function for now `_convert_old_phase_attributes` or something of the sort. Then check can just call this method.

Then in the off chance a user doesn't call check and wants to do something, we can still point them at a piece of code to help them through their issue?

mkolopanis (approved):
Thanks for adding another function for this. I'm good with this as it stands. we can swing back around if we need more changes.

PR #1246, opened 2023-01-04
Improve beam interp docs

bhazelton:
## Description

Improve the documentation around beam interpolation and add a parameter to the `interp` and `to_healpix` methods to be able to set the interpolation function directly in the call rather than on the object. Make that parameter default sensibly based on the object's pixel coordinate system.

Also deprecate the `interpolation_function` attribute on UVBeam objects.

## Motivation and Context

This was inspired by a Slack thread where a user was confused about needing to set the interpolation_function.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/uvbeam/uvbeam.py
+++ b/pyuvdata/uvbeam/uvbeam.py
@@ -2076,6 +2076,7 @@ class UVBeam(UVBase):
         self,
         az_array=None,
         za_array=None,
+        interpolation_function=None,
         az_za_grid=False,
         healpix_nside=None,
         healpix_inds=None,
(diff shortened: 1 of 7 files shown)


bhazelton:
@steven-murray Thanks, I've implemented both of those changes, so this is now a more complicated PR but I think it's a better approach.

pyuvdata/uvbeam/uvbeam.py
```diff
@@ -2485,6 +2520,13 @@ def to_healpix(
             object.
 
         """
+        if self.pixel_coordinate_system == "healpix":
+            # TODO: should this just error or just be a no-op (as implemented below)?
```
steven-murray, line 2524: I think what you've done is the right thing.

steven-murray (dismissed):
This is awesome, thanks @bhazelton !

steven-murray (approved):

PR #1250, opened 2023-01-18
perf: use lru_cache for polstr and variants

steven-murray:
## Description

Applies the `lru_cache` decorator to some utility functions, especially the `polnum2str` function and its variants. These are good functions for this decorator, as they have a very limited set of possible input parameters, so do not take much extra cache memory. 

## Motivation and Context

It turns out that these functions can take a LOT of the total time for some `hera_cal` scripts, because they are called *every time* you request a baseline-pol key from a `HERAData` object. Thus, if you loop over all baselines in a file to get their data, you end up calling `polstr2num` like 65k times. Since `polstr2num` does a deepcopy of a dict, it actually takes a non-negligible amount of time (for example, the delay filter was taking ~20min for a 2-integration file, of which about 7min was taken by the `polstr2num` function). 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ off).
 multi_phase_center objects (it was using the old `phase_center_frame` attribute).
 - Fix a bug where trying to select lsts or lst_ranges on read didn't work for some file
 types.
-
+- Severe performance hit when calling `polnum2str` and its variants for many baselines.
 ### Deprecated
 - Deprecated the `interpolation_function` attribute on UVBeams.
 - Deprecated the older phase attributes (`phase_type`, `phase_center_ra`,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ multi_phase_center objects (it was using the old `phase_center_frame` attribute)
 - Fix a bug where trying to select lsts or lst_ranges on read didn't work for some file
 types.
 - Severe performance hit when calling `polnum2str` and its variants for many baselines.
+
 ### Deprecated
 - Deprecated the `interpolation_function` attribute on UVBeams.
 - Deprecated the older phase attributes (`phase_type`, `phase_center_ra`,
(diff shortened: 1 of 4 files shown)


steven-murray:
Ech. It's not working because the polarization numbers can be input as arrays, which I didn't realize. Currently this works fine for hera_cal because we never do this, but it won't work in general for pyuvdata. Unless you have a good idea about how to do this more generally, @bhazelton or @mkolopanis, I'll just close this PR and do it in hera_cal.

steven-murray:
@bhazelton possibly, I'll have a check of the logic.

bhazelton:
@plaplant suggests creating a private function with the decorator to be called by the existing function which can convert from an array or list to a tuple as needed.

steven-murray:
@bhazelton I added @plaplant as a reviewer as well, since he had ideas about this :-) But in any case it should be ready to go.

CHANGELOG.md
```diff
@@ -30,7 +30,7 @@ off).
 multi_phase_center objects (it was using the old `phase_center_frame` attribute).
 - Fix a bug where trying to select lsts or lst_ranges on read didn't work for some file
 types.
-
+- Severe performance hit when calling `polnum2str` and its variants for many baselines.
 ### Deprecated
```
bhazelton, line 34: please add an empty line before the `Deprecated` heading.

setup.cfg
```diff
@@ -5,7 +5,7 @@ description_file = README.md
 # B905 is for using zip without the `strict` argument, which was introduced in
 # python 3.10. We should probably add this check (remove it from the ignore) when we
 # require 3.10.
-ignore = W503, E203, N806, B905, B907
+ignore = W503, E203, N806, B905, B907, B028
```
bhazelton, line 8: I think `B028` is the same as `B907`. It started out as B028 but then was converted to B907 in flake8-bugbear version 23.1.17 . We now require that version or newer, so I don't think we need `B028` in this list. See: https://github.com/PyCQA/flake8-bugbear#23117

bhazelton (commented):
Looks reasonable to me, not sure why pre-commit is so unhappy...

CHANGELOG.md
steven-murray: Ah, done.

bhazelton (approved):
Looks good. Thanks!!

PR #1252, opened 2023-01-21
Require flex_spw_id_array on UVData, UVCal and UVFlag in the future

### LINKED ISSUE #1234: Start process for always requiring the `flex_spw_id_array` to be set
This will simplify code and we can remove the `flex_spw` boolean as well.

This should be done on UVData, UVFlag and UVCal. Needs to start with deprecation warnings on those objects.

bhazelton:
## Description

Start process of requiring the flex_spw_id_array on all objects that support spectral windows.

## Motivation and Context

closes #1234

## Types of changes

- [ ] Version change

## Checklist:

Breaking change checklist:

### DIFF
--- a/pyuvdata/parameter.py
+++ b/pyuvdata/parameter.py
@@ -302,7 +302,7 @@ class UVParameter(object):
             When set to False (default), descriptive text is printed out when parameters
             do not match. If set to True, this text is not printed.
         """
-        if isinstance(other, self.__class__):
+        if isinstance(other, self.__class__) and isinstance(self, other.__class__):
             if self.value is None:
                 if other.value is not None:
                     if not silent:
@@ -519,7 +519,7 @@ class UVParameter(object):
             When set to False (default), descriptive text is printed out when parameters
             do not match. If set to True, this text is not printed.
         """
-        return not self.__eq__(other)
+        return not self.__eq__(other, silent=silent)
 
     def apply_spoof(self):
         """Set value to spoof_val for non-required UVParameters."""
@@ -1030,38 +1030,42 @@ class SkyCoordParameter(UVParameter):
             tols=(0, radian_tol),
         )
 
-    def __eq__(self, other):
+    def __eq__(self, other, silent=False):
         if not issubclass(self.value.__class__, SkyCoord) or not issubclass(
             other.value.__class__, SkyCoord
         ):
-            return super(SkyCoordParameter, self).__eq__(other)
+            return super(SkyCoordParameter, self).__eq__(other, silent=silent)
 
         if self.value.shape != other.value.shape:
-            print(f"{self.name} parameter shapes are different")
+            if not silent:
+                print(f"{self.name} parameter shapes are different")
             return False
 
         this_frame = self.value.frame.name
         other_frame = other.value.frame.name
         if this_frame != other_frame:
-            print(
-                f"{self.name} parameter has different frames, {this_frame} vs "
-                f"{other_frame}."
-            )
+            if not silent:
+                print(
+                    f"{self.name} parameter has different frames, {this_frame} vs "
+                    f"{other_frame}."
+                )
             return False
 
         this_rep_type = self.value.representation_type
         other_rep_type = other.value.representation_type
         if this_rep_type != other_rep_type:
-            print(
-                f"{self.name} parameter has different representation_types, "
-                f"{this_rep_type} vs {other_rep_type}."
-            )
+            if not silent:
+                print(
+                    f"{self.name} parameter has different representation_types, "
+                    f"{this_rep_type} vs {other_rep_type}."
+                )
             return False
 
         # finally calculate on sky separations
         sky_separation = self.value.separation(other.value).rad
         if np.any(sky_separation > self.tols[1]):
-            print(f"{self.name} parameter is not close. ")
+            if not silent:
+                print(f"{self.name} parameter is not close. ")
             return False
 
         return True
--- a/pyuvdata/uvcal/uvcal.py
+++ b/pyuvdata/uvcal/uvcal.py
@@ -380,7 +380,8 @@ class UVCal(UVBase):
         )
 
         desc = (
-            "Required if flex_spw = True. Maps individual channels along the "
+            "Required if flex_spw = True and will always be required for non-wide-band "
+            "objects starting in version 3.0. Maps individual channels along the "
             "frequency axis to individual spectral windows, as listed in the "
             "spw_array. Shape (Nfreqs), type = int."
         )
@@ -581,6 +582,7 @@ class UVCal(UVBase):
             assert (
                 self.future_array_shapes
             ), "future_array_shapes must be True to set wide_band to True."
+            # TODO think about what to test for in flex_spw_id_array
             assert (
                 not self.flex_spw
             ), "Cannot set objects with flexible spectral windows to wide_band."
@@ -594,6 +596,7 @@ class UVCal(UVBase):
         if wide_band:
             self._freq_array.required = False
             self._channel_width.required = False
+            # in version 3.0, also set _flex_spw_id_array to not be required
             self._freq_range.required = True
 
             data_shape_params = [
@@ -615,6 +618,7 @@ class UVCal(UVBase):
         else:
             self._freq_array.required = True
             self._channel_width.required = True
+            # in version 3.0, also set _flex_spw_id_array to be required
             self._freq_range.required = False
 
             if self.future_array_shapes:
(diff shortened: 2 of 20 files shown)


bhazelton:
The hera_qm errors will be fixed in this PR I made on that repo: https://github.com/HERA-Team/hera_qm/pull/439

bhazelton:
The hera_cal errors will be fixed in a PR I made on that repo: https://github.com/HERA-Team/hera_cal/pull/868

pyuvdata/parameter.py
```diff
@@ -302,7 +302,7 @@ def __eq__(self, other, silent=False):
             When set to False (default), descriptive text is printed out when parameters
             do not match. If set to True, this text is not printed.
         """
-        if isinstance(other, self.__class__):
+        if isinstance(other, self.__class__) and isinstance(self, other.__class__):
```
steven-murray, line 305: is the point here to require their classes to be exactly the same? The first `isinstance` requires `other` to be either the same class or a subclass, and the second `isinstance` turns that around, which I think means that `self.__class__ == other.__class__` would give the same thing?

pyuvdata/parameter.py
```diff
@@ -1021,38 +1021,42 @@ def __init__(
             tols=(0, radian_tol),
         )
 
-    def __eq__(self, other):
+    def __eq__(self, other, silent=False):
```
steven-murray, line 1024: Wouldn't it be better just to use the `logging` module instead of this `silent` parameter?

steven-murray: That would mean you don't have to have a parameter of `__eq__` etc, and still maintain the ability for granular output.

pyuvdata/uvcal/uvcal.py
```diff
@@ -380,7 +380,8 @@ def __init__(self):
         )
 
         desc = (
-            "Required if flex_spw = True. Maps individual channels along the "
+            "Required if flex_spw = True and will always be required for non-wide-band "
+            "objects starting in version 3.0. Maps individual channels along the "
```
steven-murray, line 384: Does the `flex_spw_id_array` have an obvious default? It seems that in many applications, it'll just be all zeros. Can we make it have a default?

pyuvdata/uvdata/ms.py
```diff
@@ -2055,11 +2055,11 @@ def read_ms(
             self.freq_array[sel_mask] = data_desc_dict[key]["CHAN_FREQ"]
             self.channel_width[sel_mask] = data_desc_dict[key]["CHAN_WIDTH"]
 
+        # Future proof: always set the flex_spw_id_array.
         if (np.unique(self.channel_width).size == 1) and (len(spw_list) == 1):
             # If we don't _need_ the flex_spw setup, then the default is not
             # to use it.
```
steven-murray, line 2061: This comment seems out of line now?

pyuvdata/uvdata/uvdata.py
```diff
@@ -300,13 +300,14 @@ def __init__(self):
             '"Stokes", but this is inaccurate as visibilities cannot be in '
             "true Stokes polarizations for physical antennas. We adopt the "
             "term pseudo-Stokes to refer to linear combinations of instrumental "
-            "visibility polarizations (e.g. pI = xx + yy)."
+            "visibility polarizations (e.g. pI = xx + yy). A value of 0 is only "
```
steven-murray, line 303: What does the value "0" represent?

pyuvdata/uvdata/uvdata.py
```diff
@@ -3206,6 +3204,19 @@ def check(
...
+                "3.0",
+                DeprecationWarning,
+            )
+        else:
+            # Check that all values in flex_spw_id_array are entries in the spw_array
+            if not np.all(np.isin(self.flex_spw_id_array, self.spw_array)):
+                raise ValueError(
+                    "All values in the flex_spw_id_array must exist in the spw_array."
+                )
+
```
steven-murray, line 3219: Going to reiterate that it would be a way easier transition if a reasonable default could be set in *most* cases.

steven-murray (commented):
Thanks @bhazelton, this is a very large PR with a lot of detail in it. I have only a few comments, the main one being that to ease the transition (and just general usability) I strongly recommend to assign a default value for the `flex_spw_id_array` and associated attributes where possible. Also, it might be useful to have a dedicated little tutorial or FAQ for the new flex id stuff?

pyuvdata/parameter.py
bhazelton: I added this because there was a case where the first one returned as True when I did not expect it, so I was getting a weird error downstream that I traced to here. I don't quite remember what all the details are, I can try to figure it out.

pyuvdata/parameter.py
bhazelton: That might be a good option. I'll need to look into that more. If we change away from this I'll need to make corresponding updates in pyradiosky.

pyuvdata/uvcal/uvcal.py
bhazelton: There's an obvious default if there is only one spectral window, in which case the obvious default is to fill it with the (single) value in the `spw_array`.

bhazelton: I added this defaulting to all the `read` methods, so users who just read in files will not notice any change. Do you want me to add defaulting to the `check` method as well?

pyuvdata/parameter.py
bhazelton: Right, I remember. If `other` is a subclass of `self` this statement returned as `True` before this modification. There's now a test that errors without this modification.

pyuvdata/parameter.py
bhazelton: Reading here: https://docs.python.org/3/howto/logging.html#when-to-use-logging
I'm uncertain. So the existing behavior is that if you call `__eq__` is that if the objects are not equal, it prints some helpful info about why they are not equal. This can be enormously helpful. But when you call `__ne__`, you don't really need that info. The worry I have about using logging is that the standard user will just stop getting this info and will not know that they can get it by adjusting the logging level. I think if you poll our users, most would not think to adjust that.

pyuvdata/parameter.py
steven-murray: That's correct. Often you want that (i.e. if it's a subclass and matches other criteria, it's fine to call it equal). Maybe here you don't though :-)

pyuvdata/parameter.py
steven-murray: Yeah, I considered that, and maybe you're right. I also don't think it's hugely obvious that you should call `this.__ne__(other, silent=False)` instead of just `this != other`. Logging is fairly standard practice. 

The fact that you _sometimes_ want to print the info and sometimes not is a bit more concerning (and not naively solvable within logging). I wonder if perhaps a better route here is to have a separate function like `uvd.print_diff(other)` where a user can explicitly print such a diff?

pyuvdata/uvdata/ms.py
bhazelton: yes, that's cruft.

pyuvdata/uvdata/uvdata.py
bhazelton: It indicates that this is a flex-pol object, which means that the spectral windows have different polarizations (which is not normally allowed) but each spectral window only has one polarization. This mode was developed to support the SubMillimeter array, which can operate in such a mode, but it turns out that we're going to use it in HERA as well as it can significantly speed up single polarization reads. There are methods for converting HERA-type data between the regular layout (single SPW with all pols) and the flex-pol layout (multiple SPWs each with one pol).

Can you suggest some language that would make this more clear to you and other users?

pyuvdata/uvdata/uvdata.py
bhazelton: There is a default set when files are read in. Do you think we should also set a default in the check method? I think that most users won't even notice, because they make objects by reading in files.

pyuvdata/parameter.py
bhazelton: So the way it's set up now, if you call `this != other` you won't get any messages but if you call `this == other` you will. In the tests I called it as `this.__ne__(other, silent=False)` in order to cover the lines of code that print things, but I wouldn't think most people would need/want to call it that way.

pyuvdata/uvcal/uvcal.py
bhazelton: I think this is another thing you'd default sensibly in a method that builds an object from the "minimal" data set. I suppose we could build in a default using `__getattr__` machinery, but that can get a bit complicated (as I've learned in pyradiosky). For now, it's just a warning, but it's heading towards being required.

steven-murray: No, let's not do a `__getattr__`! I think this is fine -- we can build it into a `.new()` classmethod, or whatever we call it.

steven-murray (approved):

steven-murray (approved):

PR #1258, opened 2023-02-02
feat: add FastUVH5Meta reader

steven-murray:
## Description

This adds a new class, `FastUVH5Meta`, which makes it possible to do partial-loading of the metadata from UVH5 files. I've threaded it through so that the `_read_header` method of `UVH5` uses this class to do its reading (although in doing so, it loses the benefit of the partial reading, because `UVData` objects themselves require all metadata to be present before a `check()`).

## Motivation and Context

The motivation here is that it is often desirable to be able to read a small fraction of a file's header to get some info. For example, say I have a thousand files on disk, all different times. I want to open each, check which times they're at, and discard them if they're not within a certain range. Currently, this involves reading a lot of unnecessary metadata on each file, which stacks up over 1000 files. By using the new `FastUVH5Meta` class _directly_, one can now access _just_ the time array, and do it in a way that is officially supported (because the `UVH5` class itself uses this reader to read).

Some notes:

* I think we can be using the `blt_order` attribute more effectively. I've tried to use it to some advantage in this code, but I think we can do better.
* In the future it might be nice to use a similar interface to enable data reading.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,13 +3,22 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- A new `FastUVH5Meta` object that enables quick partial reading of uvh5 file metadata.
+This is used by default in the `UVData.read_uvh5` method to initialize a `UVData`
+object, without any apparent change for the user. To get the benefits of the new class,
+use it directly to interface with the metadata portion of a `.uvh5` file.
+
+### Changed
+- Updated numpy requirements to `>=1.20`.
+- Updated scipy requirements to `>=1.5`.
+
 ### Fixed
 - Handling of antenna_names and antenna_numbers in `read_fhd` and `read_fhd_cal`.
 
 ## [2.3.0] - 2023-03-13
 
 ### Added
-- A new `FastUVH5Meta` object that enables quick partial reading of uvh5 file metadata.
 - A new `interpolation_function` parameter to `UVBeam.interp` and `UVBeam.to_healpix`
 to allow the function name to be passed into the methods, with sensible defaulting.
 - Support for selecting on phase center IDs, including on read.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,11 +22,13 @@ from them will inherit them. UVFlag objects will also inherit these parameters w
 they are converted between types using UVData and UVCalobjects.
 - The `UVFlag.set_telescope_params` method, similar to the ones on UVData and UVCal,
 to set several of these new parameters.
+- A new `FastUVH5Meta` object that enables quick partial reading of uvh5 file metadata.
 
 ### Changed
 - Increases the h5py minimum version to 3.1.
 - Increase the minimum compatible version for lunarsky to 0.2.1 to fix astropy
 deprecation warnings.
+
 
 ### Fixed
 - Fixed some bugs related to handling of the `UVCal.time_range` parameter in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 ## [2.3.0] - 2023-03-13
 
 ### Added
+- A new `FastUVH5Meta` object that enables quick partial reading of uvh5 file metadata.
 - A new `interpolation_function` parameter to `UVBeam.interp` and `UVBeam.to_healpix`
 to allow the function name to be passed into the methods, with sensible defaulting.
 - Support for selecting on phase center IDs, including on read.
@@ -22,13 +23,11 @@ from them will inherit them. UVFlag objects will also inherit these parameters w
 they are converted between types using UVData and UVCalobjects.
 - The `UVFlag.set_telescope_params` method, similar to the ones on UVData and UVCal,
 to set several of these new parameters.
-- A new `FastUVH5Meta` object that enables quick partial reading of uvh5 file metadata.
 
 ### Changed
 - Increases the h5py minimum version to 3.1.
 - Increase the minimum compatible version for lunarsky to 0.2.1 to fix astropy
 deprecation warnings.
-
 
 ### Fixed
 - Fixed some bugs related to handling of the `UVCal.time_range` parameter in
--- a/ci/pyuvdata_min_deps_tests.yml
+++ b/ci/pyuvdata_min_deps_tests.yml
@@ -3,9 +3,8 @@ channels:
   - conda-forge
 dependencies:
   - astropy>=5.0.4
-  - numpy>=1.19.*
-  - scipy>=1.3
   - h5py>=3.1
+  - h5py>=3.0
   - pyerfa>=2.0
   - coverage
   - pytest>=6.2.0
--- a/ci/pyuvdata_min_deps_tests.yml
+++ b/ci/pyuvdata_min_deps_tests.yml
@@ -4,6 +4,7 @@ channels:
 dependencies:
   - astropy>=5.0.4
   - h5py>=3.1
+  - scipy>=1.5
   - numpy>=1.20.*
   - pyerfa>=2.0
   - coverage
--- a/ci/pyuvdata_min_deps_tests.yml
+++ b/ci/pyuvdata_min_deps_tests.yml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - astropy>=5.0.4
   - h5py>=3.1
-  - h5py>=3.0
+  - numpy>=1.20.*
   - pyerfa>=2.0
   - coverage
   - pytest>=6.2.0
--- a/ci/pyuvdata_publish.yml
+++ b/ci/pyuvdata_publish.yml
@@ -4,8 +4,8 @@ channels:
 dependencies:
   - astropy>=5.0.4
   - numpy>=1.19.*
-  - scipy>=1.3
   - h5py>=3.1
+  - scipy>=1.5
   - pyerfa>=2.0
   - cython
   - setuptools_scm<7.0|>=7.0.3
--- a/ci/pyuvdata_publish.yml
+++ b/ci/pyuvdata_publish.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - astropy>=5.0.4
-  - numpy>=1.19.*
+  - numpy>=1.20.*
   - h5py>=3.1
   - scipy>=1.5
   - pyerfa>=2.0
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -5527,11 +5527,19 @@ def determine_blt_order(
     for i, (t, a, b, bl) in enumerate(
         zip(times[1:], ant1[1:], ant2[1:], bls[1:]), start=1
     ):
+        on_bl_boundary = i % Nbls == 0
+        on_time_boundary = i % Ntimes == 0
+
         if t < times[i - 1]:
             time_bl = False
             time_a = False
             time_b = False
             time_order = False
+
+            if not on_time_boundary:
+                bl_time = False
+                a_time = False
+                b_time = False
 
             if bl == bls[i - 1]:
                 bl_time = False
@@ -5551,12 +5559,18 @@ def determine_blt_order(
         if bl < bls[i - 1]:
             bl_time = False
             bl_order = False
+            if not on_bl_boundary:
+                time_bl = False
         if a < ant1[i - 1]:
             a_time = False
             a_order = False
+            if not on_bl_boundary:
+                time_a = False
         if b < ant2[i - 1]:
             b_time = False
             b_order = False
+            if not on_bl_boundary:
+                time_b = False
 
         if not any(
             (
@@ -5649,43 +5663,47 @@ def determine_rectangularity(time_array, baseline_array, nbls, ntimes):
     bl_first = True
 
     if time_array.size != nbls * ntimes:
+        print("time_array.size != nbls * ntimes")
         return False, False
 
     if nbls * ntimes == 1:
         return True, True
 
     # check if we're moving time first
-    for i in range(1, nbls):
+    for i in range(1, ntimes):
         if time_array[i] == time_array[0]:
             time_first = False
             break
 
     # check if we're moving bl first
     bl0 = baseline_array[0]
-    for i in range(1, ntimes):
+    for i in range(1, nbls):
         if baseline_array[i] == bl0:
             bl_first = False
             break
 
     if not time_first and not bl_first:
+        print("not time_first and not bl_first")
         return False, False
 
     # Now do a proper check.
     if time_first:
-        for i in range(nbls):
+        for i in range(ntimes):
             if (
                 np.unique(time_array[i::ntimes]).size != 1
                 or np.unique(baseline_array[i::ntimes]).size != nbls
             ):
+                print("unique times != 1 or unique baselines != nbls")
                 is_rect = False
                 break
 
-    if bl_first:
-        for i in range(ntimes):
+    elif bl_first:
+        for i in range(nbls):
             if (
                 np.unique(baseline_array[i::nbls]).size != 1
                 or np.unique(time_array[i::nbls]).size != ntimes
             ):
+                print("unique baselines != 1 or unique times != ntimes")
                 is_rect = False
                 break
 
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -5612,3 +5612,87 @@ def determine_blt_order(
         return ("time",)
 
     return None
+
+
+def determine_rectangularity(time_array, baseline_array, nbls, ntimes):
+    """Determine if the data is rectangular or not.
+
+    Parameters
+    ----------
+    time_array : array_like
+        Array of times in JD.
+    baseline_array : array_like
+        Array of baseline integers.
+    nbls : int
+        Number of baselines.
+    ntimes : int
+        Number of times.
+
+    Returns
+    -------
+    is_rect : bool
+        True if the data is rectangular, False otherwise.
+    time_first : bool
+        True if the data is rectangular and time moves first (i.e. is "last axis").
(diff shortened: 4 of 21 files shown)


steven-murray:
I've done some initial performance tests. I'm using a HERA H6C 2-integration file (many baselines, only 2 times). 

I'm running a few different scenarios:

1. Find the minimum LST in the file
2. Find the number of ant-pairs
3. Find the number of antennas

I'm running these scenarios in a few different ways:

1. With OLD pyuvdata, using `UVData` and all checks
2. With OLD pyuvdata using `UVData` and no extra checks
3. With NEW pyuvdata using `UVData` (using the meta reader) and all checks
4. With NEW pyuvdata using `UVData` (using the meta reader) and no extra checks
5. With NEW pyuvdata using bare `FastUVH5Meta`
6. With NEW pyuvdata using bare `FastUVH5Meta` and blt rectangular info

## Results

### Minimum LST
1. OLD / UVData / all checks: 49.6 ms
2. NEW / UVData / all checks: 81.5 ms
3. NEW / FastUVH5Meta / no blt info: 5.22 ms
4. NEW / FastUVH5Meta / blt info: 4.09 ms

### Number of antpairs
1. OLD / UVData / all checks: 57 ms
2. OLD / UVData / no checks: 30.4 ms
3. OLD / UVData / no extra checks: 54 ms
4. NEW / UVData / all checks: 77.8 ms
5. NEW / FastUVH5Meta / no blt info: 6.39 ms
6. NEW / FastUVH5Meta / blt info: 3.42 ms

### Number of antennas
2. OLD / UVData / no checks: 27.4 ms
6. NEW / FastUVH5Meta / blt info: 2.12 ms

## Conclusions

Generating a full `UVData` metadata-only object using the new code is about ~2 times as slow as the old code (but when reading a full file, this is typically a negligible fraction of the full cost). However, reading a small part of the data to make a decision is MUCH faster in the new code (by an order of magnitude).

pyuvdata/uvdata/uvh5.py
```diff
@@ -189,256 +194,619 @@ def _get_compression(compression):
...
+        if "blt_order" in h:
+            blt_order_str = bytes(h["blt_order"][()]).decode("utf8")
+            return tuple(blt_order_str.split(", "))
+        else:
+            if self.__blt_order == "determine":
+                return self.get_blt_order()
+            else:
+                return None
+
+    def get_blt_order(self) -> tuple[str]:
```
bhazelton, line 351: Can we move this above the `blt_order` method so that it is defined before it is called?

pyuvdata/uvdata/uvh5.py
```diff
@@ -189,256 +194,619 @@ def _get_compression(compression):
...
+                phase_center_catalog[pc_id] = pc_dict
+
+        return phase_center_catalog
+
+    @cached_property
+    def phase_type(self) -> str:
+        """The phase type of the data."""
+        h = self.header
+        if "phase_type" not in h:
+            return "drift"
```
bhazelton, line 444: I'm not sure this is what we want. In new "phase_type" will not be present (it's being replaced by the phase center catalog) but that doesn't mean it's a drift type.

I'd suggest checking if phase_center_catalog is present and if it is, this should return "drift" only if all the "cat_type"s are "unprojected", otherwise return "phased".

pyuvdata/uvdata/uvh5.py
```diff
@@ -189,256 +194,619 @@ def _get_compression(compression):
...
+            "phase_center_frame_pa",
+            "extra_keywords",
+        ]:
+            try:
+                setattr(self, attr, getattr(obj, attr))
+            except AttributeError:
+                pass
+
+            # if hasattr(obj, attr):
+            #     setattr(self, attr, getattr(obj, attr))
```
bhazelton, line 770: This should probably be removed?

pyuvdata/uvdata/uvh5.py
```diff
@@ -447,40 +815,41 @@ def _read_header(
...
+    ):
+        """
+        Read header information from a UVH5 file.
+
+        This is an internal function called by the user-space methods.
+        Properties of the UVData object are updated as the file is processed.
 
-        if proc is not None:
-            # if lsts are in the background wait for them to return
-            proc.join()
```
bhazelton, line 481: I guess the way that you've set up FastUVH5Meta makes it hard to run the lst_array conversion as a process, but it'd be nice not to lose that option when you want to read the whole header. If there's no clean way we can live with this, but running it in a process does help speed things up.

bhazelton (commented):
This is looking good to me, but I'd like @plaplant to weigh in as well.

bhazelton:
I am slightly concerned about the slowdown on reading the full metadata, it'd be nice to figure out why it is slower.

steven-murray:
I've updated a few things on the branch (one less `h5py.File` opening and a couple of tweaks making if-statements be try/excepts, which should be faster, I think). I also found that I was testing performance against the wrong version of pyuvdata -- I believe I was testing against 2.2.12 instead of `main` (which is what this PR is based off, of course). With those things updated, I get the following:


### Minimum LST
1. OLD / UVData / all checks: 66.8
2. NEW / UVData / all checks: 72.6 ms
3. NEW / FastUVH5Meta / no blt info: 4.62  ms
4. NEW / FastUVH5Meta / blt info: 3.56 ms

### Number of antpairs
1. OLD / UVData / all checks: 69.4 ms
2. OLD / UVData / no checks: 32.7 ms
3. OLD / UVData / no extra checks: 66.1 ms
4. NEW / UVData / all checks: 74.6 ms
5. NEW / FastUVH5Meta / no blt info: 5.52 ms
6. NEW / FastUVH5Meta / blt info: 3.31 ms

### Number of antennas
2. OLD / UVData / no checks: 31.7 ms
6. NEW / FastUVH5Meta / blt info: 2.1 ms

## Conclusions

Generating a full `UVData` metadata-only object using the new code is about ~5% slower than the old code. However, reading a small part of the data to make a decision is MUCH faster in the new code (by an order of magnitude). The 5% is pretty much negligible, and can probably be sped up by making *either* branch faster in certain ways.

pyuvdata/uvdata/uvh5.py
steven-murray: Good point. I added it back in.

CHANGELOG.md
```diff
@@ -17,11 +17,13 @@ from them will inherit them. UVFlag objects will also inherit these parameters w
 they are converted between types using UVData and UVCalobjects.
 - The `UVFlag.set_telescope_params` method, similar to the ones on UVData and UVCal,
 to set several of these new parameters.
+- A new `FastUVH5Meta` object that enables quick partial reading of uvh5 file metadata.
```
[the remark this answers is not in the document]
steven-murray: I always do this...

CHANGELOG.md
```diff
@@ -17,11 +17,13 @@ from them will inherit them. UVFlag objects will also inherit these parameters w
...
 
 ### Changed
 - Increase the minimum compatible version for lunarsky to 0.2.1 to fix astropy
 deprecation warnings.
 
+
```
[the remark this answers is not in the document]
steven-murray: Most likely a rebase artifact. Fixed.

pyuvdata/utils.py
```diff
@@ -5482,3 +5482,200 @@ def _index_dset(dset, indices, input_array=None):
...
+                or np.unique(time_array[i::nbls]).size != ntimes
+            ):
+                is_rect = False
+                break
+
+    if not is_rect:
+        return False, False
+
+    if time_first and bl_first:
+        raise ValueError("Something went wrong when determining rectangularity.")
```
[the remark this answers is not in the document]
steven-murray: Yeah, this is essentially a check that the logic is correct. It would be better covered by tests I guess, but I can't convince myself that I covered every single corner case, so having this will prompt a user to complain to us so we can fix the logic :-)

pyuvdata/utils.py
```diff
@@ -5482,3 +5482,200 @@ def _index_dset(dset, indices, input_array=None):
...
+        or (time_order and a_order)
+        or (time_order and b_order)
+        or (a_order and b_order)
+        or (time_order and bl_order)
+    ):
+        raise ValueError(
+            "Something went horribly wrong when trying to determine the order of "
+            "the blts axis."
+            "Please raise an issue on github, as this is not meant to happen."
+        )
```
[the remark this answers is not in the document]
steven-murray: No idea how to get here, just a final check to alert the user that something went wrong if we missed a corner case.

pyuvdata/utils.py
```diff
@@ -5482,3 +5482,200 @@ def _index_dset(dset, indices, input_array=None):
...
+    ntimes : int
+        Number of times.
+
+    Returns
+    -------
+    is_rect : bool
+        True if the data is rectangular, False otherwise.
+    time_first : bool
+        True if the data is rectangular and time moves first (i.e. is "last axis").
+        False either if baseline moves first, OR if it is not rectangular.
```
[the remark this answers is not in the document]
steven-murray: OK yeah this is confusing. I actually mean the opposite -- `time_first=True` means the time is changing faster than bl (i.e. "time changes first"), which means its the last axis. Would a variable name change be useful? I'm struggling to think of a better description.

pyuvdata/uvdata/uvh5.py
steven-murray: Yup, done.

pyuvdata/uvdata/uvh5.py
steven-murray: Thank you, I was really struggling with the logic implied by the earlier code.

steven-murray: Can you ensure the logic now is correct?

pyuvdata/uvdata/uvh5.py
steven-murray: Yes, done.

pyuvdata/utils.py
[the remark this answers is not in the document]
bhazelton: ok, then I'd advocate for an assert.

pyuvdata/utils.py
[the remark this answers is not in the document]
bhazelton: sounds good. Then I'd suggest using an assert with an error message.

pyuvdata/utils.py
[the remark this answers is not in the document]
steven-murray: Yeah, that's right. There's a difference between what "moves" first and which axis gets listed first (from left to right). I could call this `time_moves_first`?

pyuvdata/utils.py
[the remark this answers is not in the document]
steven-murray: Yup, done

pyuvdata/utils.py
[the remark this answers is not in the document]
bhazelton: I think I'd prefer `time_fastest`.

pyuvdata/uvdata/uvh5.py
bhazelton: It looks correct to me now.

steven-murray:
OK, @bhazelton, @mkolopanis and @plaplant I think this is now ready for review again. 

Since Bryna's last comprehensive review, I've added in many more tests, as well as the ability to write out the `blts_are_rectangular` and `time_axis_faster_than_bls` keywords as metadata into UVH5 files. I also added a `set_rectangularity()` method on `UVData` objects which can go and set these attributes based on the time array and baseline array in the object. It's also helpful for calling inside the `reorder_blts` and `select` methods, which can otherwise make the object be invalid wrt these attributes. 

I *don't* auto-set these attributes when reading UVH5 files (or otherwise), but it may be a good thing to do.

ci/pyuvdata_publish.yml
```diff
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - astropy>=5.0.4
   - numpy>=1.19.*
```
bhazelton, line 6: should numpy be bumped here as well?

pyuvdata/uvdata/uvh5.py
```diff
@@ -189,259 +194,689 @@ def _get_compression(compression):
...
+            "integration_time",
+            "freq_array",
+            "spw_array",
+            "channel_width",
+            "polarization_array",
+            "uvw_array",
+            "channel_width",
+            "phase_center_catalog",
+            # "blts_are_rectangular",
+            # "time_axis_faster_than_bls",
```
bhazelton, line 804: should these commented lines be removed?

bhazelton: should they actually appear below in the optional parameters?

bhazelton (commented):
This is looking pretty good to me, just a few small comments.

We should mention the increased version requirements for numpy/scipy in the changelog.

bhazelton:
We fixed the known bug with the warnings tests on main, so they should be passing now that you've rebased. Maybe just need to add some filtering?

pyuvdata/uvdata/uvdata.py
```diff
@@ -9333,6 +9417,28 @@ def select(
...
+        # elif self.blts_are_rectangular is None:
+        #     self.time_axis_faster_than_bls = None
+
+        # if self.time_axis_faster_than_bls is not None:
+        #     if self.Ntimes == 1 and self.Nbls > 1:
+        #         self.time_axis_faster_than_bls = False
+        #     elif self.Ntimes > 1 and self.Nbls == 1:
+        #         self.time_axis_faster_than_bls = True
+        #     elif self.Ntimes == 1 and self.Nbls == 1:
+        #         self.time_axis_faster_than_bls = True
```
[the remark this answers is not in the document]
steven-murray: Nope. Fixed.

pyuvdata/uvdata/uvh5.py
steven-murray: No, they always exist, so they would always be set. I just removed them from being set in this PR as it changes a few tests. They can be set later by the user on the UVData object if desired. If you think we should set them by default, I can also do that.

ci/pyuvdata_publish.yml
steven-murray: Yes, done.

steven-murray:
@bhazelton I've addressed your comments, and fixed the warnings test.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- A new `FastUVH5Meta` object that enables quick partial reading of uvh5 file metadata.
```
bhazelton, line 7: We need to detail the updated numpy & scipy requirements in the changelog. That's the thing I check when making conda packages.

pyuvdata/uvdata/uvh5.py
bhazelton: I'm confused. Why will they always exist? They wouldn't exist in older files.

bhazelton:
@steven-murray  we talked about this today in the telecon. We're comfortable moving on this with just my review. It does need to be rebased (I can help if you want!) and I had a couple more comments just now.

CHANGELOG.md
bhazelton: Also, make sure all your changelog changes end up in the `Unreleased` section after you rebase because we recently released v2.3

pyuvdata/uvdata/uvh5.py
steven-murray: So in this bit of code, we go through this list of attributes, and for each one that exists on the `FastUVH5Meta` object, it sets it on the `UVData` object. The attributes in the `FastUVH5Meta` are dynamically-defined by the `__getattr__` method for the most part, which just goes and reads them from the file itself. However, these two attributes (`blts_are_rectangular` and `time_axis_faster_than_bls`) are cached properties on the `FastUVH5Meta` object -- they're always defined, because if they're not defined, they're calculated from the other properties. It *probably* makes sense to calculate them and set them on the UVData object, since it can make other calculations faster, but I didn't want to do this by default in this PR, since it does take a little more time to init, and also it means I have to change some tests (eg. reading a file, then writing the file and reading the written one results in a *different* final object because these new attributes are set).

ci/pyuvdata_min_deps_tests.yml
```diff
@@ -3,9 +3,8 @@ channels:
   - conda-forge
 dependencies:
   - astropy>=5.0.4
-  - numpy>=1.19.*
-  - scipy>=1.3
```
bhazelton, line 7: I don't think we want to drop scipy from this yaml. Probably happened in a rebase.

pyuvdata/uvdata/uvh5.py
bhazelton: Ok, I think I follow. I've run into this kind of thing before where you can't really check if something is defined because the machinery calculates it if you ask for it. But looking at this method, I don't see where it is set on the UVH5 object if it is known from the file or has already been determined. Are we just not setting it at all even if it's in the file?

bhazelton (commented):
It looks like a few problems popped up after the rebase. The coverage has dropped and there are some errors in the warnings tests. I'm happy to help if you need, I made most of the changes you rebased on top of.

ci/pyuvdata_min_deps_tests.yml
steven-murray: ugh. Will fix.

pyuvdata/uvdata/uvh5.py
steven-murray: @bhazelton You make a good point -- I wasn't setting it, even if it existed in the file itself. I've fixed that now. It will be set if and only if it exists in the file header. Otherwise, the user can set them manually on the UVData object (or, if using the `FastUVH5Meta` object directly, can access them dynamically).

bhazelton: sounds good, thanks!

bhazelton (approved):
Looks good to me.

PR #1273, opened 2023-03-14
Fix the publish to pypi CI so that it will actually publish when there's a release

bhazelton:
## Description

I changed the way we're testing to see if the action was triggered on a new tag.

From the docs (https://docs.github.com/en/actions/learn-github-actions/contexts#github-context), `github.ref` is the correct thing to use to detect that a commit is associated with a new tag.

We were using `github.event.ref` which apparently worked when the triggering event was a push but not when the triggering event was a release.

## Motivation and Context

The last release was not actually published to pypi because the logic testing if it was a new tag failed.

## Types of changes

- [ ] Version change

## Checklist:

### DIFF
--- a/.github/workflows/publish-to-test-pypi.yaml
+++ b/.github/workflows/publish-to-test-pypi.yaml
@@ -75,14 +75,14 @@ jobs:
             rmdir $dir
           done
       - name: Upload to test PyPI
-        if: startsWith(github.event.ref, 'refs/tags')
+        if: startsWith(github.ref, 'refs/tags')
         uses: pypa/gh-action-pypi-publish@v1.5.0
         with:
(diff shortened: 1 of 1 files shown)


steven-murray (approved):
This looks fine to me, but I'm not *completely* certain it's correct. In any case, it probably won't harm things, and we should keep an eye on how releases happen in the next bit.

bhazelton:
@steven-murray do you have a suggestion to make it more correct? Or are you just having trouble parsing the docs (I did too)? I pointed to the docs reference, but I actually originally identified this solution based on this stack overflow (which didn't fill me with confidence, but the combination convinced me): https://stackoverflow.com/questions/59349905/which-properties-does-github-event-in-a-github-workflow-have

PR #1276, opened 2023-03-28
fix: allow reading of old uvh5 files that were silently broken

steven-murray:
## Description

This adds the option to tell `FastUVH5Meta` to recompute `Nbls` from the `baseline_array` (instead of just getting it directly from metadata).

## Motivation and Context

The problem was that older versions of `pyuvdata` directly set `UVData.Nbls` (from uvh5 files) using `np.unique(self.baseline_array).size`, which was guaranteed to always be the true number of baselines. This meant that one could write a UVH5 file with *incorrect metadata*, and it would still read perfectly fine. This in turn means that any old files have the possibility of being wrongly defined without knowing it. In fact, as it turns out, all HERA data before last month or so had this problem (they defined `Nbls=Nblts`). Since the new `FastUVH5Meta` code reads the file metadata correctly, those old files throw errors upon `.check()`. Thus, the newest pyuvdata on `main` right now will make reading all previous HERA files impossible. 

The fix here just allows the user to pass `recompute_nbls=True` to ensure that they are correct on read, using `len(np.unique(self.baseline_array))`. By default, this is set to None, and the Nbls are recomputed only for old HERA files. If the user passes `blts_are_rectangular`, the `Nbls` can be computed faster by `self.Nblts/self.Ntimes` (if this is asked for).

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/uvdata/tests/test_uvh5.py
+++ b/pyuvdata/uvdata/tests/test_uvh5.py
@@ -3702,6 +3702,11 @@ class TestFastUVH5Meta:
         meta4 = uvh5.FastUVH5Meta(newfl, recompute_nbls=None)
         assert meta4.Nbls == meta.Nbls
 
+        meta5 = uvh5.FastUVH5Meta(
+            self.fl, recompute_nbls=True, blts_are_rectangular=True
+        )
+        assert meta5.Nbls == meta.Nbls
+
     def test_pickleability(self):
         meta = uvh5.FastUVH5Meta(self.fl)
         meta2 = deepcopy(meta)
--- a/pyuvdata/uvdata/tests/test_uvh5.py
+++ b/pyuvdata/uvdata/tests/test_uvh5.py
@@ -3674,25 +3674,30 @@ class TestFastUVH5Meta:
         assert not meta1.blts_are_rectangular
         assert not meta1.time_axis_faster_than_bls
 
-    def test_nbl_function(self):
-        meta = uvh5.FastUVH5Meta(self.fl)
+    def test_recompute_nbls(self):
+        meta = uvh5.FastUVH5Meta(self.fl, recompute_nbls=False)
+        meta2 = uvh5.FastUVH5Meta(self.fl, recompute_nbls=True)
+
         nbls = meta.Nbls
 
         newfl = os.path.join(self.tmp_path.name, "wrong_Nbls.uvh5")
         uvd = meta.to_uvdata()
-        uvd.Nbls = nbls * 2
+        uvd.Nbls = uvd.Nblts
+        uvd.Ntimes = uvd.Nblts // nbls
+        uvd.telescope_name = "HERA"
         uvd.initialize_uvh5_file(newfl, clobber=True)
         meta.close()
 
-        # This won't run the nbl function because the version is >1.1
-        meta1 = uvh5.FastUVH5Meta(newfl, nbl_function=lambda obj: nbls)
-        assert meta1.Nbls == 2 * nbls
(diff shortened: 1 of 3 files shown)


bhazelton:
Why are you making the user pass in a function rather than setting a boolean to properly calculate it from the object metadata?

bhazelton:
I think we should also default it to do the right thing if we detect that it's HERA data with an older file version.

bhazelton:
we probably also need to add something to the error message that results from this so users have a clue how to fix it.

pyuvdata/uvdata/tests/test_uvh5.py
```diff
@@ -3673,3 +3673,31 @@ def test_rectangularity_roundtrip(self):
...
+        uvd = meta.to_uvdata()
+        uvd.Nbls = uvd.Nblts
+        uvd.Ntimes = uvd.Nblts // nbls
+        uvd.telescope_name = "HERA"
+        uvd.initialize_uvh5_file(newfl, clobber=True)
+        meta.close()
+
+        meta3 = uvh5.FastUVH5Meta(newfl, recompute_nbls=None)
+
+        print(meta.Nbls, meta2.Nbls, meta3.Nbls)
```
bhazelton, line 3693: this print should probably be removed.

pyuvdata/uvdata/uvh5.py
```diff
@@ -795,7 +875,6 @@ def _read_header_with_fast_meta(
             "ant_1_array",
             "ant_2_array",
             "phase_center_id_array",
-            "Nbls",
```
bhazelton, line 798: Why did you take `Nbls` out of this list?

pyuvdata/uvdata/uvh5.py
```diff
@@ -1577,7 +1665,7 @@ def _write_header(self, header):
             "https://github.com/RadioAstronomySoftwareGroup/pyuvdata/issues"
         )
         assert self.future_array_shapes, assert_err_msg
-        header["version"] = np.string_("1.1")
+        header["version"] = np.string_("1.2")
```
bhazelton, line 1668: thanks for fixing this.

pyuvdata/uvdata/uvh5.py
```diff
@@ -312,12 +321,56 @@ def __init__(
...
+                "datagrp",
+            )
+        }
+        return state
+
+    def __setstate__(self, state):
+        """Set the state of the object."""
+        self.__dict__.update(state)
+        self.__file = None
+        self.open()
```
bhazelton, line 351: can you explain the purpose of this to me?

bhazelton (commented):
I think this is getting close but there are a bunch of new uncovered lines of code.

pyuvdata/uvdata/tests/test_uvh5.py
steven-murray: Ah yep. Done.

pyuvdata/uvdata/uvh5.py
steven-murray: It was actually in the list twice.

pyuvdata/uvdata/uvh5.py
steven-murray: So the `__getstate__` / `__setstate__` are run before and after pickling. The `self.__file` attribute is not pickleable, so in `__getstate__` I remove it, and in `__setstate__` I restore it from the path. Having the class be pickleable is really useful for doing things like deepcopy and multiprocessing.

bhazelton (approved):
Looks good, thank you!

PR #1282, opened 2023-04-04
Improved UVData to MirParser interface

kartographer:
Updates the interface between `UVData.read` method and the `MirParser` module.

## Description

Overhauls the `read_mir` method inside of `Mir`. This includes being able to sub-select desired data on read (rather than after-the-fact), with keywords matched to those typically used inside of `UVData.read`. Additionally, some updates to the `MirParser` and `Mir` modules have been made to improve speed/memory performance (by up to a factor of two). The `UVData.read` and `UVData.select` have also had the `catalog_names` keyword added, which allows users to select sources/phase centers by their names rather than their ID numbers (the latter of which are sometimes arbitrarily assigned).

## Motivation and Context

This PR covers a final suite of changes that are needed for moving support for MIR data for general users of SMA -- in this case, primarily via improving the interface between the `UVData` and MIR-type files. Previously, in order to handle MIR files _without_ loading the whole dataset first, one needed to pass forward specific indexing codes that were not necessarily known by lay-users of the telescope.

After this PR is closed, I'd like to request a new pyuvdata version be spun up.

## Types of changes

## Checklist:

Bug fix checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/uvdata/mir.py
+++ b/pyuvdata/uvdata/mir.py
@@ -33,12 +33,10 @@ class Mir(UVData):
         antenna_nums=None,
         antenna_names=None,
         bls=None,
-        times=None,
         time_range=None,
-        lsts=None,
         lst_range=None,
         polarizations=None,
-        sources=None,
+        catalog_names=None,
         corrchunk=None,
         receivers=None,
         sidebands=None,
@@ -118,26 +116,21 @@ class Mir(UVData):
                 select_where += [
                     ("blcd", "eq", ["%i-%i" % (tup[0], tup[1]) for tup in bls])
                 ]
-            if times is not None:
-                # Have to convert times from UTC JD -> TT MJD for mIR
-                select_where += [
-                    ("mjd", "eq", Time(times, format="jd", scale="utc").tt.mjd)
-                ]
             if time_range is not None:
                 # Have to convert times from UTC JD -> TT MJD for mIR
                 select_where += [
-                    ("mjd", "btw", Time(time_range, format="jd", scale="utc").tt.mjd)
+                    (
+                        "mjd",
+                        "between",
+                        Time(time_range, format="jd", scale="utc").tt.mjd,
+                    )
                 ]
-            if time_range is not None:
-                select_where += [("lst", "eq", time_range)]
-            if lsts is not None:
-                select_where += [("lst", "eq", lsts)]
             if lst_range is not None:
-                select_where += [("lst_range", "btw", lst_range)]
+                select_where += [("lst_range", "between", lst_range)]
             if polarizations is not None:
                 select_where += [("pol", "eq", polarizations)]
-            if sources is not None:
-                select_where += [("source", "eq", sources)]
+            if catalog_names is not None:
+                select_where += [("source", "eq", catalog_names)]
             if receivers is not None:
                 select_where += [("rec", "eq", receivers)]
             if sidebands is not None:
--- a/pyuvdata/uvdata/mir.py
+++ b/pyuvdata/uvdata/mir.py
@@ -12,7 +12,7 @@ from astropy.time import Time
 from .. import get_telescope
 from .. import utils as uvutils
 from . import mir_parser
-from .uvdata import UVData, _future_array_shapes_warning
+from .uvdata import UVData
 
 __all__ = ["Mir"]
 
@@ -29,7 +29,6 @@ class Mir(UVData):
     def read_mir(
         self,
         filepath,
-        select_where=None,
         antenna_nums=None,
         antenna_names=None,
         bls=None,
@@ -40,6 +39,7 @@ class Mir(UVData):
         corrchunk=None,
         receivers=None,
         sidebands=None,
+        select_where=None,
         apply_tsys=True,
         apply_flags=True,
         apply_dedoppler=False,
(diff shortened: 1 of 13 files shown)


pyuvdata/uvdata/mir.py
```diff
@@ -589,16 +764,39 @@ def _init_from_mir_parser(
...
+                icrs_ra, icrs_dec = uvutils.transform_app_to_icrs(
+                    time_arr,
+                    mir_data.in_data["ara"][source_mask],
+                    mir_data.in_data["adec"][source_mask],
+                    self.telescope_location_lat_lon_alt,
+                )
+                self._add_phase_center(
+                    mir_data.codes_data["source"][sou_id],
+                    cat_type="sidereal",
+                    cat_lon=np.median(icrs_ra),
```
e-koch, line 783: For future proofing, I wonder about setting a check that `np.ptp` of the ra, dec varies less than some reasonable amount.

I think this would only matter for concatenated tracks (in which case, is that already handled when a phase center has the same name but different coordinates?).

pyuvdata/uvdata/mir_meta_data.py
```diff
@@ -549,14 +559,14 @@ def __iter__(self):
...
-        Calculate the number entries in the data table.
+        Calculate the number of unmasked entries in the data table.
 
         Returns
         -------
         len : int
             Number of unique entries contained within the meta data.
         """
-        return self._data.size
+        return np.sum(self._mask)
```
e-koch, line 569: How many times does this get called when creating the `UVData` object in a normal `UVData.read` call? Wondering if it's worth having this cached.

pyuvdata/uvdata/mir_meta_data.py
```diff
@@ -1144,6 +1154,8 @@ def get_value(
         idx_arr = self._index_query(use_mask, where, and_where_args, header_key, index)
 
         if isinstance(field_name, (list, set, tuple)):
+            if return_tuples is None:
```
e-koch, line 1157: Why is the default changed to `None`? Can the kwarg in the function call just be `True`?

pyuvdata/uvdata/mir_parser.py
```diff
@@ -1444,11 +1466,22 @@ def unload_data(self, unload_vis=True, unload_raw=True, unload_auto=True):
...
+            if self.vis_data is not None:
+                for item in self.vis_data.values():
+                    del item["data"]
+                    del item["flags"]
+                self.vis_data = None
             self._tsys_applied = False
-        if unload_raw:
+        if unload_raw and self.raw_data is not None:
+            for item in self.raw_data.values():
+                print(item.keys())
```
e-koch, line 1477:
```suggestion
               
```

pyuvdata/uvdata/tests/test_mir.py
```diff
@@ -679,3 +624,80 @@ def test_rechunk_on_read():
...
+        {"lst_range": [2, 2.5]},
+        {"polarizations": ["hh", "vv"]},
+        {"catalog_names": ["3c84"]},
+        {"corrchunk": [1, 2, 3, 4]},
+        {"receivers": ["230", "240"]},
+        {"sidebands": ["l", "u"]},
+    ],
+)
+def test_select_on_read(select_kwargs, sma_mir):
+    print(select_kwargs)
```
e-koch, line 645: Extra print?

e-koch (changes requested):
@kartographer -- Found just a few minor things!

pyuvdata/uvdata/mir.py
kartographer: Fair point! I've added a check immediately after the code highlighted here, which raises a warning if a change of > 1 arcmin is seen (> 1 primary beam for SMA, which is _pretty_ uncommon for standard sources).

pyuvdata/uvdata/mir_parser.py
kartographer: Deleted!

pyuvdata/uvdata/tests/test_mir.py
kartographer: Yup -- something I forgot to remove after debugging.

pyuvdata/uvdata/mir_meta_data.py
kartographer: So I'm using `None` here as a stand-in for what the typical usage is inside of `MirParser`, which is that if multiple fields have been requested, to return tuples (as they're often used for indexing via hashmap), but in cases where single field is requested, just to return the array of values. This is more of a convenience thing inside of `MirParser` rather than anything else -- it just helped to remove about two dozen extra keyword calls inside the code.

kartographer:
Thanks @e-koch! I think I've responded to everything you've commented on, but let me know if you see something that needs further attention.

pyuvdata/uvdata/mir_meta_data.py
e-koch: 👍 Makes sense!

e-koch (dismissed):

bhazelton (commented):
Looks good to me, just a couple questions

pyuvdata/uvdata/uvdata.py
```diff
@@ -12280,10 +12352,13 @@ def read(
         phase_to_pointing_center=False,
         nsample_array_dtype=np.float32,
         # MIR
-        isource=None,
-        irec=None,
-        isb=None,
```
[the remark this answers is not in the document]
kartographer: I don't think so -- these were basically options for some advanced users to be able to conduct testing, but I don't know of anyone actually using them on the regular (since they're a bit... esoteric). I'd be in favor of clearing them out completely just to avoid confusion once we start broadcasting support for MIR via pyuvdata more broadly.

pyuvdata/uvdata/uvdata.py
```diff
@@ -12219,6 +12290,7 @@ def read(
         antenna_names=None,
         ant_str=None,
         bls=None,
+        catalog_names=None,
```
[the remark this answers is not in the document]
kartographer: Yeah, I think that would be easy enough to tackle -- certainly for `uvh5` at a minimum, and I'll take a quick look at `uvfits`.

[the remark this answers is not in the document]
kartographer: Okay, added to both `uvh5` and `uvfits`. Adding it to `ms` is a bit trickier since there's actually no _current_ infrastructure for doing any sort of "select-on-read" (due in part to MS/casacore having it's own SQL-like language for doing such things). That's going to be a bigger job, so I'd favor making that into an issue separate from this PR.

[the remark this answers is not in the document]
kartographer: Issue added -- #1283

PR #1284, opened 2023-04-07
Cal init method

steven-murray:
## Description

This builds on #1281 and adds a similar initializer for `UVCal`.

## Motivation and Context

The same motivations as in #1281

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:


steven-murray:
@bhazelton I've done everything I wanted to do with this branch, except that documentation might be a bit dodgy. I think it will be a lot easier to include the `initializers` module in the docs directly, so it can be referred to from the `UVCal` docs. Otherwise, I can try to do a little magic in including one doc string into another (but this is a bit harder in this case than in the `UVData.new()` case, because here I have some parameters that have to be dealt with specially in the `initialize_from_uvdata` method, to maintain backwards compat). Otherwise, I'd have to copy-paste a whole bunch of parameter definitions into three different places. I know that's common practice in pyuvdata, but I don't like it at all.

I did try to keep backwards-compatibility as much as I could in the end. I think there are some cases now that *won't* raise an error where they did before (eg. you don't have to pass `channel_widths` when passing `freq_array` because it'll auto-calculate). So I removed some tests that are now irrelevant. But I kept all input arguments, some with new warnings.

pyuvdata/uvcal/uvcal.py
```diff
@@ -561,6 +563,45 @@ def __init__(self):
...
         super(UVCal, self).__init__()
 
+    @staticmethod
+    def new(**kwargs):
+        """
+        Create a new UVCal object.
+
+        All parameters are passed through to the
+        :func:`~pyuvdata.uvcal.initializers.new_uvcal` function.
+        """
```
bhazelton, line 573: I'd prefer to do the same thing you did in the UVData function so that the docs will show all the options here.

pyuvdata/uvcal/uvcal.py
```diff
@@ -561,6 +563,45 @@ def __init__(self):
...
+        return initializers.new_uvcal(**kwargs)
+
+    @classmethod
+    def from_uvdata(cls, uvd: UVData, **kwargs):
+        """
+        Create a new UVCal object from a UVData object.
+
+        All parameters are passed through to the
+        :func:`~pyuvdata.uvcal.initializers.new_uvdata` function, but those that can
```
bhazelton, line 582: Shouldn't this point to the `new_uvcal` function not the `new_uvdata` function? I get why you don't want to copy/paste docstrings here, my first choice would be to get all the parameters in like you did on uvdata, but if that's not possible maybe we can instead point to the `new` method (which will have all the options) so that we are just sending users to another method on the same object?

bhazelton: And are we really going to have both a `from_uvdata` and an `initialize_from_uvdata` class method on UVCal long term? I'd like us to go with one or the other. If this is the one we're going with, the other one should have a deprecation warning if it is called.

pyuvdata/uvcal/initializers.py
```diff
@@ -0,0 +1,529 @@
...
+    jones_array : ndarray of int or str
+        Array of Jones polarization integers. If a string, options are 'linear' or
+        'circular' (which will be converted to the appropriate Jones integers as
+        length-4 arrays).
+    delay_array : ndarray of float, optional
+        Array of delays in seconds. If cal_type is not provided, it will be set to
+        'delay' if delay_array is provided.
+    cal_type : str, optional
+        Calibration type. Options are 'delay', 'gain', or 'unknown'. Not required
+        if delay_array is provided (then it will be set to 'delay').
```
bhazelton, line 83: we've deprecated the `unknown` cal_type, so let's not let it be set to unknown here.

pyuvdata/uvcal/initializers.py
```diff
@@ -0,0 +1,529 @@
...
+        the antenna_numbers, but in this case the antenna_names must be strings that
+        can be converted to integers. antenna_numbers need not be provided if
+        antenna_positions is a dict with integer keys.
+    antname_format : str, optional
+        Format string for antenna names. Default is '{0:03d}'.
+    ant_array : ndarray of int, optional
+        Array of antenna numbers actually found in data (in the order of the data
+        in gain_array etc.)
+    wide_band : bool, optional
+        Whether to use wide_band. Default is False, and Trie is not yet supported.
```
bhazelton, line 105: Trie -> True

wide_band will have to be true if it's a delay style cal, right?

pyuvdata/uvcal/uvcal.py
steven-murray: I don't think I meant to leave this method in... I've deleted it. I certainly don't mind renaming `initialize_from_uvdata` to `from_uvdata`, but we don't have to, and maybe for simplicity we just leave it as-is.

pyuvdata/uvcal/initializers.py
steven-murray: Ah, I removed that option from the code, but not the docs. Good spotting.

pyuvdata/uvcal/initializers.py
steven-murray: Actually, I just forgot to remove this from the argument list. It's set automatically in the code depending on whether ``freq_array`` or ``freq_range`` is provided (False and True respectively).

pyuvdata/uvcal/initializers.py
```diff
@@ -0,0 +1,533 @@
...
+    antenna_names: list[str] | None = None,
+    antenna_numbers: list[int] | None = None,
+    antname_format: str = "{0:03d}",
+    ant_array: np.ndarray | None = None,
+    flex_spw_id_array: np.ndarray | None = None,
+    ref_antenna_name: str | None = None,
+    sky_catalog: str | None = None,
+    sky_field: str | None = None,
+    empty: bool = False,
+    data: dict[str, np.ndarray] | None = None,
```
bhazelton, line 45: data does not appear in the docstring.

pyuvdata/uvcal/initializers.py
```diff
@@ -0,0 +1,533 @@
...
+
+    # Get frequency-type info from kwargs and uvdata
+    if cal_type == "gain" and not wide_band:
+        if "channel_width" in kwargs or "freq_array" in kwargs:
+            channel_width = kwargs.pop("channel_width", None)
+            freq_array = kwargs.pop("freq_array", None)
+            flex_spw_id_array = kwargs.pop("flex_spw_id_array", None)
+        else:
+            # Just in case the input uvdata is not future_array_shapes compliant, we
+            # flatten
```
bhazelton, line 377: This confuses me, we're calling `uvdata.use_future_array_shapes()` above.

pyuvdata/uvcal/tests/test_uvcal.py
```diff
@@ -4519,7 +4497,8 @@ def test_init_from_uvdata_freqrange_errors(uvcalibrate_data):
 
     with pytest.raises(
         ValueError,
-        match="if future_array_shapes is True, freq_range must be an array shaped like",
+        match="UVParameter \\_freq\\_range is not expected shape\\. Parameter shape is "
+        "\\(1\\, 4\\)\\, expected shape is \\(1\\, 2\\)\\.",
     ):
```
bhazelton, line 4502: I think you can avoid using all the `\\`s here and below if you just call `re.escape` on the whole string.

pyuvdata/uvcal/uvcal.py
```diff
@@ -4388,421 +4378,60 @@ def initialize_from_uvdata(
...
-            Calibration times in decimal Julian date. If None, use all unique times from
-            uvdata.
-        integration_time : float or array_like of float, optional
-            Calibration integration time in seconds, an array of shape (Ntimes,)
-            or a scalar if `future_array_shapes` is False. Required if `time_array` is
-            not None, ignored otherwise.
-        time_range : array_like of float, optional
-            Range of times that calibration is valid for in decimal Julian dates,
-            shape (2,). Should only be set if `time_array` is size (1,)
+            Deprecated alias for ``time_array``. Will be removed in v3.
```
bhazelton, line 4388: Method parameter deprecations (as opposed to the future_array_shapes ones) can be set to be removed in v2.5.

bhazelton (commented):
This is looking really good. I just have a few comments plus re-working the docstrings as discussed in the telecon (or pushing that to an issue).

ci/azure-piplines.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       source ./ci/_activate_current_env.sh
       pip install --no-deps .
       cd ../hera_cal
-      pip install --no-deps .
```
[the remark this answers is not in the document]
steven-murray: We recently changed `hera_cal` to simply install in a virtual environment without conda for CI, using the direct dependencies listed in `setup.cfg`. Thus, the conda yaml in the hera_cal repo is not up to date, and actually should probably be removed because it's not used. The reason for doing this was speed.

pyuvdata/uvcal/initializers.py
```diff
@@ -0,0 +1,533 @@
...
+        Gain convention. Options are 'divide' or 'multiply'.
+    x_orientation : str
+        Orientation of the x-axis. Options are 'east', 'north', 'e', 'n', 'ew', 'ns'.
+    jones_array : ndarray of int or str
+        Array of Jones polarization integers. If a string, options are 'linear' or
+        'circular' (which will be converted to the appropriate Jones integers as
+        length-4 arrays).
+    delay_array : ndarray of float, optional
+        Array of delays in seconds. If cal_type is not provided, it will be set to
+        'delay' if delay_array is provided.
```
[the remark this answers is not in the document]
steven-murray: You're correct! Removed.

pyuvdata/uvcal/initializers.py
```diff
@@ -0,0 +1,533 @@
...
+    cal_type : str, optional
+        Calibration type. Options are 'delay', 'gain'. Forced to be 'gain' if
+        ``freq_array`` is given, and by *default* set to 'delay' if not.
+    integration_time : float or ndarray of float, optional
+        Integration time in seconds. If not provided, it will be set to the minimum
+        time separation in time_array for all times.
```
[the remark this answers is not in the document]
steven-murray: Yeah, it's doing what you're suggesting, I just hadn't updated the docstring. I've updated now to make it exactly the same as the `UVData.new()` function.

pyuvdata/uvcal/initializers.py
```diff
@@ -0,0 +1,533 @@
...
+    time_array: np.ndarray,
+    antenna_positions: np.ndarray | dict[str | int, np.ndarray],
+    telescope_location: Locations,
+    telescope_name: str,
+    cal_style: Literal["sky", "redundant"],
+    gain_convention: Literal["divide", "multiply"],
+    jones_array: np.ndarray | str,
+    x_orientation: Literal["east", "north", "e", "n", "ew", "ns"],
+    freq_array: np.ndarray | None = None,
+    freq_range: np.ndarray | None = None,
```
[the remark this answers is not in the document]
steven-murray: Ah yeah, I added this later (and took out the `delay_array`). I've added docs for this now.

pyuvdata/uvcal/initializers.py
```diff
@@ -0,0 +1,533 @@
...
+    elif cal_style == "sky":
+        uvc._set_sky()
+
+    uvc._set_future_array_shapes()
+    uvc._set_wide_band(wide_band)
+
+    uvc.Nants_data = len(ant_array)
+    uvc.Nants_telescope = len(antenna_numbers)
+    uvc.Nfreqs = len(freq_array) if freq_array is not None else 1
+    # uvc.Ndlys = len(delay_array) if delay_array is not None else 1
```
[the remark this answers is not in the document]
steven-murray: Done.

pyuvdata/uvcal/initializers.py
```diff
@@ -0,0 +1,533 @@
...
+    uvc.Ntimes = len(time_array)
+    uvc.Nspws = len(spw_array)
+    uvc.Njones = len(jones_array)
+
+    uvc.flex_spw_id_array = flex_spw_id_array
+    uvc.spw_array = spw_array
+    # We always make the following true.
```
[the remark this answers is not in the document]
steven-murray: Woops! It was vestigial.

pyuvdata/uvcal/initializers.py
steven-murray: Good point. Added.

pyuvdata/uvcal/initializers.py
steven-murray: Ah yes, also vestigial. Removed the comment and the `.flatten()` calls.

pyuvdata/uvcal/uvcal.py
steven-murray: Cool. Updated.

ci/azure-piplines.yml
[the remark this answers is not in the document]
bhazelton: ok, then I think the entire action should be updated not to use that conda file.

pyuvdata/uvcal/initializers.py
```diff
@@ -66,6 +64,12 @@ def new_uvcal(
         Telescope name.
     cal_style : str
         Calibration style. Options are 'sky' or 'redundant'.
+    freq_array : ndarray of float, optional
+        Array of frequencies in Hz. If given, the freq_range parameter is ignored, and
+        the cal type is assumed to be "gain".
+    freq_range : ndarray of float, optional
+        Frequency ranges in Hz. If given, the freq_array parameter is ignored, and the
+        calibration is assumed to be wide band. The array shape should be (Nspws, 2).
```
bhazelton, line 72: These both say that if they're set the other one is ignored. Which takes precedence?

pyuvdata/uvcal/initializers.py
```diff
@@ -74,18 +78,24 @@ def new_uvcal(
...
     integration_time : float or ndarray of float, optional
-        Integration time in seconds. If not provided, it will be set to the minimum
-        time separation in time_array for all times.
+        Integration time in seconds. If not provided, it will be derived from the
+        time_array, as the difference between successive times (with the last time-diff
+        appended). If not provided and the number of unique times is one, then
+        a warning will be raised and the integration time set to 1 second.
+        If a float is provided, it will be used for all integrations.
+        If an ndarray is provided, it must have the same shape as time_array (or
+        unique_times, if that is what is provided).
```
bhazelton, line 91: I don't think that `unique_times` is an input to this method.

pyuvdata/uvcal/initializers.py
```diff
@@ -114,6 +124,10 @@ def new_uvcal(
         gain_array). By default, this function creates a metadata-only object. You can
         pass in data arrays directly to the constructor if you want to create a
         fully-populated object.
+    data : dict of array_like, optional
+        Dictionary containing optional data arrays. Possible keys are:
+        'gain_array', 'delay_array', 'quality_array', 'flag_array',
+        and 'total_quality_array'.
```
bhazelton, line 130: What happens if only some of the keys are provided? Some of these are optional but I think some are required and we should at least tell users that if any are provided that they might get an error in the check if they don't provide at least the minimal set.

bhazelton (commented):
Looks good, just a couple things in the updated docstrings that I spotted.

pyuvdata/uvcal/initializers.py
steven-murray: Good point. I've updated to make it clear that `freq_array` takes precedence.

pyuvdata/uvcal/initializers.py
steven-murray: Good catch. Fixed.

pyuvdata/uvcal/initializers.py
steven-murray: I've added more detail to this docstring, making it clear that a subset can be provided and the others will then be set to default "empty" values.

setup.py
```diff
@@ -150,6 +150,7 @@ def is_platform_windows():
         "pyerfa>=2.0",
         "scipy>=1.5",
         "setuptools_scm!=7.0.0,!=7.0.1,!=7.0.2",
+        "docstring-parser>=0.15",
```
bhazelton, line 153: This also needs to be added to the readme and the `environment.yaml` file.

pyuvdata/uvcal/initializers.py
```diff
@@ -0,0 +1,542 @@
...
+    antname_format: str = "{0:03d}",
+    ant_array: np.ndarray | None = None,
+    flex_spw_id_array: np.ndarray | None = None,
+    ref_antenna_name: str | None = None,
+    sky_catalog: str | None = None,
+    sky_field: str | None = None,
+    empty: bool = False,
+    data: dict[str, np.ndarray] | None = None,
+    history: str = "",
+    **kwargs,
```
bhazelton, line 48: I don't see any documentation for this `kwargs`. Not sure if this should be done with your new parsing infrastructure or not.

pyuvdata/uvdata/uvdata.py
```diff
@@ -772,8 +773,8 @@ def __init__(self):
         super(UVData, self).__init__()
 
     @staticmethod
-    def new(**kwargs):
-        """Stand-in docstring for new method."""
+    @combine_docstrings(new_uvdata)
+    def new(**kwargs):  # noqa: D102
```
bhazelton, line 777: something's going wrong with this docstring's formatting currently in the RTD build. Starting part way through the `antenna_positions` parameter, the formatting goes to putting all the rest of the text into a paragraph that is out-dented so that it is no longer part of the parameters section and is not a bulleted list.

bhazelton (commented):
The docs handling is much better, just a couple minor comments.

pyuvdata/uvcal/initializers.py
steven-murray: All the kwargs are added to the `Other Parameters` section, so they should be there. It doesn't explicitly mention the kwargs, per se.

bhazelton: But there's no ` @combine_docstrings` decorator on this method, so I don't think anything will be added?

bhazelton:
@steven-murray I think this branch needs to be rebased because of the deprecation of the input flag array on the main branch.

pyuvdata/uvcal/initializers.py
steven-murray: Oh! I see what you mean. I'll check that out.

ci/azure-piplines.yml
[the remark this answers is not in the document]
steven-murray: Ah, I forgot about this comment, sorry! Yeah, let's make this an issue, because I think it's unrelated to this PR.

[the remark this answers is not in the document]
steven-murray: See https://github.com/RadioAstronomySoftwareGroup/pyuvdata/issues/1293

bhazelton:
I fixed the errors on the builds that were using astropy 5.3 by re-writing the uvh5 file that had slightly off LSTs.

@steven-murray I think this is ready to merge once the coverage issue is fixed.

CHANGELOG.md
```diff
@@ -4,9 +4,13 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- A new `UVCal.new()` method (based on new function `new_uvcal`) that creates a new,
+self-consistent `UVCal` object from scratch from a set of flexible input parameters.
```
bhazelton, line 8: Let's add something about the new `docstring_parser` dependency here. It will help me when we make the next release as well as helping users.

bhazelton (approved):

PR #1285, opened 2023-04-08
Transactional FastUVH5Meta

steven-murray:
## Description

As it stands, the `FastUVH5Meta` class will open up the HDF5 file when you instantiate it, and it will stay open unless you explicitly call `.close()`. Furthermore, each time you read something from the metadata, it will cache it (so it accesses from memory the second time). This is the fastest way to do things, and generally what you want. However, you may want to close the file every time you get something out of it, and you may not want to cache everything (you may just want to read it into a new variable that you go and use). This PR implements these features.

## Motivation and Context

The reason you might not want to generally keep the file open is that HDF5 only allows a finite number of open files at a time. In practice, we may have a list of many thousands of files that we wish to process. Making each of those a light `FastUVH5Meta` class currently breaks. Of course, you can go through and make each one in turn, get the stuff you need, then close it before opening the next one, however it is cleaner to be able to just make the whole list of files into UVH5 meta objects in one list comprehension, and this will break.

For a similar reason, you might want to only retrieve a bit of metadata without caching it in the object. Say you want to calculate the median LST for each file in a big list, and then you know you don't care about anything else in the files. Currently you could do:

```python
metas = [FastUVH5Meta(fl) for fl in file_list]
lst_medians = [np.median(meta.lsts) for meta in metas]
for meta in metas:
    del meta.lsts
```

However, after the second line, there's a bunch of memory taken up, because each of the elements of `metas` has an attached `lsts` array (also, if `file_lst` is a long list, this will also break due to number of open files).

We could instead do:

```python
lst_medians = []
for fl in file_list:
    meta = FastUVH5Meta(fl)
    lst_medians.append(np.median(meta.lsts))
```

This works fine, but in a complex pipeline, you might want to pass through the files to another function that expects them as `FastUVH5Meta` objects, and here' we've lost that association. In this PR, you can just do:

```python
metas = [FastUVH5Meta(fl) for fl in file_list]  # Files don't auto-open until you access something.
lst_medians = [np.median(meta.get_transactional('lsts', cache=False) for meta in metas]
```

You can omit the `cache=False` if you want to keep the full LSTs around in the meta objects for later *but still want to close the file after reading*. 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/docs/uvdata_tutorial.rst
+++ b/docs/uvdata_tutorial.rst
@@ -1453,8 +1453,8 @@ for each file, you can read just the time array (for example), using the special
   False
 
   # FastUVH5Meta is meant to be a read-only view into the HDF5 file, not a general
-  # data manipulation class. You should set attributes or write to the file.
-  # Use UVData for that.
+  # data manipulation class. You should *not* set attributes or use it to write to the
+  # file. Use UVData for that.
 
 .. _sorting_data:
 
--- a/docs/uvdata_tutorial.rst
+++ b/docs/uvdata_tutorial.rst
@@ -1426,7 +1426,8 @@ for each file, you can read just the time array (for example), using the special
 
   # No data is read in when creating the object, and you can access each bit of metadata
   # individually, and it will lazy-load from the file
-  >>> print(meta.time_array)
+  >>> print(len(meta.time_array))
+  200
 
   # Once loaded, the data is cached, so you can access it again without reading from disk
   >>> "time_array" in meta.__dict__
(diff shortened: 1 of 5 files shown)


pyuvdata/uvdata/uvh5.py
```diff
@@ -375,11 +374,17 @@ def close(self):
         """Close the file."""
         self.__header = None
         self.__datagrp = None
-        del self.header  # need to refresh these
+
+        try:
+            del self.header  # need to refresh these
+        except AttributeError:
+            pass
```
bhazelton, line 381: The coverage issue on this line is a known problem -- it's caused by optimizations in the compiler. You should check that you have a test that goes down this branch and if you do, you can safely mark this with a `pragma: no cover`.

docs/uvdata_tutorial.rst
```diff
@@ -1409,6 +1409,52 @@ are written to the appropriate parts of the file on disk.
...
+  >>> lsts = meta.get_transactional('lst_array')
+
+  # If you also want the data not to be cached on the object, specify cache=False
+  >>> freqs = meta.get_transactional('freq_array', cache=False)
+  >>> "freq_array" in meta.__dict__
+  False
+
+  # FastUVH5Meta is meant to be a read-only view into the HDF5 file, not a general
+  # data manipulation class. You should set attributes or write to the file.
+  # Use UVData for that.
```
bhazelton, line 1457: These last two sentences need editing. I think you're saying you cannot use it to set attributes or write to the file?

bhazelton (commented):
This is looking good, there's just one small fix needed to the tutorial. I checked the RTD build and it looks good.

docs/uvdata_tutorial.rst
steven-murray: Haha woops!

bhazelton (approved):
Looks good, thanks!

PR #1289, opened 2023-05-02
Fix error on ntimes one

steven-murray:
## Description

This just removes a `check()` error that was not useful. The error was raised if `Ntimes=1` and `time_axis_faster_than_bls=True`. But, in the case of `Ntimes=1`, the time axis could be both seen as moving faster and slower than bls without consequence, so it should not raise an error.

## Motivation and Context

This was raising errors in some code where the time axis is faster than bls, BUT I had selected on times so there was only one time left in the data.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ All notable changes to this project will be documented in this file.
 - Reading in multiple files to `UVCal` using file-type specific read methods
 (e.g. `read_calfits`) in favor of the generic `read` method.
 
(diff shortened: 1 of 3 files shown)


bhazelton:
@steven-murray I think you forgot to remove the test that exercised this error. All the CI runs are failing.

bhazelton (approved):
Looks good. Thanks!

PR #1292, opened 2023-05-23
UVW helper utilities (for use on the moon and otherwise...)

### LINKED ISSUE #1266: Lunar phasing bugs
Talking with @kartographer it looks like at least two potential lunar issues remain in the utils code.  I may not be describing these quite right but we can revisit.

(1) Line 2694 (and line 2700 in particular) seems to be using lunartopo as a drop in replacement for alt/az.  That's not right (I think).

(2) Line 2708 is using sidereal time from an astropy Time object, but it's likely not the correct sidereal time for the moon.

kartographer:
## Description

Adds some utility functions inside of `utilts` for calculating UVW coordinates _without_ needing to instantiate a full `UVData` object. Useful for calculating _uv_-coverage for a given RA/Dec and a given array. 

## Motivation and Context

Closes #1266 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

New feature checklist:

Breaking change checklist:

Version change checklist:

### DIFF
--- a/pyuvdata/parameter.py
+++ b/pyuvdata/parameter.py
@@ -659,6 +659,48 @@ class UVParameter(object):
                     )
                     return False, message
 
+    def compare_value(self, value):
+        """
+        Compare UVParameter value to a supplied value.
+
+        Parameters
+        ----------
+        value
+            The value to compare against that stored in the UVParameter object. Must
+            be the same type.
+
+        Returns
+        -------
+        same : bool
+            True if the values are equivalent (or within specified tolerances),
+            otherwise false.
+        """
+        # Catch the case when the values are different types
+        if not (
+            isinstance(value, self.value.__class__)
+            and isinstance(self.value, value.__class__)
+        ):
+            raise ValueError(
+                "UVParameter value and supplied values are of different types."
+            )
+
+        # If these are numeric types, handle them via allclose
+        if isinstance(value, (np.ndarray, int, float, complex)):
+            # Check that we either have a number or an ndarray
+            if not isinstance(value, np.ndarray) or value.shape == self.value.shape:
+                if np.allclose(
+                    value,
+                    self.value,
+                    rtol=self.tols[0],
+                    atol=self.tols[1],
+                    equal_nan=True,
+                ):
+                    return True
+            return False
+        else:
+            # Otherwise just default to checking equality
+            return value == self.value
+
 
 class AngleParameter(UVParameter):
     """
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -2440,6 +2440,7 @@ def transform_icrs_to_app(
     ra,
     dec,
     telescope_loc,
+    antenna_frame='itrs',
     epoch=2000.0,
     pm_ra=None,
     pm_dec=None,
@@ -2481,6 +2482,9 @@ def transform_icrs_to_app(
         of the array. Can either be provided as an astropy EarthLocation, or a tuple
         of shape (3,) containing (in order) the latitude, longitude, and altitude,
         in units of radians, radians, and meters, respectively.
+    antenna_frame: str, optional
+        Reference frame for latitude/longitude,altitude. Options are itrs (default) or
+        mcmf. Only used if telescope_loc is not an EarthLocation or MoonLocation.
     epoch : int or float or str or Time object
         Epoch of the coordinate data supplied, only used when supplying proper motion
         values. If supplying a number, it will assumed to be in Julian years. Default
@@ -2525,6 +2529,12 @@ def transform_icrs_to_app(
     if astrometry_library is None:
         if hasmoon and isinstance(telescope_loc, MoonLocation):
             astrometry_library = "astropy"
+        elif antenna_frame.upper() == "MCMF":
+            if not hasmoon:
+                raise ValueError(
+                    "Need to install 'lunarsky' package to work with MCMF frame."
+                )
+            astrometry_library = "astropy"
         else:
             astrometry_library = "erfa"
 
@@ -2562,6 +2572,16 @@ def transform_icrs_to_app(
         hasmoon and isinstance(telescope_loc, MoonLocation)
     ):
         site_loc = telescope_loc
+    elif antenna_frame.upper() == "MCMF":
+        if not hasmoon:
+            raise ValueError(
+                "Need to install 'lunarsky' package to work with MCMF frame."
+            )
+        site_loc = MoonLocation.from_selenodetic(
+            telescope_loc[1] * (180.0 / np.pi),
+            telescope_loc[0] * (180.0 / np.pi),
+            height = telescope_loc[2],
+        )
     else:
         site_loc = EarthLocation.from_geodetic(
             telescope_loc[1] * (180.0 / np.pi),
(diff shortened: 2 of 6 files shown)


wps2n:
I've done a direct comparison of this new helper method with the original phase method using a uvdata object obtained from a run of pyuvsim. I used an arbitrary array of three antennas at lat, lon, alt (30, 24.433, 0) and ENU for the antennas (50,0,0), (-50,0,0), and (-30,10,0). I then found the RA and DEC of zenith midway through the sim and used that point to phase to zenith ~(0.2581, 0.5219). I mention this, because the UVW tracks overlay quite nicely but there is some error between the phase method and the new helper method that may not simply be a machine precision problem. 
UV_tracks
I plot the difference (phase-helper) here, and a zoom-in shows a clear periodic pattern. 
UV_Comparison
UV_Comparison_Zoom

I am not precisely sure what is causing this difference between the two runs. I have been also attempting to run lunar sims and I am noticing that it is difficult for me to pass the same telescope_location into each method due to rounding. Is it possible that the uvdata object itself holds more decimal places but gets rounded when passed to a new variable? This might be a known fact but I'm not too familiar. This may also just be an acceptable difference between the two, I am not sure. Thoughts?

jpober:
Potential for an added feature.  Could we make it an option to have a conjugation convention when handing in antenna positions?  Right now, there doesn't seem any control over whether baselines wind up with u > 0 or u < 0.  Alternatively, could we have an option to give uvw tracks for baselines and their complex conjugates?

jpober:
Another (maybe less trivial) ask: can we have uvw_track_generator take a set of unphased (u,v,w) coordinates and phase them?  The docstrings actually suggest it can but it doesn't.

wps2n:
Ran the same test as before with the rebased branch with the improved LST handling and saw the same small-scale discrepancies between uvw_track_generator and phase. Also interesting to note is that the discrepancies between uvw_track_generator with an input uvw_array and the phase method are several orders of magnitude larger albeit the plotted tracks overlap in the same way. I imagine this is simply because the UVs are ordered differently.

I am attaching my working example below.

[uvw_track_generator_test.zip](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/12371778/uvw_track_generator_test.zip)

kartographer:
Thanks @wps2n! So the good news is that I was at least able to resolve the Earth-based discrepancy -- there is a minor difference between the LST in the file versus that which the utilities are generating. I don't know enough about the origins of the 21cmsense file, but it's magnitude and pattern look an _awful_ lot like the previous LST precision "bug" that we solved recently. It might make sense to take a look at how it's been generated for this file. In any case, when I run `set_lsts_from_time_array`, I get perfect agreement, as shown below.
UV_Comparison

Hopefully that solves the Earth-based issues, but let me know if there's anything else outstanding.

pyuvdata/parameter.py
```diff
@@ -659,6 +659,48 @@ def check_acceptability(self):
                     )
                     return False, message
 
+    def compare_value(self, value):
```
bhazelton, line 662: This is great! We should go through our tests and use it where we can.

pyuvdata/utils.py
```diff
@@ -3783,7 +3789,13 @@ def calc_sidereal_coords(
...
 
 def get_lst_for_time(
-    jd_array, latitude, longitude, altitude, astrometry_library=None, frame="itrs"
+    jd_array=None,
+    latitude=None,
+    longitude=None,
+    altitude=None,
+    astrometry_library=None,
+    frame="itrs",
+    telescope_loc=None,
```
bhazelton, line 3798: Can you add this to the docstring? It looks like it can be a tuple of lat/lon/alt or an EarthLocation or a MoonLocation?

pyuvdata/utils.py
```diff
@@ -3975,6 +4013,195 @@ def check_lsts_against_times(
...
+    ant_2_array=None,
+    uvw_array=None,
+    force_postive_u=False,
+):
+    """
+    Calculate uvw coordinates (among other values) for a given position on the sky.
+
+    This function is meant to be a user-friendly wrapper around several pieces of code
+    for effectively simulating a track. Note that the naming of things may change
+    before this hits a PR.
```
bhazelton, line 4037: This should probably be updated before we merge 😉

pyuvdata/utils.py
```diff
@@ -3975,6 +4013,195 @@ def check_lsts_against_times(
...
+        that is presently supported by astropy.
+    coord_epoch : float or str or Time object, optional
+        Epoch for ref_frame, nominally only used if converting to either the FK4 or
+        FK5 frames, in units of fractional years. If provided as a float and the
+        ref_frame is an FK4-variant, value will assumed to be given in Besselian
+        years (i.e., 1950 would be 'B1950'), otherwise the year is assumed to be
+        in Julian years.
+    time_array : float or ndarray of float or Time object
+        Times for which the apparent coordinates were calculated, in UTC JD. Must
+        match the shape of lon_coord and lat_coord.
```
bhazelton, line 4058: This confuses me. If I was phasing to a single RA/Dec I'd still expect to be able to pass a bunch of times to get my uvws for.

pyuvdata/utils.py
```diff
@@ -3975,6 +4013,195 @@ def check_lsts_against_times(
...
+    force_positive_u : bool, optional
+        If set to true, then forces the conjugation of each individual baseline to be
+        set such that the uvw coordinates land on the positive-u side of the uv-plane.
+        Default is False.
+
+    Returns
+    -------
+    obs_dict : dict
+        Dictionary containing the results of the simulation, the only key of which
+        is 'uvw' for the uvw-coordinates of the observation.
```
bhazelton, line 4099: This isn't correct, there are more entries in the dict now.

pyuvdata/utils.py
```diff
@@ -3975,6 +4013,195 @@ def check_lsts_against_times(
         )
 
 
+def uvw_track_generator(
+    lon_coord=None,
+    lat_coord=None,
+    coord_frame="icrs",
+    coord_epoch=None,
+    coord_type="sidereal",
```
bhazelton, line 4021: Please add this to the docstring.

pyuvdata/utils.py
```diff
@@ -3975,6 +4013,195 @@ def check_lsts_against_times(
         )
 
 
+def uvw_track_generator(
+    lon_coord=None,
```
bhazelton, line 4017: Let's go ahead and require all of these to be passed by name by inserting a "*," before the first parameter.

bhazelton (commented):
This is looking good. Just a few docs issues.

pyuvdata/utils.py
kartographer: Added.

pyuvdata/utils.py
kartographer: Heh, whoops -- I've taken out that bit of the docstring.

pyuvdata/utils.py
kartographer: I added the single-float case mostly because it was helpful for when I was testing/debugging (and there are instances where I can see getting a snapshot of UV-coverage would be a good thing). But in the interest of clarity,  I've taken it out of the docstring.

pyuvdata/utils.py
kartographer: Ahh, good catch -- added.

pyuvdata/utils.py
kartographer: Added!

bhazelton (approved):
Looks good. Thank you @kartographer !

PR #1306, opened 2023-06-09
Better time_range vs time_array handling in UVCal

### LINKED ISSUE #851: time_array and time_range are problematic in UVCal
time_array should be used, but time_range should be replaced by an integration_length-like array that defines what range of times the solution applies to.  This will require some work on uvcalibrate to interpret sensibly.

bhazelton:
## Description

Improve the way `time_range` is handled in UVCal. It exists to support calibration approaches that solve for cal solutions over a range of times rather than per integration. This PR makes the following changes:
- Deprecate having both a `time_array` and a `time_range` defined.
- Make `time_range` be 2D with a shape of (Ntimes, 2) where the first axis gives the number of different time solutions and the second axis gives the start/stop times for those solutions. This allows different calibration solutions that apply over ranges of times to be properly combined in a single object.
- Add a `get_time_array` method that either returns the mean of the start and stop time for each time range or the time_array (if there's a time_array and no time_range)

**Note**: this is actually a breaking change in the following ways:
- It changes the shape of the `time_range` in the object.
- It makes it so that `time_range` is not defined on an object read from a calfits file with more than one time, even if the file has a `time_range` entry in the header. This is because the header `time_range` is a single range, it doesn't have a length Ntimes axis, so it doesn't have the right shape for the new `time_range` shape.
- If a calfits file is read in with one time and it has a `time_range` entry in the header, only the `time_range` will be defined, not the `time_array`. This could be changed but it would result in many deprecation warnings because having both a `time_range` and a `time_array` is deprecated. We could add a parameter to `read_calfits` to only have it define one or the other.
- If and fhd cal file is read in (which by definition has only one time), only the `time_range` will be defined, not the `time_array`. This could be changed but it would result in many deprecation warnings because having both a `time_range` and a `time_array` is deprecated. We could add a parameter to the `read_fhd` method to only have it define one or the other.

I see how to fix the third  & fourth items above in a way that supports deprecation, but I don't see a path for allowing for a deprecation period for the first two items. So we decided to push this into version 3.0.

## Motivation and Context

Closes #851 

## Types of changes

- [ ] Version change

## Checklist:

Breaking change checklist:

### DIFF
--- a/pyuvdata/uvcal/uvcal.py
+++ b/pyuvdata/uvcal/uvcal.py
@@ -30,29 +30,31 @@ _future_array_shapes_warning = (
 )
 
 
-def _check_time_range_overlap(time_range):
+def _check_range_overlap(val_range, range_type="time"):
     """
-    Detect if any time ranges in an array overlap.
+    Detect if any val_range in an array overlap.
 
     Parameters
     ----------
-    time_range : np.array of float
-        Array of time ranges, shape (Ntimes, 2).
+    val_range : np.array of float
+        Array of ranges, shape (Nranges, 2).
+    range_type : str
+        Type of range (for good error messages)
 
     Returns
     -------
     bool
-        True if any time range overlaps.
+        True if any range overlaps.
     """
     # first check that time ranges are well formed (stop is >= than start)
-    if np.any((time_range[:, 1] - time_range[:, 0]) < 0):
+    if np.any((val_range[:, 1] - val_range[:, 0]) < 0):
         raise ValueError(
-            "The time_ranges are not well-formed, some stop times are after start "
-            "times."
+            f"The {range_type} ranges are not well-formed, some stop {range_type}s "
+            f"are after start {range_type}s."
         )
 
     # Sort by start time
-    sorted_ranges = time_range[np.argsort(time_range[:, 0]), :]
+    sorted_ranges = val_range[np.argsort(val_range[:, 0]), :]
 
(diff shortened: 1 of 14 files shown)


pyuvdata/uvcal/uvcal.py
```diff
@@ -238,36 +294,70 @@ def __init__(self):
...
+        self._lst_range = uvp.UVParameter(
+            "lst_range",
+            description=desc,
+            form=("Ntimes", 2),
+            expected_type=float,
+            tols=uvutils.RADIAN_TOL,
+            required=False,
+        )
+
+        # TODO:
```
jpober, line 357: Should we turn these into issues?

jpober (dismissed):

jpober:
I think we expect the hera_cal and hera_qm builds to fail, but is the build on pyuvdata failing expected?  Is it because this is a breaking change?

bhazelton:
That third failure is just the roll up of the set of external tests, so it fails if any of them do. I don't know why it's listed separately.

pyuvdata/uvcal/uvcal.py
bhazelton: Actually, I think those are cruft. The first one is done and I think we sorted it out reasonably without the other package.

bhazelton:
@jpober can you please re-approve? I removed the stale TODOs

jpober (approved):

PR #1308, opened 2023-06-22
Fix code that raises new numpy deprecation warnings

bhazelton:
## Description

Changes:
- Fixed a bunch of places that produced new DeprecationWarnings with numpy 1.25. I think that a few of these were real bugs.
- Improve `check_warnings` to give the file and line number where warnings occur to help with debugging.
- Don't issue a warning in efield to beam when fixing auto pols to be real if imaginary values are very small (i.e. caused by numerical precision issues). We started getting a bunch of these warnings when we didn't use to.
- Made some hard to read one-line if statements in utility methods multiline. This resulted in a drop in coverage because we haven't set `branch=True` for coverage (see #1309). So I added some tests to cover all the branches.
- Made a new conda environment for python 3.11. This is needed because we can't get h5py>=3.9 to resolve with all the other dependencies in python 3.11. We need h5py>=3.9 if we have numpy>=1.25 so we don't get flooded by deprecation warnings. So in the new env for 3.11 I restricted numpy to less than 1.25. This should be reverted in the future (#1310).

## Motivation and Context

Fix CI errors due to new numpy deprecation warnings.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/tests/test_utils.py
+++ b/pyuvdata/tests/test_utils.py
@@ -1888,7 +1888,7 @@ def test_astrometry_icrs_to_app(astrometry_args, use_extra):
         for key in extra_args:
             kwargs[key] = astrometry_args[key]
     else:
-        # don't compre to precalc if not using extra arguments
+        # don't compare to precalc if not using extra arguments
         coord_results = coord_results[:-1]
 
     for idx, name in enumerate(astrometry_list):
--- a/pyuvdata/tests/test_utils.py
+++ b/pyuvdata/tests/test_utils.py
@@ -1851,7 +1851,8 @@ def test_calc_app_fk4_roundtrip(astrometry_args, telescope_frame):
 @pytest.mark.filterwarnings('ignore:ERFA function "pmsafe" yielded 4 of')
 @pytest.mark.filterwarnings('ignore:ERFA function "utcut1" yielded 2 of')
 @pytest.mark.filterwarnings('ignore:ERFA function "d2dtf" yielded 1 of')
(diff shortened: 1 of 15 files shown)


mkolopanis (commented):
Looking good to me. I think my only question is why are we adding a lot of `[0]` indices in places? Did we used to have length 1 dimensions that we are collapsing?

bhazelton:
> I think my only question is why are we adding a lot of `[0]` indices in places? Did we used to have length 1 dimensions that we are collapsing?

Yes. Numpy now complains about converting single element arrays to scalars:
`Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)`

pyuvdata/tests/test_utils.py
```diff
@@ -1881,18 +1882,24 @@ def test_astrometry_icrs_to_app(astrometry_args):
 
     coord_results[3] = (precalc_ra, precalc_dec)
 
+    kwargs = {}
+    extra_args = ["pm_ra", "pm_dec", "vrad", "dist"]
+    if use_extra:
+        for key in extra_args:
+            kwargs[key] = astrometry_args[key]
+    else:
+        # don't compre to precalc if not using extra arguments
```
steven-murray, line 1891: typo in "compare"

steven-murray (commented):
Just the one small typo

mkolopanis (approved):

steven-murray (approved):

PR #1322, opened 2023-07-22
Add warnings in check if lst_array doesn't match expectations

bhazelton:
## Description

Add a check that the LSTs have their expected values in the `check` methods of UVData, UVCal and UVFlag. It only issues a UserWarning, not an error, and it is only run if `run_check_acceptability=True`. This is implemented via a new utility method to check the LSTs which is now called in all the places.

In moving this to the utility module, I also realized that the standard radian tolerance was being redefined in several separate places and that the one in UVFlag had not been updated when we reduced that tolerance. So I moved it to be defined in the utility module as well.

## Motivation and Context

I recently realized (as I was fixing up test files for `hera_qm` and `hera_cal`) that there was only a check that the LSTs had their expected values when reading in UVH5 and uvfits files but not on reading in any other files. I think this is a useful check, so I added it the UVData, UVCal and UVFlag `check` methods. 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/uvflag/uvflag.py
+++ b/pyuvdata/uvflag/uvflag.py
@@ -202,10 +202,6 @@ class UVFlag(UVBase):
         run_check_acceptability=True,
     ):
         """Initialize the object."""
-        # standard angle tolerance: 10 mas in radians.
-        # Should perhaps be decreased to 1 mas in the future
-        radian_tol = 10 * 2 * np.pi * 1e-3 / (60.0 * 60.0 * 360.0)
-
         desc = (
             "The mode determines whether the object has a "
             "floating point metric_array or a boolean flag_array. "
@@ -358,7 +354,7 @@ class UVFlag(UVBase):
             description=desc,
             form=("Nblts",),
             expected_type=float,
-            tols=radian_tol,
+            tols=uvutils.RADIAN_TOL,
         )
 
         desc = (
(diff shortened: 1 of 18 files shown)


pyuvdata/uvflag/uvflag.py
```diff
@@ -923,6 +923,24 @@ def _set_type_waterfall(self):
...
+            lsts,
+            rtol=self._lst_array.tols[0],
+            atol=self._lst_array.tols[1],
+        ):
+            warnings.warn(
+                "The lst_array is not self-consistent with the time_array and "
+                "telescope location. Consider recomputing with the "
+                "`set_lsts_from_time_array` method."
+            )
+
```
steven-murray, line 943: Since you're using the same function in three places, would it make sense to have this as a standalone function that is called in each class?

steven-murray (commented):
Looks good, @bhazelton. I just have the one comment which might engender a small refactor.

pyuvdata/uvflag/uvflag.py
bhazelton: Maybe. I thought about that, but it's basically already just a call to the utility method, so the only thing that'd save me from copying is the warning, which seemed like a very small savings.

steven-murray: Agreed, and I did think of that. I'm happy either way, but I guess the tipping point for me bringing it up was that it appeared not twice but three times, and the warning message should be consistent between them all. If we end up changing it for some reason in one place, we have the cognitive overhead of having to change it everywhere.

bhazelton: I made the change, I think you were right, it simplified some things. It also turned out there were two other similar checks in `uvh5` and `uvfits` that I could also replace with the utility method. I also realized that the standard radian tolerance was being redefined in several separate places and that the one in UVFlag had not been updated when we reduced that tolerance. So I moved it to be defined in the utility module as well.

Working through this also made me aware that the astrometry library is not being plumbed through all the places `get_lst_for_time` is called. I'll fix that in a future PR.

bhazelton:
@steven-murray I think this is ready for another look when you have a chance.

steven-murray (approved):
thanks @bhazelton this looks all good!

PR #1325, opened 2023-07-28
Fix bug in calc_app_coords

kartographer:
## Description

A bug in `utils.calc_app_coords` caused Time objects to not be handled correctly when supplied as the argument for the `time_array` keyword. At issue was the underlying equality-checking of the Time class (and it's returning `False` when compared against number types, even when the numeric value appears to be identical). 

## Motivation and Context

I am motivated to fix bugs in my code.

## Types of changes

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


steven-murray (approved):
Thanks @kartographer !

PR #1329, opened 2023-08-21
Add options for astrometry library to object methods, round-trip telescope frame through UVH5, UVFITS and MS files

bhazelton:
## Description

There's a choice of astrometry library for several of the utility methods (e.g. `get_lst_for_time`) that was inaccessible from the various class methods that call it. This pipes that option through to all the relevant methods.

This required updating some docstrings, so I also reduced the duplication of docstrings between the file-type objects and UVData while I was at it using docstring parser to copy the docstrings as appropriate. I also shifted to using `kwargs` in the cases where e.g. read methods were just calling a method with all the same parameters. This should make it easier to maintain docstrings and code.

Finally, there were some places where the telescope frame was not being used properly and not round tripped through file read/writes. I added support for it in uvh5, uvfits and ms files and added an error in write_miriad if the frame is not "itrs" because it appears that the miriad format implicitly only supports the ITRS frame. I also added errors on read for the uvh5, uvfits and ms formats if the telescope frame is not "itrs" or "mcmf" because otherwise there are unclear errors later related to calculating LSTs.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/uvcal/uvcal.py
+++ b/pyuvdata/uvcal/uvcal.py
@@ -1160,7 +1160,7 @@ class UVCal(UVBase):
         if self.flex_spw:
             uvutils._check_flex_spw_contiguous(self.spw_array, self.flex_spw_id_array)
 
-    def _check_freq_spacing(self, raise_errors=True, astrometry_library="erfa"):
+    def _check_freq_spacing(self, raise_errors=True):
         """
         Check if frequencies are evenly spaced and separated by their channel width.
 
@@ -1194,11 +1194,7 @@ class UVCal(UVBase):
         )
 
     def check(
-        self,
-        check_extra=True,
-        run_check_acceptability=True,
-        check_freq_spacing=False,
-        astrometry_library=None,
+        self, check_extra=True, run_check_acceptability=True, check_freq_spacing=False
     ):
         """
         Add some extra checks on top of checks on UVBase class.
@@ -1216,12 +1212,6 @@ class UVCal(UVBase):
             Option to check if frequencies are evenly spaced and the spacing is
             equal to their channel_width. This is not required for UVCal
             objects in general but is required to write to calfits files.
-        astrometry_library : str
-            Library used for running the LST acceptability check. Allowed options are
-            'erfa' (which uses the pyERFA), 'novas' (which uses the python-novas
-            library), and 'astropy' (which uses the astropy utilities). Default is erfa
-            unless the telescope_location frame is MCMF (on the moon), in which case the
-            default is astropy.
 
         Returns
         -------
@@ -1327,7 +1317,6 @@ class UVCal(UVBase):
                 longitude=lon,
                 altitude=alt,
                 lst_tols=self._lst_array.tols,
-                astrometry_library=astrometry_library,
                 frame=self._telescope_location.frame,
             )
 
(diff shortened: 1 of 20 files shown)


pyuvdata/uvcal/uvcal.py
```diff
@@ -1192,7 +1194,11 @@ def _check_freq_spacing(self, raise_errors=True):
         )
 
     def check(
-        self, check_extra=True, run_check_acceptability=True, check_freq_spacing=False
+        self,
+        check_extra=True,
+        run_check_acceptability=True,
+        check_freq_spacing=False,
+        astrometry_library=None,
```
kartographer, line 1201: Part of me feels conflicted adding this to something like `check` -- I don't know how often it's called by users, but fundamentally each of the underlying astrometry libraries are _supposed_ to give answers to much better than the UVParameter tolerances. Is there potential the "red herrings" this might produce outweigh the benefit added by including it? Maybe that's a vote for just adding a disclaimer in the docstring rather than dropping it all together, but just my 2 cents...

pyuvdata/uvdata/uvfits.py
```diff
@@ -681,14 +561,31 @@ def read_uvfits(
...
             if "FRAME" in ant_hdu.header.keys():
-                xyz_telescope_frame = ant_hdu.header["FRAME"]
+                if ant_hdu.header["FRAME"] == "ITRF":
+                    # uvfits uses ITRF, astropy uses itrs. They are the same.
+                    self._telescope_location.frame = "itrs"
+                elif ant_hdu.header["FRAME"] == "????":
+                    # default to itrs, but use the lat/lon/alt to set the location
+                    # if they are available.
+                    prefer_lat_lon_alt = True
+                    self._telescope_location.frame = "itrs"
```
kartographer, line 573: I think I would issue a warning here that a particular frame is being assumed, just so that the user is aware that it wasn't specified in the file.

kartographer (commented):
Overall I think it's looking good -- one thought that came to mind is that given how ever-present the astrometry calls are for `UVData` (and to a lesser extent, `UVFlag` and `UVCal`) if it _might_ make sense to consider adding an optional parameter which specifies the default  astrometry library, rather than having to continuously pass a function argument. 

Also, it given the number of docstrings that changed in this PR, it might be good to spot-check and make sure there's no inconsistencies, esp given how error-prone maintaining multiple identical docstrings might have been (having been guilty of making stupid copy/paste errors in this domain myself).

pyuvdata/uvcal/uvcal.py
```diff
@@ -1158,7 +1160,7 @@ def _check_flex_spw_contiguous(self):
         if self.flex_spw:
             uvutils._check_flex_spw_contiguous(self.spw_array, self.flex_spw_id_array)
 
-    def _check_freq_spacing(self, raise_errors=True):
+    def _check_freq_spacing(self, raise_errors=True, astrometry_library="erfa"):
```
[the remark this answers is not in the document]
bhazelton: Ha! No, I don't know how that slipped in there. I'll take it out.

pyuvdata/uvcal/uvcal.py
bhazelton: I think you're right, I don't think it's needed in the check given our tests. I'll take it out.

kartographer (approved):
Looks good to me @bhazelton  -- thanks!

PR #1314, opened 2023-07-06
Fix LST discontinuities

bhazelton:
## Description

Fix a numerical precision issue that led to discontinuities in LSTs when using the `erfa` or `novas` libraries. To get full precision, you need to provide the routines with the integer and fractional parts of the JD separately. These are available on astropy time objects as `jd1` and `jd2`, so this fix uses those values rather than just passing the full JD to one of the time inputs and zeros to the other.

I am a little concerned that the erfa calculation is still not quite right because it doesn't include TIO corrections (see https://github.com/astropy/astropy/blob/073d359ebd31c6a3cd10755a9717ac6243da7dd5/astropy/time/core.py#L2329), but that is an effect on the order of 8.4 microarcseconds for the times I tested with, so I think we can address that in a separate PR.

## Motivation and Context

Fixes #1313 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation change (documentation changes only)
- [ ] Version change
- [ ] Build or continuous integration change

## Checklist:

- [x] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/.github/CONTRIBUTING.md).
- [x] My code follows the code style of this project.

Bug fix checklist:
- [x] My fix includes a new test that breaks as a result of the bug (if possible).
- [x] All new and existing tests pass.
- [x] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/CHANGELOG.md).

PR #1331, opened 2023-08-24
Fix some ERFA calls

### LINKED ISSUE #1315: Include polar angle motion correction in `erfa` LST calculation?
I am a little concerned that the erfa calculation is not quite right because it doesn't include correction for polar motion in the TIO locator (see https://github.com/astropy/astropy/blob/073d359ebd31c6a3cd10755a9717ac6243da7dd5/astropy/time/core.py#L2329). It seems to be a very small effect (~8.4 microarcseconds for the times I tested with) but I think we should probably include it.

This plot shows the difference between the astropy & erfa calculation without the TIO correction:
Screen Shot 2023-07-06 at 9 59 17 AM

And this is with that correction included (basically copied it from astropy, we might want a fix that is more pure erfa). I added this block to the erfa LST calculation:
```
        # Correct for TIO
        from astropy.coordinates.builtin_frames.utils import get_polar_motion
        from astropy.coordinates.matrix_utilities import rotation_matrix
        sp = erfa.sp00(times.tt.jd1, times.tt.jd2)
        xp, yp = get_polar_motion(times)
        # Form the rotation matrix, CIRS to apparent [HA,Dec].
        r = (
            rotation_matrix(longitude, "z")
            @ rotation_matrix(-yp, "x", unit=units.radian)
            @ rotation_matrix(-xp, "y", unit=units.radian)
            @ rotation_matrix(gast_array + sp, "z", unit=units.radian)
        )
        # Solve for angle.
        angle = np.arctan2(r[..., 0, 1], r[..., 0, 0]) << units.radian

        lst_array = np.mod(angle.value, 2.0 * np.pi)[reverse_inds]
```
Screen Shot 2023-07-06 at 10 04 21 AM

### LINKED ISSUE #1318: Make sure we're getting full precision from erfa and novas
Recently we had a bug in LST calculations with erfa and novas (#1313) that was caused by the way we chose to pass times to the libraries. These libraries usually take times as two inputs. To get full precision, the first input should be the integer jd and the second should be the decimal part. These are available on astropy Time objects as `jd1` and `jd2`. That's been fixed in the LST calculation in #1314, but we should make that change throughout our astrometry utils.

bhazelton:
## Description

Fixes some ERFA calls that were missed in #1314 for calculating apparent coordinates and add one small improvement to the ERFA LST calculation.

## Motivation and Context

Fixes #1315
Fixes #1318

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Fixed
+- A bug in apparent coordinate calculation that resulted in small errors/loss of
+precision due to the way times were passed when using `erfa` astrometry library.
+- A small correction due to polar drift for LST calculation when using the `erfa`
+astrometry library.
 - Fixed a bug in `utils.calc_app_coords` that occurred when supplying an astropy `Time`
 object for the `time_array` argument.
 
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -2836,8 +2836,8 @@ def transform_icrs_to_app(
             pm_dec_use,
             parallax,
             vrad_use,
-            time_obj_array.utc.jd,
-            0.0,
+            time_obj_array.utc.jd1,
+            time_obj_array.utc.jd2,
             time_obj_array.delta_ut1_utc,
             site_loc.lon.rad,
             site_loc.lat.rad,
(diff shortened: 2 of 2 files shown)


kartographer (approved):
Thanks @bhazelton -- all looks good to me!

PR #1340, opened 2023-09-13
Add MWA antenna information

bhazelton:
## Description

Add the MWA antenna positions to our known telescope info to allow antenna positions to be filled in for old UVFlag and UVCal files that are missing that information.

Also added a new `telescope_name` option to `UVFlag.read` and `UVFlag.__init__` to set the telescope name, allowing the other metadata to be filled in from the known telescopes.

## Motivation and Context

Allows missing metadata that are now required to be easily filled in for old files for backwards compatibility.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:


pyuvdata/uvdata/mwa_corr_fits.py
mwilensky768, line 1: The diff is a little confusing here but I think what happened is @bhazelton put the antenna information collection code up towards the top before the time/frequency stuff. In the second half of the diff, it seems a convention is enforced that some variables come straight from the meta_dict, which I think makes things safer (avoids problems of variables getting renamed in loops for instance).

mwilensky768 (approved):
I've checked that this PR fixes the reading issue for old SSINS files on a local branch of SSINS (not the one tested against). I've also looked at the code and am just seeking clarification on a couple points. Other than that, I'm happy to merge.

pyuvdata/uvflag/uvflag.py
```diff
@@ -1166,6 +1176,30 @@ def set_telescope_params(self, overwrite=False, warn=True):
...
+                if self.type == "antenna":
+                    ants_to_check = self.ant_array
+                else:
+                    ants_to_check = np.union1d(self.ant_1_array, self.ant_2_array)
+
+                if not all(
+                    ant in telescope_obj.antenna_numbers for ant in ants_to_check
+                ):
+                    warnings.warn(
+                        "Not all antennas with data have metadata in the telescope "
```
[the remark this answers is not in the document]
bhazelton: This function is only called if one or more of telescope_location, antenna_names, antenna_numbers or antenna_positions is missing from the file. So this should only be called for old files.

If you have some of the antenna metadata but not the other parts (this would be very strange -- there's not a way for this to happen through the normal file writing mechanism) or if you deliberately call this function separately from the file read, then this check happens.

pyuvdata/uvdata/mwa_corr_fits.py
bhazelton: yeah, sorry. This is a little complicated because of all the parametrize options, but the code that adds a weird antenna number doesn't get called in the same parametrize where the `telescope_name` is included in the param_list.

mwilensky768: I think this was meant to address the comment in `test_uvflag`.

I don't understand why the weird antenna block was added. Are you just seeing if you can write a weird antenna into a file?

bhazelton: I'm testing that I get various warnings under various (sometimes weird) situations. There are a whole bunch of different situations (all the different parametrize options), some of which add weird antenna numbers, others don't. Almost all of these situations you can't get to via normal running of the code in either old or new versions. I have to make them on the fly in the test by modifying the hd5 file directly.

mwilensky768:
@bhazelton feel free to merge when ready.

PR #1356, opened 2023-11-04
Better handing of discrepancies for LSTs and telescope position in `check`

### LINKED ISSUE #1348: LST stringency issues
Recently, we've been seeing a lot of warnings about LST accuracy when reading in SMA datasets into pyuvdata. After poking around a bit, this seems to have arisen following #1322, which effectively increased the stringency of the LST array to 1 mas accuracy (translating to 66.7 µs timing accuracy). The issue for the SMA -- or potentially any other array that does phase tracking in real time -- is that the LSTs calculated are based on IERS-predicted values of DUT (versus calculated after the fact), and the accuracy of those predictions are typically 0.25 ms per day extrapolated out (e.g., 4 days out from the last prediction would have ~1 ms accuracy).

Obviously the warnings aren't blocking data from being read in, but I think it's fair to call them spurious, and having lots of them raised seems suboptimal (for no other reason than it sets up a "boy who cried wolf" scenario). I think the core of the issue here is that one value is being used to address two separate questions -- the first being, "absent external information, are two values of LST the 'same' given some specified precision?", and the second being "is this value accurate/consistent with what we would expect given some knowledge of the outside world?" The 1 mas limit for the former makes sense, but I think for the latter it does not, and so I think it would be good to split this somehow in `check`. There may be other similar astrometry-based values (coordinates, uvws, etc.) where we might similarly consider splitting the stringency for precision vs accuracy.

### LINKED ISSUE #1352: Handle telescope locations at the center of the earth for VLBI
Change to checking that antennas are on the surface of the earth (or moon) rather than that the telescope location is on the surface.

Change to a warning rather than an error.

kartographer:
## Description

Updates handling of `lst_array` and `telescope_position` in terms of when warnings and errors get issued inside of `check` for `UVData`, `UVCal`, and `UVFlag`. More specifically, the tolerance for when the warning about "LSTs not set correctly" has been increased to 15 milliarcsec, to better reflect the typical uncertainty in these values at the time of observation (due to the ever-changing DUT1 ). Additionally, `check` now looks at the combination of telescope location _and_ antenna positions in determining whether or not a given set of positions are "reasonable", and when a discrepancy is discovered, a warning is now issued instead of an error.

## Motivation and Context

Relevant issues below, but in both cases, the prior behavior in `check` was resulting in undesired behavior. In the case of the LSTs, a lot of spurious warnings were being generated that were confusing/concerning some users in situations where the "issues" were totally benign. In the case of the telescope position handing, it was breaking support for VLBA/VLBI-generated UVFITS file (without disabling parts of `check`), whose telescope location (phase reference center) is not neccessarily on the surface of the Earth.

Closes #1348
Closes #1352

## Types of changes

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ positions are near surface of whatever celestial body their positions are refere
 
 ### Changed
 - Increased the tolerance to 75 mas (equivalent to 5 ms time error) for a warning about
-values in `lst_array` not conforming to exceptations for `UVData`, `UVCal`, and `UVFlag`
+values in `lst_array` not conforming to expectations for `UVData`, `UVCal`, and `UVFlag`
 (was 1 mas) inside of `check`. Additionally, added a keyword to `check` enable the
 tolerance value to be user-specified.
 - Changed the behavior of checking of telescope location to look at the combination of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file.
 ## [2.4.1] - 2023-10-13
 
 ### Added
+- Added a `check_surface_based_positions` positions method for verifying antenna
+positions are near surface of whatever celestial body their positions are referenced to
+(either the Earth or Moon, currently).
 - Added a `uvw_track_generator` method within `utils` for calculating the expected
 uvws (and a few other values) without needing to instantiate a whole `UVData` object.
 - Added a convenience function called `compare_value` in `UVParameter` that enables
@@ -19,6 +22,14 @@ file to the new `mwa_metafits_file` parameter.
 - Support for recarrays in `UVParameter` objects and in `UVBase`, needed for pyradiosky.
 - Support for setting the astrometry library for various object methods including `set_lsts_from_time_array`, file read methods and others.
 - Properly round-trip the telescope frame through UVH5, UVFITS and MS files.
+
+### Changed
+- Increased the tolerance to 15 mas for a warning about values in `lst_array` not conforming
+to exceptations for `UVData`, `UVCal`, and `UVFlag` (was 1 mas) inside of `check`, additionally
+adding a keyword to enable the tolerance value to be specified.
+- Changed the behavior of checking of telescope location to look at the combination of
+`antenna_positions` and `telescope_location` together for `UVData`, `UVCal`, and `UVFlag`.
+Additionally, failing this check results in a warning (was an error).
 
 ### Fixed
 - A bug in apparent coordinate calculation that resulted in small errors/loss of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,23 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
-## [2.4.1] - 2023-10-13
-
 ### Added
 - Added a `check_surface_based_positions` positions method for verifying antenna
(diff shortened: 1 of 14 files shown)


kartographer:
One related item -- a small file from VLBA (274 kb) outside of the repo was used to check for  VLBA support, which could in principle be added to the repository. Not necessarily something for this PR, but something that was stumbled upon over the course of developing it.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file.
 ## [2.4.1] - 2023-10-13
 
 ### Added
+- Added a `check_surface_based_positions` positions method for verifying antenna
+positions are near surface of whatever celestial body their positions are referenced to
+(either the Earth or Moon, currently).
```
bhazelton, line 11: These should go under the `Unreleased` section.

CHANGELOG.md
```diff
@@ -20,6 +23,14 @@ file to the new `mwa_metafits_file` parameter.
 - Support for setting the astrometry library for various object methods including `set_lsts_from_time_array`, file read methods and others.
 - Properly round-trip the telescope frame through UVH5, UVFITS and MS files.
 
+### Changed
+- Increased the tolerance to 15 mas for a warning about values in `lst_array` not conforming
+to exceptations for `UVData`, `UVCal`, and `UVFlag` (was 1 mas) inside of `check`, additionally
+adding a keyword to enable the tolerance value to be specified.
+- Changed the behavior of checking of telescope location to look at the combination of
+`antenna_positions` and `telescope_location` together for `UVData`, `UVCal`, and `UVFlag`.
+Additionally, failing this check results in a warning (was an error).
```
bhazelton, line 32: These should go under the `Unreleased` section.

bhazelton (commented):
The changes looks great! There's just a little problem with the change log.

CHANGELOG.md
kartographer: Whoops -- my brain saw `Unreleased` and completely skipped over the version right beneath it...

kartographer: Fixed!

CHANGELOG.md
kartographer: Fixed!

kartographer:
I made one additional adjustment here, which was to increase the LST tolerance to 5 ms (75 mas in angle). This is because the estimated uncertainty in DUT1 at one week from IERS Bulletin A is about 1 ms (which are published weekly), and so I set the "detection limit" here at 5 sigma to prevent spurious warnings. I think this should be sufficient for nearly all applications, but should still detect gross errors (> 1 sec) quite easily.

pyuvdata/utils.py
```diff
@@ -115,6 +120,13 @@
                     "rr": ["r", "r"], "ll": ["l", "l"],
                     "rl": ["r", "l"], "lr": ["l", "r"]}
 
+ANGLE_TIME_EQUIV = [(units.s, units.arcsec, lambda x: x * 15.0, lambda x: x / 15.0)]
```
bhazelton, line 123: Maybe I'm missing it, but I don't see anywhere this is used...

kartographer: Ah, that's a bit of leftover code from when I was originally coding this. It's been removed.

bhazelton (approved):
Looks good, thanks!

Issue #1354, opened 2023-11-02
Utils performance for Nants > 2,147,483,648


mkolopanis:
In the largest antennas configuration there is a performance degradation compared to the other utils functions.

This is most likely from the python interactions in the large Nants utility functions.
![image](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/assets/8646829/f6d74375-8792-4380-948f-bcf40139fa13)

I've tried messing with the cython compiler directives like `cdivision`  but it does not seem to fix this issue.

I first noticed this when toying with writing the utilites in rust and noticed the factor of ~10 slow down 
![image](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/assets/8646829/c8788709-2d33-41d8-832f-b3e5df6d29e3)
compared to the 2048 implementation (and 256)

![image](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/assets/8646829/81d6660e-9c29-4ff4-b238-8601c1c05d71)

mkolopanis:
I would have expected switching to cdivision, and casting 2147483648 to a numpy.uint64_t  to resolve this issue. It helps but not fully resolved.
![image](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/assets/8646829/fdddf143-e01d-4203-b81b-181d586c9785)

PR #1357, opened 2023-11-07
Pyx datatypes

### LINKED ISSUE #1353: Unrealistic variable choice in utils extension
A lot of the baseline/ant_num conversion functions take signed 64 bit integers. These should be unsigned integers since these numbers cannot realistically be less than zero.

mkolopanis:
switches to using unsigned 64 bit integers for antenna and baseline numbers in the cython extensions.
Updates the rectangularity calculation to be compatible with unsigned integers.

Forces casting some modulus as c types in the cython extension and uses c division with those numbers (otherwise python tries to cast them as python ints anyway). I expected this would help performance for large numbered baseline (>2\*\*22 + 2\*\*16, see #1354) but it actually seems like it doesn't. Regardless this is more consistent with the other implementations in the extension. 

closes #1353

### DIFF
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -6200,11 +6200,7 @@ def determine_rectangularity(
         baseline_array = baseline_array.reshape((nbls, ntimes))
         if np.sum(np.abs(np.diff(time_array, axis=0))) != 0:
             return False, False
-        if np.any(
-            np.logical_not(
-                np.apply_along_axis(lambda x: np.allclose(x, x[0]), 1, baseline_array)
-            )
-        ):
+        if (np.diff(baseline_array, axis=1) != 0).any():
             return False, False
         return True, True
     elif bl_first:
@@ -6212,10 +6208,6 @@ def determine_rectangularity(
         baseline_array = baseline_array.reshape((ntimes, nbls))
         if np.sum(np.abs(np.diff(time_array, axis=1))) != 0:
             return False, False
-        if np.any(
-            np.logical_not(
-                np.apply_along_axis(lambda x: np.allclose(x, x[0]), 0, baseline_array)
-            )
-        ):
+        if (np.diff(baseline_array, axis=0) != 0).any():
             return False, False
         return True, False
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -6200,7 +6200,11 @@ def determine_rectangularity(
         baseline_array = baseline_array.reshape((nbls, ntimes))
         if np.sum(np.abs(np.diff(time_array, axis=0))) != 0:
             return False, False
-        if np.sum(np.abs(np.diff(baseline_array, axis=1))) != 0:
+        if np.any(
+            np.logical_not(
+                np.apply_along_axis(lambda x: np.allclose(x, x[0]), 1, baseline_array)
+            )
+        ):
             return False, False
         return True, True
     elif bl_first:
@@ -6208,6 +6212,10 @@ def determine_rectangularity(
         baseline_array = baseline_array.reshape((ntimes, nbls))
         if np.sum(np.abs(np.diff(time_array, axis=1))) != 0:
             return False, False
-        if np.sum(np.abs(np.diff(baseline_array, axis=0))) != 0:
+        if np.any(
+            np.logical_not(
+                np.apply_along_axis(lambda x: np.allclose(x, x[0]), 0, baseline_array)
+            )
+        ):
             return False, False
         return True, False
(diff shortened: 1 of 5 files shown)


mkolopanis:
annotated cython output on the changes. says it should be just c 
image

steven-murray (dismissed):
Not sure I can quite speak to all the Cython bits, but it looks reasonable. The rectangularity part looks good.

bhazelton:
It looks like there are some hera_cal errors popping up in `check` calls about ant_array (for UVCal) and ant_1_array (for UVData) being `float64` rather than `int`.

mkolopanis:
oh wow, this is all in an innit loop........ there's going to need to be more looking into it. But the changes here don't seem like they should be causing this.

mkolopanis:
I think I found what is happening. I actually wonder if this is a numpy bug. When you call intersect1d on two arrays, one is uint64 the other is int64, you get a return type of float64. Happens in uvcal initializers https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/bd03009960ba2f6aeb79cffc75e3011ef58dc310/pyuvdata/uvcal/initializers.py#L481-L484

```python
In [2]: import numpy as np

In [3]: a1 = np.arange(12, dtype=np.uint64)

In [4]: a2 = np.arange(5,10)

In [5]: a2.dtype
Out[5]: dtype('int64')

In [6]: a1.dtype
Out[6]: dtype('uint64')

In [7]: a3 = np.intersect1d(a1,a2)

In [8]: a3.dtype
Out[8]: dtype('float64')

In [9]: 
```

mkolopanis:
probably what numpy would say is the more generic form, a superset of numbers that can deal with both types.

mkolopanis:
I cannot figure out _why_ the second error is happening though. 
in hera_cal's lst bin tests there's some antpairs that are int64 and one set that is uint64.. so when it casts the antpairs to an array we get floats again

like i added this in line 2756 in hera_cal/io.py
```python
    print(f"{data._antpairs=:} {[type(a) for ap in data._antpairs for a in ap]}")
```
and get
```python
antpairs=[(1, 1), (3, 3), (14, 14), (36, 36), (52, 100), (98, 143), (102, 144), (103, 176), (127, 162), (135, 166), (140, 158), (164, 191)]
[<class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.uint64'>, <class 'numpy.uint64'>, <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.uint64'>, <class 'numpy.uint64'>, <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.uint64'>, <class 'numpy.uint64'>]
```
but can't figure out why there are different types

mkolopanis:
just found the tracking bug for this https://github.com/numpy/numpy/issues/20905

mkolopanis:
This last error seems to be from iterative operations on `lst_bin` in hera_cal. @steven-murray has disclosed this function is deprecated and will be removed in the future for hera_cal. I can propose a downstream workaround for hera_cal to fix this in their code but can't figure out the cause.

setup.cfg
```diff
@@ -12,6 +12,7 @@ max-line-length = 88
 per-file-ignores =
     pyuvdata/tests/*.py: D
     pyuvdata/*/tests/*.py: D
+    pyuvdata/uvdata/tests/test_mir_parser.py: D, B035
```
bhazelton, line 15: I don't love this because I think it actually is a sensible check. That being said, this is a test file and I'm not quite sure how to fix it immediately. Maybe we can merge this but ask @kartographer to take a look and see if he thinks ignoring it here is the best answer or if there's a way to write it that is equally good and doesn't trigger this check. The worry is that you're ignoring the whole file here, not just those lines.

bhazelton (approved):
Looks good to me. Waiting to hear if @steven-murray has any concerns.

steven-murray:
No concerns, but I would like to apply the patch @mkolopanis recommended to hera_cal to see if it comes right

setup.cfg
mkolopanis: dropped the changes here for karto's actual fix in the rebase

bhazelton (approved):

PR #1363, opened 2023-12-06
MS writer improvements

### LINKED ISSUE #1355: Reading and writing ms files does not preserve data
I am testing a CASA-based calibration and imaging pipeline that accepts .ms files as input.

Directly inputting the .ms file (in this case from the OVRO-LWA) produces this image. The colorbar scale here is 0-70 (arbitrary units):
Screenshot 2023-11-02 at 4 07 00 PM

Next, I've tried reading the .ms file in pyuvdata and immediately rewriting it with these commands:
`uv.read_ms("20230819_093023_73MHz.ms")`
`uv.write_ms("20230819_093023_73MHz_rewritten.ms")`
Now using the exact same calibration and imaging pipeline with the new .ms file gives this image (colorbar scale 0-400):
Screenshot 2023-11-02 at 4 03 00 PM

It looks like many visibilities are being flagged by CASA's calibration. The bottom line is that the `read_ms` and `write_ms` commands do not preserve the .ms file.

kartographer:
## Description

Adds some functionality to the `write_ms` method, which enforces that CASA-preferred polarization ordering is used on write, and allows the user to specify whether to flip the baseline conjugation scheme used.

## Motivation and Context

This addresses some issues that LWA users were running into when passing data through puvdata. Two issues in particular were identified in #1355. First, some CASA tasks clearly want the polarization order to be set up in a particular way, so the writer now always writes out in "CASA order". Second, there's an open question as to if some methods and functions (both internal and external to the CASA package) require different baseline conventions, which would be understandable given the conflicting documentation. Though I believe the pyuvdata standard is correct for most applications, I've added a switch that users can use to flip the convention.

In the course of addressing the above, I did make a few small tweaks to the writer and added/modified a couple of functions in `utils` to help with polarization ordering and to make indexing faster under most circumstances via use of slices.

Closes #1355.

## Types of changes

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,16 +4,19 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
-- Added a `check_surface_based_positions` positions method for verifying antenna
-positions are near surface of whatever celestial body their positions are referenced to
-(either the Earth or Moon, currently).
 - Added a switch to `UVData.write_ms` called `flip_conj`, which allows a user to write
 out data with the baseline conjugation scheme flipped from the standard `UVData`
 convention.
 - Added the `utils.determine_pol_order` method for determining polarization
 order based on a specified scheme ("AIPS" or "CASA").
+- Added a `check_surface_based_positions` positions method for verifying antenna
+positions are near surface of whatever celestial body their positions are referenced to
+(either the Earth or Moon, currently).
 
 ### Changed
+- Changed `UVData.write_ms` to sort polarizations based on CASA-preferred ordering.
+- Added some functionality to the `utils._convert_to_slices` method to enable quick
+assessment of whether an indexing array can be replaced by a single slice.
 - Increased the tolerance to 75 mas (equivalent to 5 ms time error) for a warning about
 values in `lst_array` not conforming to expectations for `UVData`, `UVCal`, and `UVFlag`
 (was 1 mas) inside of `check`. Additionally, added a keyword to `check` enable the
@@ -21,9 +24,6 @@ tolerance value to be user-specified.
 - Changed the behavior of checking of telescope location to look at the combination of
 `antenna_positions` and `telescope_location` together for `UVData`, `UVCal`, and `UVFlag`.
 Additionally, failing this check results in a warning (was an error).
-- Changed `UVData.write_ms` to sort polarizations based on CASA-preferred ordering.
-- Added some functionality to the `utils._convert_to_slices` method to enable quick
-assessment of whether an indexing array can be replaced by a single slice.
 
 ## [2.4.1] - 2023-10-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ All notable changes to this project will be documented in this file.
 - Added a `check_surface_based_positions` positions method for verifying antenna
 positions are near surface of whatever celestial body their positions are referenced to
 (either the Earth or Moon, currently).
+- Added a switch to `UVData.write_ms` called `flip_conj`, which allows a user to write
+out data with the baseline conjugation scheme flipped from the standard `UVData`
+convention.
+- Added the `utils.determine_pol_order` method for determining polarization
+order based on a specified scheme ("AIPS" or "CASA").
 
 ### Changed
 - Increased the tolerance to 75 mas (equivalent to 5 ms time error) for a warning about
@@ -16,6 +21,9 @@ tolerance value to be user-specified.
 - Changed the behavior of checking of telescope location to look at the combination of
 `antenna_positions` and `telescope_location` together for `UVData`, `UVCal`, and `UVFlag`.
 Additionally, failing this check results in a warning (was an error).
+- Changed `UVData.write_ms` to sort polarizations based on CASA-preferred ordering.
+- Added some functionality to the `utils._convert_to_slices` method to enable quick
+assessment of whether an indexing array can be replaced by a single slice.
 
 ## [2.4.1] - 2023-10-13
 
--- a/pyuvdata/utils.py
+++ b/pyuvdata/utils.py
@@ -1250,6 +1250,42 @@ def reorder_conj_pols(pols):
     return conj_order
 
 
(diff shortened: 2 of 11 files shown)


pyuvdata/utils.py
```diff
@@ -5706,88 +5742,112 @@ def _get_dset_shape(dset, indices):
...
-    Nslices = len(slices)
-    passed = (float(Nslices) / len(indices)) < max_nslice_frac
+    # determine whether slices are a reasonable representation, and determine max_nslice
+    # if only max_nslice_frac was supplied.
+    if max_nslice is None:
+        max_nslice = max_nslice_frac * Ninds
+    check = len(slices) <= max_nslice
 
-    return slices, passed
+    return [indices] if (not check and return_index_on_fail) else slices, check
```
bhazelton, line 5850: I find this (and generally these kinds of in-line if statements) a little hard to read. I'd slightly prefer the longer multi-line if statement. Open to other votes however.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ All notable changes to this project will be documented in this file.
 - Added a `check_surface_based_positions` positions method for verifying antenna
 positions are near surface of whatever celestial body their positions are referenced to
 (either the Earth or Moon, currently).
+- Added a switch to `UVData.write_ms` called `flip_conj`, which allows a user to write
+out data with the baseline conjugation scheme flipped from the standard `UVData`
+convention.
+- Added the `utils.determine_pol_order` method for determining polarization
+order based on a specified scheme ("AIPS" or "CASA").
```
bhazelton, line 14: This should be at the top (reverse chronological).

CHANGELOG.md
```diff
@@ -16,6 +21,9 @@ tolerance value to be user-specified.
 - Changed the behavior of checking of telescope location to look at the combination of
 `antenna_positions` and `telescope_location` together for `UVData`, `UVCal`, and `UVFlag`.
 Additionally, failing this check results in a warning (was an error).
+- Changed `UVData.write_ms` to sort polarizations based on CASA-preferred ordering.
+- Added some functionality to the `utils._convert_to_slices` method to enable quick
+assessment of whether an indexing array can be replaced by a single slice.
```
bhazelton, line 26: This should be at the top (reverse chronological).

pyuvdata/uvdata/ms.py
```diff
@@ -1341,12 +1381,12 @@ def write_ms(
...
         self._write_ms_antenna(filepath)
         self._write_ms_data_description(filepath)
-        self._write_ms_feed(filepath)
+        self._write_ms_feed(filepath, pol_order=pol_order)
         self._write_ms_field(filepath)
         self._write_ms_source(filepath)
         self._write_ms_spectralwindow(filepath)
         self._write_ms_pointing(filepath)
-        self._write_ms_polarization(filepath)
+        self._write_ms_polarization(filepath, pol_order=pol_order)
```
bhazelton, line 1389: I think you wouldn't need these `pol_order` arguments to these private methods (or the code to handle them in those methods) at all if you did the `reorder_pols` call before these calls. Am I missing something?

bhazelton (commented):
Thanks @kartographer ! Just a few comments.

bhazelton:
@rlbyrne we'd be happy to hear any feedback from you on this if you have time to look at it or test it.

pyuvdata/uvdata/tests/test_mir_parser.py
```diff
@@ -176,7 +176,7 @@ def test_mir_raw_data(mir_data, tmp_path):
 
     mir_data._write_cross_data(filepath)
     # Sub out the file we need to read from
-    mir_data._file_dict = {filepath: item for item in mir_data._file_dict.values()}
+    mir_data._file_dict = {filepath: list(mir_data._file_dict.values())[0]}
```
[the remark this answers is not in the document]
kartographer: Ahh, didn't realize he also was dealing with cranky flake8 messages. Does help that I'm the one who wrote the module here 😉

CHANGELOG.md
kartographer: This is either a sign that I'm writing too many or too few PRs...

pyuvdata/utils.py
kartographer: Changed in 9c862da2

CHANGELOG.md
kartographer: Fixed in 9c862da2

pyuvdata/uvdata/ms.py
bhazelton: I kind of agree, except that you are then calling the reorder function later. If you weren't then I'd totally agree.

One option is to record the order, then reorder, then reorder again to get back to the original order. But if you are calling the reorder method anywhere I don't see a need to handle the ordering in the private methods...

bhazelton: I take it back, I mis-read the diff. The reordering that I thought was being done later in the write is actually in the read. I like your approach.

rlbyrne:
@kartographer Apologies that I'm only getting around to testing this now.

I am getting an error with the new ms writer. When I read an .uvfits file and then immediately write to an .ms file, I get this message:
```Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/devel/rbyrne/envs/pyuvdata/lib/python3.8/site-packages/pyuvdata/uvdata/uvdata.py", line 12731, in write_ms
    ms_obj.write_ms(
  File "/opt/devel/rbyrne/envs/pyuvdata/lib/python3.8/site-packages/pyuvdata/uvdata/ms.py", line 1219, in write_ms
    ms.putcol(col, temp_vals)
  File "/opt/devel/rbyrne/envs/pyuvdata/lib/python3.8/site-packages/casacore/tables/table.py", line 1155, in putcol
    self._putcol(columnname, startrow, nrow, rowincr, value)
RuntimeError: ArrayColumn::putColumn - column DATA has 62128, array has 4 rows: Table array conformance error```
If I run `uv.reorder_pols(order="CASA", run_check=False)` before writing I do not get the error. Any thoughts? I can put this in an issue as well.

PR #1376, opened 2024-01-18
Mcmf enu calc

### LINKED ISSUE #1351: utils.py has error in enu_to_ecef / enu_to_ecef on the moon
There is an indexing error on lines 1508 & 1576 .  We *think* it should be [1,2,0] but aren't 100% sure. 

That said, the bigger issue is that this code is totally separate from the earth-based code.  We should let ECEF_from_ENU (in the cython) take a frame.  Right now, it has an if/else for earth vs moon, and the moon block has completely independent code.  To make this work, XYZ_from_lat_lon_alt will need to accept the frame and then use lunar values for things like gps_n etc.  It may be that we can get those values from lunarsky and, if we assume spherical symmetry, a lot of them become redundant.

mkolopanis:
## Description

Added a generic Ellipsoid class to the cython utils extension. This then calculates the necessary gps related constants for calculating lat, lon, and alt from xyz.
Currently has earth and the moon Bodies. Uses the lunarsky moon radius (matches current erfa calc)
added the moon's ellipsoid flattening parameter from https://nssdc.gsfc.nasa.gov/planetary/factsheet/moonfact.html
This in turn required updating some test quantities to account for the fact that the moon is not a perfect sphere.

There is a small computational efficiency regression in these changes as outlined in the accompanying pdf.
[xyz_calcs_update.pdf](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/files/13981420/xyz_calcs_update.pdf)

@kartographer  this seems to also now pass your small eastward delta check
```python 


In [1]: import pyuvdata.utils as uvutils
   ...: 
   ...: print("ITRS: ", uvutils.ECEF_from_ENU([0.001, 0, 0], 0, 0, 0, frame="ITRS"))
   ...: 
   ...: print("MCMF: ", uvutils.ECEF_from_ENU([0.001, 0, 0], 0, 0, 0, frame="MCMF"))

ITRS:  [6.378137e+06 1.000000e-03 0.000000e+00]
MCMF:  [1.7371e+06 1.0000e-03 0.0000e+00]
````

Still need to update the changelog but wanted to get some feedback if this is how we'd want to solve a problem like this. It does also allow some flexibility to add additional telescope worlds in the future.

## Motivation and Context

closes #1351 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### CODE REVIEW

File: pyuvdata/utils.pyx [hunk 1]
```diff
@@ -19,21 +21,35 @@ from libc.math cimport atan2, cos, sin, sqrt
...
-# expose up to python
-gps_a = _gps_a
-gps_b = _gps_b
-e_squared = _e2
-e_prime_squared = _ep2
+# keep for consistent API though these really shouldn't be used anymore
+gps_a = Body.Earth.value.gps_a
+gps_b =  Body.Earth.value.gps_b
+e_squared =  Body.Earth.value.e_squared
+e_prime_squared =  Body.Earth.value.e_prime_squared
```
bhazelton, line 52: Is anything in pyuvdata using these? Or is this for other packages?

bhazelton:
I generally like this approach.

pyuvdata/utils.pyx
mkolopanis: This is the only file I see these constants being used, just didn't want to remove it because inevitably someone somewhere is using it

bhazelton: 😆 One option would be to remove them in v3.0 where we're allowed to make breaking API changes...

mkolopanis: I'm not opposed!

aelanman:
I suppose I should update lunarsky's selenodetic coordinates to include the flattening term, though I'm not sure if there is a standard reference ellipsoid for the moon.

mkolopanis:
Yeah I couldn't find anything that looked terribly official, even that nasa site seemed more of an outreach info than published doctrine.

aelanman:
I found this paper about defining a reference ellipsoid ("selenoid") for the moon: https://link.springer.com/article/10.1007/s40328-023-00415-w

However, this elevation map uses a spherical reference: https://astrogeology.usgs.gov/search/details/Moon/LRO/LOLA/Lunar_LRO_LOLA_Global_LDEM_118m_Mar2014/cub

mkolopanis:
Yeah I saw that paper as well. not the elevation map. I think @jpober is going to help us hunt down a source that lunar people find trustworthy :laughing:

bhazelton:
We didn't find anything more authoritative and the sourcing is referenced in the code, so we'll go with this until there's better information.

bhazelton (approved):
thanks @mkolopanis !

PR #1382, opened 2024-01-23
Minor bug fix for `tests.WarningChecker` when IERS is down

kartographer:
Fixes testing bug.

## Description

I've updated the `WarningChecker` class inside of `pyuvdata.tests` to handle when the IERS service is down and when `check_warnings` is being called with the expected warning and text match being set to `None` (which effectively ensures that no warnings are being thrown).

## Motivation and Context

There are about half a dozen tests which use `check_warnings` to make sure no warnings are being thrown, which were failing due to the previously uncaught behavior.  This sems to be a fairly rare occurence that's popped up recently due to what appear to be IERS server issues.

## Types of changes

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,9 @@ fully tested and didn't work properly.
 - Having `freq_array` and `channel_width` defined on wide-band UVCal objects.
 
 ### Fixed
+- A small bug (mostly affecting continuous integration) that threw an error when the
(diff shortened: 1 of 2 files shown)


kartographer:
Note that this seems to be a different error from the one impacting the warnings ci. This problem was causing intermittent failures on the other pipelines.

bhazelton (approved):
Looks reasonable. Thanks!

Issue #854, opened 2020-06-04
Generalize x_orientation concept


jpober:
x_orientation keyword might need some adaptation to support arrays where the dishes steer/rotate.

bhazelton:
linked to #1095

PR #1383, opened 2024-01-25
Add AnalyticBeam objects and a unified interface for UVBeams and AnalyticBeams

bhazelton:
## Description

This is a refactor of some code that has existed in pyuvsim for some time with some new functionality to make it easier to work with analytic beams.

Major changes:
- Add a new AnalyticBeam base class and several specific analytic beam objects (Airy, Gaussian, Short dipole, Uniform) that support evaluating either the E-field or power response in any direction and frequency.
- Add a BeamInterface object which can have either a UVBeam or an Analytic Beam attached to it and which can be used to get beam responses in any direction and frequency via a unified method that either calls `UVBeam.interp` or the appropriate  AnalyticBeam eval method.

There are also a couple of minor changes:
- Fix a bug in the the new beam initializer method where it makes data_arrays with the wrong type when cross pols are included.
- Small refactoring in UVBeam

## Motivation and Context

This will substantially clean up code in pyuvsim, make the analytic beams more accessible for other users and make it easier for simulators by providing a unified interface for UVBeam and Analytic Beam objects.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:


steven-murray (commented):
Thanks @bhazelton, this is looking really nice. I've made a few comments on things I think could be a little tighter.

Some extra stuff:

1. I think the analytic beams could easily support YAML interfaces. There are two ways to do this. The first would be to use the `__init_subclass__` method on the base class, and fill up a private global dictionary with any subclasses, i.e.:

```python
def __init_subclass__(cls, is_abstract=False, **kwargs):
        """Provide plugin capablity and do some verification of a defined plugin."""
        super().__init_subclass__(**kwargs)

        # Plugin framework
        if not is_abstract:
            cls._plugins[cls.__name__] = cls
```

This is nice because now we globally know the possible list of AnalyticBeams. However, to be added to the list, the class has to be imported, which can't easily be ensured from a YAML file (unless you have some syntax where you have a top-level list like:

```yaml
plugins:
   - custom.module
   - other_package.submodule
```
And your YAML reader always makes sure to import those modules before doing any other loading. If you do all this, you can simply write a YAML constructor on the base class whose name is the subclass. So you could then write YAML as eg.:

```yaml
beam: !GaussianBeam
   diameter: 4.0
```
On `yaml.load(fl)` you'd get an instantiated `GaussianBeam`, and any custom-written classes would also be fine, AS LONG AS THEY ARE IMPORTED FIRST. 

If you don't want to rely on things being imported first, then you could instead allow behaviour like this:

```yaml
beam: !AnalyticBeam
   class: module.CustomBeam
   diameter: 5.0
```
To do this, you'd simply use `yaml.add_constructor("!AnalyticBeam", func_that_turns_data_into_beam, Loader=yaml.FullLoader)`. In that func, you'd do an import of the `class` and use it to construct the object. You could also support both ways if you wanted.

2. We should add some docs on how to use the new AnalyticBeam's and the beam interface.

src/pyuvdata/uvbeam/analytic_beam.py
```diff
@@ -0,0 +1,899 @@
...
+
+    @property
+    @abstractmethod
+    def basis_vector_type(self):
+        """Require that a basis_vector_type is defined in concrete classes."""
+
+    @property
+    @abstractmethod
+    def name(self):
+        """Require that a name is defined in concrete classes."""
```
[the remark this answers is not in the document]
bhazelton: I think `__repr__` can probably work for where we need a name.

src/pyuvdata/uvbeam/analytic_beam.py
```diff
@@ -0,0 +1,899 @@
...
+
+        if az_array.shape != za_array.shape:
+            raise ValueError("az_array and za_array must have the same shape.")
+
+    def _get_empty_data_array(
+        self, npts: int, nfreqs: int, beam_type: str = "efield"
+    ) -> npt.NDArray[np.float]:
+        """Get the empty data to fill in the eval methods."""
+        if beam_type == "efield":
+            return np.zeros((self.Naxes_vec, self.Nfeeds, nfreqs, npts), dtype=float)
```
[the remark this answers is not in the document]
bhazelton: hmm. I think it can in principle be complex for efield beams or for cross-pol power beams, so I will change it. Currently all our analytic beams are real, so I didn't catch it, thanks for finding it.

src/pyuvdata/uvbeam/analytic_beam.py
```diff
@@ -0,0 +1,899 @@
...
+        else:
+            return np.zeros((1, self.Npols, nfreqs, npts), dtype=float)
+
+    @abstractmethod
+    def _efield_eval(
+        self,
+        *,
+        az_array: npt.NDArray[np.float],
+        za_array: npt.NDArray[np.float],
+        freq_array: npt.NDArray[np.float],
```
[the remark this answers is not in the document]
bhazelton: I think we need it to be like this for cross-compatibility. It's also in keeping with the style elsewhere in pyuvdata. There was discussion about these names back when we first started, I remember being in a room at UW with Jonnie and Danny and Adam where we decided on several conventions like this. Hard to change now.

src/pyuvdata/uvbeam/analytic_beam.py
```diff
@@ -0,0 +1,899 @@
...
+    @combine_docstrings(UVBeam.new)
+    def to_uvbeam(
+        self,
+        freq_array: npt.NDArray[np.float],
+        beam_type: Literal["efield", "power"] = "efield",
+        pixel_coordinate_system: (
+            Literal["az_za", "orthoslant_zenith", "healpix"] | None
+        ) = None,
+        **kwargs,
+    ):
```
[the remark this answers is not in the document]
bhazelton: So all the other parameters are just passed straight through to `UVBeam.new`. I could add a pointer to that method, would that help? Or I could try to do something with `docstring_parser` to actually add them to the docstring.

src/pyuvdata/uvbeam/beam_interface.py
```diff
@@ -0,0 +1,244 @@
...
+                    print("Beams do not match.")
+                return False
+            if not self._isuvbeam and self.beam_type != other.beam_type:
+                if not silent:
+                    print(
+                        "Beam types do not match. "
+                        f"Left is {self.beam_type},"
+                        f" right is {other.beam_type}."
+                    )
+                return False
```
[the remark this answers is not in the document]
bhazelton: I'll use dataclass. Seems to do what we need without an additional import.

src/pyuvdata/uvbeam/beam_interface.py
```diff
@@ -0,0 +1,244 @@
...
+            raise ValueError(
+                "beam must be a UVBeam or an AnalyticBeam instance, not a "
+                f"{type(beam)}."
+            )
+        self.beam = beam
+        if isinstance(beam, UVBeam):
+            self._isuvbeam = True
+            if beam_type is None or beam_type == beam.beam_type:
+                self.beam_type = beam.beam_type
+            elif beam_type == "power":
```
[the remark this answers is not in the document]
bhazelton: I think it's ok. The first `if` sets captures the case where we haven't specified a beam_type or we have and it matches the input beam. The `elif` catches the case where we specified `power` but the input is `efield` (has to be `efield` or it would have been caught in the `if` above because they'd match). The `else` captures the case where we specified `efield` but the input is `power`, which isn't allowed.

src/pyuvdata/uvbeam/analytic_beam.py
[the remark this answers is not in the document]
steven-murray: You could potentially just add a note in the docstring that all other parameters go through to `UVBeam.new`, and make special reference to `axis1_array` and `axis2_array` since those will be the most commonly used (and in fact required here).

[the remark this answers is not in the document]
bhazelton: I'll add some more text to the docstring to point to `UVBeam.new` and call out the most important parameters.

`axis1_array` and `axis2_array` are not always required. They are not required if you're setting up a healpix beam. We have a test that sets up a healpix beam using this method.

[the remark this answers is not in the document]
bhazelton: Actually, looking at this more, I think I agree with you that it's confusing and I'm just going to define all the possible parameters that can be passed it and remove the `kwargs` option. It turns out that you actually cannot pass in all of the parameters supported by UVBeam and there aren't that many that you can pass in.

src/pyuvdata/uvbeam/analytic_beam.py
```diff
@@ -0,0 +1,893 @@
...
+            feed_array = ["x", "y"]
+        self.feed_array = np.asarray(feed_array)
+
+        self.Nfeeds = self.feed_array.size
+        self.x_orientation = x_orientation
+
+        self.polarization_array, _ = _convert_feeds_to_pols(
+            self.feed_array, include_cross_pols, x_orientation=self.x_orientation
+        )
+        self.Npols = self.polarization_array.size
```
steven-murray, line 100: You should not need an `__init__` with a `dataclass`.

src/pyuvdata/uvbeam/analytic_beam.py
```diff
@@ -0,0 +1,893 @@
...
+        if feed_array is not None:
+            for feed in feed_array:
+                allowed_feeds = ["n", "e", "x", "y", "r", "l"]
+                if feed not in allowed_feeds:
+                    raise ValueError(f"Feeds must be one of: {allowed_feeds}")
+        else:
+            feed_array = ["x", "y"]
+        self.feed_array = np.asarray(feed_array)
+
+        self.Nfeeds = self.feed_array.size
```
steven-murray, line 94: this can be a `@property`

src/pyuvdata/uvbeam/analytic_beam.py
steven-murray, line 100: This also can be a `@property`

src/pyuvdata/uvbeam/analytic_beam.py
```diff
@@ -0,0 +1,893 @@
...
+        self.x_orientation = x_orientation
+
+        self.polarization_array, _ = _convert_feeds_to_pols(
+            self.feed_array, include_cross_pols, x_orientation=self.x_orientation
+        )
```
steven-murray, line 99: This can either be set in the `__post_init__` or as a cached property

src/pyuvdata/uvbeam/analytic_beam.py
```diff
@@ -0,0 +1,893 @@
...
+        self.Naxes_vec = self._basis_vec_dict[self.basis_vector_type]
+
+        if feed_array is not None:
+            for feed in feed_array:
+                allowed_feeds = ["n", "e", "x", "y", "r", "l"]
+                if feed not in allowed_feeds:
+                    raise ValueError(f"Feeds must be one of: {allowed_feeds}")
+        else:
+            feed_array = ["x", "y"]
+        self.feed_array = np.asarray(feed_array)
```
steven-murray, line 92: Again, this can be managed in the `__post_init__`, but if we were using `attrs` (or `pydantic`, though I haven't used that) it would be even simpler:

```python
import attrs
from attrs import validators as vld

@attrs.define
class AnalyticBeam(ABC):
    ...
    feed_array: np.ndarray = attrs.field(
        converter=np.asarray, 
        validator=vld.deep_iterable(vld.in_(["n", "e", "x", "y", "r", "l"]))
    )

File: src/pyuvdata/uvbeam/analytic_beam.py [hunk 11]
```diff
@@ -0,0 +1,893 @@
...
+        *,
+        feed_array: npt.NDArray[str] | None = None,
+        include_cross_pols: bool = True,
+        x_orientation: Literal["east", "north"] = "east",
+    ):
+        if self.basis_vector_type not in self._basis_vec_dict:
+            raise ValueError(
+                f"basis_vector_type is {self.basis_vector_type}, must be one of "
+                f"{list(self._basis_vec_dict.keys())}"
+            )
```
steven-murray, line 82: since `basis_vector_type` is a _class_ variable, not an instance variable, it should be validated at class construction time (i.e. in the `__init_subclass__` method).

steven-murray: oh, sorry I'm wrong -- it's a property. But I think you mean it to be a class attribute?

src/pyuvdata/uvbeam/analytic_beam.py
```diff
@@ -0,0 +1,893 @@
...
+        self.Naxes_vec = self._basis_vec_dict[self.basis_vector_type]
```
steven-murray, line 83: should be a property

src/pyuvdata/uvbeam/analytic_beam.py
```diff
@@ -0,0 +1,893 @@
...
+        if uvb.pixel_coordinate_system == "az_za":
+            data_array = data_array.reshape(uvb.data_array.shape)
+
+        uvb.data_array = data_array
+
+        uvb.check()
+        return uvb
+
+
+def analytic_beam_constructor(loader, node):
```
steven-murray, line 408: should this be private?

src/pyuvdata/uvbeam/analytic_beam.py
```diff
@@ -0,0 +1,893 @@
...
+    """
+    Define a yaml constructor for analytic beams.
+
+    The yaml must specify a "class" field with an importable class and any
+    required inputs to that class's constructor.
+
```
steven-murray, line 414: I think we could add a short example of the YAML here, e.g. 

```yaml
beam: !AnalyticBeam
   class: Gaussian
   diameter: 4.0
```

src/pyuvdata/uvbeam/analytic_beam.py
```diff
@@ -0,0 +1,893 @@
...
+    class_name = class_parts[-1]
+    if len(class_parts) == 1:
+        # no module specified, assume pyuvdata
+        module = importlib.import_module("pyuvdata")
+    else:
+        module = (".").join(class_parts[:-1])
+        module = importlib.import_module(module)
+    beam_class = getattr(module, class_name)
+
+    beam = beam_class(**values)
```
steven-murray, line 441: I think we can offer slightly more ease-of-use by also implementing the plugin structure I mentioned previously. So, implement the following magic method on this class:

```python
class AnalyticBeam:
    __types__: dict = {}

    def __init_subclass__(cls):
        # consider using cls.name.lower()
        cls.__types__[cls.name] = cls
```

Then in this method, we do something similar to what you have:

```python
if len(class_parts) > 1:
    module = (".").join(class_parts[:-1])
    importlib.import_module(module)
 
# any  subclass that has been imported will work (obviously including the pyuvdata ones...)
beam_class = AnalyticBeam.__types__[class_name]
beam = beam_class(**values)
```

steven-murray: This also then means we can do more helpful validation:

```python
if class_name not in `AnalyticBeam.__types__`:
    raise NameError(f"{class_name} is not a known AnalyticBeam. Available options are: {list(AnalyticBeam.__types__.keys())}. If it is a custom beam, either ensure the module is imported, or specify the beam with dot-pathed modules included (i.e. `my_module.MyAnalyticBeam`)")
```

src/pyuvdata/uvbeam/analytic_beam.py
```diff
@@ -0,0 +1,893 @@
...
+
+    return beam
+
+
+yaml.add_constructor("!AnalyticBeam", analytic_beam_constructor, Loader=yaml.SafeLoader)
```
steven-murray, line 446: I would _also_ add it to the `FullLoader` since people sometimes use that (even if pyuvsim doesn't)

src/pyuvdata/uvbeam/analytic_beam.py
```diff
@@ -0,0 +1,893 @@
...
+        self,
+        *,
+        sigma: float | None = None,
+        sigma_type: Literal["efield", "power"] = "efield",
+        diameter: float | None = None,
+        spectral_index: float = 0.0,
+        reference_frequency: float = None,
+        feed_array: npt.NDArray[str] | None = None,
+        include_cross_pols: bool = True,
+    ):
```
steven-murray, line 560: same comment above that you shouldn't need `__init__` (it is just redundant with the fields defined above). For validation and conversion of inputs, use `__post_init__`.

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -4451,3 +4472,104 @@ def write_beamfits(self, filename, **kwargs):
...
+                "If 'path_variable' is specified, it should take the form of a "
+                "module.variable_name where the variable name can be imported "
+                "from the module."
+            )
+        else:
+            module = (".").join(path_parts[:-1])
+            module = importlib.import_module(module)
+        path_var = getattr(module, var_name)
+        for f_i in range(len(files_use)):
+            files_use[f_i] = os.path.join(path_var, files_use[f_i])
```
steven-murray, line 4519: I'm not sure I understand the point of this? Is it that you have some UVBeams installed with pyuvdata that you'd like to be able to point to easily without knowing before hand where the installation directory is? I can imagine slicker interfaces to this, but perhaps this is the simplest way to achieve it.

steven-murray: I could also imagine that, if the `filename` can be a list of files, that it might be useful to allow it to be a glob-style string.

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -4451,3 +4472,104 @@ def write_beamfits(self, filename, **kwargs):
...
+    if len(files_use) == 1:
+        files_use = files_use[0]
+    values["filename"] = files_use
+
+    beam = UVBeam.from_file(**values)
+
+    return beam
+
+
+yaml.add_constructor("!UVBeam", uvbeam_constructor, Loader=yaml.SafeLoader)
```
steven-murray, line 4530: also add to FullLoader.

src/pyuvdata/uvbeam/analytic_beam.py
bhazelton: I need an `__init__` because I have a lot of things I need to set based on the inputs, it's not just a straightforward assignment.

src/pyuvdata/uvbeam/analytic_beam.py
bhazelton: Why do you prefer a `__post_init__` to an `__init__`?

src/pyuvdata/uvbeam/analytic_beam.py
steven-murray: See my other comments for why I don't think you do

src/pyuvdata/uvbeam/analytic_beam.py
steven-murray: because writing an init means you have redundant definitions of types for the input parameters

src/pyuvdata/uvbeam/analytic_beam.py
bhazelton: It's a class attribute, but it's required to exist, so you had me write it as a property so that we could use the `abstractmethod` decorator. In principal, it can be different for different subclasses.

src/pyuvdata/uvbeam/analytic_beam.py
bhazelton: I don't think that last point will work. It seems like the `__init_subclass__` is only run when AnalyticBeam is imported, so if the new module is imported afterwards (e.g. in the function above where this is called), it won't exist in the `AnalyticBeam.__types__`. I even tried adding `AnalyticBeam.__init_subclass__()` after the new import and it doesn't update the `AnalyticBeam.__types__`.

bhazelton: I can put the whole earlier block in a try/except and use this message as the final error in the except.

steven-murray: Hmmm, I'm not understanding what won't work here. I've definitely used this pattern before, but maybe I'm misunderstanding what you're saying. Any subclass that is imported at any time will add itself to `__types__` upon definition.

bhazelton: Maybe the problem is how I'm trying to test it. I just pushed a commit that adds this infrastructure and adds a test that works without the final check but breaks with it.

I order to access the part of the constructor that works imports a class if it isn't already imported, I set `AnalyticBeam.__types__={}`. That ensures that I enter that part of the code but then it errors on the final check.

src/pyuvdata/uvbeam/analytic_beam.py
bhazelton: Seems to work to move it into the `__init_subclass__` and I agree it's cleaner there.

src/pyuvdata/uvbeam/analytic_beam.py
bhazelton: I think this is sorted out now, the problem was in the test, I came up with a better approach for that.

src/pyuvdata/analytic_beam.py
```diff
@@ -0,0 +1,1040 @@
...
+        module = importlib.import_module(module)
+        beam_class = getattr(module, class_name)
+
+    if class_name not in AnalyticBeam.__types__:
+        raise NameError(
+            f"{class_name} is not a known AnalyticBeam. Available options are: "
+            f"{list(AnalyticBeam.__types__.keys())}. If it is a custom beam, "
+            "either ensure the module is imported, or specify the beam with "
+            "dot-pathed modules included (i.e. `my_module.MyAnalyticBeam`)"
+        )
```
steven-murray, line 487:
```suggestion
    if class_name not in AnalyticBeam.__types__ and len(class_parts) > 1:
        module = (".").join(class_parts[:-1])
        module = importlib.import_module(module)
        
    if class_name not in AnalyticBeam.__types__:
        raise NameError(
            f"{class_name} is not a known AnalyticBeam. Available options are: "
            f"{list(AnalyticBeam.__types__.keys())}. If it is a custom beam, "
            "either ensure the module is imported, or specify the beam with "
            "dot-pathed modules included (i.e. `my_module.MyAnalyticBeam`)"
        )
    
    beam_class = AnalyticBeam.__types__[class_name]
```

src/pyuvdata/beam_interface.py
```diff
@@ -0,0 +1,195 @@
...
+        beam_type is "power".
+        Ignored otherwise (the cross pol inclusion is set by the beam object.)
+
+    """
+
+    beam: AnalyticBeam | UVBeam
+    beam_type: Literal["efield", "power"] | None = None
+    include_cross_pols: InitVar[bool] = True
+
+    def __post_init__(self, include_cross_pols):
```
steven-murray, line 51:
```suggestion
    def __post_init__(self, include_cross_pols: bool):
```

src/pyuvdata/beam_interface.py
```diff
@@ -0,0 +1,195 @@
...
+        Post-initialization validation and conversions.
+
+        Parameters
+        ----------
+        include_cross_pols : bool
+            Option to include the cross polarized beams (e.g. xy and yx or en and ne)
+            for the power beam.
+
+        """
+        if not isinstance(self.beam, UVBeam) and not issubclass(
```
steven-murray, line 62: why are you using `issubclass(type(self.beam))` instead of simply `isinstance(self.beam, AnalyticBeam)`?

src/pyuvdata/beam_interface.py
```diff
@@ -0,0 +1,195 @@
...
+            type(self.beam), AnalyticBeam
+        ):
+            raise ValueError(
+                "beam must be a UVBeam or an AnalyticBeam instance, not a "
+                f"{type(self.beam)}."
+            )
+        if isinstance(self.beam, UVBeam):
+            self._isuvbeam = True
```
steven-murray, line 70: I suggest making this a property:

```python
@property
def _isuvbeam(self):
    return isinstance(self.beam, UVBeam)
```

steven-murray (commented):
This is looking really great @bhazelton! 

I have a few small comments/suggestions and then I think we're good.

src/pyuvdata/beam_interface.py
bhazelton: Because that doesn't work. It fails for the subclasses because they aren't an instance on an AnalyticBeam, they're an instance of a subclass of AnalyticBeam. This does work if you prefer it: `isinstance(self.beam.__class__, AnalyticBeam)` as does `issubclass(self.beam.__class__, AnalyticBeam)`. I don't care either way.

steven-murray: But `isinstance` works on subclasses: 

image

bhazelton:
Ugg, h5py 3.12 was just released to pypi (it's not on conda-forge yet). It's breaking the Windows build wheels CI.

telegraphic:
Hi @bhazelton, finally got a chance to look at this. It's great! A few comments/questions:

* I think the tutorials should mention the `BeamInterface` (I started writing a comment about `UVBeam` and `AnalyticBeam` both being 'beamy', so unifying them would be nice -- then I noticed `beam_interface.py`)
* Could (should?) `analytic_beam.py` live in the `uvbeam` directory? 
* I know feed rotation is already on your radars (#854 and #1470), but noting `x_orientation` is still limited to either E or N. Don't let this be a hold-up though.

Building on #1470, for the lazy among us, would it be possible to include a `UserFunctionBeam(AnalyticBeam)`, e.g.

```python

def cos_squared(a, z, f):
    return np.cos(z)**2

def sin_cubed(a, z, f):
    return np.sin(z)**3

beam = UserFunctionBeam(efield_func=cos_squared, power_func=sin_cubed)

```

I'm not 100% sure how to do this with dataclass shenanigans, can we overwrite the `_efield_eval` and `_power_eval` functions via the `__post_init__()`?  E.g.

```python

import types

@dataclass(kw_only=True)
class UserFunctionBeam(AnalyticBeam):
    efield_func: types.FunctionType
    power_func: types.FunctionType

    def __post_init__(self):
        self._efield_eval = efield_func
        self._power_eval = power_func
```

This might overlap with @steven-murray's YAML plugin idea above though?

steven-murray:
Thanks for having a look @telegraphic!

As to your suggestion of a `UserFunctionBeam`, I think the simplest way to do this would simply be to subclass `AnalyticBeam` yourself -- that is what is intended. So:

```python
from pyuvdata.analytic_beam import AnalyticBeam

class MyBeam(AnalyticBeam):
    def _efield_eval(self, *, az_array, za_array, freq_array):
        return np.cos(za_array)**2

    def _power_eval(self, *, az_array, za_array, freq_array):
        return np.sin(za_array)**2
```

IMHO this is not much more writing than you need to create a new "UserFunctionBeam" as you have suggested. We could, if you really think it useful, provide a factory function, e.g. `make_custom_beam` which would construct a new class and return it, given two functions, so e.g.:

```python
def make_custom_beam(name: str, efield: types.FunctionType, power types.FunctionType):
    @dataclass
    class CustomBeam(AnalyticBeam):
        __name__ = name
        def _efield_eval(self, *, az, za, freq):
            return efield(az, za, freq)
        def _power_eval(self, *az, za, freq):
            return power(az, za, freq)

    return CustomBeam
```

which could be called by the user like so:

```python
my_new_beam_class = make_custom_beam(name='cosbeam', efield=lambda az, za, freq: np.cos(za)**2, power=lambda az,za,freq: np.sin(za)**2)
```

bhazelton:
@telegraphic Thanks for your comments! As @steven-murray outlined above, the idea was to make it easy to make new AnalyticBeams by subclassing that object, and I tried to write the tutorial to outline how to do that. I'm willing to entertain Steven's idea of a factory function, but I'd need to explore it more to understand it.

I will add some tutorial stuff on `BeamInterface` to better demonstrate how to use it.

I initially had both `analytic_beam.py` and `beam_interface.py` in the `UVBeam` folder, but I decided to move them out because I felt like that was confusing since they aren't based on `UVBeam`.

docs/analytic_beam_tutorial.rst
```diff
@@ -0,0 +1,860 @@
...
+    circularly polarized feeds.
+    This is a a dataclass field, so if it is specified, the class must have
+    ``@dataclass`` decorator and it should be specified with type annotations
+    and optionally a default (see examples below).
+
+  - ``basis_vector_type``: This defines the coordinate system for the
+    polarization basis vectors, the default is ``"az_za"``. Currently only
+    ``"az_za"`` is supported, which specifies that there are 2 vector directions
+    (i.e. ``Naxes_vec`` is 2).
+    This should be defined as a class variable (see examples below).
```
steven-murray, line 235: I don't think we need to mention these -- these are supposed to be the same for all subclasses, so shouldn't be over-written by someone writing a new beam subclass.

docs/analytic_beam_tutorial.rst
```diff
@@ -0,0 +1,860 @@
...
+                    self.sigma = np.sqrt(2) * self.sigma
+
+                if self.spectral_index != 0.0 and self.reference_frequency is None:
+                    raise ValueError(
+                        "reference_frequency must be set if `spectral_index` is not zero."
+                    )
+                if self.reference_frequency is None:
+                    self.reference_frequency = 1.0
+
+            super().__post_init__(include_cross_pols=include_cross_pols)
```
steven-murray, line 818: Rather than making the user/developer have to worry about calling the `super().__post_init__`, perhaps we should define a `validate()` method on the base class, which is called in the base-class `__post_init__`. That hides the magic.

steven-murray (commented):
Two small comments, but otherwise approved.

docs/analytic_beam_tutorial.rst
bhazelton: No, there's a warning if it's not defined on a beam that's based on the `AnalyticBeam` class. We have it there so we can support other basis vectors in the future as needed.

docs/analytic_beam_tutorial.rst
bhazelton: I don't follow. I think if the user defines a `__post_init__` method the base-class `__post_init__` will not be run.

docs/analytic_beam_tutorial.rst
steven-murray: Ah yes, that's correct.

docs/analytic_beam_tutorial.rst
steven-murray: I'm suggesting that in the base-class `__post_init__` we do:

```python
def __post_init__(self):
    ...what we have now...
    self.validate()

def validate(self):
    pass
```

Then subclasses can implement `validate` without having to call `super().__post_init__`

bhazelton: Just pushed that change.

docs/analytic_beam_tutorial.rst
```diff
@@ -597,17 +597,8 @@ method to ensure that all the normal AnalyticBeam setup has been done.
...
-
-            Parameters
-            ----------
-            include_cross_pols : bool
-                Option to include the cross polarized beams (e.g. xy and yx or en and ne)
-                for the power beam.
-
-            """
+        def validation(self):
+            """Post-initialization validation and conversions."""
```
steven-murray, line 601: I think this should be `validate(self)`

bhazelton: Yes, thank you!!

steven-murray (approved):
Wonderful, thanks @bhazelton --- this was a lot of work and will be so useful!

PR #1371, opened 2024-01-02
SMA/MirParser Improvements

kartographer:
General improvements to the `MirParser` class, and enhanced support for SMA data.

## Description

Several improvements have been made to the `MirParser` class, both to fix bugs and add new functionality in support of handling SMA data, including:

- Support for pre-2020 SMA data (so-called v0/v1/v2 MIR data sets)
- Support for older (ASIC-era) formatted visibilities (v0-era data)
- Support for importing/applying (on-the-fly) SMA COMPASS pipeline calibration products
    - Only bandpass/spectral solutions supported, "wideband" gains to be supported via UVCal objects in a future release.
- Fixes/improvements to tsys application
- Addition of handling for spectral weights inside the `MirParser` class
- Support for saving/restoring selection masks
- Added robustness against indexing errors (due to problems in files)

TODO: tutorial  improvements.

## Motivation and Context

This update addresses a number of issues that have come up in beta-testing of SMA data support via pyuvdata, and integrates a number of improvements that will be leveraged moving into the future. Most of these improvements are focused within the `MirParser` and `MirMetaData` classes, increasing the compatibility with a broader cross-section of data format versions, with the aim of making calls into these classes from the generally more user-facing `UVData` class more robust and seamless. 

~Note that this PR is being submitted as a "late-stage draft" -- ready for review, but some extra (internal) review is expected before this is ready for merge. I'm leaving the CHANGELOG unchanged until that process is done, so that being updated will serve as a sort of "green light" for merging pending review/approval.~ EDIT: major work for this PR is done, basically ready to go once it passes review.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] I have read the [contribution guide](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/.github/CONTRIBUTING.md).
- [x] My code follows the code style of this project.

Bug fix checklist:
- [x] My fix includes a new test that breaks as a result of the bug (if possible).
- [x] All new and existing tests pass.
- [x] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/CHANGELOG.md).

New feature checklist:
- [x] I have added or updated the docstrings associated with my feature using the [numpy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [x] I have updated the tutorial to highlight my new feature (if appropriate).
[...]

PR #1399, opened 2024-02-11
Minor fixes for handling of older SMA data

kartographer:
Fixes a few issues that became apparent after #1371 was merged.

## Description

There are a couple of changes that have been made here:
- Some fixtures in `test_mir_parser` that were redundant have been removed.
- Reading in of MIR data has been reordered so that data-like attributes are read in last (which is helpful in testing/debugging)
- A bug has been fixed in `MirParser._make_v3_compliant` where MJDs were not calculated correctly (forgot to convert UTC back to TT, which is the MIR standard).
- Changed the way that `Mir` determines whether or not a data set should be flex-pol. 
- When reading in MIR datasets, the corresponding value for `UVData.lst_array` is now filled by `set_lsts_from_time_array` under most circumstances. 

## Motivation and Context

The change in flex-pol handling primarily affects reading in of older, pre-V3 MIR-formatted SMA data. Specifically, "gunnLO" was not consistently filled/used in pre-V3 data, whereas "fsky" has historically always been present, hence switching from the former to the latter.

The change in LST-handling arises from nuisance warnings about errors in the LSTs, which after #1356 have reduced somewhat in frequency for MIR data sets, but still regularly arise. I believe this is due to how these values are presently calculated, as a polled average rather than directly calculated at the integration mid-point, producing errors of up to 25 ms. Given that this is a long-standing feature, the reader now defaults to using pyuvdata-calculated LST values so long as they agree w/ the recorded values to within this precision limit. This has the advantage of providing "truer-to-observed" values while reducing the erroneous warnings. When the recorded values are _not_ within the precision limit, they are plugged into `lst_array` so that a warning is appropriately raised (and the issue can be handled by the user accordingly).

## Types of changes

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvdata/uvdata/mir.py
+++ b/pyuvdata/uvdata/mir.py
@@ -343,16 +343,26 @@ class Mir(UVData):
         if len(pol_dict) == 2:
             # If dual-pol, then we need to check if the tunings are split, because
             # the two polarizations will effectively be concat'd across the freq
-            # axis instead of the pol axis.
+            # axis instead of the pol axis. First, see if we have two diff receivers
             rxa_mask = ant1_rxa_mask & ant2_rxa_mask
             rxb_mask = ~(ant1_rxa_mask | ant2_rxa_mask)
 
             if np.any(rxa_mask) and np.any(rxb_mask):
-                # If we have both VV and HH data, check to see that the tunings of the
-                # two receivers match.
-                loa_freq = np.median(mir_data.sp_data["gunnLO"][rxa_mask])
-                lob_freq = np.median(mir_data.sp_data["gunnLO"][rxb_mask])
-                pol_split_tuning = not np.isclose(loa_freq, lob_freq)
+                # If we have both VV and HH data, check to see that the frequencies of
+                # each of the spectral chunks match. If they do, then we can concat
+                # across the polarization axis, but if _not_, we should treat this as
+                # a pol-split data set.
+                fsky_vals = mir_data.sp_data["fsky"]
+                chunk_vals = mir_data.sp_data["corrchunk"]
+                loa_chunks = set(zip(fsky_vals[rxa_mask], chunk_vals[rxa_mask]))
+                lob_chunks = set(zip(fsky_vals[rxb_mask], chunk_vals[rxb_mask]))
+                pol_split_tuning = not (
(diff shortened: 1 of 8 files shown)


pyuvdata/uvdata/mir.py
```diff
@@ -419,9 +425,13 @@ def _init_from_mir_parser(
...
             # really should never happen, only one value should exist per window).
-            assert np.allclose(spw_fsky, mir_data.sp_data["fsky"][data_mask])
-            assert np.allclose(spw_fres, mir_data.sp_data["fres"][data_mask])
-            assert np.allclose(spw_nchan, mir_data.sp_data["nch"][data_mask])
+            for val, item in zip(
+                [spw_fsky, spw_fres, spw_nchan], ["fsky", "fres", "nch"]
+            ):
+                if not np.allclose(val, mir_data.sp_data[item][data_mask]):
+                    warnings.warn(
+                        "Discrepancy in %s for win %i sb %i pol %i." % (item, *spdx)
```
e-koch, line 433: Suggest adding some guidance in the warning message on whether this is a small discrepancy vs. something critical with the data.

Or would that be caught in the checks elsewhere?

pyuvdata/uvdata/mir.py
```diff
@@ -702,9 +660,22 @@ def _init_from_mir_parser(
...
+        # in some imprecision (and some nuisance warnings).  Fix this by calculating the
+        # LSTs here, checking to make sure that they agree to within the expected
+        # precision (sampling rate is 20 Hz, and the max error to worry about is half a
+        # sample: 25 ms, or in radians, 2*pi/(40 * 86400)) = pi / 1728000.
+        # TODO: Re-evaluate this if/when MIR data writer stops calculating LST this way
+        self.set_lsts_from_time_array()
+        if not np.allclose(lst_array, self.lst_array, rtol=0, atol=np.pi / 1728000.0):
+            # If this check fails, it means that there's something off w/ the lst values
+            # (to a larger degree than expected), and we'll pass them back to the user,
+            # who can inspect them directly and decide what to do.
```
e-koch, line 677: This should get printed somewhere with suggestions on what to check.

e-koch (changes requested):
@kartographer -- It wasn't clear to me from the warning messages how the lst offset warnings and what to check are getting communicated back to the user.

e-koch (approved):

PR #1403, opened 2024-02-13
Add an option to use the HERA gridding algorithm to find redundancies

bhazelton:
## Description

This adds a new option for redundancy finding to use the HERA gridding algorithm rather than the existing clustering algorithm. This has some advantages, it's faster and will not throw errors, but it might not handle non-redundant arrays well and the tolerance needs to be well specified to be less than half the minimal antenna spacing.

I also fixed what I am pretty convinced is a bug for data sets that are not rectangular in the basline-times where baselines could be compared at different times. I now use the antenna positions to get ENU baseline vectors if the object is not rectangular in baseline-times. This bug surfaced when I was testing the new HERA algorithm.

~Questions for this PR:
We could decide to just replace the existing algorithm with the HERA algorithm rather than making it an option. Or we could default it to the HERA algorithm. I'm open to any of these options. I haven't yet updated the changelog because I wasn't sure which direction we wanted to go.~
We decided to make the new method the default but with a warning message that the default had changed.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- Added a new redundancy finding algorithm based on the HERA gridding algorithm.
+This new algorithm is used by default but the older clustering algorithm is still
+available.
 - Added support for initializing `MirParser` and `MirMetaData` objects from `Path`
 objects.
 - Added a brief tutorial on how to convert SMA MIR data into MS format (with some notes
--- a/pyuvdata/tests/test_utils.py
+++ b/pyuvdata/tests/test_utils.py
@@ -2437,8 +2437,8 @@ def test_conj_pol():
     )
 
 
-@pytest.mark.parametrize("hera_alg", [True, False, None])
-def test_redundancy_finder(hera_alg):
+@pytest.mark.parametrize("grid_alg", [True, False, None])
+def test_redundancy_finder(grid_alg):
     """
     Check that get_baseline_redundancies and get_antenna_redundancies return consistent
     redundant groups for a test file with the HERA19 layout.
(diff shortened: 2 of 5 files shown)


bhazelton:
I think it probably won't error on core antennas, but it could inappropriately group them in with the hex tiles. The tolerance parameter will be key to that, I haven't messed with this much for non-redundant arrays.

baron-de-montblanc:
> @baron-de-montblanc can you take a look at this code and see if it works any better (or worse) on MWA Phase 2? It may simplify redundancy finding (i.e. no need to throw out the core antennas first).

Yes, this works well for my Phase 2 data. I set tol=3 (the minimum antenna spacing in the file was 8m) and it didn't throw any errors. In total it got rid of about 1/3 of the baselines. Interestingly enough, after the initial compression, I ran it a few more times, gradually increasing the tolerance. I increased until tol=1000 and no additional baselines were thrown out (and no errors thrown either).

bhazelton:
> Yes, this works well for my Phase 2 data. I set tol=3 (the minimum antenna spacing in the file was 8m) and it didn't throw any errors. In total it got rid of about 1/3 of the baselines. Interestingly enough, after the initial compression, I ran it a few more times, gradually increasing the tolerance. I increased until tol=1000 and no additional baselines were thrown out (and no errors thrown either).

great! Thanks for the report.

bhazelton:
@jpober I think the biggest question I have is should we just replace the old clustering method with this HERA-based gridding method or should we keep them both around? And if the latter, what should the default method be. I'm sort of leaning towards just replacing it but I haven't thought about this super deeply.

jpober:
I don't have a strong opinion.  I'm slightly reluctant to remove functionality until we have dug a little deeper and confirmed the results are (dare I say) redundant.  That said, the redundancy finder was already a little confusing since we had calculations using uvws and antenna positions as separate options.  Adding a third sounds like too much.  So I'm ok with just replacing it, but maybe add a warning that there's been an update and that it's not guaranteed to reproduce older results?

bhazelton:
On the telecon we decided to make the new method be the default but with a warning that the default has changed.

bhazelton:
@jpober I made the change we discussed in the telecon yesterday -- I defaulted it to use the HERA algorithm but with a warning.

jpober (commented):
I'm all for giving credit to the HERA team, but do we want the warnings to refer to this as the "HERA gridding based algorithm"?  I feel like for the average user this might sound like it's the algorithm for use with the HERA array, not a general algorithm from the HERA team.  Maybe we could say "Defaulting to True to use the new gridding based algorithm (developed by the HERA team) rather than the older clustering based algorithm" or something?

bhazelton:
> I'm all for giving credit to the HERA team, but do we want the warnings to refer to this as the "HERA gridding based algorithm"? I feel like for the average user this might sound like it's the algorithm for use with the HERA array, not a general algorithm from the HERA team. Maybe we could say "Defaulting to True to use the new gridding based algorithm (developed by the HERA team) rather than the older clustering based algorithm" or something?

Yep, sounds good. I'll make that change.

jpober (approved):
Looks good.  Can you confirm that the error in the build against HERA Cal is expected?  If so, I'll go ahead and merge.

PR #1416, opened 2024-03-20
Updates to support MIR v6 file format

kartographer:
Some minor updates to support the newest file format for MIR.

## Description

A total of four fields in the spectral window header object (`MirSpData`) have been updated to their new V6 names/types. Additionally, SMA targets now have their coordinates correcty labeled as being in the "ICRS" frame rather than FK5.

## Motivation and Context

The fields within the `MirSpData` class have been updated to reflect their new allocations as of the most recent update to the file format. Fields being updated were previously marked as "spare", and thus the updates should have no impact on backward compatibility of datasets (rather they simply allow various pieces of software using `MirParser` to correctly access these fields from new data sets). Note that these changes are _very_ low level from the perspective of `UVData` -- I don't expect the average user to ever encounter these changes.

The FK5/ICRS mixup was noticed only recently as part of testing related to the updated phasing memo, where thanks to the improved astrometry of both pyuvdata and the SMA online system, we were able to detect an apparent ~15 mas discrepancy between the two was traced back to this mislabeling.

## Types of changes

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvdata/uvdata/tests/test_mir_parser.py
+++ b/pyuvdata/uvdata/tests/test_mir_parser.py
@@ -2274,11 +2274,11 @@ def test_mir_remember_me_sp_data(mir_data):
 
     assert np.all(mir_data.sp_data["iddsmode"] == 0)
 
-    assert np.all(mir_data.sp_data["spareshort"] == 0)
+    assert np.all(mir_data.sp_data["gunnMult"] == 0)
 
-    assert np.all(mir_data.sp_data["spareint3"] == 0)
+    assert np.all(mir_data.sp_data["amp"] == 0)
 
-    assert np.all(mir_data.sp_data["spareint4"] == 0)
+    assert np.all(mir_data.sp_data["phase"] == 0)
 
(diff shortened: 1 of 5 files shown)


pyuvdata/uvdata/tests/test_mir_parser.py
```diff
@@ -2274,11 +2274,11 @@ def test_mir_remember_me_sp_data(mir_data):
 
     assert np.all(mir_data.sp_data["iddsmode"] == 0)
 
-    assert np.all(mir_data.sp_data["spareshort"] == 0)
+    assert np.all(mir_data.sp_data["gunnMult"] == 0)
```
e-koch, line 2277: May be noted above (and if so ignore), but a quick note in the test docstring that this is testing for the existence of the attributes rather than the data types defined in `mir_meta_data.py` would be nice to include.

kartographer: Added on 7cb8ec3d726c5b55599007b7cdc3d465d78ade89.

e-koch (approved):

PR #1417, opened 2024-03-23
Handle lunar ellipsoids

bhazelton:
## Description

Support for various lunar ellipsoids was recently added in lunar sky 0.2.2. This handles those changes and is required for pyuvsim moon simulations to work.

To keep track of this information, I added a new `lunar_ellipsoid` attribute to LocationParameter objects (similar to how we handled the `frame` attribute) and I added handling to pipe it into and out of all the necessary utility functions.

Note that lunar sky now requires astropy >= 6.0 so I updated that requirement, which lead to a bunch of other knock-on changes to our minimum dependencies. Most notably, astropy 6.0 doesn't support python 3.9, so I dropped it from our test suite.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,8 +135,6 @@ There are some optional dependencies that are required for specific functionalit
 which will not be installed automatically by conda or pip.
 See [Dependencies](#dependencies) for details on installing optional dependencies.
 
-Note that as of v2.2, `pyuvdata` is only supported on python 3.10+.
-
 ## Dependencies
 
 Required:
@@ -146,6 +144,7 @@ Required:
 * h5py >= 3.4
 * numpy >= 1.23
 * pyerfa >= 2.0.1.1
+* python >= 3.10
 * scipy >= 1.7.3
 * setuptools_scm <7.0|>=7.0.3
 
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ There are some optional dependencies that are required for specific functionalit
 which will not be installed automatically by conda or pip.
 See [Dependencies](#dependencies) for details on installing optional dependencies.
 
-Note that as of v2.2, `pyuvdata` is only supported on python 3.7+.
+Note that as of v2.2, `pyuvdata` is only supported on python 3.10+.
 
 ## Dependencies
 
--- a/README.md
+++ b/README.md
@@ -141,23 +141,23 @@ Note that as of v2.2, `pyuvdata` is only supported on python 3.7+.
 
 Required:
 
-* astropy >= 5.0.4
+* astropy >= 6.0
 * docstring_parser>=0.15
-* h5py >= 3.1
-* numpy >= 1.20
-* pyerfa >= 2.0
-* scipy >= 1.5
+* h5py >= 3.4
+* numpy >= 1.23
+* pyerfa >= 2.0.1.1
+* scipy >= 1.7.3
 * setuptools_scm <7.0|>=7.0.3
 
 Optional:
 
 * astropy-healpix >= 0.6 (for working with beams in HEALPix formats)
 * astroquery >= 0.4.4 (for enabling phasing to ephemeris objects using JPL-Horizons)
-* hdf5plugin >= 3.1.0 (for enabling bitshuffle and other hdf5 compression filters in uvh5 files)
-* lunarsky >=0.2.1 (for working with simulated datasets for lunar telescopes)
+* hdf5plugin >= 3.2.0 (for enabling bitshuffle and other hdf5 compression filters in uvh5 files)
+* lunarsky >=0.2.2 (for working with simulated datasets for lunar telescopes)
 * novas and novas_de405 (for using the NOVAS library for astrometry)
-* python-casacore >= 3.3.1 (for working with CASA measurement sets)
-* pyyaml >= 5.3 (for working with settings files for CST beam files)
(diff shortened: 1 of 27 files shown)


pyuvdata/parameter.py
```diff
@@ -846,6 +846,10 @@ class LocationParameter(UVParameter):
         Description of the data or metadata in the object.
     frame : str, optional
         Coordinate frame. Valid options are "itrs" (default) or "mcmf".
+    lunar_ellipsoid : str, optional
```
kartographer, line 849: Is it worth generalizing this to just `ellipsoid`? ITRF of course has different realizations, and  it may be helpful for this keyword to remain flexible with other frames as well (e.g., when we eventually support interferometers on Pluto in pyuvdata v9.0). Supporting this for ITRF frames may require some retooling of defaults (since "SPHERE" obviously is not the desired ellipsoid for Earth), but I think treating this as a programmable parameter for all but ITRF would help provide some future-proofing.

pyuvdata/uvdata/ms.py
```diff
@@ -293,6 +293,7 @@ def _write_ms_antenna(self, filepath):
             ant_ref_frame = "ITRF"
         else:
             ant_ref_frame = self._telescope_location.frame.upper()
+            # TODO: ask Karto what the best way is to put in the lunar ellipsoid
```
kartographer, line 296: ec3a52cd8352fe8569d00c3c22993303

pyuvdata/parameter.py
```diff
@@ -974,6 +991,7 @@ def set_lat_lon_alt_degrees(self, lat_lon_alt_degree):
                 longitude * np.pi / 180.0,
                 altitude,
                 frame=self.frame,
+                lunar_ellipsoid=self.lunar_ellipsoid,
             )
 
     def check_acceptability(self):
```
kartographer, line 997: Does `check_acceptability` here need to be updated for what values of `frame` are okay?

pyuvdata/utils.py
```diff
@@ -1490,7 +1509,9 @@ def ECEF_from_rotECEF(xyz, longitude):
     return rot_matrix.dot(xyz.T).T
 
 
-def ENU_from_ECEF(xyz, latitude, longitude, altitude, frame="ITRS"):
+def ENU_from_ECEF(
+    xyz, latitude, longitude, altitude, frame="ITRS", lunar_ellipsoid="SPHERE"
```
kartographer, line 1513: Same comment here re: ellipsoid (note that there are several in utils here that need to be updated if you wanted to make that shift).

kartographer (commented):
Overall I think this is looking good -- mostly minor comments and one potentially more significant one (assuming you agree w/ my assessment).

kartographer:
@bhazelton -- note you may need to clean up some of the CI actions here related to Python 3.9 to get past the testing checks...

pyuvdata/parameter.py
bhazelton: Sure, I wondered about that actually. What's the right default for ITRS?

pyuvdata/parameter.py
bhazelton: hmm, maybe. I think there's checking on the setter, but I'll make sure.

bhazelton: Oh, right. The setter was removed when we moved the checking for positions being on the surface to the utils. I'm a bit inclined to add a setter to check that it's allowed, because it's used in so many calculations downstream.

pyuvdata/parameter.py
bhazelton: Maybe `None` is the correct thing for ITRS?

kartographer: So I think either `None` or `"WGS84"` (the astropy default) might be sensible -- there are some subtleties here that I have yet to dig into, but I think either would be a reasonable choice while we investigate further.

README.md
```diff
@@ -135,29 +135,29 @@ There are some optional dependencies that are required for specific functionalit
 which will not be installed automatically by conda or pip.
 See [Dependencies](#dependencies) for details on installing optional dependencies.
 
-Note that as of v2.2, `pyuvdata` is only supported on python 3.7+.
```
mkolopanis, line 138: this is a big change. I think I would be in favor of making a version for this.

mkolopanis (commented):
Looking good to me so far as well. My only comment is dropping python 3.8 and 3.9 is big and we might want a version to make a clean cut.

README.md
bhazelton: note that this is really only dropping python 3.9. We already dropped 3.8 back in v2.4.2, we just forgot to update this line in the README. We did note it in the changelog.

I agree about the new version. We need one once this is merged for pyuvsim's sake anyway.

bhazelton: oh, I also just realized that I didn't update the version number here. I'm going to restructure this a little to make it easier to keep it up to date.

bhazelton: I moved it into the list of required dependencies. I think it'll be more likely to be kept up to date if it's with the other dependencies so we see it when we're changing the other ones.

mkolopanis: sounds good thank you!

kartographer (approved):

PR #1421, opened 2024-03-27
Cython3

mkolopanis:
bumps the cython version to 3.0. It's been stable since October 2023.

Not sure if it is possible to change some string thins in corr fits so i'd hold off on full PR for time being.

### DIFF
--- a/ci/pyuvdata_tests.yml
+++ b/ci/pyuvdata_tests.yml
@@ -19,7 +19,7 @@ dependencies:
   - pytest-cases>=3.8.3
   - pytest-cov
   - pytest-xdist
-  - cython
+  - cython>=3
   - setuptools_scm<7.0|>=7.0.3
   - pip
   - pip:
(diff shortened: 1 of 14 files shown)


ci/pyuvdata_tests.yml
```diff
@@ -19,7 +19,7 @@ dependencies:
   - pytest-cases>=3.8.3
   - pytest-cov
   - pytest-xdist
-  - cython
+  - cython>=3.0
```
bhazelton, line 22: This should also be updated in the readme and maybe in setup.py?

mkolopanis: Good call thanks for pointing those out.

bhazelton (commented):
Looks good, I think there's just one more place to update, the `pyuvdata_min_versions_tests.yml`. Sorry I didn't notice that one before.

PR #1422, opened 2024-03-28
Refactor the telescope-related metadata in UVData, UVCal and UVFlag into an attached Telescope object

### LINKED ISSUE #1095: Need more antenna-based tracking. Either a new Antenna object or updates to the Telescope object
Potential things:
- antenna positions (already on Telescope)
- antenna system temperatures
- primary beam pointing errors
- weather (might be just per site or maybe per antenna)
- associated UVBeam object (or filename)
- bulk delays (e.g. cable delays, correlator delays)
- UVCal object optionally associated

Note: the term telescope is potentially confusing. Karto thinks of telescope as meaning one antenna, I think of it as meaning the array. Maybe we should move to "Array" terminology.

bhazelton:
## Description

All the metadata which is really about the telescope (rather than the dataset) has been pulled out into a Telescope object which is then attached to each object as an attribute. I used `__getattr__` and `__setattr__` on UVBase to keep the old API (but with deprecation warnings). I think other the biggest difference a user might notice is that some new metadata can show up on UVCal and UVFlag (e.g. instrument and antenna_diameters).

This really streamlines a bunch of code and will make it easier to maintain.

## Motivation and Context

closes #1095 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:


pyuvdata/telescopes.py
```diff
@@ -94,212 +149,426 @@ def __init__(self):
...
+            warnings.warn(warn_str)
 
-    return obj
+        if run_check:
+            self.check(
+                check_extra=check_extra, run_check_acceptability=run_check_acceptability
+            )
+
+    @classmethod
+    def get_telescope_from_known_telescopes(
```
steven-murray, line 551: I think this should just be `from_known_telescopes`.... this is the usual syntax for object-instantiation class methods.

Also, presumably you're passing in the `known_telescope_dict` so that someone could create their own telescope dict to pass in here? But I would argue that a better interface would be to either:

1. Just let the user update `KNOWN_TELESCOPES` (either by some formal registration of new objects, or simply by doing `KNOWN_TELESCOPES.update(myscope={myscope_dict})`
2. Having a separate classmethod `.from_telescope_params()` (or similar) to which they can just pass the keys that would have been in their entry into `KNOWN_TELESCOPES`... this would be a thin layer over the `from_known_telescopes()` classmethod.

pyuvdata/uvcal/initializers.py
```diff
@@ -25,20 +24,21 @@
...
     freq_array: np.ndarray | None = None,
     freq_range: np.ndarray | None = None,
     cal_type: Literal["delay", "gain"] | None = None,
     integration_time: float | np.ndarray | None = None,
     channel_width: float | np.ndarray | None = None,
+    telescope: Telescope | None = None,
+    telescope_location: Locations | None = None,
+    telescope_name: str | None = None,
+    x_orientation: Literal["east", "north", "e", "n", "ew", "ns"] | None = None,
+    antenna_positions: np.ndarray | dict[str | int, np.ndarray] | None = None,
```
steven-murray, line 41: Here I would argue that we should deprecate the passing of the `telescope_` params... let the user just pass a `Telescope` object (which should be able to be created as a one-liner with `Telescope.from_known_telescopes()` or some other simple method whether you pass the parameters. This reduces redundant logic. You probably _could_ accept the `telescope` parameter as a string, which would be expected to match a known telescope name.

steven-murray: Obviously we can't do that without jumping a major version, but you could raise warnings from now.

steven-murray: My point is it's not harder for the user to call this:

```python
uvc = UVCal.new(
    telescope = Telescope.from_params(
        location=...,
        name='blah',
        ...
    ),
    ...
)
```

than this:

```python
uvc = UVCal.new(
    telescope_location=...,
    telescope_name=....,
    ...
)
```

But, the former reduces code logic significantly, and also subtly informs the user that what's really going on under the hood is that there's a Telescope object, which is conceptually simpler for them too.

pyuvdata/uvcal/uvcal.py
```diff
@@ -615,8 +551,136 @@ def __init__(self):
...
+        return self.telescope.instrument
+
+    @instrument.setter
+    def instrument(self, val):
+        self.telescope.instrument = val
+
+    @property
+    def telescope_location(self):
+        """The telescope location (stored on the Telescope object internally)."""
+        return self.telescope.location
```
steven-murray, line 583: Hmmm, since there are a few of these, maybe it's easier to move them to the `__getattr__` and `__setattr__` methods, since that'll be a couple of lines of code. I don't love doing that, but I think in this case it might be worth it. In the spirit of "there should be one right way of doing it" I feel like we should also discourage using these old-style names (`telescope_name` etc) in favor of `telescope.name`. We could consider adding a deprecation warning to the `__getattr__`  method if we like (or, in the future?). My other gripe with `__getattr__` is that it doesn't play well with code editors auto-completion (including jupyter). But again, in this case, I don't think we want to encourage using these attribute names anyhow.

pyuvdata/uvcal/uvcal.py
```diff
@@ -1090,76 +1162,13 @@ def set_telescope_params(self, *, overwrite=False):
...
-            raise ValueError(
-                f"Telescope {self.telescope_name} is not in known_telescopes."
-            )
+        self.telescope.update_params_from_known_telescopes(
+            overwrite=overwrite,
+            warn=warn,
+            run_check=run_check,
+            check_extra=check_extra,
+            run_check_acceptability=run_check_acceptability,
+        )
```
steven-murray, line 1171: IMHO, as stated before, this `update` method should be removed and here we should be doing `self.telescope = Telescope.from_known_telescopes()`. But hold on -- you're not even passing any actual params to update here??

steven-murray (commented):
Thanks @bhazelton, this is going in some really good directions. I've made a number of comments, but I have two more overall comments:

1. Is there a good reason that `Telescope.location` is not an astropy `Location` (either `EarthLocation` or `MoonLocation`)? There's so much mucking around throughout the code with `location_lat_lon_alt` and `location_lat_lon_alt_degrees` etc, where all of this is already taken care of by this `Location` object. Now would be the perfect time to move to such a representation (the attributes on the top-level classes can remain the same for now, since they can be computed automatically from the `Location` object). I _consistently_ find this a sticking point when using `pyuvdata` -- I never quite know where the telescope location information exists, and where to set it, and which properties will need to be manually updated if I change it, etc.
2. I think we should consider having a `.write_hdf5()` method on the Telescope object, which can write out to a Group, irrespective of any upper class to which it might belong (UVData, UVCal etc.). This would make it more consistent to read the Telescope object from any file, and can be called directly from the upper method in their write methods.

pyuvdata/telescopes.py
```diff
@@ -94,212 +149,426 @@ def __init__(self):
...
+        self._location.required = True
+        self._instrument.required = False
+        self._Nants.required = True
+        self._antenna_names.required = True
+        self._antenna_numbers.required = True
+        self._antenna_positions.required = True
+        self._antenna_diameters.required = False
+        self._x_orientation.required = False
+
+    def update_params_from_known_telescopes(
```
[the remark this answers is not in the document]
bhazelton: This is used extensively by the other objects and I consolidated a bunch of code from the other objects here. This is because we have older UVCal and UVFlag files which have some of the telescope metadata but not all of it that is now required. Since we want to be able to read older files without erroring we need this code somewhere and I think it is much cleaner to be here rather than on the top level objects because here it can be consolidated.

pyuvdata/telescopes.py
bhazelton: The known telescopes passing is basically just for tests. I can try to hide it more.

pyuvdata/uvcal/initializers.py
bhazelton: They can just pass the telescope object and if they do, all the other things are just ignored. I left them there to avoid breaking code but I can deprecate the other options if you like.

pyuvdata/uvcal/uvcal.py
bhazelton: Yeah, I wasn't loving how many there are and how they're replicated across the objects. I'm was also thinking about whether I could move them to UVBase, but not all objects that inherit from UVBase have a telescope object.

I guess we hadn't decided that we wanted to move people away from the older names, that's a very big adjustment. We were intending this change to largely be behind the scenes and to affect the users as little as possible.

pyuvdata/uvcal/uvcal.py
bhazelton: It figures out what to update based on what is available for each telescope and what is already set on the object. This replaces old code that did the same.

Some known telescopes are just from Astropy sites, so all they have is a telescope location. Others have everything down to antenna positions. It's widely variable. And when we read in old cal files and UVFlag files there's a range of what metadata is set -- for very old files they are missing almost all the telescope metadata, while on new files this is basically a no-op because everything is already set.

bhazelton:
@steven-murray I did add a new property on the Telescope object called `location_obj` which returns an EarthLocation (or a MoonLocation) so that a user can always just get the object. But I can look at making the `location` just be such an object, that might clean some things up.

pyuvdata/telescopes.py
bhazelton: I just updated the docstring to include all the parameters and to make it clear that should only be used for testing. Let me know if I should do more on this front.

kartographer (commented):
Just a small set of comments, nothing too major beyond what we talked about a couple of weeks ago.

pyuvdata/uvcal/initializers.py
bhazelton: They are now deprecated.

pyuvdata/uvcal/uvcal.py
bhazelton: These are now in `_getattr__` and `__setattr__` methods on UVBase. I haven't deprecated them yet.

pyuvdata/uvflag/uvflag.py
```diff
@@ -683,6 +634,140 @@ def __init__(
                 "list, tuple, string, pathlib.Path, UVData, or UVCal."
             )
 
+    def _set_telescope_requirements(self):
```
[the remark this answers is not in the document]
bhazelton: I think the real issue comes down to `x_orientation` as we discussed on the telecon. We can't really require it for UVData because many existing file formats do not have it or require it and there's no good way to automatically fill it in. But we do require it on UVCal. Unless we decide to back off on that requirement, we're stuck with these differences. I did make the things that are always required on all objects always be required on Telescope and removed those from these methods to make it clearer what the differences are.

pyuvdata/telescopes.py
```diff
@@ -71,6 +73,59 @@
 }
 
 
+def _parse_antpos_file(antenna_positions_file):
```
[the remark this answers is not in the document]
bhazelton: I can move it to utils.

pyuvdata/uvflag/tests/test_uvflag.py
```diff
@@ -1655,39 +1700,41 @@ def test_set_lsts(uvf_from_data, background):
...
     uvd2 = uvd.select(
         antenna_nums=ants_with_data[: uvd.Nants_data // 2],
         keep_all_metadata=False,
         inplace=False,
     )
     uvf = UVFlag(uvd2, use_future_array_shapes=True)
-    assert uvf._antenna_names == uvd2._antenna_names
-    assert uvf._antenna_numbers == uvd2._antenna_numbers
-    assert uvf._antenna_positions == uvd2._antenna_positions
+    assert uvf.telescope._antenna_names == uvd2.telescope._antenna_names
```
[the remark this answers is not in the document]
bhazelton: So it turns out that the telescope objects aren't equal because they have different sets of required parameters (UVData's requires `instrument` while UVFlag's does not). So I can make copies of the telescope objects, adjust those requirements and then test that the telescopes are equal or I can just test these attributes.

pyuvdata/uvbase.py
```diff
@@ -331,11 +331,20 @@ def __iter__(self, uvparams_only=True):
...
-        attribute_list = [
-            a
-            for a in dir(self)
-            if not a.startswith("__") and not callable(getattr(self, a))
-        ]
+        if uvparams_only:
+            attribute_list = [
+                a
+                for a in dir(self)
+                if a.startswith("_")
```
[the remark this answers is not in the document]
bhazelton: I don't remember why I went this route, but I tried just making it an isinstance and it works, so I'm changing it.

pyuvdata/uvflag/uvflag.py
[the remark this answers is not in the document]
steven-murray: One probably not very satisfying way of solving this would be to require `x_orientation` everywhere, and force people to pass it into the `read()` method for file formats that don't include it.

bhazelton:
@steven-murray @kartographer I think this ready for another look when you have time. I think I've addressed all your comments and what we discussed in the telecons.

pyuvdata/uvcal/uvcal.py
bhazelton: they are now deprecated.

pyuvdata/telescopes.py
```diff
@@ -71,21 +90,163 @@
...
+    -------
+    location : EarthLocation
+        Telescope location as an EarthLocation object.
+    citation : str, optional
+        Citation string.
+
+    """
+    astropy_sites = EarthLocation.get_site_names()
+    telescope_keys = list(known_telescope_dict.keys())
+    telescope_list = [tel.lower() for tel in telescope_keys]
```
steven-murray, line 135: you should just make a new dict here...

```python
known_telescopes = {k.lower(): v for k, v in known_telescopes_dict}
```

then later you can simply key into the dictionary rather than doing list-index kungfu. Also, this is a good way to ensure that down the function you don't accidentally edit the `KNOWN_TELESCOPES` global dict...

pyuvdata/telescopes.py
```diff
@@ -71,21 +90,163 @@
...
+                    "Bad location information in known_telescopes_dict "
+                    f"for telescope {name}. Either the center_xyz "
+                    "or the latitude, longitude and altitude of the "
+                    "telescope must be specified."
+                )
+            location = EarthLocation.from_geodetic(
+                lat=telescope_dict["latitude"] * units.rad,
+                lon=telescope_dict["longitude"] * units.rad,
+                height=telescope_dict["altitude"] * units.m,
+            )
```
steven-murray, line 167: why isn't `KNOWN_TELESCOPES` just a dictionary of `{name: Telescope()}` objects?

steven-murray: ah, is it because `Telescope` has more information in it that's required than can be reasonably specified here? In that case, at least I would use a `location` parameter in the dict, which is simply an astropy Location.

steven-murray (commented):
Thanks @bhazelton -- this is very close I think. I did have a couple of stylistic comments, but should be easy to address.

kartographer (commented):
@bhazelton looks good overall -- I don't have any detailed comments beyond what @steven-murray has already marked, the only thing that I think would be worth discussing is what will happen when we have new parameters to add to the `Telescope` object when reading/writing to HDF5-formatted files (this is of particular interest for SMA, since I've got some parameters already in mind that I'll start working on once v3 is done). Right now, everything is getting written to the main header, and I don't know if that's what we want to keep on doing, or if we want to have the new parameters grouped in some sensible way. I did this in an ad-hoc fashion w/ `phase_center_catalog`, which of course worked okay but changed shortly after that was integrated, and so if we don't want to trust my on-the-fly approach (which I never do :wink:) it might be good to discuss how we might do this in the future and make some comments in the code accordingly.

pyuvdata/uvflag/uvflag.py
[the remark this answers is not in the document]
kartographer: Okay, I think I'm content to leave it be for now -- `x_orientation` is probably where this runs aground, but I realized that I had some ideas on how to retire this (as part of addressing #854), and maybe this could be just left as is for now (though `instrument` could probably be made to be always optional -- actually not sure why it's required for `UVData` at the moment).

bhazelton:
Astropy 6.1 dropped on pypi (but interestingly not yet on conda) and a new warning is being generated:

```
FutureWarning: Setting the location attribute post initialization will be disallowed in a future version of Astropy. Instead you should set the location when creating the Time object. In the future, this will raise an AttributeError.
```

I fixed a few places where we were causing this in pyuvdata, but most of them are coming from `lunarsky`, so I made an issue there (https://github.com/aelanman/lunarsky/issues/27).

This is causing our warnings tests to fail. Not sure if I want to limit astropy on that vs just waiting for it to be fixed on `lunarsky`.

bhazelton:
> @bhazelton looks good overall -- I don't have any detailed comments beyond what @steven-murray has already marked, the only thing that I think would be worth discussing is what will happen when we have new parameters to add to the `Telescope` object when reading/writing to HDF5-formatted files (this is of particular interest for SMA, since I've got some parameters already in mind that I'll start working on once v3 is done). Right now, everything is getting written to the main header, and I don't know if that's what we want to keep on doing, or if we want to have the new parameters grouped in some sensible way. I did this in an ad-hoc fashion w/ `phase_center_catalog`, which of course worked okay but changed shortly after that was integrated, and so if we don't want to trust my on-the-fly approach (which I never do 😉) it might be good to discuss how we might do this in the future and make some comments in the code accordingly.

Good question. I'd really like this to be a larger conversation as it pertains to the UVH5 format which is being used beyond pyuvdata at this point. Currently in that format we really only differentiate between header and data items, where anything that's not the size of visibility data are lumped in the header. The phase center catalog is also in header, although it is a nested dataset. On the one hand, I understand the niceness of grouping telescope related data within a dataset, especially as we are now representing it that way internally to pyuvdata, but on the other hand since the format is more widely used I don't know that we should force our object hierarchy on the format too strictly.

pyuvdata/telescopes.py
```diff
@@ -94,212 +242,570 @@ def __init__(self):
...
+            warn=False,
+            run_check=run_check,
+            check_extra=check_extra,
+            run_check_acceptability=run_check_acceptability,
+            known_telescope_dict=known_telescope_dict,
+        )
+        return tel_obj
+
+    @classmethod
+    def from_params(
```
bhazelton, line 610: Should this be called `new` to have a similar name to the analogous methods on UVData and UVCal?

pyuvdata/parameter.py
```diff
@@ -930,43 +939,84 @@ def __init__(
...
+    def xyz(self):
+        """Get the body centric coordinates in meters."""
+        if self.value is None:
+            return None
+        else:
+            if hasmoon and isinstance(self.value, MoonLocation):
+                centric_coords = self.value.selenocentric
+            else:
+                centric_coords = self.value.geocentric
+            return units.Quantity(centric_coords).to("m").value
```
bhazelton, line 981: Note that this is not currently turned into a property like `location_xyz` on the Telescope object, unlike e.g. the `lat_lon_alt` method which is turned into the `location_lat_lon_alt` property on the Telescope object. Initially this was because I was trying to push users to just use the EarthLocation object, but there are several places in the code where I call this method because getting the xyz as a simple numpy array (rather than Quantities) is useful. Thoughts? @steven-murray @kartographer

pyuvdata/telescopes.py
kartographer: I would endorse this -- presumably that means we need to add `initializers.new_telescope` as well?

steven-murray: I also would endorse this also.

pyuvdata/parameter.py
steven-murray: I would refrain from as much bespoke functionality as possible, and go with your inclination towards pushing users to use the `EarthLocation` object. "There should be one-- and preferably only one --obvious way to do it."

steven-murray:
To add my two cents to some other issues brought up here (which we should move away into separate discussions...)...

1. I prefer the idea of a well-formed uvh5 file format where the telescope metadata is in its own sub-group that can be read completely independently. This would mean you could literally have hdf5 files with _only_ telescope metadata in them and they would make sense, and could be read directly as Telescope objects. It's just more flexible. However, the old uvh5 file format should be supported _forever_. That is, there should be legacy UVH5 readers that know how to read older formats, and can read them perfectly fine as uvdata objects. My preferred way of doing this is to have separate reading classes/functions for different versions of the _file format_, and a dispatcher that knows which function to dispatch to given the file contents. I _don't_ think we should have to worry about whether people are using older formats... the pyuvdata code itself should be how the files are read, and anyway each version of the file format should be documented, so people can use any format.
2. The movement of astropy away from setting attributes after object creation is unsurprising to me, and is how I (and I think most people) _expect_ python objects to work. This is in contrast to pyuvdata, which is jarring to me every time I use it for this reason. It might be worth thinking about (for v4...) refactoring towards this modality.

pyuvdata/parameter.py
kartographer: 👍 for having this as a convenience function.

bhazelton: The annoying thing is how hard it is to get the xyz location from an astropy EarthLocation object. I'd be happy if it'd just give it to you as a 3-element quantity array, but it doesn't. You have to ask for x, y, and z separately. Or you can do el_obj.geocentric, which gives you a list of 3 Quantities, which still takes work to get into a useful shape.

steven-murray: So, as convenient as it might be to have `.xyz` as a property, I think it's less explicit than using the `EarthLocation` -- the name doesn't tell you what frame you're in, or what the units are. The docstring does, so that's OK, but I think it adds noise to the interface to make this a top-level attribute that's less clear than other attributes. I'm perfectly fine keeping it as a convenience method of course.

kartographer: Given that I'm mostly thinking about this in the context of convenience for various utilities, we could name it `_xyz`, which I think sould resolve the "clutter" issue while knocking down need for redundant code to craft `EarthLocation` into a 3-element array for various matrix operations.

bhazelton: So it's currently available at `Telescope._location.xyz()` and I use it several places in the code. Does that work for you @kartographer?

kartographer: @bhazelton perfect!

pyuvdata/tests/test_telescopes.py
```diff
@@ -124,44 +143,183 @@ def test_known_telescopes():
...
+
+    assert len(KNOWN_TELESCOPES["MWA"]) == 1
+    for key, val in KNOWN_TELESCOPES.items():
+        assert val == _KNOWN_TELESCOPES[key]
 
 
 def test_get_telescope_center_xyz():
     ref_xyz = (-2562123.42683, 5094215.40141, -2848728.58869)
-    ref_latlonalt = (-26.7 * np.pi / 180.0, 116.7 * np.pi / 180.0, 377.8)
+    # ref_latlonalt = (-26.7 * np.pi / 180.0, 116.7 * np.pi / 180.0, 377.8)
```
kartographer, line 299: Cruft?

kartographer (commented):
Thanks @bhazelton -- one very small piece of clean-up, otherwise I think this is looking good to go.

pyuvdata/tests/test_telescopes.py
bhazelton: yes, thanks!

bhazelton: Actually, I'm not sure that test is needed. I'll do a little checking on that, I might remove the whole test.

bhazelton: I removed it entirely.

kartographer (approved):

PR #1446, opened 2024-06-13
Change to an `src` layout rather than a flat layout, use pyproj.toml as much as possible

bhazelton:
## Description

This does three things:
- change to an `src` layout
- move the `check_warnings` functionality out of `/tests/__init__.py` and into a new `testing` module to make it accessible to the other packages that use it. This was required because of the change to an `src` layout.
- Use pyproj.toml as much as possible.

## Types of changes

- [ ] Version change

## Checklist:


mkolopanis:
should the `into` branch be `prep_v3.0` not `main`?

bhazelton:
> should the `into` branch be `prep_v3.0` not `main`?

yes, thank you! fixed now.

bhazelton:
I've opened PRs on pyradiosky, pyuvsim, hera_qm and hera_cal to fix the errors that this caused in those repos. All of the errors except on hera_cal are due to the changed import for `check_warnings`. On hera_cal there was an unused import of `pyuvdata.tests` which breaks.

src/pyuvdata/telescopes.py
```diff
@@ -17,12 +17,11 @@
 from astropy import units
 from astropy.coordinates import Angle, EarthLocation
 
+from pyuvdata import hdf5_utils
+from pyuvdata import parameter as uvp
+from pyuvdata import utils as uvutils
```
steven-murray, line 22: it's not clear to me why these had to be changed? Relative imports still work in src layout

src/pyuvdata/uvbase.py
```diff
@@ -14,9 +14,9 @@
 from astropy.coordinates import SkyCoord
 from astropy.units import Quantity
 
-from . import __version__
-from . import parameter as uvp
-from .utils import _get_iterable
+from pyuvdata import __version__
+from pyuvdata import parameter as uvp
+from pyuvdata.utils import _get_iterable
```
steven-murray, line 19: same here (and many other places...)

steven-murray (commented):
Only a small comment from me. It works as-is, but I find relative imports cleaner.

src/pyuvdata/telescopes.py
bhazelton: I can put them back. I was changing them in tests and I guess got overzealous. I don't personally have a strong preference for either way, so happy to go with relative.

bhazelton: There were cases in some files that had lots of leading dots which I find challenging to read, and even in this file there was a mix of relative and absolute imports (`from pyuvdata.data import DATA_PATH`). I think I made these changes because the mix of absolute and relative felt weird so I made them all look similar. Do you have thoughts about the mixture?

bhazelton: Done.

mkolopanis (approved):
This looks good to me. I just merged the fixes for pyradiosky so we can rerun that anytime.

PR #1453, opened 2024-06-25
Speed-up of `UVBeam.interp`

kartographer:
Optimizes `UVBeam.interp` when `check_azza_domain=True`

## Description

The logic of `UVBeam._interp_az_za_rect_spline` has been modified when `check_azza_domain=True`. In the old logic, when checking that the interpolated az/za pairs were within the defined domain of the beam, the code would go point-by-point through the interpolated coordinates and check all possible az/za combos. But because the az and za are on a mesh grid, it can often be faster to check all interpolated points (in a vectorized fashion) against individual values of za and az, taking the minimum across each dimension to get the distance from a given gridpoint. The logic has been changed to do this when the number of interpolated points is large.

## Motivation and Context

Mostly it speeds up what had been a very slow test (60s -> 1.5s), with a ~25% improvement in overall testing speed . But given that `check_azza_domain=True` by default, users and downstream packages will also hopefully see some improvement in processing speed. 

## Types of changes
N.b., this is more optimziation than bugfix, but it's the category that seems to fit the best.

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ time for each time range or the time_array (if there's a time_array and no time_
 - Added new keyword handling for v.6 of the MIR data format within `MirParser`.
 
 ### Changed
+- Modified `UVBeam.interp` to speed up processing when `check_azza_domain=True`.
 - Updated minimum dependencies: setuptools>=64, setuptools_scm>=8.0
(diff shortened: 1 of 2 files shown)


bhazelton (approved):
Awesome, thanks!

PR #1455, opened 2024-07-05
Pol convention

steven-murray:
## Description

This adds a new parameter `pol_convention` to both `UVData` and `UVCal`, as well as `uvcalibrate` which applies it from one to the other. The parameter is _not_ required, but a warning is issued if it is not specified on data that is calibrated, and an error is raised if it _is_ specified when the data is not calibrated.

In `uvcalibrate`, I am forcing the parameter to be set (either inherited from the `UVCal` or set directly as a keyword to the function). 

## Motivation and Context

The primary motivation is to keep track of this convention in data, so that the combination of calibration and power-spectrum estimation can be done consistently without "external knowledge". This is something useful for HERA at the moment.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:


bhazelton (commented):
This is a good start but we need ways to round trip this new parameter through the various file formats for UVCal and UVData. I'm happy to help on that front if you need.

src/pyuvdata/utils/uvcalibrate.py
```diff
@@ -21,6 +22,7 @@ def uvcalibrate(
     undo=False,
     time_check=True,
     ant_check=True,
+    pol_convention: Literal["sum", "avg"] | None = None,
```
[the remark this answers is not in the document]
steven-murray: You're right! I've updated so that if nothing is specified, it doesn't do anything about the convention, and doesn't set it on the uvd.

steven-murray:
Thanks @bhazelton -- I've fixed up the things you suggested, and also added the read/write to uvh5. I don't know enough about fits/ms to help there, so if you could manage that part, I'd be very grateful. I added a test for uvdata to test that warnings/errors are raised appropriately. My guess is that read/write will be automatically covered in round-trip tests.

steven-murray:
However, there might be a lot of new warnings now with old file types.

jsdillon:
Is this branch at the point where I can use it to start writing files with the pol convention attribute, or should I hold off?

steven-murray:
@bhazelton we need to decide how we want to fix the breaking tests:

1. Update all the lists of expected warnings to include one more
2. Update (most of) the data files so the warnings aren't raised. 
3. something else?

bhazelton:
> @bhazelton we need to decide how we want to fix the breaking tests:
> 
> 1. Update all the lists of expected warnings to include one more
> 2. Update (most of) the data files so the warnings aren't raised.
> 3. something else?

I can accept either 1 or 2, although maybe we need a wider discussion about whether we want to warn about this every time we call `check`, especially on UVData because this is not something that exists on any of the existing formats we support, so it feels a little presumptuous to yell at all our users about it all the time. Maybe we can instead warn only in the uvcalibrate method if it's not set and then advertise the existence of this new attribute in the tutorial and docs?

kartographer (commented):
I think overall the code is looking in pretty good shape, and have nothing in terms of detailed comments to offer aside from highlighting that the doctest is still failing.

As a more general comment, I _do_ think it would be good to put a longer explanation into the tutorials about this, since I can easily see where this might trip someone up. In particular, I think most of the higher frequency interferometers (e.g., VLA, ALMA, SMA, ATCA, ATA) take the "avg" approach, so it may be good to be explicit about what's typical/where "sum" is definitely used (since I think its the less common option of the two).

bhazelton:
I have a question:
Should we error in `UVCal.check` if the `pol_convention` is set but the `gain_scale` isn’t? I feel like setting the `pol_convention` without the `gain_scale` isn’t very meaningful. This is already partially handled in `uvcalibrate` where the `pol_convention` isn't used if `gain_scale` isn't set, but I'm wondering about adding this to the UVCal check.

jsdillon:
I think a warning is probably sufficient, but I wouldn't necessarily be opposed to an error in that case.

steven-murray:
I agree with @jsdillon that a warning is the right balance here.

bhazelton:
I added the warning and updated the docstrings a bit.

src/pyuvdata/utils/uvcalibrate.py
```diff
@@ -417,5 +460,43 @@ def uvcalibrate(
         if uvcal_use.gain_scale is not None:
             uvdata.vis_units = uvcal_use.gain_scale
 
+    # Set pol convention properly
+    if uvcal_use.gain_scale is None:
+        # ignore all pol_convention stuff, because who knows???
+        warnings.warn(
+            "gain_scale is not set, so resulting uvdata will not have correct units"
```
bhazelton, line 467: I think we need to add something to this warning message to make it clear that the `pol_convention` isn't being used in this case. Maybe we should also add something about this to the docstring under the `pol_convention` parameter.

docs/uvcal_tutorial.rst
```diff
@@ -42,6 +42,23 @@ gives the beginning and end of the time range. The local sidereal times follow a
 pattern, UVCal objects should have either an ``lst_array`` or an ``lst_range`` attribute
 set.
 
+Generating calibration solutions typically requires choosing a convention concerning how
+polarized sky emission is mapped to the linear polarizations of the instrument. For
```
bhazelton, line 46: I prefer "is mapped to instrumental polarizations." Because telescopes with right and left circular polarization feeds have exactly this same issue of averaging vs summing.

I think it's fine to leave the following example with linear pols, I just don't want to cast this as limited to linear pol.

docs/uvcal_tutorial.rst
```diff
@@ -260,6 +277,64 @@ UVCal: Calibrating UVData
 Calibration solutions in a :class:`pyuvdata.UVCal` object can be applied to a
 :class:`pyuvdata.UVData` object using the :func:`pyuvdata.utils.uvcalibrate` function.
 
+Generating calibration solutions typically requires choosing a convention concerning how
+polarized sky emission is mapped to the linear polarizations of the instrument. For
```
bhazelton, line 281: same comment about "instrumental polarization" rather than "linear polarization"

docs/uvcal_tutorial.rst
```diff
@@ -260,6 +277,64 @@ UVCal: Calibrating UVData
...
+convention). This choice is generally encoded in the sky model to which the visibilities
+are calibrated. Different tools and simulators make different choices, generally following
+a standard choice for the field. When calibrating a ``UVData`` object with a ``UVCal``
+object using :func:`pyuvdata.utils.uvcalibrate`, it is *required* to specify this
+convention. At this time, the convention can be specified either on the ``UVCal`` object
+itself, or as a parameter to :func:`pyuvdata.utils.uvcalibrate`. The chosen ``pol_convention``
+will then be applied to and stored on the resulting ``UVData`` object.
+
+There are a few non-trivial combinations of parameters concerning the ``pol_convention``
+that should be noted:
```
bhazelton, line 293: This is excellent, thank you!!

src/pyuvdata/utils/uvcalibrate.py
```diff
@@ -22,7 +102,8 @@ def uvcalibrate(
     undo: bool = False,
     time_check: bool = True,
     ant_check: bool = True,
-    pol_convention: Literal["sum", "avg"] | None = None,
+    uvc_pol_convention: Literal["sum", "avg"] | None = None,
+    uvd_pol_convention: Literal["sum", "avg"] | None = None,
```
bhazelton, line 106: The docstring hasn't been updated for this change

bhazelton (commented):
Thanks for the new docs, they are great. Just a few comments.

Still working on figuring out the circleci issue...

docs/uvcal_tutorial.rst
steven-murray: Got it, good point.

src/pyuvdata/utils/uvcalibrate.py
```diff
@@ -417,5 +535,29 @@ def uvcalibrate(
...
+        correction[uvdata.polarization_array < 0] /= 2
+
+        if (undo and uvd_pol_convention == "sum") or (
+            not undo and uvd_pol_convention == "avg"
+        ):
+            correction = 1 / correction
+
+        uvdata.data_array *= correction
+
+    uvdata.pol_convention = None if undo else uvd_pol_convention
```
bhazelton, line 563: This is pretty hard to parse and be confident in, but I think your test shows that it's working properly, at least for linear pols. I don't see a test on Stokes visibilities (i.e. when uvdata.polarization_array > 0). Is there a test for that case?

Also, please move this to above `uvcalibrate` so it's defined before it's called.

steven-murray: OK cool. I moved the function, wrote a docstring for it to hopefully make the logic more clear, and also added tests for circular polarizations. However, in doing this I found that it seems that it is *not possible* to calibrate/decalibrate uvdata objects with stokes polarizations (it complains that the polarizations can't be found on the UVCal object). 

Is this to be expected?

bhazelton: I don't really know about a case where stokes calibration solutions make sense, but I feel like @kartographer ran into it when he was implementing MS cal read/write?

src/pyuvdata/utils/uvcalibrate.py
```diff
@@ -3,24 +3,163 @@
...
+
+
+def _apply_pol_convention_corrections(
+    uvdata,
+    undo: bool,
+    uvc_pol_convention: Literal["sum", "avg"] | None,
+    uvd_pol_convention: Literal["sum", "avg"] | None,
+):
+    r"""
+    Apply corrections to calibration/de-calibration from differences in convention.
```
bhazelton, line 105: Thanks, this is great!

src/pyuvdata/utils/uvcalibrate.py
```diff
@@ -406,8 +457,10 @@ def uvcalibrate(
     uvdata_feed_pols = {
         feed for pol in uvdata_pol_strs for feed in POL_TO_FEED_DICT[pol]
     }
+    print("ALL", uvdata_feed_pols)
```
bhazelton, line 460: did you mean to leave this print statement in?

src/pyuvdata/utils/uvcalibrate.py
```diff
@@ -406,8 +457,10 @@ def uvcalibrate(
...
     for feed in uvdata_feed_pols:
         # get diagonal jones str
+        print("FEED", feed)
```
bhazelton, line 463: Or this one?

bhazelton: (this one is breaking the doc tests because it wasn't expected.)

src/pyuvdata/utils/uvcalibrate.py
kartographer: So this is definitely a corner-case (that could be wrapped with a "pyuvdata does not support... if you file an issue..."), but the most common case I ran into was visibilities recorded as (pseudo-)Stokes I, which can have gains tables corresponding also to (pseudo-)Stokes I. Looking at the UVCal class though, we don't _currently_ allow for Jones > 0, which is maybe what's causing the issue here.

This got raised in the MS read/write mostly on the visibility side, since the MS format has a `FEED` table that I had to match to the data, hence why those pseudo Stokes values are supported there.

docs/uvcal_tutorial.rst
```diff
@@ -42,6 +42,23 @@ gives the beginning and end of the time range. The local sidereal times follow a
 pattern, UVCal objects should have either an ``lst_array`` or an ``lst_range`` attribute
 set.
 
+Generating calibration solutions typically requires choosing a convention concerning how
+polarized sky emission is mapped to the instrumental polarizations. For
+linear polarizations ``XX`` and ``YY``, the stokes ``I`` sky emission can be mapped to
+``I = (XX + YY)/2`` (the ``avg`` convention) or ``I = XX + YY`` (the ``sum``
+convention). This choice is generally encoded in the sky model to which the visibilities
+are calibrated. Different tools and simulators make different choices, generally following
+a standard choice for the field. In ``pyuvdata`` either of these choices are OK, but the
```
kartographer, line 51: > Different tools and simulators make different choices, generally following a standard choice for the field.

I think we need to be more explicit here about which tools do what, _especially_ if we are going to recommend that users always specify this parameter. In particular, I believe the tasks in CASA (e.g., tclean) and MIRIAD, along with WSClean, all assume the ``avg`` convention. I actually am not sure of what uses the ``sum`` convention, though I expect we can gather that information.

src/pyuvdata/utils/uvcalibrate.py
```diff
@@ -3,24 +3,163 @@
...
+    uvdata,
+    uvcal,
+    undo: bool,
+    uvc_pol_convention: Literal["sum", "avg"] | None,
+    uvd_pol_convention: Literal["sum", "avg"] | None,
+):
+    if uvc_pol_convention is None and uvcal.pol_convention is None:
+        warnings.warn(
+            message=(
+                "pol_convention is not specified on the UVCal object, and "
```
kartographer, line 23: I don't know that I'd want this to error when the pol convention is not specified on either UVCal and UVData, at least not if we're treating `pol_convention` as an optional parameter. Why can't this just stay a warning?

kartographer (commented):
Just a few additional comments...

src/pyuvdata/utils/uvcalibrate.py
steven-murray: Sorry @kartographer, I'm not following -- this is a warning?

docs/uvcal_tutorial.rst
steven-murray: I'm not very familiar with such tools, but I'll add that particular information.

src/pyuvdata/utils/uvcalibrate.py
kartographer: @steven-murray -- sorry, I mean a general `UserWarning`, not a `DeprecationWarning`. I.e., this should alert the user that there's a potential for mismatch, but I don't think this should throw an error.

docs/uvcal_tutorial.rst
bhazelton: FHD uses the sum convention and the various HERA packages are moving to it.

src/pyuvdata/utils/uvcalibrate.py
steven-murray: OK, I updated all the DeprecationWarning's to UserWarnings. Originally I had intended these to be DeprecationWarnings because we want to push people to use the keywords. However, it would be really annoying to have that warning if you really just don't care about the final units (like, you're just using calibration solutions to multiply your data by 2...)

docs/uvcal_tutorial.rst
steven-murray: Cool. I added that as well.

kartographer (approved):
Looking good to me -- thanks @steven-murray!

jsdillon:
Thanks all! 

You planning a new minor version release on pypi?

PR #1456, opened 2024-07-05
feat: add Telescope object to FastUVH5Meta

steven-murray:
## Description

Adds a convenience cached-property onto `FastUVH5Meta` that yields a `Telescope` object. 

## Motivation and Context

Since many of the `.new()` methods now prefer accepting a `Telescope` instead of the list of telescope attributes, when creating something from a `FastUVH5Meta` object it is much simpler if access to such an object is simple.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/src/pyuvdata/uvdata/uvh5.py
+++ b/src/pyuvdata/uvdata/uvh5.py
@@ -449,6 +449,11 @@ class FastUVH5Meta(hdf5_utils.HDF5Meta):
         )
         return uvd
 
+    @cached_property
+    def telescope(self):
+        """A Telescope object created from the data."""
+        return Telescope.from_hdf5(self, run_check=False)
+
 
 class UVH5(UVData):
     """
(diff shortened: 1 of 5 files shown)


src/pyuvdata/uvdata/uvh5.py
```diff
@@ -449,6 +449,11 @@ def to_uvdata(
         )
         return uvd
 
+    @cached_property
+    def telescope(self):
+        """A Telescope object created from the data."""
+        return Telescope.from_hdf5(self, run_check=False)
```
bhazelton, line 455: I think I'd prefer this go on the HDF5Meta class defined in `utils/io/hdf5.py` that FastUVH5Meta and FastCalH5Meta inherit from. That object already has some handling for telescope locations because it's needed for both the visibility data and calibrations solutions.

bhazelton: It also looks like this isn't actually called in a test?

steven-murray: I see, that makes sense. I've moved it there. And yes, I hadn't added a test. I have now.

bhazelton (commented):
This is looking good. Looks like it needs to be rebased to handle a conflict.

bhazelton (approved):

PR #1458, opened 2024-07-06
Add select on read for MWA correlator FITS files

### LINKED ISSUE #1386: Select on read for MWA correlator fits
We've got a project that is hitting a bottleneck due to the lack of select on read for MWA correlator fits files.  We should discuss the feasibility of implementing it (and if there is even a potential speedup given the nature of the file format).

bhazelton:
## Description

Add support for select on read to MWA correlator FITS files to save memory and time. This adds support for most but not all select keywords. The implementation is a bit complicated because of the structure of MWA correlator FITS files, which is also different between the files written by the legacy correlator and MWAX.

Edit: we have benchmarking now and the benchmarks didn't slow down significantly
~**I am a bit concerned that some of these changes may have slowed down the reading when not selecting on read. I think this needs to be tested/profiled more, would love some help on that front.**~

I also refactored some selection code for the baseline-time axis and the frequency axis into functions for better code reuse, leading to changes in `UVData.select` and `UVCal.select`. And I improved the way the history string is built up.

## Motivation and Context

closes #1386

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:


src/pyuvdata/uvdata/mwa_corr_fits.py
```diff
@@ -599,19 +607,29 @@ def flag_init(
             else np.reshape(self.flag_array, reshape)
         )
 
+        if freq_inds is not None:
+            nfreq_use = n_orig_freq
```
mkolopanis, line 611: should this check also look for `n_orig_freq` and not only `freq_inds`

mkolopanis (commented):
I do feel like a lot of the mwa specific stuff is lost on me, but I do have a question after a quick read

src/pyuvdata/uvdata/mwa_corr_fits.py
bhazelton: It could. The code that calls it should always pass in `n_orig_freq` if `freq_inds`
 is not None, but there's no enforcement of that in this function. I'll add that.

mkolopanis (dismissed):
Thanks for the little change. Looks good to me

mkolopanis (approved):

PR #1463, opened 2024-07-24
Benchmark

mkolopanis:
## Description

Adds CI benchmarking so we can monitor our performance sensitive areas.
## Motivation and Context

Because manually benchmarking is so 2002
## Types of changes

- [ ] Version change


mkolopanis:
well well, the output for this leaves a LOT to be sifted through

.github/workflows/macosx_windows_ci.yaml
```diff
@@ -73,3 +73,89 @@ jobs:
...
+          # Where the previous data file is stored
+          external-data-json-path: ./cache/benchmark-data.json
+          # Workflow will fail when an alert happens
+          fail-on-alert: true
+          # Enable Job Summary for PRs
+          summary-always: true
+          # Enable commenting
+          comment-always: true
+          # Comment on the PR if the branch is not a fork
+          comment-on-alert: ${{ github.event.pull_request.head.repo.fork == false }}
```
bhazelton, line 160: Is there a reason not to comment on Fork PRs?

mkolopanis: yeah there is, but I don't 100% follow it. I think it is mostly related to github pages things (you can host results on github pages with this too). So it's more a security thing against that.

bhazelton: I'd be incline to let it comment on all PRs. In the past, actions were set up so they didn't run on forks and I had to go in and click a button to actually get the actions to run on fork PRs, which is annoying.

mkolopanis:
should we add a check that only stores the benchmark data when we run on main?
like the check below on the "store benchmark data" step. Otherwise I'm worried we'll only ever be comparing against other commits on the same PR or against each other PRs
```if: github.ref == 'refs/heads/main'```

mkolopanis:
@bhazelton I think this is ready for a look now. We'll only store the benchmark results when executed from the main branch now. 

interestingly the baseline/antenna-number conversion triggered a failure this last time :laughing:

bhazelton (approved):
Awesome, thanks for this!!

PR #1464, opened 2024-07-30
Faster interpolation in `UVBeam.interp`

tyler-a-cox:
Adds new option to `UVBeam.interp` that improves the interpolation speed of `UVBeam._interp_az_za_rect_spline`

## Description
This PR adds additional functionality to `UVBeam.interp` that allows users to choose between two interpolators (`scipy.interpolate.RectBivariateSpline` and `scipy.ndimage.map_coordinates`) for interpolation along the spatial axis  when running `UVBeam._interp_az_za_rect_spline`. Users can select between the two interpolators by using the newly added `spatial_interp_func` keyword argument in `UVBeam.interp`, where the default is `RectBivariateSpline` which is currently the interpolator used in `UVBeam._interp_az_za_rect_spline`.

## Motivation and Context
When running visibility simulations with `hera_sim`, we noticed that beam interpolation was one of the most expensive steps. The current interpolator `scipy.interpolate.RectBivariateSpline` is quite flexible, but can be quite a bit slower than other gridded interpolators in `scipy`. I ran this branch on HERA's Vivaldi [beam file](https://github.com/HERA-Team/HERA-Beams/blob/master/NicolasFagnoniBeams/NF_HERA_Vivaldi_efield_beam.fits) using `spatial_interp_func="map_coordinates"` and found a ~3x improvement in the interpolation time compared to when `spatial_interp_func="RectBivariateSpline"`.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- New spatial interpolation method to `UVBeam.interp` that improves that linear interpolation speed of `_interp_az_za_rect_spline`
 - New UVParameter `pol_convention` on `UVData` and `UVCal`. This specifies the convention
 assumed for converting linear to stokes polarizations -- either "sum" or "avg". Also
 added to `uvcalibrate` to apply from the `UVCal` to the `UVData`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
-- New spatial interpolation method to `UVBeam.interp` that improves that linear interpolation speed of `_interp_az_za_rect_spline`
+- New optional spatial interpolation method, ``interpolation_function="az_za_map_coordinates"`` that improves the linear
+interpolation speed for data in ``az_za`` coordinates.
 - New UVParameter `pol_convention` on `UVData` and `UVCal`. This specifies the convention
 assumed for converting linear to stokes polarizations -- either "sum" or "avg". Also
 added to `uvcalibrate` to apply from the `UVCal` to the `UVData`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@ warnings for specific telescopes.
 telescope object from the metadata.
 - `UVData.get_enu_data_ants` method to get east, north, up positions only for
 antennas with data.
-- New spatial interpolation method to `UVBeam.interp` that improves that linear interpolation speed of `_interp_az_za_rect_spline`
 
 ### Fixed
 - A bug in reading UVH5 files with antenna names saved as variable length strings
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ warnings for specific telescopes.
 telescope object from the metadata.
 - `UVData.get_enu_data_ants` method to get east, north, up positions only for
 antennas with data.
+- New spatial interpolation method to `UVBeam.interp` that improves that linear interpolation speed of `_interp_az_za_rect_spline`
 
 ### Fixed
 - A bug in reading UVH5 files with antenna names saved as variable length strings
--- a/src/pyuvdata/uvbeam/uvbeam.py
+++ b/src/pyuvdata/uvbeam/uvbeam.py
@@ -1405,7 +1405,8 @@ class UVBeam(UVBase):
             "RegularGridInterpolator",
         ]:
             raise ValueError(
-                "interpolator must be 'RectBivariateSpline' or 'RegularGridInterpolator'"
+                "interpolator must be 'RectBivariateSpline' or"
+                " 'RegularGridInterpolator'"
             )
 
         npoints = az_array.size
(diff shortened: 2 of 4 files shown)


steven-murray (commented):
Thanks @tyler-a-cox! One thought: is `RegularGridInterpolator` exactly the same as `RectBivariateSpline` (when then grid happens to be regular)? If so, we could either just *replace* the original RectBivariateSpline with the new regular interpolator, or, if we suspect that we might get *irregular* grids from time to time, at least make the default dynamic, i.e. choose the regular grid if the grid is regular, otherwise choose the bivariate spline.

tyler-a-cox:
Thanks for the reply, @steven-murray! As we discussed offline, I've swapped the `RectangularGridInterpolator` with `map_coordinates` which produces identical results to `RectBivariateSpline` for linear interpolation. For higher order interpolation, the results are largely identical, but there are some slightly differences at the edges of the interpolation range that I think are the result of differences in smoothing the interpolators do when using higher order interpolants. I would be interested in building in the infrastructure to use a different interpolator depending on if the interpolation grid is regular or not, but might be slightly concerned about doing so if the results aren't identical. I'm also slightly concerned about the fact that the `beam.interp` allows users to provide a dictionary of parameters to pass to the interpolation function. The arguments between these functions are slightly different, and it's unclear to me how we would handle that. I'd be interested to hear your thoughts!

steven-murray (dismissed):
Thanks @tyler-a-cox. I think this is probably the way to go. We don't want to do anything API-breaking here, so it does have to be a keyword argument rather than choosing a default dynamically, because as you say the `spline_opts` are different between the two.

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -1340,6 +1341,10 @@ def _interp_az_za_rect_spline(
         check_azza_domain : bool
             Whether to check the domain of az/za to ensure that they are covered by the
             intrinsic data array. Checking them can be quite computationally expensive.
+        spatial_interp_func : str
+            The spatial interpolation function to use. Options are 'RectBivariateSpline'
+            and 'map_coordinates'. If 'map_coordinates' is chosen, the input data must
+            be on a regular grid.
```
bhazelton, line 1347: I believe that when the pixel coordinate system is `"az_za"` (which is required for this method), the beam is required to be on a regular grid. I thought there was something enforcing this in the check, but I just looked and it's only enforced in the `read_cst` method. There's also a warning (but not an error) in the `select` method if you do a select that creates an irregular grid. I guess this is something we should talk more about, but the original intent, at least, was that these beams would have regularly gridded coordinates.

steven-murray (commented):
Thanks @tyler-a-cox this all looks good to me. The only thing to still discuss I think is whether we need the check for regularity, given that as @bhazelton mentioned the data should always be regularly spaced.

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -60,6 +60,10 @@ class UVBeam(UVBase):
             "description": "scipy RectBivariate spline interpolation",
             "func": "_interp_az_za_rect_spline",
```
bhazelton, line 61: We discussed maybe changing the name of this option (with deprecation!) to include something related to RectBivariateSpline. I think maybe it's best to push that to another PR, just wanted to mention it here.

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -1303,6 +1307,161 @@ def get_lambda(real_lut, imag_lut=None):
...
 
+    def _prepare_freq_interpolation(
+        self, freq_array, *, freq_interp_kind="cubic", freq_interp_tol=1.0
+    ):
+        """
+        Perform frequency interpolation of the beam.
+
+        This helper function handles the frequency interpolation prior to performing
+        interpolation along the azimuth/zenith angle axes.
+        """
```
bhazelton, line 1318: I got a little confused here because we also have `_interp_freq` which does the frequency interpolation. I think this function is great, just maybe tweak the name and/or docstring to be more clear. Is this is specifically for az_za beams? something else?

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -1365,183 +1527,90 @@ def _interp_az_za_rect_spline(
...
-                    phi_vals,
-                    phi_vals[-1] + (np.arange(1, extend_length + 1) * axis1_diff),
-                )
-            )
+        # Prepare the data for interpolation
+        data_use, phi_use, theta_use = self._prepare_coordinate_data(input_data_array)
 
-            low_slice = input_data_array[..., :extend_length]
-            high_slice = input_data_array[..., -1 * extend_length :]
+        # Prepate basis functions
```
bhazelton, line 1568: Prepate -> Prepare?

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -1574,6 +1644,163 @@ def get_lambda(real_lut, imag_lut=None):
...
+        freq_array,
+        freq_interp_kind="cubic",
+        freq_interp_tol=1.0,
+        polarizations=None,
+        spline_opts=None,
+        check_azza_domain: bool = True,
+        reuse_spline: bool = False,
+    ):
+        """
+        Interpolate in az_za coordinate system using RectBivariateSpline.
```
bhazelton, line 1661: I thought this one didn't use RectBivariateSpline?

bhazelton (commented):
This is looking really good, thanks @tyler-a-cox! I just had a couple comments on docstrings. It looks like there are a couple of lines of missed coverage, but they're similar to lines being covered for the other interp method, so maybe a judiciously placed pytest parametrize call would take care of them pretty easily?

src/pyuvdata/uvbeam/uvbeam.py
tyler-a-cox: Oops, typo there. Thanks for catching that. That's been updated now.

src/pyuvdata/uvbeam/uvbeam.py
tyler-a-cox: Fixed.

src/pyuvdata/uvbeam/uvbeam.py
tyler-a-cox: I renamed this to `handle_input_for_freq_interpolation` and clarified the purpose of the function in the docstring.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- New spatial interpolation method to `UVBeam.interp` that improves that linear interpolation speed of `_interp_az_za_rect_spline`
```
bhazelton, line 7: Maybe update this after the refactoring, it's a little weirdly worded now and give the new name to be passed to the `interp` method (`"az_za_map_coordinates"`).

src/pyuvdata/uvbeam/uvbeam.py
bhazelton: We talked about this on the telecon yesterday and @steven-murray suggested we add the regularity checks in this PR. He said he could do it if you want, I'm also happy to help.

src/pyuvdata/uvbeam/uvbeam.py
bhazelton: moved this to an issue: #1473

bhazelton (dismissed):
Looks good to me now.

bhazelton (dismissed):

bhazelton (approved):

PR #1482, opened 2024-10-08
[WIP] Add optional near-field corrections to the phasing function

baron-de-montblanc:
## Description

Included code to perform near-field corrections through keyword arguments passed to the phase() function.

I added two new keys to the ```phase_dict``` dictionary (```cat_dist_units``` and ```cat_near_field```), allowing the user to specify the units associated with the ```cat_dist``` key, and whether or not to apply near-field corrections (boolean value). As a result, a lot of tests are breaking because these extra keys didn't exist when they were written. I'm having trouble tracking down every instance of ```phase_dict``` to patch this.

It also still isn't clear whether or not an ```unproject_near_field()``` function is necessary, so I have left in some commented-out lines of code while I work on figuring this out (to be removed or uncommented when this is resolved).

## Motivation and Context

This feature allows users to specify the RA, DEC, and distance to a near-field source, and "focus" on that source (it will then appear in focus in an image, for example). This is done in practice by calculating the geometric near-field ```w``` values for each visibility, updating those in the ```uvw_array```, computing the resulting phase differences, and updating the phase directly for each visibility. This feature is turned off by default (i.e. the default behavior for the ```phase()``` function is just to perform the traditional far-field phasing).

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:


bhazelton:
Rather than adding new phase_center_dict keys, let's use a new cat type for near-field phased objects, maybe `"near_field"? Then encourage and support passing a Quantity for distance, but if a scalar is passed interpret it as in meters for near field and parsecs for ephem cat types. Update the docs to be very clear about this.

src/pyuvdata/uvdata/uvdata.py
```diff
@@ -4623,11 +4624,65 @@ def _phase_dict_helper(
             elif (key == "cat_id") and (phase_dict[key] is not None):
                 # If this is the cat_id, make it an int
                 phase_dict[key] = int(phase_dict[key])
-            elif not ((phase_dict[key] is None) or isinstance(phase_dict[key], str)):
-                phase_dict[key] = float(phase_dict[key])
-
```
bhazelton, line 4628: Why did you remove these lines?

tests/uvdata/test_uvfits.py
```diff
@@ -584,6 +584,38 @@ def test_readwriteread(tmp_path, casa_uvfits, telescope_frame, selenoid):
...
+
+#     uv_in.filename = uv_out.filename
+#     uv_out._consolidate_phase_center_catalogs(
+#         reference_catalog=uv_in.phase_center_catalog,
+#     )
+
+#     assert uv_in == uv_out
+#     return
+
+
```
bhazelton, line 618: Let's remove this. If there's a good reason we might need it in the future it can go into an issue.

bhazelton (commented):
This is looking really good! Just a few small comments.

bhazelton (commented):
This all looks great to me now. I think we just need a changelog update and possibly an addition to the tutorial to advertise this new functionality.

kartographer (dismissed):
Hi @baron-de-montblanc -- I think this is looking in good shape, just a few substantive comments, feel free to ping me if you have any questions!

src/pyuvdata/utils/tools.py
```diff
@@ -584,3 +584,98 @@ def _where_combine(mask, inds=None, invert=False, use_and=True):
...
+    ant1 = uvd.ant_1_array
+    ant2 = uvd.ant_2_array
+
+    ind1 = []
+    ind2 = []
+
+    for i in ant1:
+        ind1.append(np.where(ants == i)[0][0])
+    for i in ant2:
+        ind2.append(np.where(ants == i)[0][0])
```
kartographer, line 614: One other thing (that I thought I put in my review but maybe I accidentally deleted?) -- there are a lot of calls to `np.where` and `list.append` here, which can be fairly slow, especially when iterating through a long list like the above. Just a suggestion, you might think of using something like a dictionary to speed up (by more than an order of magnitude -- ~30x with the code below) mapping antenna number to index position, e.g.:

```suggestion
    ant_map = {ant: idx for idx, ant in enumerate(uvd.telescope.antenna_numbers)}
    
    ind1 = [ant_map[ant] for ant in uvd.ant_1_array]
    ind2 = [ant_map[ant] for ant in uvd.ant_2_array]
```

src/pyuvdata/utils/phasing.py
```diff
@@ -2571,3 +2574,114 @@ def uvw_track_generator(
...
+    )
+
+    # Get the uvw array along the Nblts axis; select only the w's
+    old_w = uvd.uvw_array[:, -1]
+
+    # Calculate near-field delay
+    new_w = r1 - r2
+
+    # Mask autocorrelations
+    mask = np.not_equal(uvd.ant_1_array, uvd.ant_2_array).astype(int)
```
kartographer, line 2684: Hey @baron-de-montblanc, this is looking excellent! One final small comment -- you don't need the `astype` here, as the output of `not_equal` can get plugged straight into `where` (`where` wants a boolean array, so right now your going bool -> int -> bool). After that, I think this is ready for merge...

bhazelton:
Maybe also something in the tutorial?

src/pyuvdata/uvdata/uvdata.py
```diff
@@ -4650,9 +4651,45 @@ def _phase_dict_helper(
...
+        ra : ndarray
+            Right ascension of the focal point ie phase center (rad; shape (Ntimes,))
+        dec : ndarray
+            Declination of the focal point ie phase center (rad; shape (Ntimes,))
+
+        Returns
+        -------
+        None (performs operations inplace)
+        """
+        # ------- Parameters that are independent of frequency --------
```
bhazelton, line 4673: This marker line is a little bit weird now that all the other ones are gone...

bhazelton (approved):
Awesome work, thank you so much @baron-de-montblanc !

PR #1483, opened 2024-10-10
perf: don't always return basis-vector array

steven-murray:
## Description

This makes it possible to _not_ return the basis vector array, which is often not needed in simulation contexts, and takes a reasonably significant amount of memory.

It also computes the modified az/za arrays in the `map_coordinates` interpolator in-place to reduce memory (this is fine since these arrays are previously copied outside this function).

## Motivation and Context

In profiling `fftvis`, I found that the memory usage was higher than naively expected (this was multiprocessing over 128 threads, so everything was magnified). One of the unnecessary memory allocations was this basis-vector array.

## Types of changes

- [ ] Version change
- [x] Performance enhancement

## Checklist:

Performance checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+- Made it possible to *not* return the `interp_basis_vector` array from beam
+interpolators.
+
 ## [3.1.1] - 2024-10-28
 
 ### Fixed
@@ -37,10 +41,6 @@ antennas with data.
 (the azimuthal-aligned response was swapped with the zenith angle-aligned response).
 - A bug in reading UVH5 files with antenna names saved as variable length strings
 that was introduced in v3.0.0.
-
-### Changed
-- Made it possible to *not* return the `interp_basis_vector` array from beam
-interpolators.
 
 ## [3.0.0] - 2024-7-1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@ antennas with data.
 - A bug in reading UVH5 files with antenna names saved as variable length strings
 that was introduced in v3.0.0.
 
+### Changed
+- Made it possible to *not* return the `interp_basis_vector` array from beam
+interpolators.
+
 ## [3.0.0] - 2024-7-1
 
 ### Added
--- a/src/pyuvdata/uvbeam/uvbeam.py
+++ b/src/pyuvdata/uvbeam/uvbeam.py
@@ -1501,7 +1501,7 @@ class UVBeam(UVBase):
         reuse_spline=False,
         spline_opts=None,
         check_azza_domain: bool = True,
-        return_basis_vector: bool = True,
+        return_basis_vector: bool = False,
     ):
         """
         Interpolate in az_za coordinate system using RectBivariateSpline.
@@ -1532,6 +1532,9 @@ class UVBeam(UVBase):
         check_azza_domain : bool
             Whether to check the domain of az/za to ensure that they are covered by the
             intrinsic data array. Checking them can be quite computationally expensive.
+        return_basis_vector : bool
+            Whether to return the interpolated basis vectors. Prior to v3.1.1 these
+            were always returned. Now they are not by default.
 
         Returns
         -------
@@ -1684,7 +1687,7 @@ class UVBeam(UVBase):
         spline_opts=None,
         check_azza_domain: bool = True,
         reuse_spline: bool = False,
-        return_basis_vector: bool = True,
+        return_basis_vector: bool = False,
     ):
         """
         Interpolate in az_za coordinate system using map_coordinates.
@@ -1715,6 +1718,9 @@ class UVBeam(UVBase):
         check_azza_domain : bool
             Whether to check the domain of az/za to ensure that they are covered by the
             intrinsic data array. Checking them can be quite computationally expensive.
+        return_basis_vector : bool
+            Whether to return the interpolated basis vector. Default is False as of
+            v3.1.1, but was previously True.
 
         Returns
         -------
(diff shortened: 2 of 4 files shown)


bhazelton (commented):
This is great, thanks. Just a few suggestions.

tests/uvbeam/test_uvbeam.py
```diff
@@ -1260,6 +1260,38 @@ def test_interp_longitude_branch_cut(beam_type, cst_efield_2freq, cst_power_2fre
...
 
 
+@pytest.mark.parametrize(
+    "interp_func", ["az_za_simple", "az_za_map_coordinates", "healpix_simple"]
+)
+def test_interp_no_basis_vector(cst_efield_2freq, interp_func):
+    beam = cst_efield_2freq
+
+    if interp_func == "healpix_simple":
+        beam.to_healpix()
```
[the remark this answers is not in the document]
steven-murray: Good catch.

CHANGELOG.md
```diff
@@ -32,6 +32,10 @@ antennas with data.
 - A bug in reading UVH5 files with antenna names saved as variable length strings
 that was introduced in v3.0.0.
 
+### Changed
+- Made it possible to *not* return the `interp_basis_vector` array from beam
+interpolators.
+
```
bhazelton, line 38: This needs to be moved up to be in the Unreleased section.

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -1710,6 +1718,9 @@ def _interp_az_za_map_coordinates(
         check_azza_domain : bool
             Whether to check the domain of az/za to ensure that they are covered by the
             intrinsic data array. Checking them can be quite computationally expensive.
+        return_basis_vector : bool
+            Whether to return the interpolated basis vector. Default is False as of
+            v3.1.1, but was previously True.
```
bhazelton, line 1723: This is going to break things. We need a deprecation process because this keyword didn't exist in the past. I think you should default it to None and then catch that case, issue a deprecation warning and set it to True. The deprecation warning should say that it will default to False starting in version 3.3.

CHANGELOG.md
steven-murray: Ah yeah, things were released in between ;-)

src/pyuvdata/uvbeam/uvbeam.py
steven-murray: Ah yeah. So at the moment, I'm doing this deprecation stuff only in the top-level function `interp()` because these functions are private and therefore need not be API-stable. Is that OK or do you want me to do the deprecation logic everywhere?

bhazelton: Oh, sorry, I didn't see that. I think that's fine.

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -2068,6 +2084,9 @@ def interp(
             Conversely, if the passed az/za are outside of the domain, they will be
             silently extrapolated and the behavior is not well-defined. Only
             applies for `az_za_simple` interpolation.
+        return_basis_vector : bool
+            Whether to return the interpolated basis vectors. Prior to v3.1.1 these
+            were always returned. Now they are not by default.
```
bhazelton, line 2089: This isn't right. The default is still False for this method, will become True in v3.3

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -2092,6 +2111,18 @@ def interp(
...
+            # To create a new object, we always need the interpolated basis vectors.
+            return_basis_vector = True
+
+        if return_basis_vector is None:
+            return_basis_vector = True
+            warnings.warn(
+                "The default value for `return_basis_vector` is True, but in v3.3 it "
+                "will be set to False. Silence this warning by explicitly setting it "
+                " to either True or False."
+            )
```
bhazelton, line 2124: This should be a deprecation warning.

src/pyuvdata/uvbeam/uvbeam.py
steven-murray: Ah, yes.

src/pyuvdata/uvbeam/uvbeam.py
steven-murray: Good catch.

src/pyuvdata/uvbeam/uvbeam.py
steven-murray: So, I've left the default for now as True, and added the statement that in v3.3 it will be updated to be False.

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -1787,26 +1801,20 @@ def _interp_az_za_map_coordinates(
...
-        _za_array = (
-            (za_array - theta_use.min())
-            / (theta_use.max() - theta_use.min())
-            * (theta_use.size - 1)
-        )
+        az_array -= phi_use.min()
+        az_array /= (phi_use.max() - phi_use.min()) / (phi_use.size - 1)
+
+        za_array -= theta_use.min()
+        za_array /= (theta_use.max() - theta_use.min()) / (theta_use.size - 1)
```
bhazelton, line 1808: Minor nitpick: I'd find these divide lines easier to read if they were multiplied by the reciprocal. I have to think harder about what the two divides mean.

bhazelton (commented):
I think the tests will pass now if you rebase. We just released another version, so be careful with the changelog.

src/pyuvdata/uvbeam/uvbeam.py
steven-murray: Np. Updated.

bhazelton (dismissed):
Looks great, thanks! I think it can be merged once tests pass.

bhazelton:
@steven-murray There's some missing code coverage, specifically for the new deprecation warning you added.

The pyuvsim tests are failing because of the new warning, which is ok, I can patch that up after we merge this.

bhazelton (approved):

bhazelton:
Actually, the pyuvsim error is fixed now that we've merged the pyuvsim PR that uses the BeamInterface objects.

PR #1484, opened 2024-10-10
benchmark against the compiled extension explicitly

mkolopanis:
changes the baseline/antnum related benchmarking to run explicitly against the compiled extensions. 
In the previous way we were basically just benchmarking some python overhead and the inter-op with the extension. 

We may want to add something like this to the ecef calculations too


bhazelton (approved):
Looks good, thanks!!

PR #1485, opened 2024-10-10
Fix a bug in the MWA beams where the polarization responses were swapped

bhazelton:
## Description

The azimuthal-aligned response was accidentally assigned to the zenith angle basis vector index and vice versa. Thanks to @nicelmh for pointing out the problem.

I also added tests that would have caught this error if they had been in place.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ telescope object from the metadata.
 antennas with data.
 
 ### Fixed
+- A bug in the MWA beam reader that resulted in the wrong polarization response
+(the azimuthal-aligned response was swapped with the zenith angle-aligned response).
 - A bug in reading UVH5 files with antenna names saved as variable length strings
 that was introduced in v3.0.0.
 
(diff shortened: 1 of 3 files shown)


bhazelton:
@nicholebarry @rlbyrne I wanted to make sure you both saw this so you were aware of it.

bhazelton:
> Seems like a simple fix to me. Is there any change Sigma_T and Sigma_P are computed backwards/incorrectly? Or are we confident it was just an indexing/assignment issue?

We found the problem because the response wasn't physical. After fixing it, it matches our expectations much better. The response for a single dipole (all others turned off) is fairly similar to our analytic short dipole beam and it passes the new tests that are based in physical reasoning (these tests fail on the old version).

mkolopanis (approved):
Good enough for me! thanks a bunch. I'll go ahead and approve the code changes now. I'm not sure if you had more you wanted others to do about physical response or testing the otuputs

bhazelton:
Nicel has checked this, seems to be working correctly.

PR #1491, opened 2024-10-24
Add comments and code updates to clarify that the real part of the xy autos is corrected

bhazelton:
## Description

Fix a bug @d3v-null noticed in the MWA correlator FITS file reader.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/src/pyuvdata/uvdata/mwa_corr_fits.py
+++ b/src/pyuvdata/uvdata/mwa_corr_fits.py
@@ -1141,6 +1141,7 @@ class MWACorrFITS(UVData):
                         sig2_arr=sig2,
                         cheby_approx=cheby_approx,
                     )
+                    self.data_array.real[k, :, j, yx] = kap
                     # correct imaginary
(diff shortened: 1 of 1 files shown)


bhazelton:
@PyxieLouStar can you look at this and make sure it's right?

d3v-null:
I think we determined that it looks like xy aren't being corrected when you read the code, but in fact they are. It's more of a code clarity issue

bhazelton:
ok. I think we should either do this fix or add a comment so that a future reader isn't confused. What would you prefer?

PyxieLouStar:
I think my preference would be to add the fix and also add a comment noting something like: these autos are corrected in-place and the data_array assignment is there in case the in-place correction changes/stops working.

bhazelton:
@PyxieLouStar I added the comments, can you take a look when you have a chance?

PyxieLouStar:
Looks great!

PR #1492, opened 2024-10-28
Set tolerances for floating point comparisons in tests

bhazelton:
## Description

This fixes a test that suddenly started failing because of a slight change in BLAS libraries leading to a floating point comparison to fail. It turns out that the test should have been failing all along but was slipping by just under the default `np.allclose` threshold. This new test failure blocked the conda release of v3.1.0.

While we were at it, we made updates throughout our testing code base to:
- use `np.testing.assert_allclose` as much as possible because it has tighter defaults and better informational prints when it fails
- explicitly set tolerances based on our standard parameter tolerances as much as possible.

This PR also updates the changelog to bump the version so we can get a new release onto conda.

## Types of changes

- [x] Version change

## Checklist:

Bug fix checklist:

Version change checklist:
- [x] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/main/CHANGELOG.md) to put all the unreleased changes under the new version (leaving the unreleased section empty).

### DIFF
--- a/tests/uvdata/test_fhd.py
+++ b/tests/uvdata/test_fhd.py
@@ -328,8 +328,9 @@ def test_read_fhd_latlonalt_match_xyz(fhd_data_files):
 
     mwa_tel = Telescope.from_known_telescopes("mwa")
 
+    # TODO: why don't these match better?
     np.testing.assert_allclose(
-        mwa_tel._location.xyz(), fhd_uv.telescope._location.xyz()
+        mwa_tel._location.xyz(), fhd_uv.telescope._location.xyz(), rtol=0, atol=1e-1
     )
 
 
--- a/tests/uvdata/test_mir_parser.py
+++ b/tests/uvdata/test_mir_parser.py
@@ -400,7 +400,9 @@ def test_compass_bp_apply(mir_data: MirParser, compass_soln_file, muck_solns):
         if mir_data.sp_data.get_value("corrchunk", header_key=key) != 0:
             # If muck_solns is not some, then all the values should agree with our
             # temp value above, otherwise none should
-            assert np.allclose(entry["data"], tempval * (1 + (muck_solns == "none")))
+            np.testing.assert_allclose(
+                entry["data"], tempval * (1 + (muck_solns == "none"))
+            )
             assert (muck_solns != "none") == np.all(entry["flags"])
 
 
@@ -766,7 +768,7 @@ def test_apply_tsys_missing_recs(mir_data):
         mir_copy.load_data(load_cross=True, apply_tsys=True)
 
     for key in mir_data.vis_data:
-        assert np.allclose(
+        np.testing.assert_allclose(
             mir_data.vis_data[key]["data"], mir_copy.vis_data[key]["data"]
         )
         assert all(mir_copy.vis_data[key]["flags"])
@@ -810,7 +812,7 @@ def test_apply_tsys(mir_data, use_cont_det, bad_vals):
     mir_copy.load_data(load_cross=True, apply_tsys=True)
 
     for key, norm_fac in zip(mir_data.vis_data.keys(), norm_list, strict=True):
-        assert np.allclose(
+        np.testing.assert_allclose(
             norm_fac * mir_data.vis_data[key]["data"], mir_copy.vis_data[key]["data"]
         )
         assert np.array_equal(
@@ -819,7 +821,7 @@ def test_apply_tsys(mir_data, use_cont_det, bad_vals):
 
     mir_copy.apply_tsys(invert=True)
     for key in mir_data.vis_data:
-        assert np.allclose(
+        np.testing.assert_allclose(
             mir_data.vis_data[key]["data"], mir_copy.vis_data[key]["data"]
         )
         assert np.array_equal(
@@ -1617,7 +1619,7 @@ def test_generate_chanshift_kernel(
     if kern is None:
         assert exp_kern == []
     else:
-        assert np.allclose(exp_kern, kern)
+        np.testing.assert_allclose(exp_kern, kern)
 
 
 @pytest.mark.parametrize("check_flags", [True, False])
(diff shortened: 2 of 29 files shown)


tests/uvdata/test_fhd.py
```diff
@@ -328,8 +328,9 @@ def test_read_fhd_latlonalt_match_xyz(fhd_data_files):
 
     mwa_tel = Telescope.from_known_telescopes("mwa")
 
+    # TODO: why don't these match better?
```
mkolopanis, line 331: I had the same question :laughing:

tests/uvdata/test_mir_parser.py
```diff
@@ -766,7 +768,7 @@ def test_apply_tsys_missing_recs(mir_data):
         mir_copy.load_data(load_cross=True, apply_tsys=True)
 
     for key in mir_data.vis_data:
-        assert np.allclose(
+        np.testing.assert_allclose(
             mir_data.vis_data[key]["data"], mir_copy.vis_data[key]["data"]
         )
```
mkolopanis, line 773: Do we have any tols acceptable for this check? All the other tests have explicit tols. or are these values integers? i suppose in that case we won't need it.

tests/uvdata/test_mwa_corr_fits.py
```diff
@@ -1050,7 +1052,7 @@ def test_van_vleck_int():
     # read in file corrected using integrate.quad with 1e-10 precision
     uv2 = UVData()
     uv2.read(filelist[10])
-    assert np.allclose(uv1.data_array, uv2.data_array)
+    assert uv1._data_array == uv2._data_array
```
mkolopanis, line 1055: Should we change this (and the other bare `assert ..==` calls below in this file) to use `np.testing.assert_allclose` too?

mkolopanis (commented):
I was going ask why most of our testing for distance we use a tolerance of 1mm and if that was the precision we claim to be sensitive to, but looking through the pyuvdata code it is consistent with tolerances there. I think that leads me to a more general question we should probably make as an issue: is that the tolerance we actually want to be sensitive to?

Beyond that hypothetical, I think I just have a few minor questions and then this will be good to go.

tests/uvdata/test_fhd.py
bhazelton: I suspect this is due to some mismatch in where these files came from, but at this point it's unknowable without remaking the files.

tests/uvdata/test_mir_parser.py
bhazelton: This is on a mir_data object so I defer to @kartographer

tests/uvdata/test_mwa_corr_fits.py
bhazelton: So this actually does use an allclose call, I'm calling it via the UVParameter `_data_array`. But I think I did this early in my changes and I'm happy to change it to an `np.testing.assert_allclose` call on the actual arrays instead.

tests/uvdata/test_mir_parser.py
kartographer: This one _should_ be able to be treated as a straight-up equality -- it's functionally a no-op under the hood because no calibration information (namely, system temperature information) can be found, hence the check below that all vis records are flagged.

tests/uvdata/test_mwa_corr_fits.py
mkolopanis: I think in general the == comparison is good for users, but for our testing, if this ever fails, the numpy utility will help us figure out why.

tests/uvdata/test_mir_parser.py
mkolopanis: So I get the sense it doesn't really matter this line then.

tests/uvdata/test_mwa_corr_fits.py
bhazelton: I'll make the change.

bhazelton: Fixed. I also realized that I had a couple changes in files that I hadn't saved, so they didn't end up in earlier commits, so I added them too.

tests/uvdata/test_mir_parser.py
kartographer: Not _this_ time :wink:

tests/uvdata/test_mwa_corr_fits.py
```diff
@@ -1074,7 +1081,7 @@ def test_van_vleck_cheby():
     good_ants = np.delete(np.unique(uv2.ant_1_array), 76)
     uv2.select(antenna_nums=good_ants)
 
-    assert np.allclose(uv1.data_array, uv2.data_array)
+    assert uv1._data_array == uv2._data_array
```
mkolopanis, line 1084: Sorry I didn't specifically ping this one but I think it falls under the same comment about using the numpy utility.

mkolopanis (commented):
Sorry there were two other spots I should have specifically called out about using the numpy testing utility

tests/uvdata/test_mwa_corr_fits.py
```diff
@@ -1337,4 +1344,4 @@ def test_van_vleck(benchmark, cheby):
         good_ants = np.delete(np.unique(uv2.ant_1_array), 76)
         uv2.select(antenna_nums=good_ants)
 
-    assert np.allclose(uv1.data_array, uv2.data_array)
+    assert uv1._data_array == uv2._data_array
```
mkolopanis, line 1347: this too

bhazelton: ok, I got that one and another one in that file plus a few more I found in other places. I swear I searched the whole repo before but I guess I used a slightly different search this time.

mkolopanis (approved):
Awesome thank you!

PR #1495, opened 2024-11-05
fix: set include_cross_pols to False if feed_array has length 1

steven-murray:
## Description

This simply fixes behavior when setting only a single feed in an AnalyticBeam. You can manually get around this on initialization by setting `include_cross_pols=False`, but you _can't_ do this when updating an existing object, i.e.

```python
from pyuvdata.analytic_beam import GaussianBeam
from dataclasses import replace

beam = GaussianBeam(diameter=14.0)
new_beam = replace(beam, feed_array=['x'])
```

This would fail due to the issue resolved here, and you _can't_ specify `include_cross_pols` in the `replace()` call, because it's an init-only parameter.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ All notable changes to this project will be documented in this file.
 - Made it possible to *not* return the `interp_basis_vector` array from beam
 interpolators.
 
+### Fixed
+- Initialization of analytic beams with single feed.
+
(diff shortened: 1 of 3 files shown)


bhazelton (approved):
Looks good.

PR #1496, opened 2024-11-08
feat: new convenience methods on BeamInterface

steven-murray:
## Description

This adds some convenience methods to the `BeamInterface` to make it easier to treat `AnalyticBeam` and `UVBeam` objects on equal footing. 

## Motivation and Context

While updating `matvis` to use the new interface, I came across a few examples of things that would be easier if they were implemented on the `BeamInterface` instead of having to handle them in downstream code. This just implements those changes.

In particular, it adds:

- Ability to pass a `BeamInterface` to instantiate a `BeamInterface`. 
- Properties `Nfeeds`, `feed_array`, `Npols` and `polarization_array` that just expose the same properties on the underlying beam
- A new `as_power_beam` method that converts the `BeamInterface` to a power beam consistently. 
- A new `with_feeds` method that consistently down-selects the feeds in the underlying beam regardless of its type.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,10 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- New convenience methods on `BeamInterface` to simplify the handling of analytic vs UVBeam objects.
 - Added support for partial read for MWA correlator FITS files.
 - Added `antenna_names`, `time_range`, `lsts` and `lst_range` parameters to
 `UVFlag.select` to match UVData and UVCal select methods.
-- New convenience methods on `BeamInterface` to simplify the handling of analytic vs
-UVBeam objects.
 
 ### Changed
 - Made it possible to *not* return the `interp_basis_vector` array from beam
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ All notable changes to this project will be documented in this file.
 - Added support for partial read for MWA correlator FITS files.
 - Added `antenna_names`, `time_range`, `lsts` and `lst_range` parameters to
 `UVFlag.select` to match UVData and UVCal select methods.
+- New convenience methods on `BeamInterface` to simplify the handling of analytic vs
+UVBeam objects.
 
 ### Changed
 - Made it possible to *not* return the `interp_basis_vector` array from beam
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,7 +113,6 @@ select = [
 ]
 ignore = [
     "N806", # non-lowercase variable (we use N* for axes lengths)
-    "N802", # non-lowercase function name (we use N* for axes lengths)
     "B028", # no-explicit-stacklevel for warnings
     "SIM108", # prefer ternary opperators. I find them difficult to read.
     "D203", # one-blank-line-before-class. we use two.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,6 +113,7 @@ select = [
 ]
 ignore = [
     "N806", # non-lowercase variable (we use N* for axes lengths)
+    "N802", # non-lowercase function name (we use N* for axes lengths)
     "B028", # no-explicit-stacklevel for warnings
     "SIM108", # prefer ternary opperators. I find them difficult to read.
     "D203", # one-blank-line-before-class. we use two.
--- a/src/pyuvdata/beam_interface.py
+++ b/src/pyuvdata/beam_interface.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 
 import copy
 import warnings
-from dataclasses import InitVar, dataclass
+from dataclasses import InitVar, asdict, dataclass, replace
 from typing import Literal
 
 import numpy as np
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ All notable changes to this project will be documented in this file.
 - Added support for partial read for MWA correlator FITS files.
 - Added `antenna_names`, `time_range`, `lsts` and `lst_range` parameters to
 `UVFlag.select` to match UVData and UVCal select methods.
+- New convenience methods on `BeamInterface` to simplify the handling of analytic vs
+UVBeam objects.
```
bhazelton, line 11: This should go at the top of this section, it should be reverse chronological order.

pyproject.toml
```diff
@@ -113,6 +113,7 @@ select = [
 ]
 ignore = [
     "N806", # non-lowercase variable (we use N* for axes lengths)
+    "N802", # non-lowercase function name (we use N* for axes lengths)
```
bhazelton, line 116: I've typically used either in-line `# noqa N802` or added something to the `[tool.ruff.lint.per-file-ignores]` section for this, mostly to limit the situations in which we ignore this rule.

src/pyuvdata/beam_interface.py
```diff
@@ -69,7 +77,7 @@ def __post_init__(self, include_cross_pols: bool):
         if isinstance(self.beam, UVBeam):
             if self.beam_type is None or self.beam_type == self.beam.beam_type:
                 self.beam_type = self.beam.beam_type
-            elif self.beam_type == "power":
+            elif self.beam_type == "power" and self.beam.beam_type != "power":
```
bhazelton, line 80: This clause isn't required because it's guaranteed to be True. The previous `if` statement already catches the case where the beam_types are the same. Maybe add a comment instead if it's not clear?

src/pyuvdata/beam_interface.py
```diff
@@ -82,6 +90,105 @@ def __post_init__(self, include_cross_pols: bool):
...
+        return replace(self, **kw)
+
+    def as_power_beam(
+        self, include_cross_pols: bool = True, allow_beam_mutation: bool = False
+    ):
+        """Return a new interface instance that is in the power-beam mode.
+
+        If already in the power-beam mode, this is a no-op. Note that this might be
+        slighty unexpected, because the effect of `include_cross_pols` is not accounted
+        for in this case.
```
bhazelton, line 127: I think we should issue a warning in that case. We can tell what was set before because if `include_cross_pols` was False when it was initialized, Npols is equal to Nfeeds. If it was True, Npols is > Nfeeds. I guess if there's only one feed, the value of this parameter doesn't matter and the warning shouldn't be raised.

src/pyuvdata/beam_interface.py
```diff
@@ -82,6 +90,105 @@ def __post_init__(self, include_cross_pols: bool):
...
+            # Down-select polarizations based on the feeds input.
+            possible_pols = [f1 + f2 for f1, f2 in product(feeds, feeds)]
+            possible_pol_ints = upol.polstr2num(
+                possible_pols, x_orientation=self.beam.x_orientation
+            )
+            if maintain_ordering:
+                use_pols = [
+                    p for p in self.beam.polarization_array if p in possible_pol_ints
+                ]
+                print("use_pols: ", use_pols)
```
bhazelton, line 179: should this print still be here?

bhazelton (commented):
Looks pretty good, I just have a few suggestions

src/pyuvdata/beam_interface.py
```diff
@@ -82,6 +90,121 @@ def __post_init__(self, include_cross_pols: bool):
...
+                # By default, keep the value of include_cross_pols the same.
+                include_cross_pols = self.Npols > 2
+
+            if self.Npols > 1 and (
+                (include_cross_pols and self.Npols != 4)
+                or (not include_cross_pols and self.Npols != 2)
+            ):
+                warnings.warn(
+                    "as_power_beam does not modify cross pols when the beam is"
+                    "already in power mode!"
```
bhazelton, line 149: I like this. I feel like it might be more helpful to tell the user what the cross_pol status of their existing object is vs what they asked for.

tests/test_beam_interface.py
```diff
@@ -243,3 +262,94 @@ def test_compute_response_errors(param, value):
...
+
+
+def test_as_power_noop(airy):
+    """Ensure that calling as_power_beam on a power beam is a no-op."""
+    intf = BeamInterface(airy, beam_type="power")
+    intf2 = intf.as_power_beam()
+    assert intf is intf2
+
+    with pytest.warns(UserWarning, match="as_power_beam does not modify cross pols"):
+        print(intf.Npols)
```
bhazelton, line 321: is this cruft?

bhazelton (commented):
looks good. I found one more print statement and I think the warning could be more helpful.

tests/test_beam_interface.py
steven-murray: Yes, removed

src/pyuvdata/beam_interface.py
steven-murray: Great idea, done!

PR #1503, opened 2024-11-23
Minor fixes to #1502

kartographer:
## Description

Fixes a couple of tests that broke in #1502.

## Motivation and Context

I'm not quite sure _how_ it managed to pass by (both locally on laptop but also in CI), but #1502 actually broke a couple of tests since ATA was introduced as a new known telescope. I only noticed this when attempting a rebase -- not sure what to make of the CI behavior, but that's to be addressed outside of this PR.

## Types of changes

## Checklist:

Bug fix checklist:

### DIFF
--- a/src/pyuvdata/telescopes.py
+++ b/src/pyuvdata/telescopes.py
@@ -95,7 +95,7 @@ _KNOWN_TELESCOPES = {
             height=1019.222 * units.m,
         ),
         "antenna_diameters": 6.1,
-        "citation": "Private communication (2024)",
+        "citation": "Private communication (D. DeBoer to G. Keating; 2024)",
     },
 }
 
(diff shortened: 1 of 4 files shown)


src/pyuvdata/telescopes.py
```diff
@@ -95,6 +95,7 @@
             height=1019.222 * units.m,
         ),
         "antenna_diameters": 6.1,
+        "citation": "Private communication (2024)",
```
bhazelton, line 98: can we get at least a name here so we know who provided the info?

bhazelton (commented):
Looks reasonable. I'm confused about why the CI without this change manages to get the astropy site aliases as well as the site names when we don't see that behavior on our own laptops, but I've run out of ideas to run that down, so I think we should just get this in.

src/pyuvdata/telescopes.py
kartographer: Done

bhazelton (approved):

PR #1506, opened 2024-12-03
+^ look_in_catalog avoid unrelated keys for cat_type

radonnachie:
## Description
The `look_in_catalog` function never considered the cat_type when accessing keys, leading to keys unrelated to the cat_type being hard-accessed.

## Motivation and Context

When adding multiple UV objects, the `look_in_catalog` hit an error trying to access the "cat_times" key within the phase_catalog dict-entry. The entries are all of "sidereal" type though, which isn't expected to have a "cat_times" key. 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

New feature checklist:

Breaking change checklist:

Version change checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 - Bug in selecting baselines on a UVData object using `bls` keyword with 3-tuples and
 more than one polarization (introduced in 3.1.2).
+- Bug in the `look_in_catalog` utility function that expected keys which could be
+optional for some phase-center catalog-entry types. This was noticed as an issue when
+adding two uvh5 objects that were of type "sidereal" which did not have "cat_times"
+entries in their phase-center catalogs.
 
 ## [3.1.2] - 2024-11-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,12 @@ All notable changes to this project will be documented in this file.
 - ATA has been added to the list of known telescopes.
 
 ### Fixed
-- Bug in selecting baselines on a UVData object using `bls` keyword with 3-tuples and
-more than one polarization (introduced in 3.1.2).
 - Bug in the `look_in_catalog` utility function that expected keys which could be
 optional for some phase-center catalog-entry types. This was noticed as an issue when
 adding two uvh5 objects that were of type "sidereal" which did not have "cat_times"
 entries in their phase-center catalogs.
+- Bug in selecting baselines on a UVData object using `bls` keyword with 3-tuples and
+more than one polarization (introduced in 3.1.2).
 
 ## [3.1.2] - 2024-11-21
 
--- a/src/pyuvdata/utils/phase_center_catalog.py
+++ b/src/pyuvdata/utils/phase_center_catalog.py
@@ -161,6 +161,25 @@ def look_in_catalog(
         "cat_vrad": default_tols,
     }
 
+    typenonfields_dict = {
+        "sidereal": [
+            "cat_times",
+        ],
+        "ephem": [
+            "cat_pm_ra",
+            "cat_pm_dec",
+        ],
+        "unprojected": [
+            "cat_frame",
+            "cat_pm_ra",
+            "cat_pm_dec",
+            "cat_times",
+            "cat_dist",
+            "cat_vrad",
+        ],
+    }
+    typenonfields_dict["driftscan"] = typenonfields_dict["unprojected"]
+
     if target_cat_id is not None:
         if target_cat_id not in phase_center_catalog:
             raise ValueError(f"No phase center with ID number {target_cat_id}.")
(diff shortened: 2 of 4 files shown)


radonnachie:
Obviously need to run through the bug-fix checklist. Left it up to the reviewer to consider the style of the commit.

src/pyuvdata/utils/phase_center_catalog.py
```diff
@@ -196,7 +215,13 @@ def look_in_catalog(
...
                         )
             else:
-                cat_diffs += check_dict[key] is not None
+                check_type = check_dict.get("cat_type")
+                check_value = None
+                # if type is known, avoid unrelated fields
+                if (check_type is None) or (key not in typenonfields_dict[check_type]):
+                    check_value = check_dict[key]
+                cat_diffs += check_value is not None
+
```
kartographer, line 224: I think you can avoid basically all of the additional code by replacing the _old_ L199 with the following:

```suggestion
                cat_diffs += check_dict.get(key, None) is not None
```

kartographer (changes requested):
Hi @radonnachie - thanks for catching this, nominally all entries inside of the phase center dictionaries are _supposed_ to be present, although it's possible that's not actually documented in the appropriate places. Long-term there's a plan to move the catalog to an object, but for now we're trying to minimize changes to the existing stuff to make that transition easier. I think there's a much easier way to wrap the problem you've nominally come across (a 1-line replacement) which I've added below. Note that we normally also expect bug-fix PRs to include a test which fails with the older code, let me know if you need help putting that together.

wfarah:
@kartographer for the test, how should I include the data? I can include the traceback in a second

radonnachie:
I'll just setup a test with simple dicts that emulate the scenario.

radonnachie:
I'm thinking that instead of testing that `look_in_catalog` handles the ps_catalogs correctly that the test should directly add 2 sidereal UVData objects... would any of the existing data files support this test @kartographer?

radonnachie:
I tried adding a test. Not sure how robust it is, hopefully serves as an indication of what we're struggling with.

kartographer:
@radonnachie -- it looks like the test is actually failing for unrelated reasons here, but let me see if I can spin up a simple fix...

tests/uvdata/test_uvh5.py
```diff
@@ -414,6 +414,27 @@ def test_uvh5_compression_options(casa_uvfits, tmp_path):
...
+    casa_uvfits.write_uvh5(testfile, clobber=True)
+    uv = UVData()
+    uv.read(testfile)
+    uv + uv.copy()
+
+    os.remove(testfile)
+
+    return
+
+
```
kartographer, line 437:
```suggestion
@pytest.mark.filterwarnings("ignore:The uvw_array does not match the expected values")
def test_look_in_catalog_missing_entries(casa_uvfits, tmp_path):
    phase_cat = casa_uvfits.phase_center_catalog

    # Try that this works normally if we do nothing
    assert look_in_catalog(phase_cat, cat_name=phase_cat[0]['cat_name']) == (0, 5)
    
    # Now delete some keys
    for value in phase_cat.values():
        if 'cat_times' in value:
            del value['cat_times']

    # Now re-run the above and verify things work as expected
    assert look_in_catalog(phase_cat, cat_name=phase_cat[0]['cat_name']) == (0, 5)

```

kartographer (changes requested):
Okay, this is a much more focused test that I've at least verified in an ipython shell fails with the old code and works with the new code.

kartographer:
> @kartographer for the test, how should I include the data? I can include the traceback in a second

@wfarah - just to clarify here, we usually just want to see a test that should fail in the old code. We don't necessarily need to see a traceback (though it is helpful sometimes!), but usually during review whoever is looking at the PR will at least make sure that the test appears to cover the nominal scenario.

tests/uvdata/test_uvh5.py
bhazelton: This is really not specific to uvh5 files, so I'd advocate putting this in `tests/utils/test_phase_center_catalog.py` or possibly `test_uvdata.py` rather than `test_uvh5.py`

radonnachie: Righto, I actually started with something (less neat) in the `test_phase_center_catalog`. I thought though that what I really wanted to test was the actual uvh5 addition operation... but let's go with innocent until proven guilty.

bhazelton:
Looks like it's failing the pre-commit linting check. (The warning test failure and hera_cal failures are unrelated known issues and are not required to pass). There are directions in the readme on install the pre-commit hook to prevent pushing code that will fail the CI.

kartographer (changes requested):
Looking good -- just one thing missing: an entry in the change log =)

radonnachie:
Right, I've added a somewhat verbose entry to the changelog. Trust it's good enough...

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 - Bug in selecting baselines on a UVData object using `bls` keyword with 3-tuples and
 more than one polarization (introduced in 3.1.2).
+- Bug in the `look_in_catalog` utility function that expected keys which could be
```
kartographer, line 12: This should actually go at the top of the stack (entries should be in reverse chronological order).

kartographer (changes requested):
One more minor fix!

CHANGELOG.md
radonnachie: oh, 2 seconds

kartographer (approved):
Done -- thanks @radonnachie!

radonnachie:
Thank you for your help and patience. 
It's not fully solve our issue, so I'll be back before long with another PR. I'll aim to make it a bit more seamless.

kartographer:
@radonnachie - can you put in an issue so that we can follow along (and potentially help)?

Issue #154, opened 2017-05-10
float64 for uvws


adampbeardsley:
While testing the time_profiling branch, I noticed that the uvws being stored as float32 can introduce non-acceptable errors in the data when phasing/unphasing. I think we should look into always having the uvws in float64.

bhazelton:
May not really be an issue, removing for now.

PR #1508, opened 2024-12-05
Add option to write uvw_array as double precision in UVFITS when data_array is single precision

bhazelton:
## Motivation and Context

This adds the option write the uvw_array as double precision in UVFITS even when the data_array is single precision (they are always written as double precision when the data array are double precision). This is done using the same mechanism as is used for the time_array in UVFITS files, namely creating two identically named group parameters that are intended to be added together. This happens automatically in astropy's FITS reader, but it is possible this could mess up other UVFITS readers, so I added an option to allow it to be turned on or off and defaulted it to on because I think most users would prefer not to lose precision on their uvws.

This problem was noted as far back as #154 and it has come up occasionally since then.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/tests/uvdata/test_uvfits.py
+++ b/tests/uvdata/test_uvfits.py
@@ -170,9 +170,9 @@ def test_group_param_precision(tmp_path, uvw_double):
     )
 
     if uvw_double:
-        np.testing.assert_allclose(uvd.uvw_array, uvd2.uvw_array, rtol=0, atol=1e-13)
+        np.testing.assert_allclose(uvd.uvw_array, uvd2.uvw_array, rtol=1e-13)
     else:
-        assert not np.allclose(uvd.uvw_array, uvd2.uvw_array, rtol=0, atol=1e-13)
+        np.testing.assert_allclose(uvd.uvw_array, uvd2.uvw_array, rtol=1e-7)
 
     # The incoming ra is specified as negative, it gets 2pi added to it in the roundtrip
     uvd2.phase_center_catalog[1]["cat_lon"] -= 2 * np.pi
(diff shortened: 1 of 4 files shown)


kartographer (changes requested):
Overall I think this is looking good -- just had a small comment regarding the test for the new feature.

tests/uvdata/test_uvfits.py
```diff
@@ -168,6 +169,11 @@ def test_time_precision(tmp_path):
         atol=uvd2._lst_array.tols[1],
     )
 
+    if uvw_double:
+        np.testing.assert_allclose(uvd.uvw_array, uvd2.uvw_array, rtol=0, atol=1e-13)
+    else:
+        assert not np.allclose(uvd.uvw_array, uvd2.uvw_array, rtol=0, atol=1e-13)
```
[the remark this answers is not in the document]
bhazelton: done.

tests/uvdata/test_uvfits.py
```diff
@@ -168,6 +169,11 @@ def test_time_precision(tmp_path):
         atol=uvd2._lst_array.tols[1],
     )
 
+    if uvw_double:
+        np.testing.assert_allclose(uvd.uvw_array, uvd2.uvw_array, rtol=1e-13)
+    else:
+        np.testing.assert_allclose(uvd.uvw_array, uvd2.uvw_array, rtol=1e-7)
```
kartographer, line 175: @bhazelton -- I think actually holding on to both checks here would be good (both that it fails if `rtol=1e-13` and passes when `rtol=1e-7`), since that's a pretty strong indicator that double-precision _isn't_ being used. Or otherwise, adding a check that the UU2/VV2/WW2 entries aren't in there (although I'm not sure if there's a straight-forward way to do that w/ the astropy FITS interface).

bhazelton: ok, added that one back.

kartographer (approved):
Looks good, thanks @bhazelton!

PR #1521, opened 2025-01-25
Adding new Telescope attributes for feeds and mounts

### LINKED ISSUE #854: Generalize x_orientation concept

### LINKED ISSUE #1235: Top-level documentation page "Conventions"
pyuvdata by necessity uses many conventions, and these are documented throughout the code API pages and tutorials. However, it'd be good to gather all these into one reference place.

### LINKED ISSUE #1425: Adding polarization leakage handling for MS_Cal
Once #1391 is merged, there is one outstanding gains table type left to build support for, which is the polarization leakage terms (in CASA parlance, "D Jones"). I wasn't able to readily find example files for this, which is why support for it was not included in the original MS Cal PR, but it would be nice to add support for it.

### LINKED ISSUE #1460: Telescope.from_known_telescopes breaks if the known telescope doesn't have enough info
Currently there's no nice way to supply extra info when calling `Telescope.from_known_telescopes()`. For example, if I do `Telescope.from_known_telescopes('PAPER')`, it errors, because `PAPER` doesn't have the antenna position info. The easiest way to solve this would be to allow a **kwargs that can set the extra attributes.

### LINKED ISSUE #1470: Add support for feed position angles
Currently, the `UVData` object does not support arbitrary feed polarization angles. This would be useful, and UVFITS/MS files already support this (see below). My personal motivation is that the upcoming SKA-Low stations are all rotated at different angles for "polarization diversity" and to improve sidelobe response between antenna pairs. 

Currently, when writing to UVFITS the [polarization angle keywords are hardcoded](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/e2b73b4bfe047e0da889e17196ce686f3a2c425f/src/pyuvdata/uvdata/uvfits.py#L1290). I can't see anything similar for MS, which uses the `RECEPTOR_ANGLE` keyword -- presumably casacore fills these with default values.  

To make this possible, I think the following would have to happen:

1. An attribute is added to `UVData`, e.g. `UVData.feed_polarization_angle` and `__init__` is updated accordingly (with (90, 0) degrees as default values).
2. All writers are updated to use these values. 
3. All readers are updated to read the relevant keywords when creating a `UVData` object.
4. Tests are written (roundtrip conversion, e.g. UVData -> UVFITS -> UVData would exercise both read and write).
5. Documentation is updated.

I don't have enough free cycles to take all this on, but I have some example code for UVFITS + MS below.

## UVFITS keywords

From [AIPS Memo 117](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://library.nrao.edu/public/memos/aips/memos/AIPSM_117.pdf&ved=2ahUKEwjo49DZt8GHAxVc1DQHHU5qK1gQFnoECBkQAQ&usg=AOvVaw20kSnIm25f6jj2-xvR-KWS).

In UVFITS, AIPS AN antenna table, the columns `POLAA` and `POLAB` should be updated:

```
POLAA  1E  degrees  Position angle feed A
POLAB  1E  degrees  Position angle feed B
```

> The value of the `POLAA` column shall be the orientation of feed A, assumed independent
> of IF, given in degrees. Similarly, the `POLAB` column shall contain the feed orientation for feed B.

## MS keywords

Need to update `RECEPTOR_ANGLE` column of the FEED table of a measurement set:
https://safe.nrao.edu/wiki/bin/view/ALMA/ReceptorAngles

These angles are stored in radians. If there are multiple receiver bands in a measurement set, then multiple angles will be shown.

## Miriad

No idea what keywords would need to change.

## Proto-code

```python

def write_ms(uv: UVData, filename: str, *args, **kwargs):
    """Write UVData to MeasurementSet.

    Notes:
        Calls uv.write_ms(), then applies station rotation patch.
    Args:
        uv (UVData): pyuvdata object to write to file.
        filename (str): Name of output filename.
        args (list): Arguments to pass to uv.write_ms
        kwargs (dict): Keyword arguments to pass to uv.write_ms
    """
    tables = import_optional_dependency('casacore.tables', errors='raise')

    uv.write_ms(filename, *args, **kwargs)

    # Patch RECEPTOR_ANGLE column
    with tables.table(f'{filename}/FEED', readonly=False) as t:
        logger.debug('Applying station rotation (RECEPTOR_ANGLE)')
        r_ang = np.zeros(shape=t.getcol('RECEPTOR_ANGLE').shape, dtype='float64')
        x_ang = -np.pi/180 * uv.receptor_angle
        r_ang[:, 0] = x_ang
        r_ang[:, 1] = x_ang + np.pi

        t.putcol('RECEPTOR_ANGLE', r_ang)


def write_uvfits(uv: UVData, filename: str, *args, **kwargs):
    """Write UVData to UVFITS.

    Notes:
        Calls uv.write_uvfits(), then applies station rotation patch.
    Args:
        uv (UVData): pyuvdata object to write to file.
        filename (str): Name of output filename.
        args (list): Arguments to pass to uv.write_uvfits
        kwargs (dict): Keyword arguments to pass to uv.write_uvfits
    """
    uv.write_uvfits(filename, *args, **kwargs)

    # Patch POLAA/POLAB columns
    with pf.open(filename, mode='update') as hdu:
        logger.debug('Applying station rotation (POLAA/POLAB)')
        hdu[1].data['POLAA'] = uv.receptor_angle
        hdu[1].data['POLAB'] = uv.receptor_angle + 90
```

### LINKED ISSUE #1486: Coordinate of antenna position in UVFITS?
I'm converting some visibility from custom format to UVFITS/MeasurementSets and importing it to CASA. When using UVFITS as intermediate format (`write_uvfits`) and importing it using CASA task `importuvfits`, antenna positions are not expected:

```
Station   Offset from array center (m)       
             East         North     Elevation
E01       -0.0000        0.0000        0.0000
E02        1.5671      -26.6355       29.6190
E04        6.9827     -107.0955      118.4858
E05       10.6311     -160.7576      177.7517
E06       17.6700     -280.8892      311.7160
E07       19.0035     -294.5801      326.2694
E08       23.9970     -361.8501      400.0746
E09       24.4710     -375.0060      415.1973
E10       24.8243     -388.0366      430.3301
```

When using MS (`write_ms`), results are expected:

```
Station Offset from array center (m)       
           East         North     Elevation
E01      0.0000        0.0000        0.0000
E02     39.8547       -0.0939        0.8886
E04    159.8423       -0.2025        2.7755
E05    239.8672       -0.2537        4.0060
E06    419.8971       -0.5340        8.2220
E07    439.9214       -0.5419        7.9621
E08    539.9059       -0.5591        8.9570
E09    559.9360       -0.5795        9.8587
E10    579.8818       -0.6064       10.9127
```

---

After some investigation I think it is related to 
https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/d6e94366758480a31a062fa84632855701aeb521/pyuvdata/uvdata/uvfits.py#L810-L817

But it seems that in casacore, array XYZ and antenna XYZ are added without rotation (if not VLA):
https://github.com/casacore/casacore/blob/bf9e411892b3816724eb11737cc32d12f86a6436/msfits/MSFits/MSFitsInput.cc#L2228

Also found a PR related to "modifications to uvfits AN read/write rules": https://github.com/casacore/casacore/pull/400

So is that rotation required/expected? What's best practice to set antenna locations if I want to convert a custom format into UVFITS?

### LINKED ISSUE #1537: Possible unintended behavior with tols in `UVParameter` when working with integers
While working on a few things related to #1521, I "rediscovered" that the default setting for `tols` when a new `UVParameter` is instantiated, the object defaults to a relative tolerance of 1e-5, and an absolute tolerance of 1e-8, which was presumably inherited from the nominal settings of `np.isclose`. This seems fine for when `UVParameter.expected_type=float`, but I don't think this is a good default for when `UVParameter.expected_type=int`. E.g.:

```
In [115]: uvd = UVData.from_file(os.path.join(DATA_PATH, "1061316296.uvfits"))
In [116]: uvd2 = uvd.copy()
In [117]: uvd2.baseline_array[:] = 100000
In [118]: uvd.baseline_array[:] = 100001
In [119]: uvd == uvd2
Out[119]: True
In [120]: np.all(uvd.baseline_array == uvd2.baseline_array)
Out[120]: np.False_
```

Assuming others agree that the `UVParameter.__eq__` call should return `False`, I think there are two nominal paths forward:

1. Make the defaults for tols (1e-5, 1e-8) for when `expected_type=float`, and otherwise (0, 0). This can of course be overwritten by the user/etc. if there's a `UVParameter` with integer values that _should_ have non-zero tols (which I expect is the exception and not generally the rule).
2. Go through the various `UVBase` subclasses and change tols to be (0, 0) on anything that needs it.

The latter is more work (and maybe more prone to error), but doesn't cause any downstream changes in behavior, although I'm thinking the former is maybe the more preferable way to go.

### LINKED ISSUE #1540: Bug in baseline conjugation handling in `UVData.write_ms`
This issue is a re-spin of something originally reported by @rlbyrne in #1355, but is a little different in presentation, and after some digging I have a much better idea of what's going on, so I'm filing this as a new issue so that I can remember what it was I stumbled on later down the road.

As part of external testing of #1521, I have been playing around with polarization data handling, in particular using the CASA task `polcal` to try and derive d-terms/polarization leakages for the SMA antennas. The leakages (and linear polarization of the source) wasn't _awful_, but it was not consistent with what we were deriving using a "tried-but-true" method through a local compile of MIRIAD. After trying a few different things, I was finally able to get consistent results by setting `flip_conj=True` in `UVData.write_ms`, which flips the baseline conjugation scheme from "ant2-ant1" to "ant1-ant2". 

This surprised me a bit, in that I thought we had finally pinned down what CASA was doing with respect to baseline conjugation schemes. After a bit of testing and diving deep into the bowels of `casatasks`, I think I _finally_ understand what is happening, and why the documentation appears inconsistent. CASA supports *both* the "ant2-ant1" and "ant1-ant2" conjugation schemes, but it uses the antenna numbering scheme to determine which one is active. When ant1 < ant2, the "ant1-ant2" convention is assumed, whereas when ant1 > ant2, the "ant2-ant1" convention is assumed. Or rather, it's the antenna numbers themselves (rather than their placement in either the `"ANTENNA1"` and `"ANTENNA2"` columns) that determines the conjugation.

I haven't fully traced down the underlying code in the main CASA repository, but I can see that there is some code for reversing the conjugation scheme at will, which matches the observed behavior that I have seen here. Fortunately this should not affect imaging, just calibration tasks within CASA (which mostly impacts SMA, although potentially not the only one).

kartographer:
Adds feed orientation and mount information to Telescope class, deprecates `x_orientation`.

## Description

A few new attributes have been added to the Telescope class: `Nfeeds`, which describes the number of polarizations captured by each antenna; `feed_array`, which records the polarizations for each antenna (one of 'x', 'y', 'l', 'r');`feed_angle`, which records the feed position angle (where 0 is aligned to vertical polarization); and `mount_type` which records the optics of the antenna.  Several related changes have been made here, which are discussed below.

~Note: This is a late-stage WIP, but I'm putting in a PR now so that interested parties can review while we conduct some additional testing. As per usual, me updating the CHANGELOG will denote when I think everything is wrapped up.~ At this point I'm calling this no longer WIP but ready to merge upon approval (RMUA?).

## Motivation and Context

SMA is in the process of adding polarimetric support in CASA, and adding the new attributes to the `Telescope` object are needed to facilitate that (and more generally, for accurately translating polarization on-sky to polarization as seen by an arbitrary detector/receiver). To facilitate this, several related changes have been made within this PR:

- `feed_array` and `feed_angle` are superseding the `x_orientation` attribute, and now must be set when `x_orientation` was previously required. Since it is integral to the interpretation of the feed polarization response, `mount_type` is also required when `feed_array` and `feed_angle` are set.
- Polarization ("D Jones") and non-polarization ("T Jones") handling for MSCal has been added/expanded.
- `Telescope` generator methods that populate the object based on "known telescopes" now allow for the specification of individual parameters that can be used to fill in missing information or otherwise override the `pyuvdata` defaults.
- `Telescope` now contains the `__add__` and `__iadd__` methods, to better facilitate combining other `UVBase` objects which contain a `Telescope` object as a parameter.
- Some new setter/getter methods have been added to `UVParameter` which allow one to select based on a particular "named" axis in `UVParameter.form` (e.g., "Nants", "Nblts", "Nfeed"), which were introduce to help facilitate the new aforementioned addition methods on the `Telescope` object.
-  A new "Conventions" page has been added to the docs, to help describe what these new parameters mean (and document other conventions found within `pyuvdata`)

Some other bug fixes have  

Closes #1470
Closes #854
Closes #1235 
Closes #1425 
Closes #1486
Closes #1537 
Closes #1540 
Closes #1460

Open items:
- [x] Figure out where documentation for `mount_type` needs to live.
- [x] Confirm that deprecation handling is suitable for downstream processes.
- [x] Extra credit: get d-terms working in CASA.

## Types of changes

## Checklist:

Breaking change checklist:


src/pyuvdata/telescopes.py
```diff
@@ -426,6 +429,61 @@ def __init__(self):
             tols=1e-3,  # 1 mm
         )
 
+        desc = (
+            'Antenna mount type, shape (Nants,). Options are: "alt-az", "equitorial", '
+            '"orbiting", "x-y", "alt-az+nasmyth-r", "alt-az+nasmyth-l", and "phased".'
```
bhazelton, line 434: I don't know what all of these mean. Is there a good reference? Is an LWA dipole "x-y"? Is MWA "phased"? What would HERA be?

src/pyuvdata/telescopes.py
```diff
@@ -426,6 +429,61 @@ def __init__(self):
...
+            description=desc,
+            required=False,
+            expected_type=str,
+            form=("Nants", "Nfeeds"),
+            acceptable_vals=["x", "y", "n", "e", "r", "l"],
+        )
+
+        desc = (
+            "Position angle of a given feed, shape (Nants, Nfeeds), units of radians. "
+            "Optional parameter, required if feed_array is set."
```
bhazelton, line 476: What angle is this? What is it referenced to? And how does it interact with the feed_array above?

src/pyuvdata/utils/io/hdf5.py
```diff
@@ -679,6 +679,20 @@ def antenna_names(self) -> list[str]:
...
+        return np.char.decode(
+            self.header["mount_type"][:].astype("|S"), encoding="utf8"
+        )
+
+    @cached_property
+    def feed_array(self) -> np.ndarray[str]:
+        """The antenna names in the file."""
+        return np.char.decode(
+            self.header["feed_array"][:].astype("|S"), encoding="utf8"
+        ).astype(np.object_)
```
bhazelton, line 694: should feed_angle be here too?

src/pyuvdata/utils/io/ms.py
```diff
@@ -1358,17 +1374,58 @@ def read_ms_feed(filepath):
...
     """
     _ms_utils_call_checks(filepath + "/FEED")
+    filepath += "::FEED"
 
-    raise NotImplementedError("Reading of MS FEED tables not available yet.")
+    with tables.table(filepath, ack=False) as feed_table:
+        if "pyuvdata_has_feed" in feed_table.getkeywords() and not (
+            feed_table.getkeyword("pyuvdata_has_feed")
+        ):
+            feed_array = feed_angle = Nfeeds = antenna_numbers = None
```
bhazelton, line 1383: What is this handling? Does this mean that this file was written by pyuvdata?

bhazelton (commented):
This is looking interesting, I have some questions.

src/pyuvdata/telescopes.py
kartographer: There's not a good singular reference (even [Wikipedia](https://en.wikipedia.org/wiki/Telescope_mount) seems to only cover the first two in the main article on telescope mounts), but to cover the rest here:

- [Nasmyth](https://en.wikipedia.org/wiki/Nasmyth_telescope) mounts (which the SMA is) have a mirror in the path that directs the light such that the detector can point at a constant elevation (rather than being rotated as the telescope moves in elevation). Good for big instruments and thing w/ cryogenics...
- Orbiting, I assume, orbits. Reading through the [UVFITS memo](https://library.nrao.edu/public/memos/aips/memos/AIPSM_117.pdf), looks like this dates back to [VSOP](https://en.wikipedia.org/wiki/HALCA), the very first space-based VLBI mission. There's history in the metadata table options, apparently.
- [X-Y](http://www.antesky.com/features-of-x-band-x-y-type-pedestal-ground-station/) is kind of a funky setup, easiest way I could describe it is 'imagine if for an alt-az mount, you rotated it so that "north" pointed toward the ground, and you could point toward negative elevations". The mount is able to track objects through zenith much better than a typical alt-az mount,  which makes it more popular in the LEO community.
- "Phased" is, strictly speaking, what MWA is, where a single antenna is the composite of multiple electrical elements. 

LWA and HERA are weird, in that they actually don't have mounts, on account of no motion. I _suspect_ that the most correct option of the above is probably phased, in that phased assumes no physical motion of the optics, although that may be a somewhat software-package-dependent thing.

src/pyuvdata/telescopes.py
kartographer: Conventions differ, but the one I've seen most commonly used (and maybe the one we should adopt) is the angle with respect to the great circle at the present elevation, similar to how paralactic angle is defined. That is at least how I've presently set some defaults.

src/pyuvdata/utils/io/hdf5.py
kartographer: No -- this only needs to be here b/c of string handling.

src/pyuvdata/utils/io/ms.py
kartographer: Yes -- written by pyuvdata, but _before_ we necessarily had correct handling of the feed information. I don't know that this is the correct way to handle this, but I did at least want to catch when this information might be at least suspect.

src/pyuvdata/telescopes.py
bhazelton: Fascinating. We need to write this up in our docs.

src/pyuvdata/telescopes.py
bhazelton: Ok, so how does this interact with `x_orientation`?

bhazelton: The current plan for SKA low has stations made of arrays of dipoles (so are phased feeds) where the orientations of the dipoles are the same within a station but are different station to station. Would this parameter handle that kind of metadata?

kartographer: So, I'm two minds about this, but my current preference is that it should be defined separate from `x_orientation`, and potentially supersede it (i.e., since if I know `x_orientation`, I can calculate everything that goes into `feed_angle`). It is, in principle, the "more general" solution to what I think `x_orientation` was made to fix, a la #854. Given that both parameters in principle describe one underlying aspect of the instrument, my preference in the long run would be for there to be only one parameter which describes it. But knowing that this is a bit of a paradigm shift, I expect it's something we'll need to discuss.

The good news is that, if we decide to go the route of preserving`x_orientation`, it isn't too much work to implement here.

bhazelton: Ok, let's talk about this in our next telecon. I need to understand it more and we'll definitely need deprecation and file backwards compatibility if it supersedes `x_orientation`.

src/pyuvdata/telescopes.py
bhazelton: Maybe we can add a "fixed" for LWA & HERA?

Also, ASKAP is an alt-az + spinning the dish + feed. how would we represent that?

kartographer: We can add "fixed", although it won't have a corresponding match in UVFITS/CASA. I don't know what impact that would have, but we can warn about it. Same thing for whatever we call ASKAP (although it's _possible_ from a software package perspective the right thing there is "equatorial", but I'd need to know more about it before I could say it for certain).

kartographer: Noting this as another good document to point to in the docs: https://arxiv.org/abs/2210.13381

kartographer:
Noting here that we've had our first successful external test using CASA's `polcal` to derive polarimetry for a pol calibrator w/ SMA data. Also noting that CASA sends a spooky message about unknown feeds because the "antenna 0" entries are blank -- probably worth either fixing or noting somewhere in the docs...

src/pyuvdata/telescopes.py
kartographer: Also thinking about this further post-telecon discussion -- I think in addition to deprecating `x_orientation`, it probably _also_ makes sense to have `feed_array` only support the options x/y/r/l, effectively dropping the n/e as allowed options (though things like `select` can and will still support supplying those options, without plans to deprecate). My reasoning here: n/e is just a special case of x/y, BUT without a clear mapping back to `polarization_array` or `jones_array`. Because `feed_array` is a new attribute, there's no backwards compatibility to worry about here as long as various methods can handle "n" and "e" as options.

One small issue is that this will cause `UVBeam` and `Telescope` to be slightly different in what's permitted, but in my head I can justify why that should be (tying back to fixed vs relative offsets) and so I don't think that's a problem. But again, we can discuss more at length at the next telecon.

kartographer: Alright, I've added the explanation -- I'm gonna resolve this comment for now, left open an item to talk about the path for deprecating `x_orientation`.

src/pyuvdata/telescopes.py
kartographer: One question for closing this out @bhazelton -- where do you want this to live? There's no Telescope tutorial per se, and I don't know if we want to throw it on to the docstring for the parameter. UVH5/CalH5 memo?

kartographer:
Alright, I'm moving this out of draft since major work here is completed. Still need to do some documentation work, which I could use some guidance on _where_ to put that, but otherwise I'm at a point where I'd like to make sure that folks are on board with what is effectively the proposed deprecation of `x_orientation`.

src/pyuvdata/telescopes.py
bhazelton: We have an open issue to add a top-level doc page on conventions used throughout pyuvdata: https://github.com/RadioAstronomySoftwareGroup/pyuvdata/issues/1235. Maybe this could go there? We could start the doc in this PR and add more later to close the issue?

src/pyuvdata/telescopes.py
```diff
@@ -400,30 +402,79 @@ def __init__(self):
...
             required=False,
+            form=("Nants",),
+            expected_type=float,
+            tols=1e-3,  # 1 mm
+        )
+
+        desc = (
+            'Antenna mount type, shape (Nants,). Options are: "alt-az", "equitorial", '
+            '"orbiting", "x-y", "alt-az+nasmyth-r", "alt-az+nasmyth-l", "phased", '
+            '"fixed", and "other".'
```
bhazelton, line 420: If we're not going to say more than this I think we need to point to the documentation you're planning to add on what these mean. I find these names confusing and even misleading (especially "x-y") so it's important to tell users where to get more info..

src/pyuvdata/telescopes.py
```diff
@@ -400,30 +402,79 @@ def __init__(self):
...
+        )
+
+        desc = (
+            "Position angle of a given feed, shape (Nants, Nfeeds), units of radians. "
+            "The feed angle is defined with respect to the great circle going through "
+            "constant azimuth and zenith (for pointed telescopes) or north (for fixed "
+            "telescopes like HERA or MWA), where a feed angle of 0 would correspond to "
+            "vertical polarization for linear feeds, and pi/2 to horizontal "
+            "polarization (as seen by the observer with reference to a horizontal "
+            "coordinate system). Optional parameter, required if feed_array is set."
```
bhazelton, line 469:
```
            "where a feed angle of 0 would correspond to "
            "vertical polarization for linear feeds, and pi/2 to horizontal "
            "polarization (as seen by the observer with reference to a horizontal "
            "coordinate system)
```

I find this very confusing. I don't consider either MWA dipole to have a vertical polarization. I wasn't sure what you meant by a horizontal coordinate system, Wikipedia say that's equivalent to alt/az coordinates. In that frame, both North and East aligned dipoles would be horizontal, not vertical...

src/pyuvdata/utils/pol.py
```diff
@@ -812,3 +812,181 @@ def _check_jones_spacing(*, jones_array, strict=True, allow_resort=False):
             "make it impossible to write this data out to calfits."
         )
         tools._strict_raise(err_msg=err_msg, strict=strict)
+
+
+def get_feeds_from_pols(polarization_array):
+    """
+    Return a list of expected feeds based on polarization values.
+
+    Translates values in polarization_array or jones_array into
```
bhazelton, line 821: sentence fragment.

src/pyuvdata/utils/uvcalibrate.py
```diff
@@ -440,26 +440,28 @@ def uvcalibrate(
             if len(uvcal_freqs_to_keep) < uvcal.Nfreqs:
                 downselect_cal_freq = True
 
-    # check if uvdata.telescope.x_orientation isn't set (it's required for uvcal)
-    uvd_x = uvdata.telescope.x_orientation
+    # check if x_orientation-equivalent in uvdata isn't set (it's required for uvcal)
+    uvd_x = uvdata.telescope.get_x_orientation_from_feeds()
```
bhazelton, line 444: I think maybe this should move to requiring the feed_array & feed_angle rather than the xorientation.

This code is trying to make sure that we apply the right calibration to the data, for that you need to know which feed the calibrations apply to.

bhazelton (commented):
This is looking really good to me, I'm comfortable with the approach you took for the x_orientation deprecation. Just a few other comments.

src/pyuvdata/telescopes.py
kartographer: Agreed - I'm gonna expound on this a bit in the docstring and point toward the conventions page for further details (writing that now)...

src/pyuvdata/telescopes.py
kartographer: I can try another pass here, but I think I would call the polarization aligned with north-south plane as being "vertical". This gets weird though when talking about zenith (as all good spherical geometry problems do), so maybe this calls for a more extended discussion _also_ in the "conventions"  page.

bhazelton: North-South could be vertical in some frames but not in the horizontal reference plane. I'd like to just stay away from the word vertical and go with something more like 0 means North for fixed mounts and aligned with the great circle for pointed telescopes, 90 means East for fixed mounts and perpendicular to the great circle for pointed telescopes.

bhazelton:
I will work on PRs for the external tests that are failing, but I may not get them done until week after next when I'm back from vacation.

src/pyuvdata/utils/pol.py
kartographer: Fixed in 11d373ecdf270b073d714514c43f9a1bd55e21d0.

src/pyuvdata/telescopes.py
kartographer: Okay, I've updated it in 11d373ecdf270b073d714514c43f9a1bd55e21d0.

src/pyuvdata/utils/uvcalibrate.py
kartographer: So the requirements are indeed set for `feed_array` and `feed_angle` similar to how they used to be set for `x_orientation` -- i.e., required for `UVCal`, and not for `UVData`. Are you suggesting that we additionally require it for `UVData`, or is the above sufficient?

Note that I'm checking for this string mostly for the sake of trying to identically replicate the previous behavior. I could instead check `feed_array` and `feed_angle`, although the logic would mostly just be replicating code that's already in the utilities.

src/pyuvdata/telescopes.py
kartographer: Okay, added in 11d373ecdf270b073d714514c43f9a1bd55e21d0.

src/pyuvdata/telescopes.py
kartographer: Okay, added in 2ebbd0faea01a23ea1cc23dd5a74edb482dbdde4, which includes info for this and some of the other major conventions that stood out in my mind. I think I'm gonna close the issue w/ this PR, baring objection/major missing elements to that page.

src/pyuvdata/utils/uvcalibrate.py
bhazelton: I guess I was just worried about the case that `feed_array` and `feed_angle` were set but `get_x_orientation_from_feeds` returned None because it wasn't a n/e arrangement. In that case I don't think we want to ignore what's on uvdata.

docs/make_index.py
```diff
@@ -55,6 +55,7 @@ def write_index_rst(readme_file=None, write_file=None):
         "   fast_calh5_meta\n"
         "   utility_functions\n"
         "   developer_docs\n"
+        "   conventions\n"
```
bhazelton, line 58: I think this should go higher up. Maybe right after the tutorial?

docs/conventions.rst
```diff
@@ -0,0 +1,151 @@
...
+    moves up and down in elevation. For this "right-handed" Nasmyth variant, a feed angle
+    of 0 degrees will be aligned to sum of the parallactic and elevation angles.
+  - ``"alt-az+nasmyth-l"``: The "left-handed" variant of the Nasmyth mount, where a
+    feed angle of 0 degrees is aigned to the **difference** of the parallactic and
+    elevation angles.
+  - ``"phased"``: Denotes an instrument where the an individual antenna input is
+    a phased array of detectors that are "beamformed" into a single voltage stream
+    (otherwise sometimes referred to as "electronically steered"). For this mount, a
+    0-degree feed angle is aligned to the great circle that connects local north,
+    local south, and the azimuth and elevation position of the phased antenna beam. Note
```
bhazelton, line 141: This feels wrong to me because it would imply that we had to write different feed angles into files depending on the beamformer settings. I think if the physical antennas do not move (e.g. MWA, SKA-low) that the feed angles should always be the same and should be recorded in the same way as they are for fixed antennas, so 0 should be north and 90 should be east.

bhazelton (commented):
LOVE the new convention page! @steven-murray can you take a look and see if anything is missing for you?

I also really like the condensed way you treated the antenna_mount in the docstring!

Just had a couple comments:

docs/conventions.rst
kartographer: I think you're right -- I managed to convince myself of a subtlety that in hindsight doesn't actually exist, and 0 degrees should be parallel w/ local north. I'll fix that here shortly...

docs/make_index.py
kartographer: Bumped up in a4b4bfa160f5d0bb923b28970bb457c333554526

src/pyuvdata/utils/uvcalibrate.py
kartographer: Okay, I've at least put a check inside of `uvcalibrate` for now so that the x-orientation doesn't get "overwritten" if it returns `None`. There's probably a longer conversation to be had about `uvcalibrate` _should_ behave when the feed information doesn't "match" -- presumably we can chat about this on a telecon in the coming weeks.

steven-murray:
@kartographer that convention doc page is fantastic, thanks! It'd be great to add a polarization convention section there as well (i.e. `I=XX +YY` vs `I=0.5(XX + YY)`).

bhazelton:
I started poking at pyuvsim to figure out what needed to be done there and realized we need to harmonize UVBeam with these changes. UVBeam doesn't have a telescope object on it, in large part because it doesn't need anything to do with antenna layout information. But it does have `x_orientation`, and pyuvsim uses that to set the x_orientation on the output UVData object.

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -668,6 +803,7 @@ def _set_power(self):
         self._Ncomponents_vec.required = False
         self._Nfeeds.required = False
         self._feed_array.required = False
+        self._feed_angle.required = False
```
bhazelton, line 806: I think we talked about always having feed_array being required, even for power beams.

src/pyuvdata/uvbeam/cst_beam.py
```diff
@@ -189,13 +187,17 @@ def read_cst_beam(
...
                     self.feed_array = np.array(["y", "x"])
+                    self.feed_angle = np.array([0.0, np.pi / 2])
             else:
                 if feed_pol == "x":
                     self.feed_array = np.array(["x"])
+                    self.feed_angle = np.array([np.pi / 2])
                 else:
                     self.feed_array = np.array(["y"])
+                    self.feed_angle = np.array([0.0])
             self.Nfeeds = self.feed_array.size
```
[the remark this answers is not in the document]
bhazelton: I think we also need to add a way to set the mount type. Both of these need to be optional inputs to `read_cst` and need to be supported being added to the yaml specification file for CST beams.

src/pyuvdata/uvbeam/initializers.py
```diff
@@ -24,6 +25,7 @@ def new_uvbeam(
     model_name: str = "default",
     model_version: str = "0.0",
     feed_array: npt.NDArray[str] | None = None,
+    feed_angle: npt.NDArray[float] | None = None,
```
bhazelton, line 28: mount type need to be added here as well.

bhazelton: Actually, it might be ok to leave it to be picked up in kwargs, but it needs to be tested to work.

tests/uvbeam/test_initializers.py
```diff
@@ -40,7 +41,7 @@ def uvb_healpix_kw():
 
 @pytest.fixture()
 def uvb_efield_kw():
-    return {"feed_array": ["x", "y"]}
+    return {"feed_array": ["x", "y"], "feed_angle": [np.pi / 2, 0]}
```
bhazelton, line 44: Need a test for initializing with a mount_type.

src/pyuvdata/uvbeam/mwa_beam.py
```diff
@@ -617,12 +617,11 @@ def read_mwa_beam(
         ):
             self.history += self.pyuvdata_version_str
 
-        self.x_orientation = "east"
-
         self._set_efield()
         self.Naxes_vec = 2
         self.Ncomponents_vec = 2
         self.feed_array = np.array([str(pol.lower()) for pol in pol_names])
+        self.feed_angle = np.where(self.feed_array == "x", np.pi / 2, 0.0)
```
bhazelton, line 624: should we specify a mount type? I'm confused about whether it should be fixed or phased...

src/pyuvdata/analytic_beam.py
```diff
@@ -898,24 +1079,59 @@ class ShortDipoleBeam(AnalyticBeam):
...
-        The orientation of the dipole labeled 'x'. The default ("east") means
-        that the x dipole is aligned east-west and that the y dipole is aligned
-        north-south.
+        Feeds to define this beam for, e.g. x & y.
+    feed_angle : array-like of float
+        Position angle of a given feed, units of radians. A feed angle of 0 is
+        typically oriented toward zenith for steerable antennas, otherwise toward
+        north for fixed antennas (e.g., HERA, LWA). More details on this can be found
+        on the "Conventions" page of the docs.
+    mount_type : str
```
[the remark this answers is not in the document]
kartographer: Just for clarification -- you want that to be the default, or you want that to be the _only_ option?

src/pyuvdata/parameter.py
```diff
@@ -756,6 +767,150 @@ def setter(self, inst):
...
+        ----------
+        form_dict : dict
+            Dictionary which maps axes to index arrays, with keys that are matched
+            against entries within UVParameter.form, and values which demark which
+            indices should be selected (must be 1D). Values can also be given as None,
+            in which case no selection is performed along that axis.
+        values : list or ndarray
+            Values based on the positions selected, as specified in form_dict. Note that
+            for singleton values or if no entries in form_dict match the form parameter,
+            the whole of UVParameter.value is set.
```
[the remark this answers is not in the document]
kartographer: So the `form_dict` is meant to identify which axes have subsets of indices to assign data to, where no matching entry is treated as "all the indices along this access should effectively be selected". To that end, I saw the above behavior as being consistent with that usage pattern, although I can see that this might come as an unexpected surprise to someone. I'd like to retain the above functionality if possible -- what if I were to add a `strict` keyword that behaves in the same way that it does in other places? Nominally I'd raise a warning, but we could make the default an error.

src/pyuvdata/uvbeam/mwa_beam.py
kartographer: I think "phased" is the most appropriate here, or at least the one that matches the limited documentation on it UVFITS (an "electronically-steered" primary beam).

src/pyuvdata/analytic_beam.py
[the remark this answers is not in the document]
kartographer: I've set this as the default for now (as of 4bd22a9260ec0b4845b31f0bc4d84d7f61e23587)

src/pyuvdata/telescopes.py
```diff
@@ -767,6 +975,9 @@ def new(
         instrument: str | None = None,
         x_orientation: Literal["east", "north", "e", "n", "ew", "ns"] | None = None,
         antenna_diameters: list[float] | np.ndarray | None = None,
+        feeds: Literal["x", "y", "l", "r"] | list[str] | None = None,
+        feed_array: np.ndarray | None = None,
+        feed_angle: np.ndarray | None = None,
```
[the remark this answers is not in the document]
kartographer: Added in 4bd22a9260ec0b4845b31f0bc4d84d7f61e23587.

src/pyuvdata/analytic_beam.py
```diff
@@ -617,36 +752,82 @@ class UnpolarizedAnalyticBeam(AnalyticBeam):
...
+            "orbiting",
+            "x-y",
+            "alt-az+nasmyth-r",
+            "alt-az+nasmyth-l",
+            "phased",
+            "fixed",
+            "other",
+        ]
+        | None
+    ) = field(default=None, repr=False, compare=False)
```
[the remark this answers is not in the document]
kartographer: Updated in 4bd22a9260ec0b4845b31f0bc4d84d7f61e23587.

src/pyuvdata/uvbeam/initializers.py
kartographer: Added test in 4bd22a9260ec0b4845b31f0bc4d84d7f61e23587 -- it does catch correctly w/ kwargs

src/pyuvdata/uvbeam/mwa_beam.py
kartographer: Added in 4bd22a9260ec0b4845b31f0bc4d84d7f61e23587.

tests/uvbeam/test_initializers.py
kartographer: Added in 4bd22a9260ec0b4845b31f0bc4d84d7f61e23587.

src/pyuvdata/uvbeam/uvbeam.py
kartographer: Sorry -- wasn't clear if we wanted to also fold that into this PR, changed in 0c9851936cf3aae64c073b4df8b527903988922f.

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -578,6 +613,105 @@ def __init__(self):
...
+            warnings.warn(
+                'Support for physically oriented feeds (e.g., "n", and "e") has been '
+                'deprecated, with the only allowed options now available being "x", '
+                '"y", "l", or "r". The physical orientation of the feed is now '
+                "recorded in the UVBeam.feed_angle parameter. This will become an "
+                "error in version 3.4.",
+                DeprecationWarning,
+            )
+            feed_map = {"e": "x", "n": "y"}
+            self.feed_array = [feed_map.get(item, item) for item in self.feed_array]
```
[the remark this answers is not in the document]
kartographer: Added in 0c9851936cf3aae64c073b4df8b527903988922f.

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -267,7 +270,53 @@ def __init__(self):
             required=False,
             expected_type=str,
             form=("Nfeeds",),
-            acceptable_vals=["x", "y", "n", "e", "r", "l"],
+            acceptable_vals=["x", "y", "r", "l"],
+        )
+
+        desc = (
+            "Array of feed orientations. shape (Nfeeds). "
+            'options are: n/e or x/y or r/l. Not required if beam_type is "power".'
```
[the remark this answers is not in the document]
kartographer: Fixed in 0c9851936cf3aae64c073b4df8b527903988922f.

src/pyuvdata/uvbeam/cst_beam.py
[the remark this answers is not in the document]
kartographer: Added keywords for setting of `feed_angle` and `feed_array` in 0c9851936cf3aae64c073b4df8b527903988922f. Still need to add docstrings...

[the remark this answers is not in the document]
kartographer: Added docstrings in 28bd931ab254404a794a74bdc82ec17c4d87827a.

src/pyuvdata/parameter.py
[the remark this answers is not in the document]
kartographer: Strict keyword added in a977e1520f4b282c9801a1cd24a0ee4ca1fce071.

kartographer:
Making an additional external testing note here -- following the fix to the CASA conjugation scheme, we've been able to successfully validate polarization data handling via pyuvdata -> CASA (versus MIR -> MIRIAD). D-terms are not _yet_ externally validated (since they're derived in CASA directly), but everything else checks out.

CHANGELOG.md
```diff
@@ -4,6 +4,35 @@ All notable changes to this project will be documented in this file.
...
+gains) and "D Jones" (polarization leakages) has been added to `MSCal`.
+- The `UVBeam.x_orientation` parameter has been deprecated, superseded by two new
+required parameters: `UVBeam.feed_array` and `UVBeam.feed_angle`, which describe the
+polarization and orientation of the beam (the former of which was already present but
+only required for efield beams).
+- The `Telescope.x_orientation` parameter has been deprecated, superseded by two new
+parameters: `Telescope.feed_array` and `Telescope.feed_angle`, which describe the
+polarization and orientation of the detectors for a given antenna.
+- The `Telescope.mount_type` and `UVBeam.mount_type` parameters have been added, which
+describe the mount and optics of a given antenna (e.g., HERA/MWA are "fixed", VLA/ALMA
```
bhazelton, line 24: MWA is set to "phased" in mwa_beam, not fixed. Maybe just take MWA out of this parenthetical?

src/pyuvdata/uvdata/uvdata.py
```diff
@@ -5701,6 +5687,9 @@ def __add__(
                 raise ValueError(msg)
 
         # Begin manipulating the objects.
+        # Note that this will check to see if we can merge the telescopes (if they are
+        # different, otherwise the underlying checking is the same).
+        this.telescope += other.telescope
```
bhazelton, line 5692: What happens if the telescopes are the same (probably the most common situation)? Is this a no-op?

src/pyuvdata/uvflag/uvflag.py
```diff
@@ -2092,6 +2091,26 @@ def __add__(
...
+            if (
+                this.telescope.feed_angle is not None
+                and other.telescope.feed_angle is not None
+            ):
+                temp_feed_angle = np.concatenate(
+                    [this.telescope.feed_angle, other.telescope.feed_angle], axis=0
+                )
+                this.telescope.feed_angle = temp_feed_angle[unique_inds]
+            else:
+                this.telescope.feed_angle = None
```
bhazelton, line 2113: could this be replaced by a call to the telescope add function like you're doing in the UVData add?

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -3678,8 +3879,14 @@ def read_cst_beam(
...
-            if x_orientation is None and "x_orientation" in settings_dict:
-                x_orientation = settings_dict["x_orientation"]
+            if x_orientation is None:
+                x_orientation = settings_dict.get("x_orientation", None)
+            if feed_array is None:
+                feed_array = settings_dict.get("feed_array", None)
+            if feed_angle is None:
+                feed_angle = settings_dict.get("feed_angle", None)
+            if mount_type is None:
+                mount_type = settings_dict.get("mount_type", None)
```
bhazelton, line 3889: I think these should also be added to the `overriding_keywords` so there's a warning if they are set in the yaml and when the method is called.

bhazelton (commented):
Looking good! A couple of high-level comments:
- I think that we need to require mount_type if feed_angle is set because the meaning of feed_angle depends on mount_type.
- Should we set the feed_array/feed_angle/mount_type in our known telescopes dicts?

src/pyuvdata/parameter.py
```diff
@@ -20,6 +20,19 @@
...
+
+allowed_location_types = [EarthLocation]
+try:
+    from lunarsky import MoonLocation
+
+    allowed_location_types.append(MoonLocation)
+
+    hasmoon = True
+except ImportError:
+    hasmoon = False
```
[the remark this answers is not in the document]
kartographer: I think this may have just been leftover from a rebase gone awry -- neither `allowed_location_types` or `hasmoon` are referenced anywhere else, so I should be able to nuke this snippet.

CHANGELOG.md
```diff
@@ -28,6 +57,12 @@ default is `True`), such that most warnings about frequency/polarization/time sp
 will not normally be raised.
 
 ### Changed
+- `feed_array` (and by association, `feed_angle` and `Nfeeds`) is now a required
+parameter for `UVBeam`.
```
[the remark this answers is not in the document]
kartographer: Yes -- updated in 0ede0cb9b4bbc4628136211eb13c25fded476c26.

src/pyuvdata/parameter.py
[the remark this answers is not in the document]
kartographer: Fixed in 0ede0cb9b4bbc4628136211eb13c25fded476c26.

CHANGELOG.md
kartographer: Changed to the right type in 0ede0cb9b4bbc4628136211eb13c25fded476c26.

src/pyuvdata/telescopes.py
```diff
@@ -511,7 +679,11 @@ def update_params_from_known_telescopes(
         run_check: bool = True,
         check_extra: bool = True,
         run_check_acceptability: bool = True,
-        known_telescope_dict: dict = _KNOWN_TELESCOPES,
+        feeds: str | list[str] | None = None,
+        polarization_array: np.ndarray | None = None,
+        flex_polarization_array: np.ndarray | None = None,
+        override_known_params: bool = True,
+        **kwargs,
```
bhazelton, line 686: should this kwargs be documented? I don't know what can be passed to it...

src/pyuvdata/telescopes.py
```diff
@@ -709,7 +1021,7 @@ def from_known_telescopes(
         run_check: bool = True,
         check_extra: bool = True,
         run_check_acceptability: bool = True,
-        known_telescope_dict: dict = _KNOWN_TELESCOPES,
+        **kwargs,
```
bhazelton, line 1024: This one is also not documented.

src/pyuvdata/telescopes.py
```diff
@@ -948,3 +1331,439 @@ def get_enu_antpos(self):
...
+        if run_check:
+            self.check(
+                check_extra=check_extra, run_check_acceptability=run_check_acceptability
+            )
+
+    def __add__(
+        self,
+        other: Telescope,
+        *,
+        warning_params: list | None = None,
```
bhazelton, line 1456: This parameter is not in the docstring here or in `__iadd__`

bhazelton (commented):
Looks like external tests are failing again, I'll try to take a look soon.

src/pyuvdata/telescopes.py
```diff
@@ -493,6 +652,15 @@ def check(self, *, check_extra=True, run_check_acceptability=True):
             check_extra=check_extra, run_check_acceptability=run_check_acceptability
         )
 
+        # If using feed_angle, make sure feed_array is set (and visa-versa)
+        param_check = [
+            p is not None for p in [self.feed_array, self.feed_angle, self.mount_type]
+        ]
+        if any(param_check[0:2]) and not all(param_check):
```
bhazelton, line 659: This is causing errors in pyuvsim when we read in an existing uvfits files with a made-up telescope with no feed_array/feed_angle/mount_type information. It seems that mount_type is getting defaulted to "fixed" but the other parameters are not set, so we get an error in the check. I think it should be ok to have the mount_type set even if the feed_array and feed_angle are not set.

bhazelton: I think this may also be responsible for the errors in hera_cal.

kartographer: On pyuvsim at least I've confirmed that the issue at hand is that `mount_type` is in fact not being set -- there is a nominal default for `UVBeam`, but not one for `Telescope` (which is affecting `UVData` and `UVCal`). There are a couple of ways to approach this:

- Do nothing (i.e., tell the user to set `run_check=False` when loading; not my preferred option)
- Set `mount_type` to a default if the check above fails, in the same way as is done w/ `UVBeam`
- Set `mount_type` to a default on read if one is not found and the telescope is not present in `KNOWN_TELESCOPES` (there's already keyword for facilitating this).

All else being equal, I probably favor the last option since it feels like the best long-term solution, but I'm open to the second option if folks think that'll be easier to deal with.

kartographer: One additional option -- we can set the feed information to `None` if it's not required and `mount_type` is set to None.

src/pyuvdata/telescopes.py
kartographer: Added in 1b8340609b362fc9b9225a3eb7e9535b7793952d.

src/pyuvdata/telescopes.py
kartographer: Added in 1b8340609b362fc9b9225a3eb7e9535b7793952d.

src/pyuvdata/telescopes.py
kartographer: Added in 1b8340609b362fc9b9225a3eb7e9535b7793952d.

src/pyuvdata/uvflag/uvflag.py
```diff
@@ -942,6 +947,8 @@ def set_lsts_from_time_array(self, *, background=False, astrometry_library=None)
     def set_telescope_params(
         self,
         *,
+        x_orientation=None,
+        mount_type=None,
```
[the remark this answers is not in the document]
kartographer: Updated in 50395cc7a20e9d386e6213b15e50fd01b45f6532.

src/pyuvdata/uvdata/uvdata.py
```diff
@@ -1476,6 +1478,8 @@ def known_telescopes(self):
     def set_telescope_params(
         self,
         *,
+        x_orientation=None,
+        mount_type=None,
```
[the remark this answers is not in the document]
kartographer: Updated in 50395cc7a20e9d386e6213b15e50fd01b45f6532.

src/pyuvdata/telescopes.py
kartographer: Just to prevent one from having to dive into the code -- I've modified the code to effectively do options 3 and 4 above (default on read is now "other", which is typical behavior for multiple file formats, and the feed info is set to `None` if not required as is the case for `UVData` and `UVFlag`). That should at least hopefully prevent this from being a breaking change (even if the new warnings are breaking some tests).

src/pyuvdata/uvbeam/uvbeam.py
kartographer: Alright, finally tackled this orphaned comment in 602c12b9fe74c2c43eda7a5595486ae763c6fa69

kartographer:
Making a note here (mostly for @bhazelton to look when she's back) - the Conventions page has been updated per our conversion on the weekly call.

bhazelton (approved):
Looking good, just a couple comments

bhazelton:
@kartographer if you rebase the cibuildwheel jobs should pass.

src/pyuvdata/uvdata/uvfits.py
```diff
@@ -1251,14 +1286,36 @@ def write_uvfits(
         # ADD the ANTENNA table
         staxof = np.zeros(self.telescope.Nants)
 
-        # 0 specifies alt-az, 6 would specify a phased array
+        # If unknown, default to alt-az (0) as the default (as it was before)
```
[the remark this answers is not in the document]
kartographer: Yeah, this was something of a "I should do this for backwards compatibility, but I'm not mad because I like arrays of steerable dishes and this would be my personally preferred default" 😉

src/pyuvdata/telescopes.py
```diff
@@ -948,3 +1526,451 @@ def get_enu_antpos(self):
...
+
+        Parameters
+        ----------
+        other : Telescope object
+            Another Telescope object which will be added to self.
+        warning_params : list of str
+            By default, when parameters are from the two objects are incompatible, an
+            error is thrown if the parameter is required, otherwise the parameter on
+            the returned object is set to None and a warning is raised. However, one can
+            a list of strings to explicitly specify which parameters to allow to raise a
```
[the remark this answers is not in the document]
kartographer: Fixed in 58882eb674c55536753e72aacb5e20235f6e9dce. Definitely would not be opposed, though it's not clear it would work out of the box since the arguments are ever-so-slightly different (most notably, `__iadd__` has no `inplace` keyword, for obv reasons). But maybe this is a reflection of my limited understanding of the docstring parser stuff...

src/pyuvdata/telescopes.py
```diff
@@ -948,3 +1526,451 @@ def get_enu_antpos(self):
...
```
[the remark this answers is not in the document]
kartographer: Fixed in 58882eb674c55536753e72aacb5e20235f6e9dce.

src/pyuvdata/uvcal/uvcal.py
```diff
@@ -4738,13 +4732,18 @@ def read(
         run_check=True,
         check_extra=True,
         run_check_acceptability=True,
+        default_mount_type="other",
         # file-type specific parameters
+        # All types
```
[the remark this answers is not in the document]
kartographer: Yup  -- removed in 58882eb674c55536753e72aacb5e20235f6e9dce.

src/pyuvdata/telescopes.py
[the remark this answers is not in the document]
bhazelton: ah, that might require some tweaking of the code. Something to keep in mind for the future.

bhazelton (approved):

PR #1535, opened 2025-03-14
only import lunarsky if it's actually needed

bhazelton:
## Description

Lunarsky does a bunch of calculations and caching on import. So we should only import it if it's actually needed.

I also cleaned up some `pragma: no cover` marks on code that I could find ways to write tests for and fixed some pytest warnings about impending changes to the default behavior of pytest.importorskip.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ default is `True`), such that most warnings about frequency/polarization/time sp
 will not normally be raised.
 
 ### Changed
+- Only import lunarsky if needed.
 - `UVData.select`, `UVBeam.select`, `UVFlag.select`, `UVFlag.select` have been
 significantly refactored and made to behave more uniformly.
 - Allowing `UVParameter.__eq__` to use `UVParameter.compare_value` if the item being
--- a/src/pyuvdata/parameter.py
+++ b/src/pyuvdata/parameter.py
@@ -20,17 +20,6 @@ import numpy as np
 from astropy import units
 from astropy.coordinates import EarthLocation, SkyCoord
 
-allowed_location_types = [EarthLocation]
-try:
-    from lunarsky import MoonLocation
-
-    allowed_location_types.append(MoonLocation)
-
-    hasmoon = True
-except ImportError:
-    hasmoon = False
-
-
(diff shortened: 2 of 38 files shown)


bhazelton:
@mkolopanis can you please take a look at the handling in `coordinates.pyx`?

mkolopanis:
So I've been thinking about this issue, with how the `Body` enum is implemented, I don't think we currently can move the attempt to import lunarsky to a call at run time. This enum is created at compile time for the coordinates extension. I know how to do this in pure python (with the [aenum](https://github.com/ethanfurman/aenum) package), but I think this code lives in the pyx file to prevent a cyclic import. 

maybe we could just have the coordiate functions take an `Ellipsoid` instance, and a function that allows a name as an input.. basically re-creating a dictionary at that point.

bhazelton:
@mkolopanis Fascinating. I don't think I'd really understood that this depended on what was available at compile time. Does this mean that people using our conda package (which is built without lunarsky) can't actually work with moon locations even if they install lunarsky?

mkolopanis:
Well there is a bit of magic because cython can do python stuff too :thinking: So i'm not totally sure but we should check if this is true or not.

mkolopanis:
We decided we'd like to have functions that require an Ellipsoid input (or Body) in the cython to instead take the equatorial radius and flattening parameters and construct the Ellipsoid at runtime.

My personal notes is that this does incur an increased runtime hit but it may not be very much if at all. Got to do what we got to do.

mkolopanis:
I now wonder if we should we cache the "get_selenoids" utility function. Welcome any thoughts

bhazelton:
> I now wonder if we should we cache the "get_selenoids" utility function. Welcome any thoughts

I'm totally open to it.

steven-murray (approved):
@bhazelton this all looks good to me! I didn't check the .pyx file because 🤷🏻‍♂️

mkolopanis:
Luckily the pyx files only had things deleted from them really so sounds like it should be fine?

steven-murray (approved):

steven-murray:
Re-approved!

PR #1541, opened 2025-04-02
Fix bug in constructing off-zenith MWA beams

bhazelton:
## Description

Off zenith beams made no sense. This was initially reported to me by Jack Line, and he suggested it might be a dipole sorting issue. Turns out he was exactly correct. I was sorting the dipoles as strings rather than integers, so dipoles 10-15 were sorting before dipole 2. This fixes that problem and adds tests to prevent this kind of thing from happening again.

Here are plots for delays of [0 2 4 6 0 2 4 6 0 2 4 6 0 2 4 6] (eastward pointing).

Prior to fix:

Screenshot 2025-04-02 at 2 19 14 PM
Screenshot 2025-04-02 at 2 19 19 PM

After fix:

Screenshot 2025-04-02 at 2 17 50 PM
Screenshot 2025-04-02 at 2 18 00 PM

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/src/pyuvdata/uvbeam/mwa_beam.py
+++ b/src/pyuvdata/uvbeam/mwa_beam.py
@@ -265,7 +265,7 @@ class MWABeam(UVBeam):
 
         pol_names = sorted(pol_names)
 
-        dipole_names = np.array(sorted(dipole_names))
+        dipole_names = np.sort(np.asarray(list(dipole_names), dtype=int)).astype(str)
 
         freqs_hz = np.array(sorted(freqs_hz))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@ compared and `UVParameter.value` share the same class.
 and `UVBeam.check`.
 
 ### Fixed
+- Bug in MWA beams that caused beams pointed away from zenith to be wrong because
+the delays were not assigned to the right dipoles.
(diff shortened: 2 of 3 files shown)


src/pyuvdata/uvbeam/mwa_beam.py
```diff
@@ -265,7 +265,7 @@ def _read_metadata(self, h5filepath):
 
         pol_names = sorted(pol_names)
 
-        dipole_names = np.array(sorted(dipole_names))
+        dipole_names = np.sort(np.asarray(list(dipole_names), dtype=int)).astype(str)
```
mkolopanis, line 268: I believe, this could also have been accomplished as the following. Specifying `key` in the sorted method will call that function on each element, sort based on the output. Up to you if you want to keep the double cast.

```suggestion
        dipole_names = np.asarray(sorted(dipole_names, key=int)))
```

mkolopanis (commented):
Good find. I have one comment about the sorting that really just comes down to your personal preference.

src/pyuvdata/uvbeam/mwa_beam.py
bhazelton: Thank you! I thought there was a way to do this without actually converting to ints and back but I couldn't remember what it was.

PR #1560, opened 2025-05-08
Initial editorial tweaks for JOSS paper

warrickball:
This is an initial set of editorial changes to the JOSS paper for v3 following the [successful review](https://github.com/openjournals/joss-reviews/issues/7482). These are mostly minor and many in the bibliography, rather than text, but you should double check that everything looks acceptable.

Note that I changed "dataset" to "data set" in the title for consistency with the text but you may prefer to go the other way and keep "dataset" in the title and change "data set" to "dataset" in the text (under "Major updates in this version").

Either way, note that the title has changed slightly ("python" → "Python"), so you'll need to update the Zenodo archive title to match.

Once these changes are all in I'll launch another test run of the publication pipeline.

### DIFF
--- a/paper.bib
+++ b/paper.bib
@@ -1,6 +1,6 @@
 @TECHREPORT{aipy,
   author = {Aaron Parsons},
-  title = {Astronomical Interferometry in PYthon},
+  title = {Astronomical Interferometry in {Python}},
   year = {2007},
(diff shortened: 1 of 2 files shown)


kartographer (approved):
Looks good, thanks for the editing work!

Issue #1543, opened 2025-04-03
Re-evaluate Cython Dependencies


mkolopanis:
Cython is not the most fun language to work in. There are some advantages to moving out compiled extensions to another language (pybind11, rust, etc), but for now we decided we'd like to move what we can to [numba](https://numba.pydata.org/).

Numba appears to be asymptotically equivalent in execution to our compiled extensions for large number of inputs. The only real advantage other languages seem to have is overhead at the nano-second level and the users probably will never notice this.

We should evaluate all the compiled extensions to see if we can refactor them to numba for now.
The main advantage we see to numba is the barrier for entry for others to maintain the code and not need experience with the compiled languages.

PR #1561, opened 2025-05-09
Pin version of cython for installation and CI jobs

kartographer:
## Description

Pins the version of cython used to `3.0.*`.

## Motivation and Context

A new release of cython appears to be breaking the build for pyuvdata -- this PR is meant to be a "quick-fix", and arguably one would actually go in and update the cython code itself, but given that we are in the process of re-evaluating our use of cython (a la #1543), I think this is a sensible solution for the immediate term. 

## Types of changes

## Checklist:

- [x] If required or optional dependencies have changed (including version numbers), I have updated the readme, pyproject.toml, environment.yaml and CI conda yaml files to reflect the changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ you will need the following packages:
 * pytest >= 6.2
 * pytest-cases >= 3.8.3
 * pytest-cov
-* cython >=3.0
+* cython ==3.0
 * coverage
 * pre-commit
 * matplotlib
--- a/ci/pyuvdata_min_deps_tests.yml
+++ b/ci/pyuvdata_min_deps_tests.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - astropy>=6.0
-  - cython>=3.0
+  - cython==3.0.*
   - docstring_parser>=0.15
   - h5py>=3.1
   - numpy>=1.23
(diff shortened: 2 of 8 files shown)


bhazelton (commented):
Looks reasonable. I think we discussed that we will need cython in the foreseeable future for Miriad, so I think we will need to update the code and unpin at some point. Can you make an issue for that so we don't forget?

We do also list dependency versions in the readme, should be updated too?

kartographer:
> Looks reasonable. I think we discussed that we will need cython in the foreseeable future for Miriad, so I think we will need to update the code and unpin at some point. Can you make an issue for that so we don't forget?

Done

> We do also list dependency versions in the readme, should be updated too?

Yeah -- it actually was only listed as a dependency for the developer build, though insofar as I can tell, it seems to be actually required for the package as a whole, so I've nominally moved it up (but obviously correct me here if I'm wrong @bhazelton)

bhazelton:
> it actually was only listed as a dependency for the developer build, though insofar as I can tell, it seems to be actually required for the package as a whole, so I've nominally moved it up (but obviously correct me here if I'm wrong @bhazelton)

I am not the expert, but I thought that when the package was built the code was compiled so no longer required cython to run (e.g. when installing from pypi). @mkolopanis ?

kartographer:
> I am not the expert, but I thought that when the package was built the code was compiled so no longer required cython to run (e.g. when installing from pypi). @mkolopanis ?

Okay, after a bit of testing, I think you're right - I've changed the README accordingly.

bhazelton (approved):
Looks good.

PR #1566, opened 2025-05-16
First adventures in numba

### LINKED ISSUE #1544: Translate baseline extension to numba
@kartographer and I have done some work that shows moving the baseline utilities to numba will be asymptotically equivalent for the user and feel that the barrier to entry is lower for future maintainers.

Therefore we should translate this utility's extension modules into numba

kartographer:
First numba-based refactor.

## Description

Integrates numba, replaces the baseline utils originally developed in Cython.

## Motivation and Context

This PR is meant to be a "first pass" at refactoring some of the old cython code using numba, integrating some work that @mkolopanis and I did while exploring if we could get the numba-based routines to perform as quickly as the cython-based ones. Since they were relatively straightforward to test and translate, we focused on the `utils.bls` module first. I'm attaching a notebook that @mkolopanis primarily put together for testing -- the good news is that I think I've got the numba versions of routines actually working at close to parity with the cython code (and for large-N, actually running _better_ than cython). Note that the inputs for the numba routines are slightly different, in that I've slightly refactored the higher level utils to be a bit smarter (e.g., finding the min/max values once, using those for both input-checking and choosing the cipher), so for fairness I've run timing tests with and without this "optimized overhead".

Note that for the moment, I haven't deleted the old cython routines, although they are no longer being used by any of the existing code at the moment.

Closes #1544

## Types of changes

- [x] Other

## Checklist:

Other:
- [x] I have updated any docstrings associated with my change using the [numpy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [x] I have updated the readme and/or tutorial to reflect my changes (if appropriate).
- [x] I have added/updated tests to cover my changes (if appropriate).


kartographer:
[bls_speed_test.pdf](https://github.com/user-attachments/files/20242431/bls_speed_test.pdf)

src/pyuvdata/utils/bls.py
```diff
@@ -5,15 +5,59 @@
...
 
 
+@numba.njit()
+def _max_ant(ant1: npt.NDArray[np.int64], ant2: npt.NDArray[np.int64]) -> np.int64:
+    return np.maximum(np.max(ant1), np.max(ant2))
+
+
+@numba.njit()
+def _min_ant(ant1: npt.NDArray[np.int64], ant2: npt.NDArray[np.int64]) -> np.int64:
+    return np.minimum(np.min(ant1), np.min(ant2))
```
mkolopanis, line 25: @calling `np.ufunc` should be equivalent to `array.ufunc` right?

kartographer: It _should_ be - curious to see if it makes any difference w.r.t. speed (it shouldn't, but doesn't hurt to measure)

kartographer: Yup, no difference

mkolopanis: thanks for checking at least. Glad they use the same backend

kartographer:
FWIW, codecov is balking on account of the new numba code not being able to be traced by `coverage` right now. This seems to be a [known issue w/ numba](https://github.com/numba/numba/issues/4268), with a couple of potential workarounds (including doing nothing at all), but no clear definitive solutions.

kartographer:
> FWIW, codecov is balking on account of the new numba code not being able to be traced by `coverage` right now. This seems to be a [known issue w/ numba](https://github.com/numba/numba/issues/4268), with a couple of potential workarounds (including doing nothing at all), but no clear definitive solutions.

Following discussion on the weekly telecon, I've shuffled the new numba stuff over to a new module, and set up a series of tests with a new specialized CI workflow that tests only that module with the numba JIT compiler disabled (`NUMBA_DISABLE_JIT=1`). This _doesn't_ work for for `@vectorize` decorator, although that's a small subset of the new code.

bhazelton (approved):
Looks good to me! We'll just need to keep in mind as we review future PRs that anything marked with the `vectorize` decorator is not included in coverage reports.

PR #1582, opened 2025-06-17
Fix UVBase._select_along_param_axis for SkyCood

bhazelton:
## Description

When I tried to use the new `UVBase._select_along_param_axis` in pyradiosky it errored because `UVBase.get_from_form` doesn't handle `SkyCoordParameter` objects properly. SkyModel also has Quantity parameters which should be tested against too. This updates both `UVBase.get_from_form` and `UVBase.set_from_form` to work for `SkyCoordParameter`, `Quantity` and `Time` objects. We don't currently have any code using Time objects, so we could drop that testing if we want.

## Types of changes

- [ ] Version change
- [ ] Other

## Checklist:

Bug fix checklist:

### DIFF
--- a/src/pyuvdata/parameter.py
+++ b/src/pyuvdata/parameter.py
@@ -836,8 +836,6 @@ class UVParameter:
                 extra_axes[axis] = val_slice[axis]
                 val_slice[axis] = slice(None)
 
-        # use issubclass ShapedLikeNDArray to handle Time and SkyCoord objects
-        # Quantities are instances of np.ndarray.
         if isinstance(
             self.value, np.ndarray | np.ma.MaskedArray | SkyCoord | Time | Quantity
         ):
@@ -917,8 +915,6 @@ class UVParameter:
                 extra_axes[axis] = val_slice[axis]
                 val_slice[axis] = slice(None)
 
-        # use issubclass ShapedLikeNDArray to handle Time and SkyCoord objects
-        # Quantities are instances of np.ndarray.
         if isinstance(self.value, Quantity | SkyCoord | Time):
             if extra_axes:
                 # Quantity, SkyCoord, and TIme have some extra caching that causes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- A bug in `UVParameter.get_from_form` and `UVParameter.set_from_form` that caused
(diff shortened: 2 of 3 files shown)


src/pyuvdata/parameter.py
```diff
@@ -917,18 +927,35 @@ def set_from_form(
...
                 # Make sure this is a view that is connected in memory
-                assert temp_arr.base is (
-                    self.value if self.value.base is None else self.value.base
-                ), (
-                    "Something is wrong, slicing self.value does not return a view "
-                    "on the original array. Please file an issue in our GitHub "
-                    "issue log so that we can fix it."
-                )
+                if issubclass(self.value.__class__, ShapedLikeNDArray):
+                    for attr in self.value._extra_frameattr_names:
```
bhazelton, line 931: This actually probably only works for SkyCoord objects, not Time or Quantity objects. Need to add testing for those.

src/pyuvdata/parameter.py
```diff
@@ -834,13 +835,19 @@ def get_from_form(self, form_dict: dict, strict: bool | None = None):
...
+            self.value.__class__, ShapedLikeNDArray
+        ):
             # If we're working with an ndarray, use take to slice along
             # the axis that we want to grab from.
             value = self.value[tuple(val_slice)]
             if extra_axes:
                 ind_arr, new_shape = _multidim_ind2sub(extra_axes, value.shape)
-                value = value.flat[ind_arr].reshape(new_shape)
+                if issubclass(self.value.__class__, ShapedLikeNDArray):
+                    value = (value.flatten())[ind_arr].reshape(new_shape)
```
bhazelton, line 848: maybe use ravel instead of flatten to avoid unnecessary copies

src/pyuvdata/parameter.py
bhazelton: I've added testing and I can't figure out how to make this view testing work for Time and Quantity objects. @kartographer any ideas? I'm also wondering whether `np.put` or `np.put_along_axis` might help here?

kartographer: Okay, finally had a chance to poke around with this - yes, it does appear as though for `Quantity` and `Time`, slices and whatnot do not return a view of the original object but rather a fresh new copy.  I don't think this is a situation where `put`/`put_along_axis` is going to help, as the thing that's at issue here is that you've got arrays that are NDArray-like in some ways but not others. ***BUT*** I am able to force the behavior I want if I grab `obj.value` first, which grabs the underlying NDArray and behaves more like as I expect.

PR #1583, opened 2025-06-20
Implement a check of pyuvsim cache for fallback file inside _uvbeam_constructor

burdorfmitchell:
## Description

Implemented check of pyuvsim cache for uvbeam file inside _uvbeam_constructor as a fallback if the file isn't existent otherwise. Updated docstring of _construct_uvbeam to reflect this change.

This change is to support the downloading and handling of large data files -- gleam, uvbeam, etc... (currently just beams and catalogs) -- using astropy. The intention is to allow the passing of the cache url corresponding to a file downloaded or added to the astropy cache into a simulation configuration yaml file directly without having to specify any special path information. Individuals can then simply download the relevant data files to the cache -- in this case UVBeam files -- and they will be recognized and read appropriately while running a simulation.

Where this functionality is immediately important is in the reference simulations of pyuvsim. The current approach to running the new reference simulations is to install the larger data files to the pyuvsim cache specified by astropy using download_file. There is currently no method to appropriately target cached UVBeam files besides manually editing the yaml file to insert the absolute path to the file, a problem that this pr solves.

## Types of changes

- [ ] Version change
- [ ] Other

## Checklist:

New feature checklist:

### DIFF
--- a/src/pyuvdata/uvbeam/uvbeam.py
+++ b/src/pyuvdata/uvbeam/uvbeam.py
@@ -14,6 +14,7 @@ from typing import Literal
 import numpy as np
 import yaml
 from astropy import units
+from astropy.utils.data import cache_contents, is_url_in_cache
 from docstring_parser import DocstringStyle
 from scipy import interpolate, ndimage
 
@@ -4745,7 +4746,8 @@ def _uvbeam_constructor(loader, node):
     Define a yaml constructor for UVBeam objects.
 
     The yaml must specify a "filename" field pointing to the UVBeam readable file
-    and any desired arguments to the UVBeam.from_file method.
+    and any desired arguments to the UVBeam.from_file method. If the file does not
+    exist, checks pyuvsim cache for file treating the "filename" as the cache url.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -4784,6 +4786,12 @@ def _uvbeam_constructor(loader, node):
         for f_i in range(len(files_use)):
             files_use[f_i] = os.path.join(path_var, files_use[f_i])
 
+    for i, file in enumerate(files_use):
+        # if file does not exist, check pyuvsim cache defined from astropy prescription
+        # treat file as download url to check astropy cache for file
+        if not os.path.exists(file) and is_url_in_cache(file, pkgname="pyuvsim"):
+            files_use[i] = cache_contents("pyuvsim")[file]
```
bhazelton, line 4793: It feels a bit weird to hardcode this to pyuvsim. If we're going to hardcode it, I think it should be "pyuvdata" or possibly "rasg". The options I see for setting it to pyuvsim without hardcoding it is adding an option to the yaml, which feels clunky or if we can autodetect that the call is originating in pyuvsim (not sure that's easy/possible).

Another option is to just move all this yaml constructor stuff to pyuvsim...

bhazelton: Sorry, I keep getting confused, this is just checking for a cache, not putting a file there. I think it's ok to check the pyuvsim cache, but I think it should check the pyuvdata cache first.

burdorfmitchell: Would checking the pyuvdata cache just be checking the pyuvdata DATA_PATH, or is there a separate cache for large files that should be targeted? I've looked at the directory and docs and only see mention of object / attribute caching and some astropy caching in testing.

bhazelton: I think we can wait on this for now. We can add it when we tackle the issue of moving our test data out of the pyuvdata repo. Sorry, I've been slow to grok all this. 

I think this looks pretty good as soon as we solve the coverage issue. If there's no good way to trigger it I'm even ok with a `# pragmatic: no cover` since this is tested by pyuvsim ref sims tests.

bhazelton (approved):

PR #1606, opened 2025-08-18
Use UVParameter forms in UVData.__add__ and UVData.fast_concat to prevent missing parameters.

### LINKED ISSUE #1595: Parameter scan_number_array is not updated during concatenation
Related to closed issue #1159 

Combining UVData objects fails when parameter `scan_number_array` is supplied. I am trying to concatenate two time-adjacent UVData objects read from .ms files. Each have 62128 blts, so the combined object should have Nblts=124256. The concatenation fails with error `ValueError: UVParameter _scan_number_array is not expected shape. Parameter shape is (62128,), expected shape is (124256,).`. Setting `uv.scan_number_array=None` for each object before combining bypasses the error.

bhazelton:
## Description

This is following some of the ideas in UVBase._select_along_axis. This could probably be generalized to UVBase, but I wanted to get UVData's version done to work out the ideas before moving it up UVBase. And I wanted to at least get a draft up for @rlbyrne to try out.

## Motivation and Context

fixes #1595

## Types of changes

- [ ] Version change
- [ ] Other

## Checklist:

Bug fix checklist:

### DIFF
--- a/src/pyuvdata/utils/tools.py
+++ b/src/pyuvdata/utils/tools.py
@@ -9,6 +9,8 @@ from collections.abc import Iterable, Iterable as IterableType
 
 import numpy as np
 
+from .types import FloatArray, IntArray, StrArray
+
 
 def _get_iterable(x):
     """Return iterable version of input."""
@@ -687,3 +689,43 @@ def _ntimes_to_nblts(uvd):
         inds.append(np.where(unique_t == i)[0][0])
 
     return np.asarray(inds)
+
+
+def flt_ind_str_arr(
+    *,
+    fltarr: FloatArray,
+    intarr: IntArray,
+    flt_tols: tuple[float, float],
+    flt_first: bool = True,
+) -> StrArray:
+    """
+    Create a string array built from float and integer arrays for matching.
+
+    Parameters
+    ----------
+    fltarr : np.ndarray of float
+        float array to be used in output string array
+    intarr : np.ndarray of int
+        integer array to be used in output string array
+    flt_tols : 2-tuple of float
+        Tolerances (relative, absolute) to use in formatting the floats as strings.
+    flt_first : bool
+        Whether to put the float first in the out put string or not (if False
+        the int comes first.)
+
+    Returns
+    -------
+    np.ndarray of str
+        String array that combines the float and integer values, useful for matching.
+
+    """
+    prec_flt = -2 * np.floor(np.log10(flt_tols[-1])).astype(int)
+    prec_int = 8
+    flt_str_list = ["{1:.{0}f}".format(prec_flt, flt) for flt in fltarr]
+    int_str_list = [str(intv).zfill(prec_int) for intv in intarr]
+    list_of_lists = []
+    if flt_first:
+        list_of_lists = [flt_str_list, int_str_list]
+    else:
+        list_of_lists = [int_str_list, flt_str_list]
+    return np.array(["_".join(zpval) for zpval in zip(*list_of_lists, strict=True)])
--- a/src/pyuvdata/utils/tools.py
+++ b/src/pyuvdata/utils/tools.py
@@ -691,11 +691,11 @@ def _ntimes_to_nblts(uvd):
     return np.asarray(inds)
 
 
-def flt_ind_str_arr(
+def float_int_to_str_array(
     *,
     fltarr: FloatArray,
     intarr: IntArray,
-    flt_tols: tuple[float, float],
+    flt_tol: tuple[float, float],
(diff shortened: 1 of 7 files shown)


tests/uvdata/test_mir.py
```diff
@@ -592,9 +592,11 @@ def test_flex_pol_add(sma_mir_filt):
     sma_yy_copy._make_flex_pol()
 
     # Add the two back together here, and make sure we can the same value out,
-    # modulo the history.
+    # modulo the history and sorting.
     sma_check = sma_yy_copy + sma_xx_copy
 
+    sma_mir_filt.reorder_freqs(channel_order="freq")
+
```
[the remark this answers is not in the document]
bhazelton: Nevermind, I figured out a way to preserve the old behavior.

src/pyuvdata/utils/tools.py
```diff
@@ -687,3 +689,43 @@ def _ntimes_to_nblts(uvd):
...
+
+def flt_ind_str_arr(
+    *,
+    fltarr: FloatArray,
+    intarr: IntArray,
+    flt_tols: tuple[float, float],
+    flt_first: bool = True,
+) -> StrArray:
+    """
+    Create a string array built from float and integer arrays for matching.
```
steven-murray: Would be good to have an example like "E.g. for float 3.7 and int 3 this will create an entry in the output array as '3.7_3'"

src/pyuvdata/uvbase.py
```diff
@@ -808,3 +866,135 @@ def _select_along_param_axis(self, param_dict: dict):
...
+            The axis name (e.g. "Nblts", "Npols").
+        add_len : int
+            The extra length to be padded on for this axis.
+
+        """
+        update_params = self._get_param_axis(axis_name)
+        multi_axis_params = self._get_multi_axis_params()
+        for param, axis_list in update_params.items():
+            if param not in multi_axis_params:
+                continue
```
steven-murray: Why are we not also padding single-axis objects? This is confusing

steven-murray (commented):
Thanks @bhazelton and sorry for the very slow review! I have a few comments. Mostly I think we can make some of these functions a bit more clear in what they are doing. They seem like they will become quite central in how UVxxx objects operate, so we should make them as clear as possible for our own future reference. 

It might also be useful to do a quick profiling for these functions, as I expect they might be bottleneck functions in some applications (especially in terms of memory)

bhazelton:
@steven-murray I finally got memray up and running. I started with a 9 GB MWA uvfits file, which I split along the frequency axis into two objects which I saved out to uvh5 files (each file was ~3 GB -- not sure why that's so much less than half the uvfits size). Then I used memray to run two scripts on this branch and on main. The two scripts just read in the two files into a UVData object using `from_file` and passing a list containing the two file names, the only difference was that in one script I specified `axis="freq"` so it used the `fast_concat` method rather than the `__add__` method.

I'm attaching the memray flame graph html files for all 4 runs as well as the pngs of the memory graph for all 4 runs (the png is just made from the html, use the html to get tooltip readings). My take away is that this refactor causes a small improvement on the memory usage for the `__add__` method. There is very little difference for `fast_concat` but this refactor might be using very slightly more memory. I'm interested in your read of these:

[memray-flamegraph-output_add_main.html](https://github.com/user-attachments/files/25429706/memray-flamegraph-output_add_main.html)
[memray-flamegraph-output_add_refactor.html](https://github.com/user-attachments/files/25429708/memray-flamegraph-output_add_refactor.html)
[memray-flamegraph-output_concat_main.html](https://github.com/user-attachments/files/25429709/memray-flamegraph-output_concat_main.html)
[memray-flamegraph-output_concat_refactor.html](https://github.com/user-attachments/files/25429710/memray-flamegraph-output_concat_refactor.html)

main `__add__`:
memray_add_main

this branch `__add__`:
memray_add_refactor

main `fast_concat`:
memray_concat_main

this branch `fast_concat `:
memray_concat_refactor

steven-murray:
Cool. I think you answered my other points -- though this is still in draft so I can't be approving yet :-)

steven-murray (dismissed):
Thanks @bhazelton I think I am good for this to be merged now :-)

kartographer (commented):
@bhazelton - I think this is looking overall good to me. Most of the places where I would have commented are flags that @steven-murray already raised, and everything else is probably out of scope here. I think you need a rebase, but otherwise I'm happy.

bhazelton:
@kartographer Thanks for the review, I just rebased.

jpober (approved):
Approving following the rebase

PR #1607, opened 2025-08-20
Don't use asserts in non-testing code

bhazelton:
## Description

Asserts can be stripped from python code when it is compiled, so using them is a problem identified by bandit in pre-commit. In the past we ignored those errors, but after some reading I think it's probably something we should address. This PR turns that check back on (except in test code) and converts assert statements to explicitly raising an error. Some of them are not things I can figure out how to trigger, so they have been marked to be ignored in coverage reports for now, but I'm interested in any suggestions for covering them.

## Types of changes

- [ ] Version change
- [x] Other

## Checklist:

Other:
- [ ] I have updated any docstrings associated with my change using the [numpy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ ] I have updated the readme and/or tutorial to reflect my changes (if appropriate).
- [ ] I have added/updated tests to cover my changes (if appropriate).


src/pyuvdata/testing/warning_check.py
```diff
@@ -106,7 +106,11 @@ def __enter__(self):
             raise RuntimeError(f"Cannot enter {self!r} twice")
         _list = super().__enter__()
         # record=True means it's None.
-        assert _list is not None
+        if _list is None:  # pragma: no cover
+            raise RuntimeError(
+                "Something is wrong, _list should not be None. Please file an "
+                "issue in our GitHub issue log so that we can fix it."
```
mkolopanis, line 112: considering this is the warning checker, I cannot imagine many scenarios where this context would be entered without the user knowing. That said, if this error _somehow_ triggered on a test in a package like uvsim which uses this checker it could be ambiguous where to open an issue.

Should we mention something like the following? 

 - Something went wrong in pyuvdata's warning checker. Please file an issue in our GitHub issue log so that we can help: https://github.com/RadioAstronomySoftwareGroup/pyuvdata/issues. Developer info: _list is None

that's pretty verbose though 😕

src/pyuvdata/uvbeam/uvbeam.py
```diff
@@ -2312,10 +2281,22 @@ def interp(
...
+        if az_array is not None:
+            az_array = np.asarray(az_array)
+            arr_to_check["az_array"] = az_array
+        if za_array is not None:
+            za_array = np.asarray(za_array)
+            arr_to_check["za_array"] = za_array
+
+        for name, arr in arr_to_check.items():
+            if arr.ndim != 1:
+                raise ValueError(f"{name} must be one-dimensional")
```
mkolopanis, line 2297: I was wondering where all those little checks went. FOUND THEM!

mkolopanis (commented):
Looking really good. I am pretty comfortable with all the "let's make an issue" posts but do wonder if we should attach the issue log url more, but that's a lot of work. The only one I really have opinions about is the warning checker because we do use that in other testing? maybe that means we should try less to make that one good. I feel like I talk myself back and forth even as I write this. Let me know what you want to do.

mkolopanis (approved):
Thanks for making those tweaks. I think it definitely reads better for the user now if something terrible does happen

PR #1634, opened 2025-11-03
Add `flip_conjugation` method and user info if it would reduce the uvw discrepancy

### LINKED ISSUE #1632: In `uvd.check`, UVW orientation can be flipped even if it leads to a discrepancy >1m
When reading a measurement set into a `UVData` object, the [UVW orientation](https://pyuvdata.readthedocs.io/en/latest/conventions.html#baseline-conjugation-and-uvw-direction) is wrong.
 
I am loading a measurement set with `uvd.read_ms()` and get the warning `The uvw_array does not match the expected values given the antenna positions. The largest discrepancy is 224341.28399469642 meters [...]`.
224km is a large discrepancy. It turns out that the UVW orientation is wrong and should be flipped: once flipped, the discrepancy is "only" ~4m. 

The reason the flip is not performed is that `uvd.check()` compares the `uvd.uvw_array` with the UVW computed from antenna positions (`temp_obj.set_uvws_from_antenna_positions()`) with an absolute tolerance of 1. If the condition is not satisfied, it tries flipping the convention (`uvd.uvw_array *= (-1)`) and re-does the comparison with an absolute tolerance of 1 (`np.allclose(-temp_obj.uvw_array, self.uvw_array, atol=1)`. In my case, the max discrepancy after flipping is 4 which means the flip is not performed. However, the match would be better if the flip happened.
 
I would recommend checking which UVW orientation leads to the smallest discrepancy and apply it to the data, but still print the warning.

Thanks!

bhazelton:
## Description

Pulled the conjugation flip code out as a user-callable method. Added info to the uvw check warning message to inform a user when a conjugation flip might reduce uvw discrepancies.

## Motivation and Context

fixes #1632 

## Types of changes

- [ ] Version change
- [ ] Other

## Checklist:

New feature checklist:

### DIFF
--- a/tests/uvdata/test_ms.py
+++ b/tests/uvdata/test_ms.py
@@ -1017,22 +1017,32 @@ def test_ms_bad_history(sma_mir, tmp_path):
         (
             True,
             "The uvw_array does not match the expected values given the antenna "
-            "positions. The largest discrepancy is 2042.176844761807 meters. "
+            "positions. The largest discrepancy is max_diff meters. "
             "This is a fairly common situation but might indicate an error in the "
             "antenna positions, the uvws or the phasing. It is possible that the "
             "conjugation convention does not match pyuvdata's convention. Flipping "
             "the conjugation with the `flip_conjugation` method would result in "
-            "a maximum uvw discrepancy of 1.1000000000000227 meters.",
+            "a maximum uvw discrepancy of max_flip_diff meters.",
         ),
     ],
 )
 def test_flip_conj(nrao_ms, tmp_path, add_error, msg):
     filename = os.path.join(tmp_path, "flip_conj.ms")
     nrao_ms.set_uvws_from_antenna_positions()
+
+    orig_uvws = nrao_ms.uvw_array.copy()
     if add_error:
         nrao_ms.uvw_array += 1.1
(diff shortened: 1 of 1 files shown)


bhazelton:
It turns out this is causing a lot of test errors. Digging in, there are flips happening when the improvement to the uvw matching to the expected value is quite marginal e.g. 44m max diff instead of 50m. This is giving me bad feelings because it feels quite stochastic -- e.g whether or not it wants to flip can depend on which set of baselines is read in on a partial read.

I think we need a different approach here. Options I can think of:
- Add a method to flip the conjugation and allow the user to call it directly. Update the current warning to tell users what the max uvw difference is if they flip the conjugation and point them to the method if they want to call it themselves.
- Add a parameter to set the tolerance to use when checking for flipped conjugation that defaults to 1m.

Thoughts? @jpober @mkolopanis @adeliegorce

adeliegorce:
Hi! Thank you for having a look at this. I do agree point #3 implies a hardcoded upper limit which can be sub-optimal. Personally I think option #1 is best - or a variation: the flip is an additional input to the `read_ms` method such that it can be done serially in a heavy job importing a lot of datafiles.

PR #1643, opened 2025-12-04
Making uvcalibrate work w/ wideband gains solns

### LINKED ISSUE #1124: Make uvcalibrate support wide-band calibrations
Wide-band calibrations were introduced with UVCal future array shapes, but the `uvcalibrate` function does not support them yet.

kartographer:
_"I'm going, going... back, back... to cali-bration..."_

## Description

`uvcalibrate` has been modified to allow for wide-band `UVCal` gains to be applied to `UVData` objects. As part of that, some minor clean-up has been completed within `uvcalibrate` related to spectral window handling.

## Motivation and Context

Aside from being a requested feature, the above is one of two updates required for directly calibrating SMA data within pyuvdata itself (rather than relying on external programs such as CASA). 

Closes #1124.

N.b., a minor modification has been made to the `Mir` module that causes SMA data so no longer be labeled with a polarization convention due to an error thrown in `uvcalibrate`. It's unclear whether that error is the desired behavior -- this would be good to get clarification on prior to merging this PR.

## Types of changes

## Checklist:

New feature checklist:

### DIFF
--- a/src/pyuvdata/utils/uvcalibrate.py
+++ b/src/pyuvdata/utils/uvcalibrate.py
@@ -200,7 +200,7 @@ def uvcalibrate(
     freq_range_check : bool
         Option to check that frequency ranges on the UVCal object matches the channel
         frequencies given in the UVData object for a given spectral window. Only
-        applicable for wide-band UVCal objects, defaul is True.
+        applicable for wide-band UVCal objects, default is True.
     uvc_pol_convention : str, {"sum", "avg"}, optional
         The convention for how instrumental polarizations (e.g. XX and YY) are assumed
         to have been converted to Stokes parameters in ``uvcal``. Options are 'sum' and


src/pyuvdata/utils/uvcalibrate.py
```diff
@@ -402,43 +398,44 @@ def uvcalibrate(
...
+            if uvcal.wide_band:
+                freq_array_spw = uvdata.freq_array[uvdata.flex_spw_id_array == spw]
+                min_freq = min(uvcal.freq_range[idx])
+                max_freq = max(uvcal.freq_range[idx])
+                if all((freq_array_spw < min_freq) | (freq_array_spw > max_freq)):
+                    raise ValueError(
+                        "SPWs between UVData and UVCal overlap, but the frequencies of "
+                        "channels are inconsistent with frequency ranges."
+                    )
+                uvcal_spws_to_keep.append(spw)
```
bhazelton: It seems like either all the spws end up in this list or it errors, so why bother building it up this way? Why not just set it equal to uvdata.spw_array after doing this check per spw?

kartographer: The original logic I had set up wasn't _quite_ right -- it now better mimics what's present for the frequency axis in non-wideband gains.

bhazelton (approved):
Looks good. Thanks @kartographer !

PR #1651, opened 2026-02-24
Improve memory usage in uvfits reader

bhazelton:
## Description

This makes a few memory handling improvements to the uvfits reader (and maybe the MWA correlator FITS reader).

It turns out that when using `astropy.io.fits.open` with `memmap=True`, the python garbage collector does not release the memory promptly upon exiting the context manager. We were always using `memmap=True`, but it's really only needed when doing partial reads. This PR changes the calls to the `astropy.io.fits.open` to only use `memmap=True` when doing partial reads. This leads to significant memory improvements for the uvfits reader when reading the whole file. It's not clear if there are any improvements to the MWA correlator FITS reader in terms of memory usage, but it seemed better to be consistent.

I also refactored the uvfits reader to remove a temporary array that was being used to store and manipulate the data stored in the primary uvfits HDU before assigning parts of it to the data_array, nsample_array and flag_array. I now have it just assign the full data from the HDU to the data_array and do all the manipulations in place in that array. It feels a bit yucky but leads to much better memory usage.

Below are plots of the memory usage as measured by `memray` for a small script that just reads in an MWA uvfits file (the blue resident size is most important). There is some run-to-run variability to these plots that I don't fully understand, so I'm including 2 plots for each codebase, from different days but all three codebase runs were run back to back on the 2 different days (all on my laptop):

main branch:
uvfits_read_main
uvfits_read_main_2

After using `memmap=False`:
uvfits_read_fix1
uvfits_read_fix1_2

After refactoring out the temporary array:
uvfits_read_fix2
uvfits_read_fix2_2

Note that the peak memory use doesn't change much (and is seems quite variable in the last set of plots), but there's a consistent big change in the final memory usage at the end of the script. If you are doing anything with the object after reading it, that becomes your baseline memory usage for the next step. I actually stumbled on this when trying to profile something else (frequency averaging) and discovered that the memory usage was much higher when I used a script that started by reading in a uvfits file vs when I started by reading in the same data saved as a uvh5 file.

Similar plots for the MWA Correlator FITS reader seem to be dominated by run-to-run variability -- I don't see a consistent improvement with these changes, but I don't see it getting materially worse either.

We could certainly make similar changes in the calfits and beamfits readers, but I wasn't sure I should cram them into this PR.

## Types of changes

- [ ] Version change
- [x] Other

## Checklist:

Other:
- [ ] I have updated any docstrings associated with my change using the [numpy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ ] I have updated the readme and/or tutorial to reflect my changes (if appropriate).
- [ ] I have added/updated tests to cover my changes (if appropriate).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to this project will be documented in this file.
 ### Changed
 - Added improved handling of a few warning messages from h5py and numpy.
 
+### Fixed
+- Significant reduction in memory when reading UVFITS files, particularly when
+reading the entire file (no selection on read).
+
 ## [3.2.5] - 2025-12-11
 
 ### Changed
--- a/src/pyuvdata/parameter.py
+++ b/src/pyuvdata/parameter.py
@@ -1183,7 +1183,7 @@ class LocationParameter(UVParameter):
         """Detect if this location is on the moon."""
         # figure out if this is on the moon. Doing it this way limits
         # attempted imports of lunarsky
-        if isinstance(self.value, EarthLocation):
+        if self.value is None or isinstance(self.value, EarthLocation):
             return False
         else:
             try:
--- a/src/pyuvdata/uvdata/mwa_corr_fits.py
+++ b/src/pyuvdata/uvdata/mwa_corr_fits.py
@@ -48,7 +48,7 @@ def read_metafits(
     telescope_info_only=False,
 ):
     # get information from metafits file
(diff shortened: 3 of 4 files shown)


steven-murray (dismissed):
Sorry this took so long to get to... this solution seems totally reasonable to me!

steven-murray (approved):

PR #1671, opened 2026-05-15
remove spiceypy dependency with updated lunarsky

aelanman:
## Description

The latest version of lunarsky drops the dependency on spiceypy. As a result, the try blocks that catch SPICEUnknownFrameErrors in unit tests can be removed. This also bumps the minimum version of lunarsky.

## Motivation and Context

This cleans up code that will no longer be needed.

- [x] If this is a new CI setup, I have added the associated badge to the readme and to references/make_index.py (if appropriate).


bhazelton:
Thanks @aelanman!  Obviously tests won't pass until the new lunarsky version is released but it's nice to have this ready to go when it is released. There's several other places in the repo where the lunarsky dependency is set, so we'll need to update those as well. Off the top of my head that includes:
- CI conda yamls
- environment.yaml
- readme

Maybe just search the repo to see if there are any more. We also usually put something in the changelog when we increase dependency versions, you can see examples in there from earlier lunarsky version bumps.

aelanman:
Thanks for the reminder about the changelog @bhazelton ! I've added to that (and for pyuvsim) and updated the README and environment.yaml.

bhazelton:
The pyuvsim test errors are caused by pyuvsim assuming an older version of lunarsky. That will be fixed by a separate PR in that repo.

I don't understand the hera_cal failures, I made an issue in that repo to ask about them: https://github.com/HERA-Team/hera_cal/issues/1052

bhazelton (approved):

PR #1691, opened 2026-08-08
Time-based interpolation of UVCal calibration tables

kartographer:
## Description

A new method has been added to UVCal (`interpolate_in_time`) that allows for calibration tables to be "upsampled", effectively allowing gains/delay measurements taken at one set of times to be interpolated/extrapolated to another set of times. Code associated with performing the interpolations has also been added to a new module, `utils.gain_interpolation`, which supports several different interpolation schemes, including nearest, linear, and cubic (PCHIP) interpolation strategies. An additional strategy -- "poly" -- allows for fitting the calibration solution with a variable-order Chebyshev polynomial and then performing interpolation/extrapolation using the fitted polynomial.

A separate small change has been made to `uvcalibrate`, which also allows users to reweight `nsamples_array` based on the amplitude calibration being applied to `data_array`.

## Motivation and Context

Higher frequency interferometers like the SMA frequently interleave observations of science targets with observations of bright quasars (or other point sources) that can act as gain calibrators for the data. Because those observations are disjoint in time with the observations of the "source", interpolation between the derived gain solutions is typically used to derive effective gains corrections during science target observations. This PR effectively provides that functionality, allowing `uvcalibrate` to be used for SMA observations (among other facilities) .

## Types of changes

## Checklist:

New feature checklist:

### DIFF
--- a/src/pyuvdata/utils/gain_interpolate.py
+++ b/src/pyuvdata/utils/gain_interpolate.py
@@ -1,0 +1,1292 @@
+# Copyright (c) 2026 Radio Astronomy Software Group
+# Licensed under the 2-clause BSD License
+"""Gain interpolation related utilities."""
+
+from typing import Literal
+
+import numba as nb
+import numpy as np
+
+from .types import BoolArray, ComplexArray, FloatArray, IntArray
+
+_MIN_CUBIC_SAMPLES = 4
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+def _flag_outlier_cal(cal_array, tol_lim, is_gain_amp=False, is_gain_pha=False):
+    """
+    Flag individual outliers in a set of calibration solutions.
+
+    Marks a sample as bad where the deltas to both of its neighbors exceed the
+    tolerance but the delta that skips it does not, which is the signature of an
+    isolated outlier rather than a genuine excursion.
+
+    Parameters
+    ----------
+    cal_array : ndarray of float or complexfloat
+        Calibration solutions to check, shape (Nants, Nfreqs, Ntimes, Njones), dtype
+        is either float or complex (assumed complex if `is_gain_phase` or `is_gain_amp`
+        is set to True).
+    tol_lim : float
+        Largest delta between samples that is considered acceptable. Units match
+        those of the deltas being tested, which depend on `is_gain_amp` and
+        `is_gain_pha`.
+    is_gain_amp : bool
+        If True, treat `cal_array` as complex gains and test fractional changes in
+        amplitude. Default is False.
+    is_gain_pha : bool
+        If True, treat `cal_array` as complex gains and test changes in phase, in
+        radians. Default is False.
+
+    Returns
+    -------
+    new_flags : ndarray of bool
+        Flags for `cal_array`, same shape as `cal_array`. True where the sample is
+        NaN (or zero and `is_gain_amp` or `is_gain_pha` is set), or looks like an
+        isolated outlier.
+    """
+    # Set the initial flags based on nans (and zeros if the solns are complex).
+    new_flags = np.isnan(cal_array)
+    if is_gain_amp or is_gain_pha:
+        if not np.issubdtype(cal_array.dtype, np.complexfloating):
+            raise ValueError(
+                "cal_array must be complex if is_gain_amp or is_gain_pha is True."
+            )
+        mask = cal_array == 0.0
+        if np.any(mask):
+            cal_array = np.where(mask, np.nan, cal_array)
+            new_flags |= mask
+
+    # Dividing through the NaNs when flagged is deliberate, so suppress the warnings.
+    with np.errstate(invalid="ignore"):
+        if is_gain_amp:
+            delta_12 = np.abs(cal_array[:, :, 1:, :] / cal_array[:, :, :-1, :]) - 1
+            delta_13 = np.abs(cal_array[:, :, 2:, :] / cal_array[:, :, :-2, :]) - 1
+        elif is_gain_pha:
+            delta_12 = np.angle(cal_array[:, :, 1:, :] / cal_array[:, :, :-1, :])
+            delta_13 = np.angle(cal_array[:, :, 2:, :] / cal_array[:, :, :-2, :])
+        else:
+            delta_12 = np.real(cal_array[:, :, 1:, :] - cal_array[:, :, :-1, :])
+            delta_13 = np.real(cal_array[:, :, 2:, :] - cal_array[:, :, :-2, :])
+
+    # Where a delta between adjacent points exceeds the threshold, mark it as "bad"
+    bad_12 = (delta_12 > tol_lim) | (delta_12 < -tol_lim)
+    # Where a delta that skips a point is within the threshold, mark it as "good"
+    good_13 = (delta_13 <= tol_lim) & (delta_13 >= -tol_lim)
+    # Separately capture where the delta is NaN
+    nan_12 = np.isnan(delta_12)
+
+    # Where two bad deltas are adjacent, and the larger delta that skips the midpoint
+    # is good, assume the midpoint is bad and flag it.
+    new_flags[:, :, 1:-1, :] |= (bad_12[:, :, 1:, :] & bad_12[:, :, :-1, :]) & good_13
+
+    # If an outlier delta is next to a NaN, assume its bad and flag it too.
+    new_flags[:, :, 1:-1, :] |= bad_12[:, :, 1:, :] & nan_12[:, :, :-1, :]
+    new_flags[:, :, 1:-1, :] |= nan_12[:, :, 1:, :] & bad_12[:, :, :-1, :]
+
+    # Use the "goodness" of the second and second-to-last points to determine the
+    # flags for the first and last points (if the deltas exceed the threshold).
+    new_flags[:, :, 0, :] |= bad_12[:, :, 0, :] & (
+        np.abs(delta_12[:, :, 1, :]) <= tol_lim
+    )
+    new_flags[:, :, -1, :] |= bad_12[:, :, -1, :] & (
+        np.abs(delta_12[:, :, -2, :]) <= tol_lim
+    )
+
+    return new_flags
+
+
+def build_interp_flags(
+    cal_array,
+    flag_array,
+    old_times,
+    new_times,
+    tol_lim=0.0,
+    is_gain_amp=False,
+    is_gain_pha=False,
+    flag_adjacent=True,
+    max_time_delta=0.0,
+    allow_extrapolation=True,
+):
+    """
+    Build the flags that apply to a set of interpolated calibration solutions.
+
+    Parameters
+    ----------
+    cal_array : ndarray
+        Calibration solutions being interpolated, shape (Nants, Nfreqs, Ntimes, Njones),
+        dtype is either float or complex (assumed complex if `is_gain_phase` or
+        `is_gain_amp` is set to True).
+    flag_array : ndarray of bool
+        Flags for `cal_array`, shape (Nants, Nfreqs, Ntimes, Njones), dtype bool.
+    old_times : ndarray of float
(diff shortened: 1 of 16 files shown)


src/pyuvdata/utils/gain_interpolate.py
```diff
@@ -0,0 +1,1296 @@
...
+    Returns
+    -------
+    new_flags : ndarray of bool
+        Flags for the interpolated solutions, shape (Nants, Nfreqs, Nnew, Njones), dtype
+        bool.
+    """
+    # Capture any NaN values in cal_array and flag them.
+    flag_array = flag_array | np.isnan(cal_array)
+
+    # If this is a gain amplitude array, also flag any zero values (which are invalid).
```
bhazelton: This comment combined with the code below confused me because this comment implies we're working with amplitudes but the code below requires the cal_array to be complex. Also the code below applies when checking phases as well, so I think this comment is just out of date?

src/pyuvdata/utils/gain_interpolate.py
```diff
@@ -0,0 +1,1296 @@
...
+            np.nan,
+            dtype=old_cal.dtype,
+        )
+
+    # Number of usable samples for each antenna/freq/jones solution.
+    n_good_times = n_old_times - np.sum(old_flags, axis=2)
+
+    # Every unflagged solution shares the same time grid, so the whole array can be
+    # fit in one pass, which is typically 1-2 orders of magnitude faster than going
+    # slice by slice. Worst case, this adds a ~10% overhead of all solutions are flagged
```
bhazelton: "of" -> "if"  I think...

src/pyuvdata/utils/gain_interpolate.py
```diff
@@ -0,0 +1,1296 @@
...
+    # (though the logic check below traps this worst case scenario).
+    if np.any(n_good_times == n_old_times):
```
bhazelton, line 676: Question on this: It's common in processing for EoR to have the same time/frequency flags across all antennas. When that's the case, I think you could do the batch processing but I'm not quite sure if that's handled here or not. Basically the question is: does n_old_times include times that are fully flagged across all antennas and frequencies?

src/pyuvdata/utils/gain_interpolate.py
```diff
@@ -0,0 +1,1296 @@
...
+        Times at which to evaluate the interpolation, shape (Nnew,), dtype is float,
+        units are Julian days.
+    interp_mode : str
+        Quantity to interpolate. Options are "ampphase" (amplitude and phase,
+        interpolated separately), "complex" (real and imaginary parts, interpolated
+        separately), "real" (real-values only), "imag" (imaginary values only), "amp"
+        (amplitude only), and "phase" (phase only). Default is "ampphase".
+    kind : str
+        The type of interpolation to use. Options are "nearest", "linear", "cubic",
+        and "poly" (polynomial fit using Chebyshev polynomials). Default is "linear".
```
bhazelton: I might favor using a different name for Chebyshev polynomials because there are several kinds of polynomials and I could imagine a future scenario where someone wants some other kind of polynomial and the "poly" name would be taken (we often use low order "standard" polynomials in our cal fits).

src/pyuvdata/utils/gain_interpolate.py
```diff
@@ -0,0 +1,1296 @@
...
+        with each other or with the time arrays.
+
+    """
+    valid_modes = {"real", "imag", "amp", "phase", "ampphase", "complex"}
+    if interp_mode not in valid_modes:
+        raise ValueError(
+            f"Unrecognised interp_mode '{interp_mode}'. Choose from {valid_modes}."
+        )
+
+    if not (np.issubdtype(old_cal.dtype, np.complexfloating) or interp_mode == "real"):
```
bhazelton, line 1196: This logic seems slightly off to me, but I might just be confused...

src/pyuvdata/utils/uvcalibrate.py
```diff
@@ -213,6 +312,33 @@ def uvcalibrate(
...
+        Keywords describing which data to calibrate, using the same vocabulary
+        `UVData.select` accepts -- e.g. `{"catalog_names": ["3c279"]}` to calibrate
+        one source, or `{"antenna_nums": [1, 2]}`, `{"time_range": [...]}`,
+        `{"blt_inds": [...]}`. Default is None, which calibrates everything. See the
+        `UVData.select` docstring for further details.
+    uvc_select_kwargs : dict, optional
+        Keywords passed to `UVCal.select` to choose which solutions to calibrate
+        with, e.g. `{"catalog_names": ["3c279"]}` to use the solutions derived from
+        one source. Default is None, which uses all of the solutions. See the
+        `UVCal.select` docstring for further details.
```
bhazelton: This is interesting. It sounds like this can end up with a mixture of calibrated and uncalibrated data on the same object and I'm not sure there's a way to know after the fact which pieces of data were calibrated and which weren't. I can see that this could be useful for an expert user but even for them I worry about how to track the provenance of what was done to what part of the data. Is there some way to capture the information about what was and wasn't calibrated in the history if no where else? Also, I think we should warn people about doing this kind of thing cavalierly.

Another question arises for me too with the `uvc_select_kwargs`: if I use this, does the data that are not matched in the UVData object get flagged?

src/pyuvdata/utils/uvcalibrate.py
```diff
@@ -324,6 +450,41 @@ def uvcalibrate(
...
+        if "inplace" in interp_kwargs:
+            raise ValueError(
+                "Cannot set inplace via interp_kwargs, since uvcalibrate needs to "
+                "leave the UVCal object it was handed alone."
+            )
+        if uvcal.time_range is not None:
+            raise ValueError(
+                "Cannot interpolate a UVCal object that uses time_range rather than "
+                "time_array, since there is no single time to interpolate from."
+            )
```
bhazelton: This is interesting. We commonly compute a single calibration solution from a range of times to improve the signal to noise on the calibration solution and then record it using the `time_range` parameter to indicate that it is valid across that range of times. In that case I'd argue that the center of the time_range is an appropriate time to use for interpolation.

src/pyuvdata/uvcal/uvcal.py
```diff
@@ -4037,8 +4037,9 @@ def _select_preprocess(
             list of spw indices to keep. Can be None (to keep everything).
         freq_inds : list of int
             list of frequency indices to keep. Can be None (to keep everything).
-        pol_inds : list of int
-            list of polarization indices to keep. Can be None (to keep everything).
+        jones_inds : list of int
+            list of Jones polarization indices to keep. Can be None (to keep
+            everything).
```
bhazelton, line 4042: good catch.

src/pyuvdata/utils/uvcalibrate.py
bhazelton: Maybe an option (default True) which flags any uncalibrated data would be helpful?

src/pyuvdata/uvcal/uvcal.py
```diff
@@ -4400,6 +4401,224 @@ def _convert_to_filetype(self, filetype):
...
+            Option to check optional parameters as well as required ones. Default is
+            True.
+        run_check_acceptability : bool
+            Option to check acceptable range of the values of parameters after
+            interpolating. Default is True.
+
+        Returns
+        -------
+        UVCal object or None
+            The interpolated object, or None if `inplace` is True.
```
bhazelton, line 4527: I think this method should update the history with relevant information about how it was interpolated/extrapolated.

bhazelton: I think maybe that information should also be copied to the uvdata history when we call uvcalibrate?

bhazelton (commented):
This is a pretty major enhancement and looks super useful. I just have a few comments and questions.

src/pyuvdata/utils/gain_interpolate.py
kartographer: The trap here is to avoid having real-only values interpolated via amp/phase/imaginary `interp_mode` settings, since these are sort of nonsensical for real-only data (and I managed to get myself stupidly tripped up when testing where "ampphase" was being used for the delays). So either the cal table has to be complex, _or_ if not complex, `interp_mode` has to be doing real-only interpolation.

src/pyuvdata/utils/gain_interpolate.py
kartographer: That's a good point -- the typical case for SMA is that we often have an entire antenna flagged for all times (and that's partially trapped here in the loop below, same for a bad freq channel), but I didn't think about the case of one bad time soln. Let me see if I can implement something better here...

src/pyuvdata/utils/uvcalibrate.py
kartographer: I can add that for the UVData object -- the idea here was that if you needed to take an iterative approach to applying the gains, either because different sources (e.g., calibrators versus target fields) requiring different strategies or different solutions (e.g., self-cal _after_ applying the gains from the calibrators, which would nominally only be applied to the science targets). I think I would prefer to keep the default False though -- if you are doing this in an iterative fashion (as I expect we will w/ SMA) it becomes a bit of a nuisance to keep setting the flag and an opportunity, whereas if you're doing this in one shot adding the flag in there. I'll try to make this clear in the docstring though so as to ensure it gets used w/ care. 

For `uvc_select_kwargs` I think the case is a little simpler -- it only matches what's in UVCal, so it's as though the cal records weren't there to begin with when calling `uvcalibrate`. For time-based selection I think that's fine, although I realized in the way I constructed the code a subtle bug could creep in if, e.g., running antenna/pol selection that'll be fixed here shortly. I think this now fails in a sensible way (fails if you select something on UVCal, not on UVData but is present in UVData), but happy to chat further on this if need be.

src/pyuvdata/utils/uvcalibrate.py
kartographer: That's fair -- I didn't want to assume that at the start since SMA doesn't use time ranges in this way, but if you're happy with that, I'm happy to code it that way!

src/pyuvdata/utils/gain_interpolate.py
kartographer: Okay, I clarified it in 4dc0e1c -- the comment was a bit of a holdover, but yes, anytime the gain is `0 + 0j`, that means that the soln is implicitly "bad".

src/pyuvdata/utils/gain_interpolate.py
kartographer: Yup, fixed in 4dc0e1c.

src/pyuvdata/utils/gain_interpolate.py
kartographer: Okay, I optimized this a bit further in 4dc0e1c -- I know screen out time solns that are all bad.

src/pyuvdata/utils/gain_interpolate.py
kartographer: Okay, I made the keyword "chebyshev" instead just to make it clear.

src/pyuvdata/utils/gain_interpolate.py
kartographer: Added a comment in the code in 4dc0e1c to address this.

src/pyuvdata/utils/uvcalibrate.py
kartographer: Added in 4dc0e1c -- default is not to flag BUT to raise a warning.

src/pyuvdata/utils/uvcalibrate.py
kartographer: Added in 4dc0e1c

src/pyuvdata/uvcal/uvcal.py
kartographer: Added this in 4dc0e1c

bhazelton (approved):
This looks good to me now, thanks!

PR #1694, opened 2026-08-15
Numba-enhanced Data Averaging (and more!)

### LINKED ISSUE #730: Frequency averaging
It would convenient to be able to average the visibilities and weights as a function of frequency (i.e. go from 80kHz resolution to 160kHz resolution). There currently isn't an option to do this. However, time averaging has already been implemented, so it should be straightforward to implement frequency averaging.

### LINKED ISSUE #1311: Make UVData.frequency_average more flexible
The current idea would be to add an option to pass in an integer array that specifies which final channel to put each input channel into (and allowing the user to drop channels by specifying -1 in this array). This would give a power user full flexibility to average in complicated ways.

### LINKED ISSUE #1624: High memory usage when frequency averaging MWA correlator fits files
I have read in a ~13 GB set of MWA correlator fits files (from one 2 minute observation at 40kHz) whose frequency resolution I'd like to reduce. However, when I run `frequency_average(n_chan_to_avg=2)` on this data, the process consumes more than my available system memory (96 GB) and fails. Averaging over 4 channels succeeds but still uses almost all available memory. As a workaround, I can manually process a few files at a time and concatenate the results, but it might be nice to have a built-in process that doesn’t load as much data into memory at once.

kartographer:
## Description

A new numba kernel for averaging data (with arbitrary grouping along one axis) has been added and integrated into `UVData.frequency_average` and `UVData.downsample_in_time`, which should substantially increase its speed (via multithreaded processing) and significantly reduce the memory footprint of the averaging step.

## Motivation and Context

The primary motivation for this was addressing the speed and memory issue with frequency averaging, given the prior report about MWA data suffering a ~6x increase in memory footprint, although because the new kernel is somewhat flexibly designed to handle irregularly grouped arrays, I took the opportunity to add a few additional features to `frequency_average` (including an option to sum across `nsample_array`, which is appropriate for when it's being used to carry weights, as is the case in a few different filetypes). 

Closes #730 
Closes #1311 
Closes #1624

## Types of changes

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ were calibrated when a selection was used, whether the rest were flagged, and ho
 solutions were interpolated if `interpolate` was set.
 - Updating code to no longer set shapes directly on ndarrays (deprecated in NumPy 2.5)
 - Require lunarsky>1.0 to drop spiceypy requirement
+- Refactored `UVData.frequency_average` and `UVData.downsample_in_time` to use the new
+numba-based `utils.averaging.mapped_average` routine, which is significantly faster and
+uses less memory (substantially so for `UVData.frequency_average`).
 
 ### Added
 - A new method `UVCal.interpolate_in_time`, which allows for UVCal calibration tables
@@ -33,6 +36,15 @@ and feed_aligned_projection beams which result from an E-field beam decompositio
 used in some analysis codes.
 - Added more options for controlling labels on beam plots.
 - Support for MWA Average Embedded Element beams in UVBeam.
+- A new utility function `utils.averaging.mapped_average`, which performs a weighted
+average of an array along a single axis using an arbitrary map of input to output
+positions.
+- A `chan_map` keyword to `UVData.frequency_average`, which allows for mapping input to
+output channels, allowing for more sophisiticated averaging schemes (e.g., unevenly
+sized bins).
+- The keywords `weight_by_nsample` and `sum_nsample` to `UVData.frequency_average`,
+which control whether channels are weighted by `nsample_array` in the average, and
+whether `nsample_array` records the summed or the fractional number of samples.
 
 ### Fixed
 - A bug where `UVCal.pol_convention` was not being properly set when reading in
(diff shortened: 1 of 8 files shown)


src/pyuvdata/utils/averaging.py
```diff
@@ -0,0 +1,193 @@
...
+    they need be neither contiguous nor evenly sized, and elements can be dropped from
+    the result entirely.
+
+    Parameters
+    ----------
+    data : ndarray
+        Array of values to average, shape (..., Naxis, ....).
+    index_map : ndarray of int
+        Array of shape (Naxis,), giving the index of the bin that each element along
+        the middle axis that each value contributes to.
```
bhazelton: What should this have in it for an element that should be excluded entirely? Also, what do you mean by "middle axis"?

bhazelton: edit: looks like the "middle axis" is a carry over from the numba function, but I think it should be abstracted here since this function takes an "axis" parameter?

CHANGELOG.md
```diff
@@ -35,6 +47,9 @@ used in some analysis codes.
 - Support for MWA Average Embedded Element beams in UVBeam.
 
 ### Fixed
+- A bug where `UVData.frequency_average` would either raise an IndexError or zero out
+fully-flagged channels in all spectral windows but the first, rather than recording
+the average of the flagged values.
```
bhazelton, line 52: oof, glad you found and fixed this!

bhazelton (commented):
Looks fantastic, thank you! I just have a couple comments on docstrings. My student is going to test this soon on a full MWA dataset.

src/pyuvdata/utils/averaging.py
kartographer: Okay, I took another pass at it -- hopefully it's a little clearer now.

bhazelton (approved):
@ldress reports that it works excellently for her. Thank you so much @kartographer !

Issue #1710, opened 2026-09-03
Support moving near-field phase centers


wfarah:
## Summary

Near-field phasing currently supports only a focal point fixed at scalar `cat_lon`, `cat_lat` and `cat_dist`. I'd like to allow `cat_times` on `near_field` catalog entries so that the focus can move during an observation.

## Motivation

We want to be able to represent a moving source, like a satellite whose path is known from telemetry, in our data. There's no way to do that in a single `UVData` object today: `generate_phase_center_cat_entry` rejects `cat_times` for anything but `ephem`, and all other types coerce the coordinates to scalars. You have to split the observation into per-integration objects, phase each one separately, and concatenate.

## Proposed approach

Extend the existing `near_field` type rather than adding a new one. With `cat_times` absent, behavior is unchanged. With `cat_times` present, equal-length `cat_lon`, `cat_lat` and `cat_dist` arrays describe a moving focus interpolated onto the integration times, reusing the existing ephemeris interpolator. The near-field geometry itself (`w = r1 - r2`) doesn't change.

## Plan

Two PRs, following CONTRIBUTING's one-feature-per-PR guidance:

1. Make the private near-field correction selection-aware, adding `select_mask` and `update_vis` to `_apply_near_field_corrections()`. No behavior change apart from one bug fix that falls out of it.
2. The timed near-field feature on top.

Does extending `near_field` sound like the direction you'd want before I open the first one?

PR #1711, opened 2026-09-03
Near field select mask

wfarah:
## Description

Adds keyword-only `update_vis` and `select_mask` arguments to the private `UVData._apply_near_field_corrections()`, and threads `select_mask` through `_get_focus_xyz`, `_get_nearfield_delay` and `_ntimes_to_nblts`.

`select_mask` fixes a bug: `phase()` honors `select_mask` everywhere except the near-field correction, which took no mask and rewrote `uvw_array[:, -1]` for every record before applying the w-projection to the whole data array. Phasing a subset of records to a near-field target therefore also overwrote the w-coordinate and rotated the visibilities of records belonging to *other* phase centers, silently — in the added test the w of unselected records moves from 0.28 m to 26.41 m while their u and v stay identical.

`update_vis` gates only the w-projection, so the exact near-field w can be assigned without modifying data that has not been phased to the focal point. It has no caller in this PR — it is groundwork for time-dependent near-field phase centers (#1710), where `set_uvws_from_antenna_positions()` applies the correction to objects whose visibilities may not be phased. Happy to split it out if you'd rather not carry an unused argument.

Behavior with no selection is unchanged and bit-identical in both `uvw_array` and `data_array`. This also replaces the index loop in `_ntimes_to_nblts` with `np.unique(..., return_inverse=True)`, which computes the same mapping without a Python loop over `Nblts`; happy to revert that if you'd prefer the diff stay narrower.

## Motivation and Context

Groundwork for moving (time-dependent) near-field phase centers, proposed in #1710. A moving focus applies to its own records and its `cat_times` need not span the whole file, so the near-field correction has to be selection-aware. Fixing the ignored `select_mask` is the same code change, so it comes along here.

## Types of changes

## Checklist:
- [x] I have read the contribution guide.

Bug fix checklist:
- [x] I have updated the CHANGELOG.

New feature checklist:
- [x] I have added or updated the docstrings associated with my feature using the numpy docstring format.
- [ ] I have updated the tutorial to highlight my new feature (if appropriate). — not applicable, all changes are to private methods
- [x] I have updated the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- A bug in `UVData.phase` where near-field phasing ignored `select_mask` and applied the
+near-field correction to every record, overwriting the w-coordinate and rotating the
+visibilities of records belonging to other phase centers.
+
 ## [3.2.7] - 2025-08-20
 
 ### Added
--- a/src/pyuvdata/utils/phasing.py
+++ b/src/pyuvdata/utils/phasing.py
@@ -2521,7 +2521,7 @@ def uvw_track_generator(
     }
 
 
-def _get_focus_xyz(uvd, focus, ra, dec):
+def _get_focus_xyz(uvd, focus, ra, dec, select_mask=None):
     """
     Return the x,y,z coordinates of the focal point.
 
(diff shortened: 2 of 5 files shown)


wfarah:
I was hoping a [stacked PR](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests) would work here, but it seems I can't do that from a forked repo/branch. In any case, this [PR](https://github.com/wfarah/pyuvdata/pull/1) is the followup one that implements the actual major changes.

bhazelton:
> I was hoping a [stacked PR](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests) would work here, but it seems I can't do that from a forked repo/branch. In any case, this [PR](https://github.com/wfarah/pyuvdata/pull/1) is the followup one that implements the actual major changes.

Wow, I didn't know about Stacked PRs! I need them! Thanks for the link to the other PR, the context is helpful for interpreting this one.

bhazelton:
This looks really good to me, thanks for the bug fix! I'd like to get @kartographer to take a look at it, he's our resident phasing expert.

kartographer (approved):
I think this guy looks good -- the next PR is the one I'll have a bit more commentary on.

wfarah:
Thanks! I'll rebase and open a new PR

PR #1713, opened 2026-09-04
Moving near field

wfarah:
## Description

Lets the focal point of a `cat_type="near_field"` phase center move during an observation, by allowing `cat_times` on near-field catalog entries.

The contract is purely additive:

- Without `cat_times`, behavior is unchanged: scalar `cat_lon`, `cat_lat` and `cat_dist` describe a fixed focus.
- With `cat_times`, equal-length 1-D `cat_lon`, `cat_lat` and `cat_dist` arrays describe a track, interpolated onto the integration times.
- The near-field geometry itself is untouched — the path difference is still exactly `w = r1 - r2`. No new `cat_type`.

Interpolation reuses `interpolate_ephem()` unchanged, so a timed near-field entry behaves like an `ephem` entry: nearest-neighbour for exactly matching samples, a spline otherwise, and `interp1d`'s out-of-range `ValueError` if the track does not span the data.

The main changes:

- `generate_phase_center_cat_entry()` accepts `cat_times` for `near_field`, requires `cat_lon`/`cat_lat`/`cat_dist` alongside it, rejects duplicate `cat_times` (which would make the interpolation ambiguous), and coerces the coordinates to `(Npts,)` arrays.
- A new private `_resolve_near_field_focus()` in `utils.phasing` returns the focus for a set of times, resolving fixed versus moving entries and converting the stored parsec `cat_dist` into meters.
- `calc_app_coords()` routes a timed `near_field` entry through the `ephem` branch.
- `phase()` accepts array `lon`/`lat` for a timed near-field entry, and resolves the focus through the new helper.
- `set_uvws_from_antenna_positions()` re-applies the near-field correction after recalculating the uvws — see below.

### Bug fix: `set_uvws_from_antenna_positions()` silently dropped the focus

`calc_uvw()` only knows far-field geometry, so recalculating the uvws replaced the near-field w with the far-field one and rotated the visibilities to match, losing the focus with no warning. This also reached `UVData.new()` and `UVData.update_antenna_positions()`, which call that method. The fix re-applies the correction once per near-field phase center afterwards, keeping the same two-stage order that `phase()` uses: ordinary uvws first, then the near-field correction. This is what consumes the `update_vis` and `select_mask` arguments added in #1711.

### Bug fix: `phase(lookup_name=True)` used the caller's `cat_type`, not the resolved one

On a lookup, `_phase_dict_helper()` resolves the type from the catalog but rebinds its own local `cat_type`, so the resolved value never reaches `phase()`. Two guards at the end of `phase()` read that stale value, while everything around them already uses `phase_dict["cat_type"]`. A looked-up `near_field` entry therefore received only far-field phasing and the adjacent `unprojected` guard let the w-projection rotate visibilities that should have been left alone. Both now test the resolved type; happy to split the `unprojected` half out.

### Tests

Fourteen new tests: catalog validation, the resolver, moving-focus phasing checked against phasing each integration separately with a fixed focus, interpolation onto times that deliberately do not coincide with the track samples, uvw recalculation for both fixed and moving focus, a UVH5 round trip, rephasing (near-field to sidereal, and refocusing to a new distance), and the two `lookup_name` paths above. I sanity-checked the suite by mutation: not assigning w, skipping the visibility rotation, flipping the rotation sign, ignoring `cat_times`, snapping the interpolation to nearest-neighbour, and reverting either `lookup_name` guard are each caught.

### Known limitations and possible follow-ups

- `update_antenna_positions(delta_antpos=True)` is not near-field aware. It bypasses `set_uvws_from_antenna_positions()` and adds a linear `delta_uvw`, which cannot reproduce `r1 - r2`, so the near-field w comes out approximate. I left it alone rather than widen this PR. Separately, and independently of near-field, its `_apply_w_proj()` call passes `select_mask=self._check_for_cat_type("unprojected")`, which rotates the *unprojected* records; that looks backwards to me, but I haven't touched it. Happy to file it separately if it is unintended.
- `interpolate_ephem()` interpolates RA straight through the 0/2pi branch cut, which sends the target the long way around the sky — I measured phase errors up to 163 degrees, and with a cubic spline it corrupts the whole track rather than just the crossing. It affects `ephem` entries equally, so this PR only leaves a `TODO`; the fix (unwrap before interpolating, re-wrap after) is small and belongs on its own.
- Angles are interpolated independently rather than Cartesian position, which is slightly unfaithful for sparse, fast-moving tracks. I would rather document that callers supply densely-sampled tracks than add a second interpolator, but happy to revisit.
- Two further pre-existing near-field problems that I ran into while scoping this, and deliberately left alone: the parsec `cat_dist` is handed to the far-field astrometry as a parallax distance, which distorts `phase_center_app_ra`/`app_dec` and therefore `u` and `v` (only `w` is corrected afterwards); and `cat_frame` is ignored by `_get_focus_xyz()`, which builds its `SkyCoord` with no frame argument and so always assumes ICRS. Neither is filed yet — happy to open issues for them if they are unintended.

## Motivation and Context

Proposed in #1710. The path of a moving near-field source, such as a satellite pass, is known from telemetry as a time series of positions, and there is currently no way to represent that in a single `UVData` object — the observation has to be split into per-integration objects, phased separately, and concatenated.

Builds on #1711.

## Types of changes

## Checklist:
- [x] I have read the contribution guide.

Bug fix checklist:
- [x] I have updated the CHANGELOG.

New feature checklist:
- [x] I have added or updated the docstrings associated with my feature using the numpy docstring format.
- [x] I have updated the CHANGELOG.


src/pyuvdata/utils/phase_center_catalog.py
```diff
@@ -780,6 +811,15 @@ def generate_phase_center_cat_entry(
         cat_dist = None if cat_dist is None else float(cat_dist)
         cat_vrad = None if cat_vrad is None else float(cat_vrad)
 
+    # A focal point has to sit at a real, positive range. A non-finite distance
+    # propagates NaNs into the w-coordinate and the visibilities without complaint,
+    # and a non-positive one has no geometric meaning.
+    if (cat_type == "near_field") and (cat_dist is not None):
```
kartographer: `cat_dist` technically should be positive across the board, so if we're going to input check here we might as well do it regardless of `cat_type`.

src/pyuvdata/utils/phasing.py
```diff
@@ -1969,6 +1969,10 @@ def interpolate_ephem(
         if ephem_vel is not None:
             vel_vals += ephem_vel
     else:
+        # TODO: RA samples that cross the 0/2pi branch cut look like a 2pi jump to the
```
kartographer: This is a good catch, but is there a reason this couldn't be incorporated into `_interpolate_ephem`? Seems like it should be a two-line addition (just `np.unwrap` and `np.mod`).

src/pyuvdata/utils/phasing.py
```diff
@@ -2521,6 +2528,64 @@ def uvw_track_generator(
     }
 
 
+def _resolve_near_field_focus(*, phase_dict, time_array):
```
kartographer: So this helper function is called only one time, and it's _highly_ specialized. I would integrate the logic here back into `UVData.phase`.

kartographer:
So I am in the process of reviewing this, but I'm wondering if there's a further parameter related to phase center distance that should be carried w/ UVData (like how `phase_center_app_ra` is carried). Let me talk to @bhazelton about this...

@wfarah  -- FWIW, if you are looking at supporting sats, there's two things worth keeping in mind. First, there's no ToF handling on the pyuvdata side, so all of this should be assuming that element of bookkeeping is being held in ephem generation. Second, a potentially missing piece that you may need to add here, which is handling of relative differences in doppler/LOS velocity that is not presently handled. For the prior/first deployment of the near-field phasing w/ MWA, this wasn't really a leading factor, but if you're thinking about looking at, e.g., LEOs, it's a piece that's missing and you'll need some additional TLE info to be stored (fortunately, `cat_vel` is already a supported field here, it just makes the handling a but more complex).

PR #1717, opened 2026-09-12
A batch of fixes to MeasurementSet vis/cal data reading and writing

### LINKED ISSUE #1714: MeasurementSets fail validation check when DATA's QuantumUnits is a list
Hi,

I ran into an issue with loading MeasurementSets via pyuvdata@main ("b4dba02") following a change in how the MS writer wrote out metadata. In this case, `pyuvdata` started raising validation errors on the `vis_units` parameter, derived from the DATA column of the MeasurementSet (full stack trace here can be found here [0]).

As defined by UVData, that parameter is only allowed to be a string [1], however the code for loading MeasurementSets is loading from the QuanumUnits value [2], which the 3.0β definition says is a "String Vector", so it can be loaded as a list in Python

I'm not sure how you'd want to handle this case, but a hack that got the files loading for me was to just check if the list is a length of 1, and take that value as the entry for vis_units [3]. If that's considered to be a fine approach (given I hope someone isn't mixing units in there), I can open an MR with that fix

Cheers,
David

[0] https://gist.github.com/David-McKenna/97e8a6aff3ee26a296efddb9ead7fb6f
[1] https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/b4dba0227b008951dc4ac5155a7c9ed4323fe520/src/pyuvdata/uvdata/uvdata.py#L90-L97
[2] https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/b4dba0227b008951dc4ac5155a7c9ed4323fe520/src/pyuvdata/uvdata/ms.py#L446
[3] https://github.com/David-McKenna/pyuvdata/commit/09ca3692215bf9b2428ce42e7aa53daffbb12dbf

kartographer:
## Description

A series of changes were made to address a number of bugs that were uncovered in poking around with the MeasurementSet reading/writing routines for both visibilities and calibration tables -- mostly related to data reordering as required for CASA. With the exception of a change to `UVData._convert_from_filetype` (which now drops extra parameters that are added to subclasses of `UVData`), all changes have been confined to the `MS` and `MSCal` classes.

To facilitate some of the ordering handling, I've added two new parameters for the `MS` class -- `MS.corrected_data` and `MS.model_data`, which now carries the inputs to the keywords of the same name on write, and allows for the various reordering operations to also act on those tables as well. Note that we may want to promote these to (optional) `UVData` parameters at some point, but I've decided to keep things confined here to where they're needed.

## Motivation and Context

The original motivation for this branch was to deal with an issue where `UVData.vis_units` wasn't being set correctly when the corresponding header from a MeasurementSet produced a length-1 list. Separately in doing some testing of the new `uvcalibrate` features (added in #1691), I discovered some bugs that arose from the fact that polarization and blt ordering were being inconsistently applied in a few places. Separately, I _also_ discovered that the conjugation scheme nominally used for MS-formatted calibration tables is flipped from what `uvcalibrate` expects, and from that determined that we've been likely reading/writing in these tables incorrectly. The good news is that the error cancelled itself on roundtrip, I ran into it because I'm trying to stuff my own bespoke format into `UVCal` to plumb through to MS (and my conjugation scheme is the same as MS).

Closes #1714 

## Types of changes

## Checklist:

Bug fix checklist:


src/pyuvdata/uvcal/ms_cal.py
```diff
@@ -387,7 +387,10 @@ def read_ms_cal(
 
         # Use if this is a delay soln
         if self.cal_type == "gain":
-            self.gain_array = ms_cal_soln
+            # CASA gains conjugation scheme is the opposite of pyuvdata's, so fix that
```
bhazelton, line 389: Is this about the conjugation convention of the visibilities (i.e. the thing discussed here: https://pyuvdata.readthedocs.io/en/latest/conventions.html#baseline-conjugation-and-uvw-direction)?

If so, can you maybe update that a little bit? Do MS calibration solutions always use the other convention or just sometimes?

src/pyuvdata/uvdata/ms.py
```diff
@@ -43,6 +44,26 @@ class MS(UVData):
 
     """
 
+    def __init__(self):
+        # Define these internal to MS so that the data can be ordered accordingly.
+        # Maybe at some point we can port these out to the UVData class more generally.
```
bhazelton, line 58: Hmm, this is the exception to the rule that users should only interact with UVData, not with the file type specific classes...

Do we have info on this in our tutorial or other docs?

kartographer: I think in this case, there is no user interaction here (or rather the interaction is w/ the keywords, I don't really want users to have direct interactions with these parameters) -- these are effectively "helper parameters" that just allow the MS class to leverage the same tools as the other parameters. I could arguably do the proper thing of propagating these up to `UVData`, but that seemed like a bigger change then maybe we want to deal with here. I could duplicate the sorting code, but that also seems sub-ideal.

I hadn't planned on putting in items into the docs about it, but I could if needed.

bhazelton: Ah, ok. I misunderstood their purpose.

src/pyuvdata/uvcal/ms_cal.py
kartographer: It's slightly different, but definitely related. But I can update the docs accordingly.

kartographer: Updated in [bcf2741](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1717/commits/bcf27413390ce7a285b524a094f8043ae2e663ad).

docs/conventions.rst
```diff
@@ -23,13 +23,16 @@ the position of the first (after which various rotations are applied). The secon
...
+MIRIAD; with UVFITS, FHD, Mir, and MeasurementSet (used within CASA) formats using the
+opposite convention. Note that this also applies to calibration solutions derived from
+the dataset, with the implication being that gains solved in the *ant2-ant1* convention
+are the complex conjugates of those solved in the *ant1-ant2* convention.
+
+Note that separately from the conjugation convention, some platforms have requirements
+for which antenna is listed first in the baseline (e.g., AIPS needs UVFITS files
+recorded with :math:`\text{ant1}<\text{ant2}`), separate from the conjugation
+convention. pyuvdata is able to support so-called "mixed" conventions in this respect,
+with the antenna ordering convention being independent of the conjugation convention.
```
bhazelton, line 35: Thank you, this is great.

src/pyuvdata/uvdata/ms.py
```diff
@@ -475,13 +516,9 @@ def _read_ms_main(
         if flip_conj is None:
             # if we got to this point, it means that the conjugation scheme has not
             # been encoded into the dataset, which is _either_ and old pyuvdata written
```
bhazelton: and -> an

src/pyuvdata/uvdata/ms.py
```diff
@@ -475,13 +516,9 @@ def _read_ms_main(
...
-            # file or written external to pyuvdata. CASA's convention is not 100% clear,
-            # but testing of the code base reveals that CASA supports both conventions.
-            # Which convention is used is dependent on antenna numbering, i.e. whether
-            # ant1 >= ant2 or ant1 <= ant2 (flip_conj=False for the former and True for
-            # the latter). This seems to explain the apparent contradictions in the
-            # documentation, and the inconsistent results we have seen w/ importuvfits.
-            flip_conj = (not pyuvdata_written) and np.all(ant_1_arr <= ant_2_arr)
+            # file or written external to pyuvdata. If the former, it means that the
+            # conjugation scheme is pyuvddata's, so _don't_ flip the data.
```
bhazelton: pyuvddata -> pyuvdata

CHANGELOG.md
```diff
@@ -4,10 +4,31 @@ All notable changes to this project will be documented in this file.
...
+`flag_array` cross-polarization entries.
 - A bug in `UVData.phase` where near-field phasing ignored `select_mask` and applied the
 near-field correction to every record, overwriting the w-coordinate and rotating the
 visibilities of records belonging to other phase centers.
 
+### Changed
+- `UVData.read_ms` and `UVData.write_ms` now always flip conjugation scheme on read and
+write for non-pyuvdata generated MeasurementSets (it was previously believed that when
+ant1 > ant2, the conjugation scheme may be flipped, but that was a misreading of CASA's
+codebase).
```
bhazelton, line 30: Good find. It's amazing how complicated this stuff is!

bhazelton (commented):
Looks great, thank you! I just found a couple typos in comments.

src/pyuvdata/uvdata/ms.py
kartographer: Oy, well that's just downright embarrassing!

src/pyuvdata/uvdata/ms.py
kartographer: Fixed in b2ca9952.

src/pyuvdata/uvdata/ms.py
kartographer: Fixed in b2ca9952, next time I'm going to write a PR to rename the repo to save face :wink:

CHANGELOG.md
kartographer: It has definitely made me appreciate the conventions page a lot more!

bhazelton (approved):

NickSwainston:
Could we please get a release for this?

bhazelton:
> Could we please get a release for this?

Sure. See #1719.

PR #1720, opened 2026-09-19
Unwrap interpolate_ephem RA before spline interpolation

### LINKED ISSUE #1712: RA wrapping in `interpolate_ephem`
##  Problem

[`interpolate_ephem`](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/500800eaf0d02d4c48128f2b04bcdd70c81e699a/src/pyuvdata/utils/phasing.py#L1889) interpolates `ephem_ra` directly. A track crossing RA of 0 is stored wrapped into `[0, 2pi)`, so the samples contain a 2pi jump that the interpolant reads as real motion, and it sends the target the "long way" around the sky. Found while implementing moving near-field support (#1710). Deliberately **not** fixed in that PR series because it is a pre-existing bug that affects `ephem` phase centers and deserves its own PR.

Because the interpolatant is a cubic spline for 4 or more samples, this is not a local error at the crossing; the discontinuity rings across the entire track.

## Evidence

A target drifting steadily at 0.05 rad/sample through RA = 0, with 5 samples:

    stored ephem_ra   : [6.2    6.25   0.0168 0.0668 0.1168]
    continuous form   : [6.2    6.25   6.3    6.35   6.4   ]

Interpolated at every midpoint between samples, compared as angles on the circle so the wrap itself is not counted as error:

| midpoint | expected RA | interpolated RA | error |
|---|---|---|---|
| 1 | 6.2250 | 1.8071 | 106.9 deg |
| 2 | 6.2750 | 2.8389 | 163.1 deg |
| 3 (crosses 0h) | 0.0418 | 5.6378 | 39.4 deg |
| 4 | 0.0918 | 0.7790 | 39.4 deg |

All four midpoints are wrong, including even the two before the crossing.

## Fix

Interpolate an unwrapped copy of `ephem_ra`, and fold the result back into `[0, 2pi)` range on return.

Add tests of `ephem` and `near_field` tracks crossing RA of 0 are interpolated correctly, assert non-wrapping track results remain unchanged, and that interpolated returned RA is in `[0, 2pi)` range.

shaneraphel:
`interpolate_ephem` interpolates `ephem_ra` as stored. A track that crosses RA 0 is wrapped into [0, 2π), so cubic/linear interpolation treats the 2π jump as real motion and rings the whole track, not only the crossing sample.

Interpolate an unwrapped copy of `ephem_ra`, then fold the result back with `np.mod(..., 2π)`. Dec interpolants are unchanged.

Fixes #1712


kartographer:
So I'm going to reject/dismiss this with prejudice -- the account here seems to be AI-created and spamming a bunch of open-source repos, and the account that created the PR appears to be different than the one that did the work, which makes me a little suspicious of what's going on here. If the PR confirmed to pyuvdata standards, I might be inclined to give it a pass, but it'd be less work to spin up a separate PR than to try to coach the agent to do what it is that we need it to do.

shaneraphel:
thanks for the review.i am human author;i designed the logic and use my own formal language system to coding,i agree move this code to independence pr,and i have no employer,those fix just my nonprofit work,and not generation by llm or any pytorch mod.

